PR #6, opened 2016-04-15
First iteration of utils module

ghost:
### What was wrong?

The utils module was not implemented.
### How was it fixed?

This is the first iteration, which is not ready to merge yet. Some functions may be removed, as they are nonsensical in a Python environment. Tests are required as well.
#### Cute Animal Picture

xslgjwa


pipermerriam:
Can you split up ` web3/utils/utils.py` into something like the following:

* `utils/encoding.py` for string operation
* `utils/currency.py` for currency stuff
* `utils/abi.py` for some of the event name stuff.
* etc as you see fit.

Otherwise the utils file is going to be HUGE

void4:
Done. We now have:

- `utils/utils.py` for general string formatting and utility functions
- `utils/address.py` for address utilities
- `utils/currency.py` for currency unit converters
- `utils/abi.py` ABI for call formatters
- `utils/encoding.py` for string encodings and numeric representations

web3/utils/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    Returns true if object is BigNumber, otherwise false
+    """
+    raise NotImplementedError()
+
+
+def isString(obj):
+    """
+    Returns true if object is string, otherwise false
+    """
+    return isinstance(obj, str)
```
pipermerriam: for things like this we'll need to use the `six` library

```
import six
isinstance(obj, six.string_types)
```

web3/utils/utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def isFunction(obj):
+    """
+    Returns true if object is function, otherwise false
+    """
+    return hasattr(obj, "__call__")
```
pipermerriam: You can use `return callable(obj)` for this.

void4:
This is mostly complete, I think we can merge this now.

tests/test_encoding.py
```diff
@@ -0,0 +1,72 @@
...
+import web3.utils.encoding as encoding
+
+
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+    ('myString', '0x6d79537472696e67'),
+    ('myString\x00', '0x6d79537472696e6700'),
+    #('\u0003\u0000\u0000\u00005èÆÕL]\u0012|Î¾\u001a7«\u00052\u0011(ÐY\n<\u0010\u0000\u0000\u0000\u0000\u0000\u0000e!ßd/ñõì\f:z¦Î¦±ç·÷Í¢Ëß\u00076*
\bñùC1ÉUÀé2\u001aÓB', '0x0300000035e8c6d54c5d127c9dcebe9e1a37ab9b05321128d097590a3c100000000000006521df642ff1f5ec0c3a7aa6cea6b1e7b7f7cda2cbdf07362a85088e97f19ef94331c955c0e9321ad386428c')
```
pipermerriam: How come this is commented out?

pipermerriam:
Looks great.  No blockers.  Merge away whenever you're ready.

tests/test_encoding.py
void4: Python2 tries to decode the unicode characters from the input anyway, then throws an exception when they are not in the ASCII range. Python3 can ignore them when the "ignore" parameter on str.decode("ascii") is added, but creates different results than Javascript. So we can either:

- modify the function to work with this edge case
- simply treat it as invalid input and throw an exception as it does now (it is called fromAscii after all)

Despite this, I will merge this branch now, because some other PRs depend on it (I don't want to merge between non-master branches too often.

PR #10, opened 2016-04-16
Add method.py

ghost:
### What was wrong?

method.py was missing.
### How was it fixed?

The file was added.

Still missing:
- tests
- callback handling/removal
#### Cute Animal Picture

12140183346_eeb042d2dc_z

### DIFF
--- a/web3/web3/method.py
+++ b/web3/web3/method.py
@@ -1,0 +1,127 @@
+import utils.utils
+import errors
+
+
+class Method:
+
+    def __init__(self, options):
+        self.name = options["name"]
+        self.call = options["call"]
+        self.params = options["params"] if "params" in options else 0
+        self.inputFormatter = options["inputFormatter"]
+        self.outputFormatter = options["outputFormatter"]
+        self.requestManager = None
+
+    def setRequestManager(self, rm):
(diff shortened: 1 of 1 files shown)


web3/web3/method.py
```diff
@@ -0,0 +1,127 @@
+import utils.utils
+import errors
+
+
+class Method:
```
pipermerriam: Writing classes like this comes from an older python that didn't have `object`.

The current *one-way-to-do-things* in python regarding classes is to inherit from `object` (unless you have a compelling reason not to).  Mind updating this?

void4: This is only necessary because we want to support Python 2 as well, correct? (Because in Python 3, _object_ is inherited from implicitly if it is omitted). http://stackoverflow.com/questions/4015417/python-class-inherits-object/9448136#9448136

pipermerriam: Fancy, I wasn't aware of that.

I would however like to support python 2.7 so can we add it?

pipermerriam:
Since nothing is using this yet I'd prefer not to merge until this is actually in use.

void4:
Closing this and merging with the web3 branch.

PR #11, opened 2016-04-16
Add web3.py

void4:
### What was wrong?

web3.py was missing

### How was it fixed?

It was added.

This module depends on various utils/ modules and should probably be added only once these are finalised.

#### Cute Animal Picture

adwklhj


pipermerriam:
This also doesn't look ready to be merged.  Tests, Documentation, and it seems like a lot of what it depends on isn't actually in place yet?  Am I missing anything?

void4:
Not yet, this is the difficult part. See the my monologue on gitter.

PR #13, opened 2016-04-18
Add solidity parameter types, encoder, decoder

ghost:
### What was wrong?

The entire web3/solidity folder was not implemented.
### How was it fixed?

I translated it. To be done:
- comments
- flaking
- tests
#### Cute Animal Picture

nature-bird-flying-red


void4:
Reminder: fix web3.eth.defaultAddress

void4:
Should the IPCProvider automatically connect to the testnet IPC socket when the mainnet socket was not found or only do so explicitly, when testnet=True and no ipcpath is given?

PR #14, opened 2016-04-18
Implemented formatter.py

void4:
### What was wrong?

formatter.py was not implemented.

### How was it fixed?

I did exactly that.
It still requires
- the fix of the "WTF" location (see Gitter, forEach return)
- comments (the original source isn't too specific)
- testing

#### Cute Animal Picture

hamster


void4:
The tests for this module are complete now. As it depends on the utility modules, which are implemented in another branch, either a merge with that branch or with master are necessary. If the latter is preferred, this merge will be delayed until we have resolved the multithreaded architecture.

void4:
Flake will fail until I have fixed the WTF section (see the Javascript snippet on gitter).

tests/test_formatter.py
```diff
@@ -0,0 +1,335 @@
...
+    [
+    ('0x0c5496aee77c1ba1f0854206a26dda82a81d6d8'),
+    ('0x0c5496aee77c1ba1f0854206a26dda82a81d6d8'),
+    ('00c5496aee77c1ba1f0854206a26dda82a81d6d'),
+    ('XE7338O073KYGTWWZN0F2WZ0R8PX5ZPPZE'),
+    ('0x'),
+    ]
+)
+def test_inputAddressFormatter2(value):
+    with pytest.raises(Exception):
```
pipermerriam: *nitpick*

It may be better to catch the specific exception class this raises to be sure this test isn't passing due to some other unexpected exception being raised.

web3/web3/formatter.py
```diff
@@ -0,0 +1,192 @@
...
+        return blockNumber
+    return encoding.toHex(blockNumber)
+
+
+def inputCallFormatter(options):
+    """
+    Formats the input of a transaction and converts all values to HEX
+    """
+
+    options["from"] = options.get("from") or config["defaultAccount"]
```
pipermerriam: the *python* way to do this would be `options.setdefault('from', config["defaultAccount"])`

web3/web3/formatter.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Formats the input of a transaction and converts all values to HEX
+    """
+    options["from"] = options.get("from") or config["defaultAccount"]
+    options["from"] = inputAddressFormatter(options["from"])
+
+    if options.get("to"):
+        options["to"] = inputAddressFormatter(options["to"])
+
+    for key in ["gasPrice", "gas", "value", "nonce"]:
```
pipermerriam: *nitpick*

You could change this list to a `set` and make this a tiny bit faster (and slightly more correct in the purist sense)

pipermerriam:
:+1: no blockers.

:shipit: whenever you want.

Issue #18, opened 2016-06-09
Bugs in code for creating a new contract on the blockchain


michailbrynard:
* Version: 0.1.9
* Python: 2.7/3.4/3.5
* OS: osx/linux/win

### What was wrong?
The contract.new() function fails. I've fixed most of the obvious errors, will submit as a pull request.

#### Cute Animal Picture

![image](https://cloud.githubusercontent.com/assets/7025263/15943908/5f9747d4-2e8b-11e6-9e15-dc5b1607138a.png)

PR #19, opened 2016-06-09
Fix bugs in contract.new()

michailbrynard:
### What was wrong?
The contract.new() function failed due to bugs in the code. Issue #18 

### How was it fixed?
I fixed the obvious errors in the code. The new() function now creates a contract on the blockchain. It returns the transaction hash rather than the contract address, as the checkForContractAddress() function is still not implemented.

#### Cute Animal Picture
image


void4:
Thank you!

PR #33, opened 2016-07-27
Added test and patch to support addresses as constructor args.

XertroV:
### What was wrong?

Addresses could not be used as constructor arguments

### How was it fixed?

Added relevant code to `utils/abi.py` to check the address type (was omitted).
Also wrote relevant tests (basically copying what was written for other contracts)

#### Cute Animal Picture

SoCute!

### DIFF
--- a/web3/utils/abi.py
+++ b/web3/utils/abi.py
@@ -78,6 +78,12 @@ def is_encodable(_type, value):
 
         max_length = int(sub)
         return len(value) <= max_length
+    elif base == 'address':
+        if not is_string(value):
+            return False
+        if len(value) != 42:
+            return False
+        return True
     else:
         raise ValueError("Unsupported type")
 
--- a/web3/utils/abi.py
+++ b/web3/utils/abi.py
@@ -13,6 +13,9 @@ from .types import (
     is_integer,
     is_boolean,
 )
+from .address import (
+    is_address,
+)
 
 
 def filter_by_type(_type, contract_abi):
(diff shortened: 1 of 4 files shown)


web3/utils/abi.py
```diff
@@ -78,6 +78,12 @@ def is_encodable(_type, value):
 
         max_length = int(sub)
         return len(value) <= max_length
+    elif base == 'address':
+        if not is_string(value):
```
pipermerriam: Do you mind changing this to use the `is_address` utility function found at `web3.utils.address.is_address`?

pipermerriam:
Thank you for the contribution.  Just the one small change and I'll happily get this merged and released.

web3/utils/abi.py
XertroV: Done! Have also removed the length check because it's implied by `is_address`.

XertroV:
You're welcome :)
All updated.

XertroV:
The last travis job failed because the regex checks have the patterns encoded as strings, but the address was bytes. I've added `address = force_text(address)` in the `is_address` method to fix this, though not sure if it's the best approach. Another approach would be making the patterns into bytes instead of strings, but that will probably cause all the address tests to fail (as they're strings).

This approach is the smallest way to solve the problem I can think of.

pipermerriam:
This change is available in the `1.4.0` release on pypi or here in the releases section of this repository.  Thanks again for your contribution.

PR #34, opened 2016-07-27
adjust text coersion for is_address

pipermerriam:
### What was wrong?

The manner in which the `address` argument was being coerced to text in the `is_address` helper wasn't ideal.

### How was it fixed?

adjusted it to use the decorator pattern found elsewhere (as well as implementing the appropriate decorator)

#### Cute Animal Picture

> put a cute animal picture here.

babypanda_b

### DIFF
--- a/web3/utils/address.py
+++ b/web3/utils/address.py
@@ -11,6 +11,7 @@ from .encoding import (
 )
 from .string import (
     force_text,
+    coerce_args_to_text,
 )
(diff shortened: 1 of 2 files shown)


XertroV:
Nice, way more elegant.

pipermerriam:
Thank you for doing the heavy lifting.  I'll get you a release with this in it tomorrow.

PR #48, opened 2016-08-01
Added information how to work around not implemented functionality.

miohtama:
### What was wrong?

Web3 does not yet support all of JavaScript. However this is not a limitation of the underlying protocol itself, but the state of web3.py project.

### How was it fixed?

Document how you can access RPC methods that are not yet implemented in web3 if you need to move fast.

#### Cute Animal Picture

turtle

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,13 +101,34 @@ web3.config.defaultBlock = "latest"
 
 ### Timeouts, blocking and nonblocking requests
 
-
-All function and property requests block until a respose is received.  Asynchronous function calling has not yet been implemented.
+All function and property requests block until a response is received.  Asynchronous function calling has not yet been implemented.
 
 ```python
 # Blocks indefinitely
 >>> web3.eth.getBalance("0xaddress", timeout=None)
 23423234
+```
+
+### Not implemented functionality
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,15 +101,36 @@ web3.config.defaultBlock = "latest"
...
 
 ```python
 # Blocks indefinitely
 >>> web3.eth.getBalance("0xaddress", timeout=None)
 23423234
 ```
 
+### Not implemented functionality
+
+web3.py does not implement all JavaScript API yet. In the case of missing functionality you can directly call the underlying RPC API using [ethereum-rpc-client](https://github.com/pipermerriam/ethereum-rpc-client). For example you can use this to access the filter logs for events (topics) of smart contracts.
```
pipermerriam: I would prefer to route people to just directly use `web3.request_manager.request_blocking("eth_mining", [])` as I don't think I plan on supporting the `ethereum-rpc-client` and `ethereum-ipc-client` libraries going forward.

pipermerriam:
In general, +1 on this.  Also, filters are really really close.

https://github.com/pipermerriam/web3.py/pull/32

Once that is green I'll be merging and releasing which opens up the filtering events api for use.

miohtama:
Ok. I can change the example how to directly use underlying `request_manager`.

pipermerriam:
Filters and the Events interface for contracts are now in master.  Will probably release tonight or tomorrow depending on how much time I can carve out.

pipermerriam:
I don't believe this pull request still applies.

PR #65, opened 2016-08-15
fix top level module imports

pipermerriam:
### What was wrong?

The `README` suggested that `from web3 import TestRPCProvider` should work, when actually it isn't available form there.

### How was it fixed?

I added `TestRPCProvider` to the things that can be imported from the top level `web3` module.

#### Cute Animal Picture

dog-toys


pipermerriam:
Fixes #65

PR #72, opened 2016-08-30
Remove implicit gevent.

miohtama:
### What was wrong?

Using web3.py in the same Python interpreter as other stuff causes Python interpreter to hang.

### How was it fixed?

Never do implicit gevent.


pipermerriam:
fixed via #73

PR #90, opened 2016-09-08
Fixes KeyError when TypeError is raised in merge_args_and_kwargs

4gn3s:
### What was wrong?
When `merge_args_and_kwargs` is passed an incorrect argument count, it raises a `KeyError` instead of `TypeError` because of a typo: there is no 'input' in `function_abi`.

### How was it fixed?
Fixed the typo.

#### Cute Animal Picture

Cute animal picture


pipermerriam:
Thanks.  THis is now available in `2.7.1` on pypi.

PR #92, opened 2016-09-09
Add support for SSL connections

Arachnid:
### What was wrong?
SSL connections weren't supported

### How was it fixed?
Added the ssl flag for HTTPClient

#### Cute Animal Picture
Not this time.


pipermerriam:
Looks like the test failure was just a timeout.  I restarted the test run.  Will merge when green.

PR #97, opened 2016-09-10
Fix `Filter.get()` to work if no log entries are returned

Arachnid:
### What was wrong?
`eth_getFilterLogs` and `eth_getFilterChanges` return None, rather than an empty list, if no results are returned. This causes an exception in the following code that tries to iterate over it.

### How was it fixed?
Check if the result is falsy, and replace it with an empty list if so.


pipermerriam:
Would you be willing to add a test for this?

PR #100, opened 2016-09-12
speed up travis test runs

pipermerriam:
### What was wrong?

Tests are slow

### How was it fixed?

Parrallelize them

#### Cute Animal Picture

tumblr_lqau71vilc1qd4n9xo1_250


pipermerriam:
Not faster.

PR #101, opened 2016-09-16
default to data=fn_selector in encodeABI

pdobacz:
### What was wrong?

getting call data using `encodeABI` required to call private functions of web3.Contract
### How was it fixed?

immediate fix = data arg of `encodeABI` will default to fn_selector from `_get_function_info`. After this change, cleanup is required https://gitter.im/pipermerriam/populus?at=57dbe1c1c3e7045a30549051
#### Cute Animal Picture

Cute animal picture


pdobacz:
<sigh> I've tried getting back to this and fixing, but found myself a bit confused about the `data` argument purpose (looking at the failing tests and `encodeABI` and `_encode_abi` functions)...

For now it seems that `fn_selector` is not the expected default when `data=None`

pipermerriam:
So what you are seeing is failures related to the previous behavior.  All of those tests need to be updated to the new behavior.  What I would recommend is to update those tests to make the call with `data=''` and they *should* begin passing again.

Then if you can please write new tests showing the new behavior when the `data` parameter is not provided.

pipermerriam:
Oh, so the purpose of the `data` argument is so that the ABI encoding can be isolated from the function selector or deployment bytecode.  For example, if you take a look at this line https://github.com/pipermerriam/web3.py/blob/master/web3/contract.py#L620 you can see that we build the deployment transaction payload by passing in the contract code as the `data` parameter.

PR #103, opened 2016-09-23
Correct use of geventhttpclient and RPC connection

miohtama:
### What was wrong?

`RPCProvider` did not correctly use underlying geventhttpclient connection pool. This may easily cause the leakage of TCP/IP connections in long running processes.

Furthermore the old implementation did not correctly use HTTP Keep-Alive, leading to extra overhead when making RPC calls.

### How was it fixed?

Created a new class `KeepAliveRPCProvider` that maintains in-process per-host connection pool.  Old `RPCProvider` is still available for legacy API compatibility.

Documented the intended use of `KeepAliveRPCProvider´

#### Cute Animal Picture

random

### DIFF
--- a/tests/providers/conftest.py
+++ b/tests/providers/conftest.py
@@ -3,7 +3,7 @@ import pytest
 from gevent import socket
 
 from web3.providers.ipc import IPCProvider
-from web3.providers.rpc import TestRPCProvider, RPCProvider
+from web3.providers.rpc import TestRPCProvider, RPCProvider, KeepAliveRPCProvider
 
 
 def get_open_port():
@@ -15,7 +15,7 @@ def get_open_port():
     return port
 
 
-@pytest.fixture(params=['tester', 'rpc', 'ipc'])
+@pytest.fixture(params=['tester', 'rpc', 'ipc', 'keep-alive-rpc'])
 def disconnected_provider(request):
     """
     Supply a Provider that's not connected to a node.
@@ -31,6 +31,8 @@ def disconnected_provider(request):
         return provider
     elif request.param == 'rpc':
         return RPCProvider(port=9999)
+    elif request.param == 'rpc':
+        return KeepAliveRPCProvider(port=9999)
     elif request.param == 'ipc':
         return IPCProvider(ipc_path='nonexistent')
     else:
--- a/tests/providers/conftest.py
+++ b/tests/providers/conftest.py
@@ -31,8 +31,8 @@ def disconnected_provider(request):
         return provider
     elif request.param == 'rpc':
         return RPCProvider(port=9999)
-    elif request.param == 'rpc':
-        return KeepAliveRPCProvider(port=9999)
+    elif request.param == 'keep-alive-rpc':
+        return KeepAliveRPCProvider(port=9998)
     elif request.param == 'ipc':
         return IPCProvider(ipc_path='nonexistent')
     else:
--- a/web3/providers/rpc.py
+++ b/web3/providers/rpc.py
@@ -7,6 +7,9 @@ import logging
 from .base import BaseProvider  # noqa: E402
 
 
+logger = logging.getLogger(__name__)
+
+
 class RPCProvider(BaseProvider):
     """Create a RPC client.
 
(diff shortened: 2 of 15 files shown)


tests/providers/conftest.py
```diff
@@ -31,6 +31,8 @@ def disconnected_provider(request):
         return provider
     elif request.param == 'rpc':
         return RPCProvider(port=9999)
+    elif request.param == 'rpc':
+        return KeepAliveRPCProvider(port=9999)
```
pipermerriam: I don't think this is going to work with assigning the same port.  I believe it would be best to use the `get_open_port` function (feel free to patch the previous hard-coded port as well)

web3/providers/rpc.py
```diff
@@ -57,6 +74,94 @@ def make_request(self, method, params):
         return response_body
 
 
+class KeepAliveRPCProvider(BaseProvider):
```
pipermerriam: Is there a reason this doesn't inherit from the `RPCProvider`?  Looks like there is a decent amount of duplication between the two classes.

web3/providers/rpc.py
```diff
@@ -57,6 +74,94 @@ def make_request(self, method, params):
...
+        super(KeepAliveRPCProvider, self).__init__(*args, **kwargs)
+
+        self.client = self.get_or_create_client()
+
+    def get_or_create_client(self):
+        from web3 import __version__ as web3_version
+
+        key = "{}:{}".format(self.host, self.port)
+        client = KeepAliveRPCProvider.clients.get(key)
+        if client:
```
pipermerriam: Here is a pattern that I've used a lot in the past and liked for doing caching like this

```python

def my_cached_method(self):
    if key not in cache:
        ... do things
        cache[key] = thing
    return cache[key]
```

I've found the control flow is easier to follow.  Just a suggestion.

tests/providers/conftest.py
```diff
@@ -31,6 +31,8 @@ def disconnected_provider(request):
         return provider
     elif request.param == 'rpc':
         return RPCProvider(port=9999)
+    elif request.param == 'rpc':
```
pipermerriam: And I'm assuming this needs to be changed to `'keep-alive-rpc'`

web3/providers/rpc.py
miohtama: I'd like to drop `RPCProvider` away if possible and just do `RPCProvider = KeepAliveRPCProvider` at the end of file. Then for `TestRPCProvider` do it directly from `BaseProvider`. I feel the classes are so small that chaining inheritance hierarchy at the moment will add more complexity with little benefit, as basically `make_request` is the only function.

web3/providers/rpc.py
miohtama: Edited function for better control flow readability.

web3/providers/rpc.py
pipermerriam: I'm fine with that design decision.

pipermerriam: I think it might be prudent to adjust `RPCProvider` to be a subclass of the `KeepAliveRPCProvider` that raises a [`DeprecationWarning`](https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning) at instantiation.  I've been meaning to start being nicer about breaking changes and now is as good a time as any.

pipermerriam:
Looks like I'm seeing some upstream commits.  Can you potentially try a `git rebase master` after pulling down the latest master to clean up commit history?

miohtama:
I'll close this PR and squash a new rebased single commit PR out of this to clean this up.

web3/providers/rpc.py
```diff
@@ -57,6 +74,94 @@ def make_request(self, method, params):
...
+
+    HTTP client and underlying TCP/IP network connection is recycled across requests.
+    :class:`HTTPClient` connection pooling is used for connections.
+
+    Preferably create only one instance of KeepAliveProvider per process,
+    though the class has internal in-process cache for clients.
+    """
+
+    #: In-process client cache keyed by host:port -> HTTPClient
+    clients = {}
```
pipermerriam: I *think* it would clean things up by moving this to a global variable at the module level which would allow extracting the `get_or_create_client` logic entirely from this class.  That would allow writing a very basic test for it as well without requiring much overhead.  Just a thought.  Not sure it's the right design decision.

pipermerriam:
Merging via https://github.com/pipermerriam/web3.py/pull/120

PR #117, opened 2016-10-13
Whisper support

shravan-shandilya:
### What was wrong?
Shh delegations were not implemented.
### How was it fixed?
Added missing implementations and basic tests.
#### Cute Animal Picture
pr


shravan-shandilya:
@pipermerriam How do I configure Travis to run Shh related testcases?

tests/shh-module/test_shh_filter.py
```diff
@@ -0,0 +1,21 @@
+import random,string,gevent
+
+def test_shh_filter(web3,skip_if_testrpc):
+    skip_if_testrpc(web3)
+    topic = "".join(random.choice(string.ascii_uppercase + string.digits + string.ascii_lowercase) for _ in range(10))
```
pipermerriam: Instead of using `random` here can we use something deterministic or non-random.  I have sound that unless the test is intentionally [fuzzing](https://en.wikipedia.org/wiki/Fuzz_testing) the inputs that randomness can lead to infuriating hard to reproduce test failures.

tests/shh-module/test_shh_filter.py
```diff
@@ -0,0 +1,21 @@
...
+    payloads.append("".join(random.choice(string.ascii_uppercase + string.digits + string.ascii_lowercase) for _ in range(20)))
+    web3.shh.post({"topics":[web3.fromAscii(topic)],"payload":web3.fromAscii(payloads[len(payloads)-1])})
+    gevent.sleep(1)
+    
+    payloads.append( "".join(random.choice(string.ascii_uppercase + string.digits + string.ascii_lowercase) for _ in range(20))) 
+    web3.shh.post({"topics":[web3.fromAscii(topic)],"payload":web3.fromAscii(payloads[len(payloads)-1])})
+    gevent.sleep(1)
+
+    for message in recieved_messages:
+        assert web3.toAscii(message["payload"]) in payloads
```
pipermerriam: So this will pass silently if there are zero received messages.  I think you need to add another assertion that ensures that at least *some* messages were received.

web3/shh.py
```diff
@@ -1,21 +1,60 @@
...
+)
+
+
 class Shh(object):
     """
     TODO: flesh this out.
     """
     def __init__(self, web3):
         self.web3 = web3
+        self.request_manager = web3._requestManager
```
pipermerriam: Instead of storing this, can you either make a property that accesses this or access it directly through the `web3` object.  This pattern will break some of the new managers that I've been working on that distribute requests across multiple providers or managers.

```python
@property
def request_manager(self):
    return self.web3._requestManager
```

tests/shh-module/test_shh_filter.py
```diff
@@ -0,0 +1,21 @@
+import random,string,gevent
+
+def test_shh_filter(web3,skip_if_testrpc):
```
pipermerriam: *nitpick*

Can you add a space after the `,` here and in the other test cases as well.

tests/shh-module/test_shh_filter.py
shravan-shandilya: Agreed.

web3/shh.py
shravan-shandilya: Sure.I'll go with the property approach.electric

shravan-shandilya:
@pipermerriam Not sure why https://travis-ci.org/pipermerriam/web3.py/jobs/167550202 is failing.

andylockran:
Yeah, that's weird - just checked it out locally and it's worked for me - was hoping I'd be able to contribute. 🎱

shravan-shandilya:
@andylockran Thanks for checking!

pipermerriam:
This just happens sometimes.  Been meaning to increase that timeout.  I restarted the run.  Should pass and I'll get this merged.

pipermerriam:
Doh, ok, one more hoop if you're willing to jump through it.  Can you add/update the docs @ [docs/web3.shh.rst](https://github.com/pipermerriam/web3.py/blob/master/docs/web3.shh.rst) with basic explanations of how to use these methods?

Bonus points for functional example code for a common use case 😈

shravan-shandilya:
Sure.Will update soon.

shravan-shandilya:
@pipermerriam Am I missing anything?

pipermerriam:
This looks great.  Thank you for doing this.  Very well rounded PR.  Going to get this merged now but likely not released until early next week.

shravan-shandilya:
@pipermerriam Thanks to you!

PR #119, opened 2016-10-24
fix for issue: 106

shravan-shandilya:
### What was wrong?
We had to explicitly yield the context after creating the filter

### How was it fixed?
Added a _gevent.sleep(0)_ inside _Filter.watch()_ method

#### Cute Animal Picture
baby_panda

### DIFF
--- a/tests/filtering/test_filter_against_latest_blocks.py
+++ b/tests/filtering/test_filter_against_latest_blocks.py
@@ -16,9 +16,8 @@ def test_filter_against_latest_blocks(web3_empty, wait_for_block, skip_if_testrp
 
     wait_for_block(web3, current_block + 3)
 
-    with gevent.Timeout(5):
-        while len(seen_blocks) < 2:
-            gevent.sleep(random.random())
+    while len(seen_blocks) < 2:
+        pass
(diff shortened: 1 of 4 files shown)


tests/filtering/test_filter_against_latest_blocks.py
```diff
@@ -16,9 +16,8 @@ def test_filter_against_latest_blocks(web3_empty, wait_for_block, skip_if_testrp
 
     wait_for_block(web3, current_block + 3)
 
-    with gevent.Timeout(5):
-        while len(seen_blocks) < 2:
-            gevent.sleep(random.random())
+    while len(seen_blocks) < 2:
+        pass
```
pipermerriam: I don't like that this will hang indefinitely if there is an underlying failure.  Can you re-add the timeout and just increase the wait time?

shravan-shandilya: Oops..I will revert it.

pipermerriam:
I pushed one commit to add a simple test for this.  Will merge when tests are done running (and hopefully green 🐢 )

PR #122, opened 2016-10-27
added is_same_address utility method

shravan-shandilya:
### What was wrong?
Added a utility method to check whether two addresses are same.Requirement from https://github.com/pipermerriam/populus/issues/191

### How was it fixed?
Added a utility method and a test

#### Cute Animal Picture

kiwi


pipermerriam:
Thank you.

PR #129, opened 2016-12-06
Piper/fix contract methods with unamed arguments

### LINKED ISSUE #127: Bug in calling methods automatically generated by solidity for public contract attributes
* Version: 656903b34473a8cb658996a3dd8877ad9c734a12 (current master)
* Python: 2.7/3.4/3.5
* OS: osx/linux/win

### What was wrong?
For solidity code:
```
contract A {
    mapping (address => bytes32[]) public somename;
}
```
solc will generate function with ABI:
```
{
            "constant": True,
            "inputs": [
                {
                    "name": "",
                    "type": "address"
                },
                {
                    "name": "",
                    "type": "uint256"
                }
            ],
            "name": "somename",
            "outputs": [
                {
                    "name": "",
                    "type": "bytes32",
                    "value": "0x"
                }
            ],
            "payable": False,
            "type": "function"
        },
```
Notice that both `name` keys of inputs are empty.

The code in https://github.com/pipermerriam/web3.py/blob/master/web3/utils/abi.py#L162
Will squash two arguments pased to Contract.call().somename(addr, index) into one argument because it uses name as unique identifier. 

(Note, above examples are edited from my code so I haven't actually compile this exact code to reproduce the issue)

#### Cute Animal Picture

Cute animal picture

pipermerriam:
Fixes #127 

### What was wrong?

When Solidity generates accessor functions their `'name'` values are set to be empty string `''`.  The handling of these was causing positional arguments to be collapsed into a single argument.

### How was it fixed?

Simple early return when no keyword arguments are passed in

#### Cute Animal Picture

ottermom

Cute animal picture

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,62 +9,9 @@ before_install:
   - travis_retry sudo apt-get update
 env:
   matrix:
-    # admin
-    - TOX_ENV=py27-admin
-    - TOX_ENV=py34-admin
-    - TOX_ENV=py35-admin
-    # eth
-    - TOX_ENV=py27-eth
-    - TOX_ENV=py34-eth
-    - TOX_ENV=py35-eth
-    # mining
-    - TOX_ENV=py27-mining
-    - TOX_ENV=py34-mining
-    - TOX_ENV=py35-mining
-    # providers
-    - TOX_ENV=py27-providers
-    - TOX_ENV=py34-providers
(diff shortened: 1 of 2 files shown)


pipermerriam:
Breakage is due to changes to the geth RPC API that now need to be accounted for.  Blarg.

Issue #130, opened 2016-12-07
string and byte conversion in events.py corrupts data


brenzi:
* Version: 3.2.0
* Python: 3.5
* OS: linux

### What was wrong?
My strings passed in contract events get corrupted. They look like this when I listen to events:
`'_ipfsaddr': '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00`'`

### solution

```
--- events.py   2016-12-07 14:21:27.523182611 +0100
+++ events-new.py       2016-12-07 14:18:24.779178254 +0100
@@ -116,8 +116,8 @@
     entries using the correct types.
     """
     return [
-        'bytes32' if arg_type in {'bytes', 'string'} else arg_type
-        for arg_type in input_types
+        #'bytes32' if arg_type in {'bytes', 'string'} else arg_type
+        arg_type for arg_type in input_types
     ]
```
now I get my ipfs address correctly:
`'_ipfsaddr': 'QmSL2Zj3QmQFNcF6pethSYDF8fGYpz1LDeFiaLSzkCtAQ2'`

4gn3s:
I am experiencing the same behaviour with web3.py 3.2. The minimal test case:
```solidity
contract Test {
    function createEvent(string input) {
        MyEvent(input);
    }
    event MyEvent(string str);
}
```

and the test in populus:
```python
def test_event(chain):
    test_contract = chain.get_contract('Test')
    test_string = "i love food"
    event_filter = test_contract.pastEvents('MyEvent', {'toBlock': 'latest'})
    txn_hash = test_contract.transact().createEvent(test_string)
    chain.wait.for_receipt(txn_hash, timeout=120)
    event_logs = event_filter.get()[0]["args"]
    assert event_logs["str"] == test_string
```

which returns:
```python
>       assert event_logs["str"] == test_string
E       assert u'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@'' == 'i love food'
```

pipermerriam:
[...]

PR #142, opened 2016-12-30
Piper/fix dynamic event type decoding

pipermerriam:
### What was wrong?

Event arguments with non-indexed dynamic types are not being decoded correctly.  Current code incorrectly assumes they will be a 32 bytes sha3 hash of the value.

### How was it fixed?

Change logic to decode them as their real type

#### Cute Animal Picture

unlikely-animal-friendships-22


pipermerriam:
Fixes #130

PR #149, opened 2017-01-12
Human friendly ABI input parameter error

miohtama:
### What was wrong?

It was hard to figure out why the contract call input parameter encoding failed, esp. with all type coercion going on.

Now you get:

    web3.exceptions.ArgumentEncodeError: Cannot encode 1. argument of the function, expected type: b'string', inbound value None, full signature is {'inputs': [{'name': b'_value', 'type': b'string'}], 'type': b'constructor'

### How was it fixed?

Added a new TypeError exception subclass that comes with a human friendly error message.

For backwards compatibility, I kept old `check_if_arguments_can_be_encoded` function around, as there might be subtle differences in the behavior. However this function can be marked to be deprecated in the future release.

#### Cute Animal Picture

```
   /\     /\
  {  `---'  }
  {  O   O  }
  ~~>  V  <~~
   \  \|/  /
    `-----'__
    /     \  `^\_
   {       }\ |\_\_   W
   |  \_/  |/ /  \_\_( )
    \__/  /(_E     \__/
      (  /
       MM
```

### DIFF
--- a/tests/contracts/test_bad_abi.py
+++ b/tests/contracts/test_bad_abi.py
@@ -1,0 +1,13 @@
+import pytest
+
+from web3.exceptions import ArgumentEncodeError
+
+
+def test_bad_constructor_input_encoding(StringContract):
+    """Deploying a contract with bad ABI signature should give a friendly exception."""
+
+    with pytest.raises(ArgumentEncodeError):
+        # web3.exceptions.ArgumentEncodeError: Cannot encode 1. argument of
+        StringContract.deploy(args=[None])
+
+
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -46,7 +46,6 @@ from web3.utils.abi import (
     function_abi_to_4byte_selector,
     merge_args_and_kwargs,
     normalize_return_type,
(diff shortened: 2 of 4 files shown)


pipermerriam:
This looks like a really nice usability improvement.  Can you explain your choice to write this as a new function rather than modifying the existing function?

miohtama:
It is subtly different regarding exceptions as the old one. So I was thinking maybe have a deprecation path and avoid breaking the library users was nice. But I can make it this one function if you think we can do compromises here with the backwards compatibility.

miohtama:
Actually it doesn't make sense as this is called from Contract internals. Let me change this to the one function?

pipermerriam:
It isn't always the case, but generally I consider anything in `utils/...` to be non-public API an subject to change without notice.

carver:
@miohtama Are you still interested in doing this improvement? It looks like the next step was merging to a single function. Some things have changed in the last eight months :) so you would need to rebase against master.

carver:
Some other merged PRs have improved this situation. Closing, but feel free to reopen or comment if the current state is still unsatisfactory.

PR #156, opened 2017-02-07
Add RTD theme

void4:
### What was wrong?

The docs need to look better.

### How was it fixed?

I added the blue RTD theme.


carver:
@pipermerriam I think this would look better, too. Objections?

pipermerriam:
go for it

PR #165, opened 2017-02-27
IPC via named pipe on Windows

### LINKED ISSUE #17: Windows IPC/Named pipe does not work
Right now web3.py use sockets on GNU/Linux to connect to the IPC. This wont work on Windows since named pipes on windows does not make use of sockets. From what google says you can use a simple file handle to write and read on the named pipe (\.\pipe\geth.ipc). I have tried using simple popen/open to read and write data, but no luck. 

I have also tried to use this library for python, http://docs.activestate.com/activepython/2.6/pywin32/win32file.html

Next step is to check how the JS-version of web3 is connecting to the named pipe, but i can imagine that it uses a more low level connection directly via ethereum code base.

chfast:
This PR implements IPC on Windows by named pipe. The design is not very beautiful, but works for basic cases we tested. Unfortunately, there are no functional test that check communicate with real Ethereum client.

Definitely, Windows support is better than it was.

Resolves #17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,6 @@ language: python
 python:
   - "3.5"
 dist: trusty
-sudo: required
-before_install:
-  - travis_retry sudo add-apt-repository -y ppa:ethereum/ethereum
-  - travis_retry sudo apt-get update
 env:
   matrix:
     - TOX_ENV=py27-stdlib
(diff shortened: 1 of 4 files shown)


chfast:
Works also after your changes @pipermerriam.

Issue #164, opened 2017-02-23
Implement support for "function" abi type


ericssonaxel:
* Version: 3.6.1
* Pyhon: 2.7
* OS: macOS

### What was wrong?
In (https://github.com/ethereum/solidity/pull/1122) a "function" abi type was added. This is not yet implemented in web3.py

Example:
In my_contract:
`
function truebool(address a) external constant returns (bool){ return true;}`

`
function get_function() external constant returns(
function(address) external constant returns(bool)
){ return (this.truebool); }`

In python:
`
func = my_contract.call().get_function() #returns function type
my_other_contract.transact({...}).other_function(func) #expects function type as arg`

Currently, func will have value `None`. It seems like the function type is similar to bytes24 per (http://solidity.readthedocs.io/en/develop/types.html) under function types. 

It is possible to trick web3.py to return the function by changing the output type to "bytes24" in the abi for the contract, but this doesn't seem to work for input types; parity gives a **Transaction execution error** when trying to call `other_function(func)`.

#### Cute Animal Picture
![cute-polar-bear](https://cloud.githubusercontent.com/assets/3268707/23246669/28e1c4b4-f94a-11e6-8e72-bbbe3b392634.jpg)

ericssonaxel:
Pull request created -> 
https://github.com/pipermerriam/web3.py/pull/169
https://github.com/pipermerriam/ethereum-abi-utils/pull/12

pipermerriam:
Merged via a different PR.  Web3 will need to bump its requirement to `ethereum-abi-utils>=0.4.1`.

carver:
Version bump in #322

PR #169, opened 2017-03-19
Added function type support 

ericssonaxel:
Issue #164

### What was wrong?
The function variable type (introduced here ethereum/solidity#1122) was not supported by web3.py. 

### How was it fixed?
Added a new "function" type into web3/utils/abi.py modelled after the bytes24 type (as specified in http://solidity.readthedocs.io/en/develop/types.html)

### DIFF
--- a/web3/utils/abi.py
+++ b/web3/utils/abi.py
@@ -112,6 +112,12 @@ def is_encodable(_type, value):
         if not is_address(value):
             return False
         return True
+    elif base == 'function':
+        if not is_string(value):
+            return False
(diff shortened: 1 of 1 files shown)


pipermerriam:
Can you please add a few tests for this.

https://github.com/pipermerriam/web3.py/blob/master/tests/utilities/test_abi_is_encodable.py

carver:
This seems like something we want to support, but I think it belongs in ethereum-abi-utils as an alias from `function` -> `bytes24`

PR #174, opened 2017-04-07
Fix merge_and_kwargs for kwargs and contract constructors.

miohtama:
### What was wrong?

Passing (invalid) kwargs to contract constructor was not handled properly.

### How was it fixed?

Constructor does not have `abi["name"]`. Default to empty string in this case.

#### Cute Animal Picture

```
              (__)                   (__)
              [##]                   (@o)
       /-------\/             /-------\/              /-------  (__)
      / |     ||             / |     ||              / |     || (oo)
     *  ||----||            *  ||----||             *  ||----|---\/
        ^^    ^^               ^^    ^^                ^^    ^
  This cow belonged      This cow lived with      This cow belonged to
   to Flash Gordon       the Little Rascals       the Headless Horseman
```

### DIFF
--- a/web3/utils/abi.py
+++ b/web3/utils/abi.py
@@ -143,6 +143,7 @@ def check_if_arguments_can_be_encoded(function_abi, args, kwargs):
     )
 
 
+
 @coerce_args_to_text
 def merge_args_and_kwargs(function_abi, args, kwargs):
     if len(args) + len(kwargs) != len(function_abi['inputs']):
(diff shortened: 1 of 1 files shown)


pipermerriam:
@miohtama can you provide a test case for the problem you're fixing?

pipermerriam:
So my understanding is that there are two valid cases that should be covered.

* All inputs have a `name` field.
* Some or all inputs don't have a name field.

In the case that they all have a name, the current logic is appropriate.

The current approach you have implemented is likely to cause some un-expected errors.  By defaulting to empty string we will end up dropping arguments.

```python
>>> '': i for i in range(10)}
{'': 9}
```

So to handle missing names fields I think the following should be done.

* Map positional args to their inputs.  
* If there are any remaining inputs which have not been covered by positional arguments that are missing a `name` field, throw an exception since it is impossible to map keyword arguments to these inputs.

The remaining logic in this function will likely need some massaging because currently it assumes it can turn everything into keyword arguments and then turn them back into positional arguments.  I think this is flawed now, so it needs to be changed to just immediately turn everything into positional arguments.

carver:
What problem is this trying to solve? A code example would help.

pipermerriam:
Closing under the assumption this is no longer a thing.  Will re-open if it is.

PR #178, opened 2017-05-01
add eth_verify to eth

cle1000:
### What was missing?

web3.eth.verifiy

### How was it fixed?

i added a method which checks if a signature, generated with web3.eth.sign is valid


web3/eth.py
```diff
@@ -357,3 +373,26 @@ def contract(self,
 
     def getCompilers(self):
         return self.web3._requestManager.request_blocking("eth_getCompilers", [])
+
+    def _extract_ecdsa_signer(self, msg_hash, signature):
+        msg_hash_bytes = decode_hex(msg_hash) if msg_hash.startswith(b'0x') else msg_hash
```
pipermerriam: I would prefer we just require that the message be passed in as a hex encoded value.

web3/eth.py
```diff
@@ -262,6 +270,14 @@ def sign(self, account, data):
             "eth_sign", [account, data_hash],
         )
 
+    def verify(self, account, data, signature):
+        eth_data_hash = self.getEthSignHash(data)
+        signer = self._extract_ecdsa_signer(force_bytes(eth_data_hash), force_bytes(signature))
+        return signer == account
+
+    def getEthSignHash(self, msg):
+        return self.web3.sha3('\x19Ethereum Signed Message:\n' + str(len(msg)) + msg, encoding="utf8")
```
pipermerriam: Can you remove the hashing from this and change it to be something like `prefixMessageForSigning`, and then do the hashing outside of this function?

carver:
Since this is more than a year old, and all the functionality seems to be covered by `w3.eth.account` now, I'm closing. If there's something that `w3.eth.account` doesn't do that you would like it to, please keep the conversation going, and we'll try to get it sorted out.

PR #182, opened 2017-05-28
Update contracts docs with correct function params

XertroV:
Contract.deploy() takes `args` not `arguments` as an argument.

### What was wrong?

Contract.deploy() takes `args` not `arguments` as an argument.

### How was it fixed?

Updated docs


pipermerriam:
@XertroV sorry for the long delay in getting this merged.  Thank you for your contribution.

PR #191, opened 2017-06-30
new docs and test cleanup for a successful fresh checkout

carver:
### What was wrong?

Overall, it was difficult to get a fresh checkout to pass all tests with `pytest`. Specifically:

1. There were no README instructions for setting up the dev environment
2. If pip installs `ethereum` v2+, a bunch of tests break
3. Two tests fail because the `LogTopics` fixture is not defined (but the variable also apparently isn't used)
4. Two tests fail because the `web3_empty` fixture is not defined (variable is used, but test is skipped on all other usages throughout the repo) 

### How was it fixed?

1. Added some README instructions to help get started, focused on Debian-flavor, because that's what I can easily test
2. Modify requirements-dev.txt to force ethereum <v2
3. Remove unused variable
4. Skipped failing tests. This doesn't seem like the best solution, but I didn't get a chance to dig into the intention of the test to fix it. I would love some feedback for a better solution.

#### Cute Animal Picture

Cute animal picture


README.md
```diff
@@ -10,3 +10,19 @@ A python implementation of [web3.js](https://github.com/ethereum/web3.js)
...
+```
+sudo apt-get install libssl-dev
+# ^ This is for Debian-like systems. TODO: Add more platforms
+
+git clone git@github.com:pipermerriam/web3.py.git
+cd web3.py
+virtualenv venv
+. venv/bin/activate
+pip install -r requirements-dev.txt
+python setup.py install
```
pipermerriam: Two things:

1. Lately I've been recommending `pip install -e .` as I've had it work in cases where `python setup.py ..` fails.
2. To develop, this should be `python setup.py develop`.  If you use `install` then your local changes won't be represented in the runtime code as `install` actually creates a copy of all of the modules in your python environment.

carver: Sure thing. Just to be clear, the `pip install -e .` ask overrides the `python setup develop` one, right?

As I understand, the first makes the second unnecessary.

carver: Specifically, `pytest tests/` passes for me in a fresh virtualenv, with:

```
pip install -r requirements-dev.txt
pip install -e .
```

PR #201, opened 2017-07-12
Transaction equality, hashing, and attributes

carver:
Note that this is not a final draft ready for merging, it's meant as a reference implementation for discussion.

Some key discussion points:
1. Are my goals incompatible with other project goals?
2. Is there an alternative approach that satisfies my goals? 
3. If returning a new Tx object is acceptable, are there other important implementation issues to consider that I missed?

### What was wrong?

These were some key benefits I was looking for:
* equality test using the transaction hash value (For example, in my first several use cases, I wanted: `transaction_before_block_inclusion == transaction_after_block_inclusion`)
* can be used in a set, to prevent duplicate transactions with the same hash
* can access the transaction values as attributes (It just feels more natural to use `tx.hash` than `tx['hash']`)
* transaction object should be immutable, to eliminate a whole class of bugs at very low cost

In short, I want to use transactions as a first class object.

### How was it fixed?

I updated the output formatter to return an object of a new class that acts like a dictionary for a lot of backwards compatibility (except for equality tests, since changing that was an explicit goal).

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/transaction.py
+++ b/web3/transaction.py
@@ -1,0 +1,63 @@
+
+
+class Transaction:
+    IMMUTABILITY_WARNING = "Transactions are immutable, create a new one instead of modifying this"
+
+    def __init__(self, values):
+        '''
+        @param values a dictionary of transaction values
+        '''
+        self.__dict__ = values
+        try:
+            self._unique_txid()
+        except AttributeError as e:
+            assert False, "Transaction must be uniquely defined. Failed because: %s" % e
+
+    def _unique_txid(self):
+        'currently the hash works as a unique identifier, but someday maybe not'
+        return getattr(self, 'hash')
+
+    def __getitem__(self, key):
+        return getattr(self, key)
+
+    def __setitem__(self, key, val):
+        raise NotImplementedError(Transaction.IMMUTABILITY_WARNING)
+
+    def __delitem__(self, key):
+        raise NotImplementedError(Transaction.IMMUTABILITY_WARNING)
+
+    def __iter__(self):
+        return iter(self.__dict__)
+
+    def todict(self):
+        return dict(self.__dict__)
+
+    def items(self):
+        return self.__dict__.items()
+
+    def iteritems(self):
+        return self.__dict__.iteritems()
+
+    def __setattr__(self, key, val):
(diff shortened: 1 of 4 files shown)


web3/transaction.py
```diff
@@ -0,0 +1,66 @@
+
+
+class Transaction(object):
```
pipermerriam: A few things.

1. Lets move this object to `web3.utils.datastructures`.
2. Rename to `AttributeDict` and remove any language that is transaction specific.
3. Remove the immutability functionality (will explain elsewhere)
3. Inherit the class from `collections.Mapping`

web3/transaction.py
```diff
@@ -0,0 +1,66 @@
...
+        @param values a dictionary of transaction values
+        '''
+        self.__dict__ = values
+        try:
+            self._unique_txid()
+        except AttributeError as e:
+            assert False, "Transaction must be uniquely defined. Failed because: %s" % e
+        self._locked = True
+
+    def _unique_txid(self):
```
pipermerriam: Lets remove this until it's actually needed.

web3/transaction.py
```diff
@@ -0,0 +1,66 @@
...
+    def __setitem__(self, key, val):
+        raise NotImplementedError(Transaction.IMMUTABILITY_WARNING)
+
+    def __delitem__(self, key):
+        raise NotImplementedError(Transaction.IMMUTABILITY_WARNING)
+
+    def __iter__(self):
+        return iter(self.__dict__)
+
+    def todict(self):
```
pipermerriam: I'd like to remove this and simply rely on `dict(txn_obj)` for casting to a dictionary.

web3/transaction.py
```diff
@@ -0,0 +1,66 @@
...
+        return dict(self.__dict__)
+
+    def items(self):
+        return self.__dict__.items()
+
+    def iteritems(self):
```
pipermerriam: Is this still required?  I've been under the impression that for `2.7/3.4/3.5` that merely implementing `.items()` was sufficient.

carver: This was actually from me poking around trying to get dict(txn_obj) working. It looks like the ideal solution is inheriting from `collections.Mapping`, so I'm happy to remove this.

PR #202, opened 2017-07-14
Access fields using attributes, in transactions and blocks

carver:
First draft was at #201 - see there for previous commentary. Things changed enough that I wanted a fresh branch, instead of backing out the commits or using a force push.

Additional benefits:
 * Objects become hashable for use in sets
 * Immutability by default

Currently applied to these output objects:
 * Transactions
 * Transaction Receipts
 * Blocks
 * eth.syncing

Also, I added gitignore for vim swap files and the virtualenv folder name used in the setup instructions.

### What was wrong?

These were some key benefits I was looking for:
* can be used in a set, to prevent duplicate transactions with the same hash
* can access the transaction values as attributes (It just feels more natural to use `tx.hash` than `tx['hash']`)
* transaction object should be immutable, to eliminate a whole class of bugs at very low cost

In short, I want to use transactions as a first class object. At request, I extended this concept to several other objects.

### How was it fixed?

A new AttributeDict data structure is applied to the relevant output formatters. Note that `AttributeDict(None)` will raise an exception, so `@apply_if_not_null` must precede the AttributeDict wrapper.

There is also a MutableAttributeDict that is currently unused, and a base class to share implementation details.

### What was implemented contrary to feedback?

* I used the shorter name for the Immutable version, because
** it was a bit long/unwieldy
** it was the only one in current use in the project
** it mirrors the python collections naming of: Mapping / MutableMapping
* I didn't create a separate hashable variation
** A hashable, mutable class is disallowed by python spec
** I didn't see a need for a non-hashable, immutable class

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -28,12 +28,19 @@ class ReadableAttributeDict(Mapping):
     def __len__(self):
         return len(self.__dict__)
 
-    def __str__(self):
-        return repr(self)
-
     def __repr__(self):
-        dict_lines = ['\n\t' + repr(k) + ': ' + repr(self[k]) + ',' for k in sorted(self.__dict__)]
-        return self.__class__.__name__ + '({' + ''.join(dict_lines) + '\n})'
+        return self.__class__.__name__ + "(%r)" % self.__dict__
+
+    def _repr_pretty_(self, builder, cycle):
+        """
+        Custom pretty output for the IPython console
+        """
+        builder.text(self.__class__.__name__ + "(")
+        if cycle:
+            builder.text("<cycle>")
+        else:
+            builder.pretty(self.__dict__)
+        builder.text(")")
 
 
 class MutableAttributeDict(MutableMapping, ReadableAttributeDict):
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -59,16 +59,11 @@ class AttributeDict(ReadableAttributeDict, Hashable):
     def __delattr__(self, key):
         raise TypeError('This data is immutable -- create a copy instead of modifying')
 
-    def _sorteditems(self):
-        return tuple(sorted(self.items()))
-
     def __hash__(self):
-        return hash(self._sorteditems())
+        return hash(tuple(sorted(self.items())))
 
     def __eq__(self, other):
-        if hasattr(other, '__dict__'):
-            return self.__dict__ == other.__dict__
-        elif isinstance(other, Mapping):
-            return other == self.__dict__
+        if isinstance(other, Mapping):
+            return self.__dict__ == dict(other)
         else:
             return False
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -1,9 +1,9 @@
 
 from collections import (
-        Mapping,
-        MutableMapping,
-        Hashable,
-        )
+    Mapping,
+    MutableMapping,
+    Hashable,
+)
 
 # Hashable must be immutable:
 # "the implementation of hashable collections requires that a key's hash value is immutable"
@@ -11,7 +11,9 @@ from collections import (
 
 
 class ReadableAttributeDict(Mapping):
-    'the read attributes for the AttributeDict types'
+    """
+    The read attributes for the AttributeDict types
+    """
 
     def __init__(self, dictionary, *args, **kwargs):
         self.__dict__ = dict(dictionary)
@@ -44,7 +46,9 @@ class MutableAttributeDict(MutableMapping, ReadableAttributeDict):
 
 
 class AttributeDict(ReadableAttributeDict, Hashable):
-    'This provides superficial immutability, someone could hack around it'
+    """
+    This provides superficial immutability, someone could hack around it
+    """
 
     def __setattr__(self, attr, val):
         if attr == '__dict__':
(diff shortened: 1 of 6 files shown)


web3/utils/datastructures.py
```diff
@@ -0,0 +1,70 @@
+
+from collections import (
+        Mapping,
+        MutableMapping,
+        Hashable,
+        )
```
pipermerriam: Can you fix this indentation?

web3/utils/datastructures.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Hashable must be immutable:
+# "the implementation of hashable collections requires that a key's hash value is immutable"
+# https://docs.python.org/3/reference/datamodel.html#object.__hash__
+
+
+class ReadableAttributeDict(Mapping):
+    'the read attributes for the AttributeDict types'
```
pipermerriam: *nitpick*

Can you change this formatting to:

```
class Thing(object):
   """
   the doc string
   """
```

web3/utils/datastructures.py
```diff
@@ -0,0 +1,70 @@
...
+        return iter(self.__dict__)
+
+    def __len__(self):
+        return len(self.__dict__)
+
+    def __str__(self):
+        return repr(self)
+
+    def __repr__(self):
+        dict_lines = ['\n\t' + repr(k) + ': ' + repr(self[k]) + ',' for k in sorted(self.__dict__)]
```
pipermerriam: I'd ask that you remove the *fancy* linebreaking as to keep this consistent with the `repr` output from a normal dictionary.

web3/utils/datastructures.py
```diff
@@ -0,0 +1,70 @@
...
+    def __setattr__(self, attr, val):
+        if attr == '__dict__':
+            super(AttributeDict, self).__setattr__(attr, val)
+        else:
+            raise TypeError('This data is immutable -- create a copy instead of modifying')
+
+    def __delattr__(self, key):
+        raise TypeError('This data is immutable -- create a copy instead of modifying')
+
+    def _sorteditems(self):
```
pipermerriam: Since this API isn't use elsewhere and it's only used once within the `__hash__` method what do you think about removing it and just inlining the sort code into the `__hash__` method.

web3/utils/datastructures.py
```diff
@@ -0,0 +1,70 @@
...
+        return tuple(sorted(self.items()))
+
+    def __hash__(self):
+        return hash(self._sorteditems())
+
+    def __eq__(self, other):
+        if hasattr(other, '__dict__'):
+            return self.__dict__ == other.__dict__
```
pipermerriam: This approach seems like it could lead to unexpected objects being considered equal to each other.  What do you think about the following logic for equality.

```python
def __eq__(self, other):
    if isinstance(other, Mapping):
        return dict(other) == dict(self)
    else:
        return False
```

It's not immediately apparent to me if there are cases where either this would fail or not produce the desired result.

web3/utils/datastructures.py
carver: That was my first implementation until I saw it in the interpreter/docs, which will get uglier. Showing a block/transaction in REPL is a pretty common use case. How about leaving it in for prettier interpreter usage and documentation?

Old docs:
```
>>> web3.eth.getBlock('latest')
{
  'number': 4023147,
  'hash': "0xf3eb83d76f27eddf21ca9db87c297612c07fad00807a601e03990feb5b6c9ad4"
  ... etc ...
}
```

Current docs:
```
>>> web3.eth.getBlock('latest')
AttributeDict({
  'number': 4023147,
  'hash': "0xf3eb83d76f27eddf21ca9db87c297612c07fad00807a601e03990feb5b6c9ad4"
  ... etc ...
})
```

Proposed docs:

```
>>> web3.eth.getBlock('latest')
AttributeDict({'number': 4023147, 'hash': "0xf3eb83d76f27eddf21ca9db87c297612c07fad00807a601e03990feb5b6c9ad4", ... etc ... })
```

web3/utils/datastructures.py
carver: That seems reasonable. It rubs me the wrong way a little bit to make two new dictionary objects on every `==` test, but the cost is tiny in practice. I believe we can save at least one of the two allocations by using `self.__dict__`, which should be exactly equivalent to `dict(self)` for equality testing.

web3/utils/datastructures.py
pipermerriam: I'm 👍 on pretty docs examples but those can be massaged by hand or generated using the `pprint` module.  I'm even inclined to not have it be wrapped in `AttributeDict(....)` and to just pass it off a normal dictionary when repr'd in the console for simplicity sake.  I'm -0 on this which should be interpeted as "I don't agree but not strongly enough that we need to reach consensus.".

Your call 😛

pipermerriam:
This looks good to me.  There's the one thing about repr formatting that's your call.  Let me know when you're good with this being merged and I'll get it in and released.

carver:
Leaving it as-is wasn't sitting quite right with me, either. My favorite solution now is to have a classic repr, plus a custom ipython output.

I looked at hooking into pprint with a custom dispatch, but it was getting ugly (and wasn't working for ipython).

I don't love tying the solution to one package, but ipython is the de-facto standard, and the implementation only affects aesthetics. We should steer people away from the under-featured default python REPL, anyway.

I'm inclined to say this is good enough, and if someone finds a better solution, we can fix forward.

PR #206, opened 2017-07-19
[Issue 203] Address Checksum validation

djrtwo:
### What was wrong?
Issue [203](https://github.com/pipermerriam/web3.py/issues/203)

### How was it fixed?
Created utils/validations.py and moved some contract validations there.
Added `validate_address_checksum` function to validations.py
Inserted `validate_address_checksum` into methods that have address as param

### Questions
`is_encodable` in abi.py -- should this check the checksum on address input?
`construct_event_filter_params` in utils/filters.py -- should this check checksum on address(es) and contract address? Seems like it could be loose because it is just a filter

#### Cute Animal Picture
Cute animal picture


web3/contract.py
```diff
@@ -62,6 +61,10 @@
     construct_event_filter_params,
     PastLogFilter,
 )
+from web3.utils.validations import (
```
pipermerriam: Can you rename the `validations` model to `validation`.  Maintaining consistency with the naming conventions from some of my other repositories.

djrtwo: will do. writing some unit tests for the validation functions too

djrtwo:
@pipermerriam updated for review

PR #207, opened 2017-07-20
[issue 161] Better message when missing contract on function call

djrtwo:
### What was wrong?
Issue [161](https://github.com/pipermerriam/web3.py/issues/161)

### How was it fixed?
Added checks for type of error and added specific message when missing contract.
Added tests for both types of errors:
- Wrong function call on contract
- Missing contract at address entirely

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -77,6 +77,8 @@ DEPRECATED_SIGNATURE_MESSAGE = (
     "'Contract.factory(...)' class methog."
 )
 
+ACCEPTABLE_EMPTY_STRINGS = ["0x", b"0x", ""]
+
 
 class Contract(object):
     """Base class for Contract proxy classes.
@@ -778,9 +780,11 @@ def call_contract_function(contract,
     except DecodingError as e:
         # Provide a more helpful error message than the one provided by
         # eth-abi-utils
-        empty_strings = ["0x", b"0x", ""]
-        if (return_data in empty_strings
-                and contract.web3.eth.getCode(contract.address) in empty_strings):
+        is_missing_code_error = (
+            return_data in ACCEPTABLE_EMPTY_STRINGS and
+            contract.web3.eth.getCode(contract.address) in ACCEPTABLE_EMPTY_STRINGS
+        )
+        if is_missing_code_error:
             msg = (
                 "Could not transact with/call contract function, is contract "
                 "deployed correctly and chain synced?"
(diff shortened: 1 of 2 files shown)


web3/contract.py
```diff
@@ -778,14 +778,22 @@ def call_contract_function(contract,
...
         # eth-abi-utils
-        msg = (
-            "Could not decode contract function call {} return data {} for "
-            "output_types {}".format(
-                function_name,
-                return_data,
-                output_types
+        empty_strings = ["0x", b"0x", ""]
+        if (return_data in empty_strings
+                and contract.web3.eth.getCode(contract.address) in empty_strings):
```
pipermerriam: *nitpick*

I have a somewhat gigantic aversion to multi-line `if` statements.  Can you reformat this to something like the following or an alternative that doesn't split the `if` statement lines?

```python
is_missing_code_error = (
    return_data in empty_strings and
    contract.web3.eth.getCode(contract_address) in empty_strings
)
if is_missing_code_error:
    ...
```

web3/contract.py
```diff
@@ -778,14 +778,22 @@ def call_contract_function(contract,
     except DecodingError as e:
         # Provide a more helpful error message than the one provided by
         # eth-abi-utils
-        msg = (
-            "Could not decode contract function call {} return data {} for "
-            "output_types {}".format(
-                function_name,
-                return_data,
-                output_types
+        empty_strings = ["0x", b"0x", ""]
```
pipermerriam: Can you move this to the top level of the module so it doesn't have to be re-allocated everytime.  Also, rename to cap-cased and something bytecode specific like `EMPTY_BYTECODE_VALUES`

web3/contract.py
djrtwo: not a problem. I was struggling with figuring out how python standards wanted me to format it. Good approach with the boolean assignment :)

web3/contract.py
djrtwo: 👍

djrtwo:
@pipermerriam submit changes. ready for review

kivo360:
I'm getting the same error. The main problems is that I see the abi of the contract, yet I can't access the method.

image

image


As you can see the function is inside of the contract, but it's not registering in web3. I see that it says the chain is not synced correctly, but I don't believe that is the main problem.

carver:
@kivo360 you will need your chain to be synced, otherwise the contract won't have been deployed in your node yet. Also, please open a new issue, otherwise your question is likely to get lost. Finally, you may find that ethereum.stackexchange.com has the best response time for general questions, until they have confirmed that it's a web3.py bug.

mamoona-aslam:
@kivo360  did u solve it??/

Issue #210, opened 2017-07-25
`web3.eth.syncing` throws `'bool' object is not iterable`


reverendus:
* Version: 3.10.0
* Python: 3.6
* OS: osx

### What was wrong?

Attempting to get the value of `web3.eth.syncing`:
```
web3 = Web3(HTTPProvider(endpoint_uri=f"http://localhost:8545"))
print(web3.eth.syncing)
```

causes the following error:
```
Traceback (most recent call last):
  File "/Users/abc/Projects/test.py", line 30, in <module>
    print(web3.eth.syncing)
  File "/Users/abc/.pyenv/versions/3.6.1/lib/python3.6/site-packages/web3/utils/functional.py", line 15, in inner
    return formatter(value)
  File "/Users/abc/.pyenv/versions/3.6.1/lib/python3.6/site-packages/eth_utils/functional.py", line 11, in <lambda>
    return lambda x: f(g(x))
  File "/Users/abc/.pyenv/versions/3.6.1/lib/python3.6/site-packages/web3/formatters.py", line 54, in inner
    return fn(value)
  File "/Users/abc/.pyenv/versions/3.6.1/lib/python3.6/site-packages/web3/formatters.py", line 78, in wrapped
    return wrapper(fn(*args, **kwargs))
  File "/Users/abc/.pyenv/versions/3.6.1/lib/python3.6/site-packages/web3/utils/datastructures.py", line 19, in __init__
    self.__dict__ = dict(dictionary)
TypeError: 'bool' object is not iterable
```

Node: `Parity/v1.6.8-beta-c396229-20170608/x86_64-macos/rustc1.17.0`

Used to work in 3.8.0.

carver:
A fix was pushed in 3.11.0

@reverendus can you confirm?

reverendus:
Just checked, seems to be fine in 3.11.0.
Thanks!

PR #211, opened 2017-07-25
Don't wrap eth.syncing in AttributeDict when False

carver:
### What was wrong?

`web3.eth.syncing` was crashing when returning `False` --  see #210

### How was it fixed?

Only wrap eth.syncing with an AttributeDict when the result is a dictionary

#### Cute Animal Picture

Cute animal picture


pipermerriam:
Will get this released tomorrow.

Issue #215, opened 2017-08-02
web3.toHex(b'0x\xde\xad') gives '0xÞ\xad' instead of '0xdead'


carver:
* Version: 3.11.0
* Python: 3.5
* OS: linux
* Node: geth

### What was wrong?

`Web3.toHex()` doesn't convert bytes back to hex if they happen to start with '0x'

```
example = '0x3078DEAD'
binary = web3.toAscii(example)
b'0x\xde\xad'
reverted = web3.toHex(binary)
'0xÞ\xad'
# whoops
assert example == reverted, "I expect X == web3.toHex(web3.toAscii(X)), but it's not this time :("
```

### How can it be fixed?

If `bytes` are passed into web3.toHex, don't try to auto-detect that it's already hex, just convert it.

PR #216, opened 2017-08-02
Do not guess if a string is already hex encoded

carver:
Also, skip flake8 tests in several unnecessary folders. This significantly sped up my local `flake8` and `tox` run.

### What was wrong?

See issue #215 

### How was it fixed?

Stop inspecting the string for a leading '0x'

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/contracts/test_normalization_of_return_types.py
+++ b/tests/contracts/test_normalization_of_return_types.py
@@ -28,7 +28,9 @@ from web3.utils.abi import normalize_return_type
                 '0xbb9bc244d798123fde783fcc1c72d3bb8c189413',
             ],
         ),
-    )
+    ),
+    ids=['nullbyte', 'soloaddr', 'addrlist']
+
 )
 def test_normalizing_return_values(data_type, data_value, expected_value):
     actual_value = normalize_return_type(data_type, data_value)
--- a/tests/utilities/test_encoding.py
+++ b/tests/utilities/test_encoding.py
@@ -8,6 +8,7 @@ from hypothesis import (
 from web3.utils.encoding import (
     from_decimal,
     to_decimal,
+    to_hex,
 )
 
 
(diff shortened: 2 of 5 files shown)


carver:
So I rebuilt my tox and the error started reproducing locally. There are a few libraries that changed between [this](https://travis-ci.org/pipermerriam/web3.py/jobs/260289788) and a recent [successful](https://travis-ci.org/pipermerriam/web3.py/jobs/257412685):

* certifi
* hypothesis
* pytest 3.1.3 to 3.2.0

Since the primary error is coming from an internal pytest plugin, I'm guessing it's the pytest upgrade. Investigating further...

carver:
Confirmed that tox run is successful again after forcing pytest back to 3.1.3

carver:
Yup, [3.2.0 bug](https://github.com/pytest-dev/pytest/issues/2644). Will be fixed in 3.2.1

Pinning to 3.1.8 for now, then we can go to >=3.2.1 after it's released.

tests/contracts/test_normalization_of_return_types.py
```diff
@@ -28,7 +28,9 @@
                 '0xbb9bc244d798123fde783fcc1c72d3bb8c189413',
             ],
         ),
-    )
+    ),
+    ids=['nullbyte', 'soloaddr', 'addrlist']
```
pipermerriam: Assuming this is the fix for pytest 3.2, I think I'd rather just pin the dependency to `pytest<3.2.0` with a comment linking to the issue in the pytest repo.

pipermerriam: Assuming this is the fix for pytest 3.2, I think I'd rather just pin the dependency to `pytest<3.2.0` with a comment linking to the issue in the pytest repo.

carver: Yes, it fixes the pytest issue, but it also makes the pytest output somewhat more readable:

```
tests/contracts/test_normalization_of_return_types.py::test_normalizing_return_values[nullbyte] PASSED
tests/contracts/test_normalization_of_return_types.py::test_normalizing_return_values[soloaddr] PASSED
tests/contracts/test_normalization_of_return_types.py::test_normalizing_return_values[addrlist] PASSED
```

But I'm only mildly in favor of this approach. If you are more worried about consistency between tests than test output readability for this test, I'm happy to go the version pinning route.

pipermerriam: I'm convinced.  feel free to leave this in.

PR #229, opened 2017-08-13
Allow local signing of contract method transactions

jannikluhn:
### What was wrong?

It was not possible to send transactions calling contract methods if the RPC client doesn't have access to the private key (e.g., Infura).

### How was it fixed?

Add the option to pass a private key to `contract.transact`. If a key is given, the transaction is created locally, signed with the key and sent with `eth_sendRawTransaction`.

For this, I had to add pyethereum to the dependencies. If it is too heavy we could also make it optional.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -27,6 +27,15 @@ from eth_abi.exceptions import (
     DecodingError,
 )
 
+from ethereum.transactions import (
+    Transaction
+)
+from ethereum.utils import (
+    privtoaddr
+)
+
+import rlp
+
 from cytoolz.functoolz import (
     compose,
 )
@@ -53,6 +62,9 @@ from web3.utils.decorators import (
 )
 from web3.utils.empty import (
     empty,
+)
+from web3.utils.encoding import (
+    to_hex
 )
 from web3.utils.events import (
     get_event_data,
@@ -511,11 +523,12 @@ class Contract(object):
     def transact(self, transaction=None):
         """
         Execute a contract function call using the `eth_sendTransaction`
-        interface.
+        or `eth_sendRawTransaction` interface.
 
         You should specify the account that pays the gas for this transaction
-        in `transaction`. If no account is specified the coinbase account of
-        web3 interface is used.
+        in `transaction`. Alternatively, you can pass a private key to sign
+        the transaction locally. If neither is specified the coinbase account of
+        web3 interface is used
 
         Example:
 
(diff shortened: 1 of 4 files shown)


carver:
Very cool. Could you add a first pass on the docs, too?

jannikluhn:
Done. Now it also uses `estimateGas` if `gas` is not provided (learned about that in the docs, reading them really helps!)

web3/contract.py
```diff
@@ -861,6 +859,11 @@ def transact_with_contract_function(contract=None,
             sender = transact_transaction['from']
             nonce = contract.web3.eth.getTransactionCount(sender)
             transact_transaction['nonce'] = nonce
+        if 'gas' not in transact_transaction:
+            gas_estimation_transaction = {k: v for k, v in transact_transaction.items()
+                                          if k not in ('private_key', 'nonce')}
+            gas = contract.web3.eth.estimateGas(gas_estimation_transaction)
+            transact_transaction['gas'] = gas
```
carver: For consistency and generality, this estimate can go in `sendRawTransaction()` (similar to the estimate in `sendTransaction()`)

https://github.com/pipermerriam/web3.py/blob/1d551de901d60a08c8cb51170cda44809b67e6af/web3/eth.py#L253

pipermerriam:
@jannikluhn I am not comfortable adding this functionality to web3.  If you'll take a look at the middleware API being implemented in https://github.com/pipermerriam/web3.py/pull/224 this would be the right place to implement this functionality.  I expect to get that API merged here in the coming days and then you can create this as a 3rd party package which supplies a web3 middleware which provides this functionality.

I think this is a really useful feature, but I can't accept the necessary dependencies that this introduces into the main library.

pipermerriam:
@jannikluhn I'd like to point you at the new middleware API

http://web3py.readthedocs.io/en/latest/internals.html#middlewares

If you would like to re-implement your feature as a middleware I could accept it into the core library under the following constraints.

1. It cannot add any new hard dependencies.
2. It *must* support a pluggable ECC signing mechanism to allow users to bring-their-own ECC as to not require a new ECC dependency in the library.
3. For testing you can introduce `coincurve` as a dependency for ECC signing.
4. For transaction serialization, please create a local `Transaction` class using `rlp.Serializable` rather than including the one from the pyethreum library.

Sorry for all the constraints, but I have to balance the ease of people installing this tool with the usefullness of the feature.

Alternatively, if you'd like to *go-wild* and do this outside of the constraints, you can now easily just write a 3rd party middleware that does that which people can install separately.

jannikluhn:
Thank you for the suggestions. I've already started working on a [new package](https://github.com/jannikluhn/ethereum-accounts) (still WIP). It provides not only the signing middleware, but also keystore handling.

I'm generally open of integrating it (or parts of it) into web3. However, I guess keystores don't fit well into the scope of web3 (and introduce even more dependencies), but signing without access to private keys doesn't make a lot of sense either. Therefore I slightly lean towards keeping the two packages separate for now.

PR #231, opened 2017-08-14
Added changelog link to README

miohtama:
### What was wrong?

Changelog was "hard" to find.

### How was it fixed?

Added  a link to README.


carver:
The docs seems like a reasonable place for this, but the README seems a bit overkill. Since the changelog is just another file in the same place as the readme, it doesn't seem so hard to find.

pipermerriam:
I somewhat agree but it doesn't hurt.

PR #232, opened 2017-08-14
Added change log link to docs

miohtama:
### What was wrong?

Changelog was hard to find.

### How was it fixed?

Link it in the docs, too.


pipermerriam:
I'd like to also do what I've already done in populus and migrate this from a flat file to just only be present in `docs/releases.rst` so that it gets built with the docs.

PR #233, opened 2017-08-15
[Issue 218] Solidity sha3

djrtwo:
### What was wrong?
[Issue 218](https://github.com/pipermerriam/web3.py/issues/218)

### How was it fixed?

- Added helper functions and tests in web3/utils/abi.py for discerning general abi_type.
- Added `soliditySha3` method to Web3.
- Added test file tests/web3-module/test_web3_soliditySha3.py.
- Solidity sha3 values to test against come from deployed contract at https://kovan.etherscan.io/address/0x0d96c4b1a7a340cae48e8a889ad303a83f79cde3#readContract

#### Cute Animal Picture

Cute animal picture


djrtwo:
@pipermerriam I'm having trouble figuring out how to encode arrays. I assume I'm missing some sort of header bits about length. I'm currently just tightly packing the elements of the arrays/subarrays.

For example my code is encoding the following  (`[uint8[2]]`, `[[8, 9]]`)  as `0x0809` but this output does not match the output of `[8, 9]` to the solidity keccak256 function.

web3/main.py
```diff
@@ -180,5 +180,7 @@ def twos_compliment_of_negative(value, bit_size):
     return remove_0x_prefix(hex_value)
 
 
-def string_to_hex(string):
-    return codecs.getencoder('hex')(string)[0]
+def bytes_to_hex(byte_string):
+    if type(byte_string) == str:
+        byte_string = str.encode(byte_string)
+    return codecs.getencoder('hex')(byte_string)[0].decode("utf-8")
```
carver: Explicit encoding is better than implicit here, like `byte_string.encode('utf-8')` instead of `str.encode(byte_string)`

You could also leverage some of the tools from the utlis library here, like: `return remove_0x_prefix(to_hex(byte_string))`

web3/main.py
```diff
@@ -125,3 +147,48 @@ def createBatch(self):
...
+            return twos_compliment_of_negative(value, bit_size)
+        else:
+            return positive_int_to_hex(value, bit_size)
+    elif is_address_type(abi_type):
+        return remove_0x_prefix(value)
+    elif is_bytes_type(abi_type):
+        if is_0x_prefixed(value):
+            return remove_0x_prefix(value)
+        else:
+            return bytes_to_hex(value)
```
carver: I believe you'll run into the same issue as https://github.com/pipermerriam/web3.py/issues/215 . Try input `b'\x30\x78\xde\xad'`.

Perhaps you could assume that value of type `str` is hex-encoded, and value of type `bytes` is the raw bytes. Then you're hosed in Python 2.x, but maybe this could be a Python 3+ only feature (of course, I'm in favor of forcing the whole lib to py3).

web3/main.py
```diff
@@ -125,3 +147,48 @@ def createBatch(self):
...
+    value = (1 << bit_size) + value
+    hex_value = hex(value)
+    hex_value = hex_value[:-1] if hex_value[-1] == "L" else hex_value
+    return remove_0x_prefix(hex_value)
+
+
+def bytes_to_hex(byte_string):
+    if type(byte_string) == str:
+        byte_string = str.encode(byte_string)
+    return codecs.getencoder('hex')(byte_string)[0].decode("utf-8")
```
carver: Explicit encoding is better than implicit here, like `byte_string.encode('utf-8')` instead of `str.encode(byte_string)`

You could also leverage some of the tools from the utils library here, like: `return remove_0x_prefix(to_hex(byte_string))`

djrtwo:
@carver Made the suggested changes. Not really sure how I should handle the ambiguous "0x" in strings in 2.7 because of not being able to tell the difference between `bytes` and `str`. Should I put in an explicit check for python version and just do no "0x" stripping for Python 2.7?

Also, do you know how arrays are encoded for the solidity keccak256 function? I'm having trouble reverse engineering it and can't find any explicit documentation anywhere.

djrtwo:
@pipermerriam Ready for review. 

Only failing tests are those fixed in [PR 235](https://github.com/pipermerriam/web3.py/pull/235). I'll merge those fixes in once they're in master.

carver:
@djrtwo 

**On using '0x' to decide if the value is hex-encoded**

We definitely shouldn't use the first two letters to decide if it is hex-encoded. I could maybe be convinced to use `str` = hex and `bytes` = raw.

For python2 compatibility, you could import `bytes` into python2, with `from builtins import bytes` ([source](http://python-future.org/compatible_idioms.html#byte-string-literals))

Then you could follow this path
* if is string:
  * if 0x prefixed: return with 0x stripped
  * if no 0x prefix: raise ValueError  # force callers to be explicit: "bytes" or "hex-encoded str with 0x"
* elif is bytes:
  * return hex encoded

**On array encoding**

Looks like the only thing you need to add is to left-pad the int256 with zeros. So:

```
def encodeint(val):
    return val.to_bytes(32, byteorder='big')
web3.sha3(encodeint(8) + encodeint(9))
```

At least that matches my array test in remix...

djrtwo:
@carver Yeah, I figured out the array stuff. Every type is padded to 256bits regardless of type. A little counter intuitive based on documentation but got it.

Did a trace on etherscan...

djrtwo:
Will work look into bytes update in the morning. Thanks!

web3/utils/validation.py
```diff
@@ -18,6 +38,59 @@ def validate_abi(abi):
...
+    elif is_bool_type(abi_type):
+        correct = is_boolean(value)
+    elif is_uint_type(abi_type):
+        correct = is_integer(value) and value >= 0
+    elif is_int_type(abi_type):
+        correct = is_integer(value)
+    elif is_address_type(abi_type):
+        correct = is_address(value)
+    elif is_bytes_type(abi_type):
+        if sys.version_info[0] >= 3 and is_bytes(value):
```
djrtwo: @carver logic implements requiring either python3 bytes type or '0x' prefixed string.

carver:
Ok, looks like that addressed the issues I brought up. @pipermerriam is wrapping up a significant refactor to using middleware, and I'm not sure whether this will go in before or after that.

djrtwo:
Got, thanks. Coding around python2/3 incompatibilities is no fun...

web3/utils/encoding.py
```diff
@@ -1,29 +1,117 @@
...
+        return pad_hex(value, data_size)
+    elif is_bytes_type(abi_type):
+        if sys.version_info[0] >= 3:
+            if is_bytes(value):
+                return to_hex(value)
+            return value
+        else:
+            return value
+    elif is_string_type(abi_type):
+        return bytes_to_hex(value)
```
pipermerriam: I think I'd like to see one more `else` clause here that raises an exception to be sure that we don't every accidentally return `None` from this function.

web3/utils/encoding.py
```diff
@@ -1,29 +1,117 @@
...
+
+
+def bytes_to_hex(byte_string):
```
pipermerriam: I believe that `eth_utils.encode_hex` will do this for you.  Is there a reason why you can't use that?

web3/utils/encoding.py
```diff
@@ -1,29 +1,117 @@
...
+    """
+    return pad_hex(to_hex(value), bit_size)
+
+
+def pad_hex(value, bit_size):
+    """
+    Pads a hex string up to the given bit_size
+    """
+    value = remove_0x_prefix(value)
+    return "0x{hex_string}".format(hex_string=value.zfill(int(bit_size / 4)))
```
pipermerriam: Would you mind changing this to use the `eth_utils.add_0x_prefix` function for consistency.

web3/utils/validation.py
```diff
@@ -18,6 +38,59 @@ def validate_abi(abi):
...
+        raise ValueError("Unrecognized abi_type: {abi_type}".format(abi_type=abi_type))
+
+
+def validate_abi_value(abi_type, value):
+    """
+    Helper function for validating a value against the expected abi_type
+    Note: abi_type 'bytes' must either be python3 'bytes' object or '0x'
+    prefixed string
+    """
+    correct = True
```
pipermerriam: I'd like to see the mutability removed from this function and instead have `return` statements inside the `if/elif` blocks.  Then, if the function has not exited by the point where the `if/elif/else` block finishes you can raise the exception there or in the final `else` statement.

pipermerriam:
Not sure if it helps but there is an [`int_to_big_endian`](https://github.com/pipermerriam/ethereum-utils#int_to_big_endianvalue---bytes) function available in the `ethereum-utils` library.

web3/utils/encoding.py
```diff
@@ -1,29 +1,117 @@
...
+            return value
+    elif is_string_type(abi_type):
+        return bytes_to_hex(value)
+
+
+def bytes_to_hex(byte_string):
+    """
+    Converts byte_string to hex representation
+    """
+    if type(byte_string) == str:
```
pipermerriam: I think that the `eth_utils.is_text` function does the same thing.  If not, then can you convert this to `isinstance(byte_string, str)`.

djrtwo:
@pipermerriam ready for review

PR #235, opened 2017-08-15
[issue 234] fix issues related to address validation

djrtwo:
### What was wrong?
[issue 234](https://github.com/pipermerriam/web3.py/issues/234)

### How was it fixed?
- `validate_address` checks if the issue was checksum or not to discern as to whether it should raise a TypeError or ValueError
- eth.contract ensures that user gets ValueError if passes in an address with a bad checksum. This isn't super clean because of new functionality from `is_address`. It would be cleaner if we had something like `is_address_like` that didn't check the checksum, so we could check it ourselves and easily raise the relevant error.

#### Cute Animal Picture

Cute animal picture


web3/eth.py
```diff
@@ -338,15 +338,18 @@ def contract(self,
             len(args) >= 2 and is_address(args[1]),
         ))
 
+        for potential_address in args:
+            validate_address_checksum(potential_address)
+
         if has_address:
             if 'address' in kwargs:
                 address = kwargs.pop('address')
+                validate_address_checksum(address)
```
djrtwo: 👍

carver:
:shipit:

web3/utils/validation.py
```diff
@@ -23,8 +23,8 @@ def validate_address(address):
     Helper function for validating an address
     """
     if not is_address(address):
+        validate_address_checksum(address)
         raise TypeError("'address' is not an address")
```
pipermerriam: This line seems like it is missing some string formatting.  CAn you fix it?

web3/utils/validation.py
```diff
@@ -23,8 +23,8 @@ def validate_address(address):
     Helper function for validating an address
     """
     if not is_address(address):
+        validate_address_checksum(address)
```
pipermerriam: This seems weird.  What am I missing.  This is now inside of the block of code that should only be reachable if it is not an address..  Why would this be here?

djrtwo: `validate_address_checksum` to raises a ValueError while `validate_address` used to raise a TypeError. Because `is_address` now can return false due to wrong checksum, it was an attempt to catch whether that was because the checksum failed or otherwise to throw the ValueError vs TypeError.

When calling `validate_address`, if the address has a bad checksum, it will never get to `validate_address_checksum` now. It will simply throw the generic ValueError without any details about checksum.

I know the `validate_address_checksum` under the block of code looks strange, but I think it gives us the desired functionality.

carver: This got lost in the merge fixes. I opened #240 to patch it up.

PR #240, opened 2017-08-18
Raise error specific to address checksum failure

carver:
@djrtwo Can you confirm this is what you were intending with the address validation?

### What was wrong?

Addresses that are correctly formatted, but have an invalid checksum, would cause a generic "not an address" failure.

### How was it fixed?

Move the custom checksum check before the more general "is this an address" check.

#### Cute Animal Picture

Cute animal picture


djrtwo:
Yes, and that placement is way more logical.

Issue #243, opened 2017-08-23
The `web3.utils.datastructures.AttributeDict` should offer an API for recursive converstion to attribute dicts.


pipermerriam:
### What was wrong?

```python
>>> x = AttributeDict({'y': {'z': 3}})
>>> x.y
{'z': 3}
>>> x.y.z
AttributeError(...)
```

This behavior is correct because we don't want to *magically* convert everything because explicit is better than implicit, **but**, it would be good to have an API to allow recursive attribute dict creation.

### How can it be fixed?

New api `AttributeDict.recursive(value)` which recursively walks down into the value and converts things to attribute dicts.

This should probably be done in two pieces, a `classmethod` on the `AttributeDict` object and a stand-alone function that does the heavy lifting.

```python
import collections

from eth_utils import (
    is_dict,
    is_list_like,
)

def wrap_dicts_with_attrdict(value, _type):
    if isinstance(value, _type):
        return value
    elif is_dict(value):
        return _type({
            key: wrap_dicts_with_attrdict(item, _type)
            for key, item
            in value.items()
        })
    elif is_list_like(value) or isinstance(value, collections.Set):
        return type(value)(
            wrap_dicts_with_attrdict(item, _type) for item in value
        )
    else:
        return value
```

Then in the base class`ReadableAttributeDict` class, add a method that looks like this.

```python
class ReadableAttributeDict(...):
    @classmethod
    def recursive(cls, value):
        return wrap_dicts_with_attrdict(value, cls)
```

You will also need to write tests for this which which can be done using the simple example from above with a single nested dictionary as a starting point, and then expanding to be sure that values with nested lists, tuples, sets, strings, and integers are all converted as expected.

carver:
There shouldn't be any circular references in normal operation, but I'll add a test for that too

PR #244, opened 2017-08-23
New method to recursively create AttributeDict

carver:
Plus a few helpers to map a function to a collection and return the same
type.

Largely punts on cycle references, by returning None when discovering a
cycle.

Addresses Issue #243 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/utilities/test_attributedict.py
+++ b/tests/utilities/test_attributedict.py
@@ -74,3 +74,34 @@ def test_attributedict_inequality(dict1, dict2):
     assert AttributeDict(dict1) != dict2
     assert AttributeDict(dict1) != AttributeDict(dict2)
     assert dict1 != AttributeDict(dict2)
+
+
+def test_attributedict_recursive_dict():
+    w = AttributeDict.recursive({'x': {'y': {'z': 8}}})
+    assert w.x.y.z == 8
+
+
+@pytest.mark.parametrize('sequence', [list, tuple])
+def test_attributedict_sequence_with_dict(sequence):
+    data = sequence(['a', {'found': True}, 'c'])
+    dict_in_sequence = AttributeDict.recursive(data)
+    assert dict_in_sequence[1].found is True
+
+
+def test_attributedict_dict_in_list_in_dict():
+    data = {'instructions': [
+        0,
+        1,
+        'neither shalt thou count, excepting that thou then proceedeth to three',
+        {'if_naughty': 'snuff it'},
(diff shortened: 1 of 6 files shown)


tests/utilities/test_formatters.py
```diff
@@ -0,0 +1,38 @@
+
+import pytest
+
+from web3.utils.formatters import map_collection, recursive_map
```
pipermerriam: nitpick: split imports to their own lines

```python
from x import (
    y,
    z,
)
```

web3/utils/decorators.py
```diff
@@ -13,3 +13,20 @@ def _wrapper(*args, **kwargs):
             else:
                 return self.method(objtype, *args, **kwargs)
         return _wrapper
+
+
+def reject_recursive_repeats(to_wrap):
+    '''
+    Prevent simple cycles by returning None when called recursively with same instance
```
pipermerriam: fancy, I was planning on just ignoring this case.  Glad you covered it.

web3/utils/formatters.py
```diff
@@ -52,3 +61,29 @@ def apply_formatters_to_dict(formatters, value):
...
+
+def map_collection(func, collection):
+    '''
+    Apply func to each element of a collection, or value of a dictionary.
+    If the value is not a collection, return it unmodified
+    '''
+    datatype = type(collection)
+    if isinstance(collection, Mapping):
+        return datatype((key, func(val)) for key, val in collection.items())
+    elif isinstance(collection, Iterable):
```
pipermerriam: I suspect this might misbehave as I believe this will return true for strings.

tests/utilities/test_attributedict.py
```diff
@@ -74,3 +74,8 @@ def test_attributedict_inequality(dict1, dict2):
     assert AttributeDict(dict1) != dict2
     assert AttributeDict(dict1) != AttributeDict(dict2)
     assert dict1 != AttributeDict(dict2)
+
+
+def test_attributedict_recursive():
```
pipermerriam: I think this warrants an expanded test suite now that I've seen the implementation.  Try to cover the following.

* nested dicts
* lists and tuples with dicts
* dicts with lists with dicts
* stuff with integers, bools, strings
* something with a set.

carver:
Raising an exception on the cycle was a good tweak. It exposed a bug in py2 with unicode input. Ready for review/merge.

PR #245, opened 2017-08-24
Piper/middleware api tweaks

pipermerriam:
### What was wrong?

The middleware API needed some tweaks.

### How was it fixed?

* Removed the `request_id` from the API as it was no longer needed.
* Middleware can now be stateful as they are only instantiated once.
* Added `AttributeDict` wrapping middleware.

#### Cute Animal Picture

baby lion wallpaper


pipermerriam:
TODO:

* test middleware and providers setting on manager correctly updates wrappers.
* test middleware can be stateful

PR #250, opened 2017-08-24
[issue 248] add validations for abi_type array lengths

djrtwo:
### What was wrong?
[issue 248](https://github.com/pipermerriam/web3.py/issues/248)

### How was it fixed?
- Added support for validating array lengths like 'bool[4][2]' to `validate_abi_value`.
- Added relevant tests

#### Cute Animal Picture

Cute animal picture


djrtwo:
@pipermerriam ready for review

web3/utils/validation.py
```diff
@@ -51,6 +52,16 @@ def validate_abi_value(abi_type, value):
     Note: abi_type 'bytes' must either be python3 'bytes' object or ''
     """
     if is_array_type(abi_type) and is_list_like(value):
+        # validate length
+        specified_length = length_of_array_type(abi_type)
+        if specified_length and (specified_length != len(value)):
```
pipermerriam: Should this be `if specified_length is not None and ...` to account for the edge case of zero length arrays?  or are those even a thing?

djrtwo: not sure if you can specify 0 (empty array) as length. either way we should probably handle it. I'll investigate.

djrtwo: Appears you cannot define a fixed-size array in solidity.
```
browser/keccak_test.sol:20:31: TypeError: Unable to deduce common type for array elements.
    bool[0] public coolBool = [];
```

djrtwo:
@pipermerriam updated to handle cases like 'bool[0]'

pipermerriam:
Thank you for writing this.  I'm going to get it merged, but I also think that in the short/medium term I may extract it into the `ethereum-abi-utils` library.  It's a lot of code and I suspect it won't need to change often and I'd love to move it out of web3 and deeper into the lower level libraries.

djrtwo:
@pipermerriam I'm looking into extracting this into `ethereum-abi-utils` but am not sure how to approach it. I'm happy to do it with a bit of guidance.

PR #252, opened 2017-08-24
fix call to `blockNumber` on quickstart docs

djrtwo:
### What was wrong?
Docs said to run `web3.blockNumber` to demonstrate that web3 was connected to node.
This fails and might turn newbies off to using the library.
```
>>> web3.blockNumber
AttributeError: 'Web3' object has no attribute 'blockNumber'
```

### How was it fixed?
change call to `web3.eth.blockNumber`

#### Cute Animal Picture

Cute animal picture


djrtwo:
@pipermerriam minor change to docs. ready for review

PR #253, opened 2017-08-24
Test suite improvements

pipermerriam:
### What was wrong?

The current test suite for web3 is severely lacking.  It only really tests against eth-testrpc which is not a very robust library.

### How was it fixed?

Some of this will be improved when the `ethereum-tester` library is ready (which it almost is).

However, if we really want to *know* that everything works as expected, we need to refactor the entire test suite in the following way.

* Core test suite which tests the *internals* such as that the `Filter` objects behave as expected, that the utility functions work as expected.
* Integration tests against each of the supported backends which run a generic suite of smoke tests which ensure that web3 doesn't blow up.

#### Cute Animal Picture

screen-shot-2013-05-11-at-3 55 19-pm


web3/manager.py
```diff
@@ -48,7 +48,10 @@ def middlewares(self, value):
     def _generate_request_functions(self):
         self._wrapped_provider_request_functions = {
             index: combine_middlewares(
-                middlewares=self._middlewares,
+                middlewares=tuple(itertools.chain(
+                    self._middlewares,
+                    provider.get_middlewares(),
+                )),
```
carver: perhaps just `middlewares=self.middlewares + provider.get_middlewares(),` ?

pipermerriam: A fair simplification.  I get a little carried away with my *functional* approach at times and need these kinds of reminders.

pipermerriam:
@carver If you wanted to extract any of the following changes into independent pull requests, go for it.

* Any of the changes to `web3.middleware.pythonic`
* The introduction of provider level middlewares.
* The new utility in `web3.eth.blocks.select_method_for_block_identifier` and the related use of it in `web3.eth`
* The move of `web3.utils.module` to `web3.module`

pipermerriam:
This thing has become a monster.  I think it's *done* enough to be merged but not before it's cut into smaller pieces as laid out in my other comment.

pipermerriam:
Ok, I split this up.  Just need to merge a *bunch* of other small stuff first.

pipermerriam:
Ok, this has been rebased with all the extra stuff merged upstream.

PR #255, opened 2017-08-24
New Contract Factory: ConciseContract

carver:
### What was wrong?

I want a contract with a shorter/cleaner style for reading out data.

### How was it fixed?

The pull adds a new Contract Factory: `ConciseContract`. The most notable feature is being able to call  `contract.owner()` instead of `contract.call().owner()`

It uses the standard factory mechanism:

```
from web3.contracts import ConciseContract
web3.eth.contract(..., ContractFactoryClass=ConciseContract)
```

All short contract calls will be issued as read-only, unless it is modified first. You can switch back to a transaction like so:

```
contract.withdraw(amount, transact={'from': eth.accounts[1], 'gas': 100000, ...})
```

Which is equivalent to this classic approach:

```
contract.transact({'from': eth.accounts[1], 'gas': 100000, ...}).withdraw(amount)
```

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/contracts/test_concise_contract.py
+++ b/tests/contracts/test_concise_contract.py
@@ -1,0 +1,48 @@
+
+import sys
+
+import pytest
+
+from web3.contract import (
+    ConciseContract,
+    ConciseMethod,
+)
+
+if sys.version_info >= (3, 3):
+    from unittest.mock import Mock
+
+
+@pytest.mark.skipif(sys.version_info < (3, 3), reason="needs Mock library from 3.3")
+def test_concisecontract_call_default():
+    contract = Mock()
+    sweet_method = ConciseMethod(contract, 'grail')
+    sweet_method(1, 2)
+    contract.call.assert_called_once_with({})
+    contract.call().grail.assert_called_once_with(1, 2)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 3), reason="needs Mock library from 3.3")
+def test_concisecontract_custom_transact():
+    contract = Mock()
+    sweet_method = ConciseMethod(contract, 'grail')
+    sweet_method(1, 2, transact={'holy': 3})
+    contract.transact.assert_called_once_with({'holy': 3})
+    contract.transact().grail.assert_called_once_with(1, 2)
+
+
+def test_class_construction_sets_class_vars(web3,
+                                            MATH_ABI,
+                                            MATH_CODE,
+                                            MATH_RUNTIME):
+    MathContract = web3.eth.contract(
+        abi=MATH_ABI,
+        bytecode=MATH_CODE,
+        bytecode_runtime=MATH_RUNTIME,
+        ContractFactoryClass=ConciseContract,
+    )
+
+    math = MathContract()
+    classic = math._classic_contract
+    assert classic.web3 == web3
+    assert classic.bytecode == MATH_CODE
+    assert classic.bytecode_runtime == MATH_RUNTIME
--- a/tests/contracts/test_concise_contract.py
+++ b/tests/contracts/test_concise_contract.py
@@ -30,6 +30,22 @@ def test_concisecontract_custom_transact():
     contract.transact().grail.assert_called_once_with(1, 2)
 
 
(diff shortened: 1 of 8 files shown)


tests/contracts/test_concise_contract.py
```diff
@@ -0,0 +1,48 @@
...
+import sys
+
+import pytest
+
+from web3.contract import (
+    ConciseContract,
+    ConciseMethod,
+)
+
+if sys.version_info >= (3, 3):
```
carver: When do we get to stop supporting py2? I'm so ready.

carver:
If this approach looks good, I'll add a small section in the docs

tests/contracts/test_concise_contract.py
pipermerriam: I'm going to formally announce it at devcon after which **all-the-things** will get warnings that support is being dropped.

I suppose we could drop some `PendingDeprecationWarning` warnings in sooner.  I'm game.

web3/contract.py
```diff
@@ -753,6 +753,53 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+        self.__function = function
+
+    def __call__(self, *args, **kwargs):
+        return self.__prepared_function(**kwargs)(*args)
+
+    def __prepared_function(self, **kwargs):
+        if not kwargs:
+            modifier, modifier_dict = 'call', {}
+        elif len(kwargs) == 1:
+            modifier, modifier_dict = kwargs.popitem()
```
pipermerriam: It'd be nice if the API explicitly didn't allow `modifier` to be something weird if someone say, mispells `trasnact`.

Could this be changed to something like

```python
if 'transact' in kwargs:
    ...  # transact semantics
elif kwargs:
    ...  # explicit error that kwargs is being misused.
else:
    ...  # call semantics
```

web3/contract.py
```diff
@@ -753,6 +753,53 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+        else:
+            raise ValueError("Only use one keyword argument at a time, eg~ transact or call")
```
pipermerriam: I've not been consistent about this but I think `TypeError` is the right exception when the error is related to how something was called like bad keywords, not enough arguments, etc.

web3/contract.py
carver: Good idea. I'll limit modifier to one of `('call', 'estimateGas', 'transact')`

web3/contract.py
carver: Ah yes, that is a bug. `TypeError` is the right choice here.

carver:
@pipermerriam I may be using requirements-docs.txt wrong, but it seemed like `sphinx` should be in there. A couple more pieces had to be added for the build to succeed. I also added `py-geth` to requirements-dev. Does all of that look ok to you?

Also `open` fails in bash on linux (which causes the last line of `make docs` to file). I punted on fixing that because I'm unfamiliar with Makefiles and my first attempt at switching to `xdg-open` for linux failed.

pipermerriam:
I'm inclined to punt on the Makefile issue as well.  If you want to just add an extra `make linux-docs` or something as a stop-gap solution you've got my blessing.

One other thing.  Do you want to add a convenience API to the `Web3` object.  Something to allow a user to set the default class to use for the `ContractFactory` so that you don't always have to specify it in the `web3.eth.contract` API and you can just set it once?  If so, I'm 👍

web3/contract.py
```diff
@@ -753,6 +753,56 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+
+    def __init__(self, contract, function):
+        self.__contract = contract
+        self.__function = function
+
+    def __call__(self, *args, **kwargs):
+        return self.__prepared_function(**kwargs)(*args)
+
+    def __prepared_function(self, **kwargs):
+        allowed_modifiers = ('call', 'estimateGas', 'transact')
```
pipermerriam: nitpick:  

* slightly more *correct* if this was a `set` rather than a tuple
* slightly more *correct* if this was defined once, maybe at the top level of this module.

pipermerriam:
This is good to merge whenever unless you make significant enough changes that you want more feedback.

PR #265, opened 2017-08-26
pythonic formatter fixes

pipermerriam:
### What was wrong?

The `pythonic_middleware` was missing some conversions that were needed for compliance with go-ethereum

### How was it fixed?

Added them.

#### Cute Animal Picture

1735_cute-spicebush-bug


pipermerriam:
rolled into #267

PR #268, opened 2017-08-26
block-identifier util

pipermerriam:
### What was wrong?

There was repeated and generally not great code to determine which RPC method to use for some of the `web3.eth` APIs

### How was it fixed?

Abstracted it into a more robust generic helper function.

#### Cute Animal Picture

851159422_562cf6467a_z

### DIFF
--- a/web3/utils/blocks.py
+++ b/web3/utils/blocks.py
@@ -1,10 +1,44 @@
 from eth_utils import (
+    is_hex,
     is_string,
+    is_integer,
+    remove_0x_prefix,
     force_text,
 )
 
 
-def is_predefined_block_number(block_number):
-    if not is_string(block_number):
+def is_predefined_block_number(value):
+    if not is_string(value):
         return False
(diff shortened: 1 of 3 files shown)


pipermerriam:
This needs tests written against it.

web3/utils/blocks.py
```diff
@@ -1,10 +1,44 @@
...
+def is_hex_encoded_block_number(value):
+    if not is_string(value):
+        return False
+    elif is_hex_encoded_block_hash(value):
+        return False
+    try:
+        value_as_int = int(value, 16)
+    except ValueError:
+        return False
+    return 0 <= value_as_int < 2**256 - 1
```
carver: off by one, it looks like: `0 <= value_as_int < 2**256`

pipermerriam: 👍

pipermerriam:
Tests added.

PR #272, opened 2017-08-29
Bugfix for unknown blocks

pipermerriam:
### What was wrong?

Errors coming from when blocks are not found.

### How was it fixed?

It isn't yet... WIP

#### Cute Animal Picture

Cute animal picture


web3/providers/tester.py
```diff
@@ -48,8 +49,25 @@ def is_testrpc_available():
 )
 
 
+def static_return(value):
+    def inner(*args, **kwargs):
+        return {'result': value}
+    return inner
```
carver: This would probably be useful to other middlewares. Maybe it belongs in `web3/utils/formatters.py` or a new `web3/middleware/utils.py`?

carver:
:+1:

web3/providers/tester.py
pipermerriam: Done.  `web3.utils.formatters` for now.

Side note. I named that module and I'm growing to hate the terms formatter/formatting.  Wanting to move back towards the term `normalizer/normalizing` as it has more semantic meaning....

PR #280, opened 2017-09-01
New stalecheck middleware

carver:
### What was wrong?

An example:

It can be pretty bad news to resolve an ENS name from old blockchain data, so an ENS tool would likely prefer an exception over silent, stale success.

### How was it fixed?

Users can add this new middleware to raise an exception if the blockchain is too old.

Added a new stalecheck middleware. It's generated and added like so:

```
stalecheck_middleware = make_stalecheck_middleware(hours=36)
web3.add_middleware(stalecheck_middleware)
```

TODO for me: add a section in the docs for Middleware, focusing on examples for the optional middleware.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/middleware/test_stalecheck.py
+++ b/tests/core/middleware/test_stalecheck.py
@@ -1,0 +1,104 @@
+
+import pytest
+import sys
+
+from web3.middleware.stalecheck import (
+    _isfresh,
+    make_stalecheck_middleware,
+    StaleBlockchain,
+)
+
+if sys.version_info >= (3, 3):
+    from unittest.mock import Mock, patch
+
+
+pytestmark = pytest.mark.skipif(sys.version_info < (3, 3), reason="needs Mock library from 3.3")
+
+
+@pytest.fixture
+def now():
+    return 3141592653
+
+
+@pytest.fixture
+def days_fresh():
+    return 3
+
+
+@pytest.fixture
+def request_middleware(days_fresh):
+    middleware = make_stalecheck_middleware(days=days_fresh)
+    make_request, web3 = Mock(), Mock()
+    initialized = middleware(make_request, web3)
+    # for easier mocking, later:
+    initialized.web3 = web3
+    initialized.make_request = make_request
+    return initialized
+
+
+class BlockStub():
+    def __init__(self, _timestamp):
+        self.timestamp = _timestamp
+        self.number = 123
+
+
+def test_is_not_fresh_with_no_block():
+    assert not _isfresh(None, 1)
+
+
+def test_is_not_fresh(now):
+    with patch('time.time', return_value=now):
+        SECONDS_ALLOWED = 2 * 86400
+        stale = BlockStub(now - SECONDS_ALLOWED - 1)
+        assert not _isfresh(stale, SECONDS_ALLOWED)
+
+
+def test_is_fresh(now):
+    with patch('time.time', return_value=now):
+        SECONDS_ALLOWED = 2 * 86400
+        stale = BlockStub(now - SECONDS_ALLOWED)
+        assert _isfresh(stale, SECONDS_ALLOWED)
+
+
+def test_stalecheck_pass(request_middleware):
+    with patch('web3.middleware.stalecheck._isfresh', return_value=True):
+        method, params = Mock(), Mock()
+        request_middleware(method, params)
+        request_middleware.make_request.assert_called_once_with(method, params)
+
+
+def test_stalecheck_fail(request_middleware, now):
+    with patch('web3.middleware.stalecheck._isfresh', return_value=False):
+        method, params = Mock(), Mock()
+        request_middleware.web3.eth.getBlock.return_value = BlockStub(now)
+        with pytest.raises(StaleBlockchain):
+            request_middleware(method, params)
+
+
+@pytest.mark.parametrize(
+    'rpc_method',
+    [
+        'eth_getBlockByHash',
+        'eth_getBlockByNumber',
+        'eth_getBlockTransactionCountByHash',
+        'eth_getBlockTransactionCountByNumber',
+        'eth_getTransactionByBlockHashAndIndex',
+        'eth_getTransactionByBlockNumberAndIndex',
+        'eth_getUncleCountByBlockHash',
+        'eth_getUncleCountByBlockNumber',
+    ]
+)
+def test_stalecheck_ignores_get_by_block_methods(request_middleware, rpc_method):
+    # This is especially critical for getBlock('latest') which would cause infinite recursion
+    with patch('web3.middleware.stalecheck._isfresh', return_value=True):
+        params = Mock()
+        request_middleware(rpc_method, params)
+        assert not request_middleware.web3.eth.getBlock.called
+
+
+def test_stalecheck_calls_isfresh_with_block_and_time(request_middleware, days_fresh):
+    with patch('web3.middleware.stalecheck._isfresh', return_value=True) as freshspy:
+        method, params, block = Mock(), Mock(), Mock()
+        request_middleware.web3.eth.getBlock.return_value = block
+        request_middleware(method, params)
+        freshspy.assert_called_once_with(block, 86400 * days_fresh)
--- a/tests/core/middleware/test_stalecheck.py
+++ b/tests/core/middleware/test_stalecheck.py
@@ -2,9 +2,9 @@
 import pytest
 import sys
 
+from web3.middleware import make_stalecheck_middleware
 from web3.middleware.stalecheck import (
     _isfresh,
-    make_stalecheck_middleware,
     StaleBlockchain,
 )
 from web3.utils.datastructures import AttributeDict
--- a/tests/core/middleware/test_stalecheck.py
+++ b/tests/core/middleware/test_stalecheck.py
@@ -97,3 +97,21 @@ def test_stalecheck_calls_isfresh_with_empty_cache(request_middleware, allowable
         cache_call, live_call = freshspy.call_args_list
         assert cache_call[0] == (None, allowable_delay)
         assert live_call[0] == (block, allowable_delay)
+
+
+def test_stalecheck_adds_block_to_cache(request_middleware, allowable_delay):
+    with patch('web3.middleware.stalecheck._isfresh', side_effect=[False, True, True]) as freshspy:
+        block = object()
+        request_middleware.web3.eth.getBlock.return_value = block
+
+        # cache miss
+        request_middleware('', [])
+        cache_call, live_call = freshspy.call_args_list
+        assert freshspy.call_count == 2
+        assert cache_call == ((None, allowable_delay), )
+        assert live_call == ((block, allowable_delay), )
+
+        # cache hit
+        request_middleware('', [])
+        assert freshspy.call_count == 3
+        assert freshspy.call_args == ((block, allowable_delay), )
--- a/tests/core/middleware/test_stalecheck.py
+++ b/tests/core/middleware/test_stalecheck.py
@@ -64,45 +64,36 @@ def test_is_fresh(now):
 
 def test_stalecheck_pass(request_middleware):
     with patch('web3.middleware.stalecheck._isfresh', return_value=True):
-        method, params = Mock(), Mock()
+        method, params = object(), object()
         request_middleware(method, params)
         request_middleware.make_request.assert_called_once_with(method, params)
 
 
 def test_stalecheck_fail(request_middleware, now):
     with patch('web3.middleware.stalecheck._isfresh', return_value=False):
-        method, params = Mock(), Mock()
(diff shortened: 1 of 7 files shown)


web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+    'eth_getBlockTransactionCountByNumber',
+    'eth_getTransactionByBlockHashAndIndex',
+    'eth_getTransactionByBlockNumberAndIndex',
+    'eth_getUncleCountByBlockHash',
+    'eth_getUncleCountByBlockNumber',
+])
+
+
+def _isfresh(block, allowable_delay):
+    return block and time.time() - block.timestamp <= allowable_delay
```
pipermerriam: Can we change this to use `block['timestamp']`.  I think it might be a bad idea for us to rely on the `AttributeDict` wrapper in our tests.

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+                last_block = web3.eth.getBlock('latest')
+                if not _isfresh(last_block, allowable_delay):
+                    raise StaleBlockchain(last_block, allowable_delay)
+
+            return make_request(method, params)
+        return middleware
+    return stalecheck_middleware
+
+
+class StaleBlockchain(Exception):
```
pipermerriam: Currently all of the exceptions exposed by web3 live in `web3.exceptions`  I think I'd like to keep it that way.

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, block, allowable_delay):
+        last_block_date = datetime.datetime.fromtimestamp(block.timestamp).strftime('%c')
```
pipermerriam: Instead of doing this, what do you think about just storing the `block`.  Not very informative to only get the date back as I imagine many people will want to know the exact block number/hash/etc for debugging.

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
+import datetime
+import time
+
+
+SKIP_STALECHECK_FOR_METHODS = set([
+    'eth_getBlockByHash',
+    'eth_getBlockByNumber',
+    'eth_getBlockTransactionCountByHash',
+    'eth_getBlockTransactionCountByNumber',
+    'eth_getTransactionByBlockHashAndIndex',
```
pipermerriam: Why are the `getTransactionByBockXXXXAndIndex` methods included?

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+    middleware will raise a StaleBlockchain exception.
+
+    `make_stalecheck_middleware(...)` takes the same keyword arguments as datetime.timedelta()
+    '''
+    allowable_delay = datetime.timedelta(**kwargs).total_seconds()
+
+    def stalecheck_middleware(make_request, web3):
+        def middleware(method, params):
+            if method not in SKIP_STALECHECK_FOR_METHODS:
+                last_block = web3.eth.getBlock('latest')
```
pipermerriam: So I think we can cache this.

* we only allow `D` seconds of *staleness*.
* Let `B` be the last block time we saw.

We should only need to fetch the latest block *if*.

* This is our first call and `B` is not populated.
* `time.time() - B['timestamp'] > D`

The idea is that as long as the time difference between now and the last block we saw is under the allowed delta, then even if the chain froze after retrieving the last block, we are still safe.  Only once the last block has expired (the timestamp is now beyond the current allowable delta) should we need to fetch the new latest block.

Also, we can be clever and try using the result of the RPC call to re-populate the latest block anytime we see that `method == "eth_getBlockByNumber" and params == ["latest"]`.

web3/middleware/stalecheck.py
carver: I waffled a bit on this locally, so am very flexible.

I figured any call that specifically references the block number/hash seemed safe, since it will fail anyway if the block isn't found. This seemed like the cleanest way to define the methods to blacklist. The next-best alternative seemed to be just the "getBlock*" methods, but that seemed more like patching my bug than logically sound.

web3/middleware/stalecheck.py
carver: Sure, are you thinking for flexibility to move back away from `AttributeDict` in the future? Or because an `AttributeDict` failure would cause test failures in unrelated places?

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+    This middleware takes an argument, so unlike other middleware, you must make the middleware
+    with a method call.
+    For example: `make_stalecheck_middleware(hours=12)`
+
+    If the latest block in the chain is older than 12 hours ago in this example, then the
+    middleware will raise a StaleBlockchain exception.
+
+    `make_stalecheck_middleware(...)` takes the same keyword arguments as datetime.timedelta()
+    '''
+    allowable_delay = datetime.timedelta(**kwargs).total_seconds()
```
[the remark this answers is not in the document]
carver: On a second look, I do see that this is an ugly use of `kwargs`. For supporting a `datetime.datetime`, I think the added complexity outweighs the value. I'm having trouble imagining when that would be the best way to specify the allowable delay.

Here is an attempt to succinctly and clearly cover the first and third point (in spirit):

`def make_stalecheck_middleware(seconds=0, minutes=0, hours=0, days=0):`

What do you think?

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def stalecheck_middleware(make_request, web3):
+        def middleware(method, params):
+            if method not in SKIP_STALECHECK_FOR_METHODS:
```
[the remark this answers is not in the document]
carver: I'd like to keep it as simple as possible, until we know what use cases there are for customizing it.

If the main objection is that it's weird that there are calls besides getBlock that are included in the blacklist, then I'd rather just remove them.

web3/middleware/stalecheck.py
carver: Absolutely. I was thinking the same thing, but hadn't gotten around to it. I'll add it.

I might punt that last bit until the more general caching middleware. Trying to be clever usually gets me in trouble. :)

carver: Ugh, calling `web3.setProvider()` should invalidate the cache. I'm working on how to do that cleanly.

carver: I indexed into the cache with the providers found in web3. As long as people don't mutate the provider, we will be good. :crossed_fingers:

Outside of this PR, I'm tempted to say that we should remove `setProviders()` and force people to always start a new `Web3()`. Anything stateful is going to have to address this same problem.

web3/middleware/stalecheck.py
carver: I really like having a verbose error message for people at the shell, but I totally agree that including the block in the exception is a good idea. I'll do both.

web3/middleware/stalecheck.py
pipermerriam: So here's how I'm thinking about this.

The job of this middleware is explicitly cause failures when the chain goes out of sync.  In my little web3 world this is intended to allow me to differentiate between failure caused by a non-existent block hash and failure caused by being out of sync. 

But that raises the question of whether the stale checking middleware should only throw an error *if* the result is `None`...  Honestly not sure.  Your call.

web3/middleware/stalecheck.py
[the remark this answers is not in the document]
pipermerriam: I still think it's a cleaner API to allow any of the three because they serve the most common use cases.

* I would use seconds in cases where my allowable delay is small (an hour or less)
* I would use `datetime.timedelta` or seconds when my delay gets bigger.
* I would use `datetime.datetime` if I knew the exact moment where I needed to be able to read the data (this is the weakest use case).

If you don't want to do negotiation of the three types and you want a simple API, I'd recommend just leaving it as seconds only and making people do the conversion outside of this function.

web3/middleware/stalecheck.py
carver: Ah, interesting.

For ENS, at least, it's bad to get valid but old data (you might be resolving to an old address, for example). So it's best to raise an exception no matter what on old data.

It seems like the least contentious option then is to switch to only "getBlockBy*", and arguably, only if the parameter is `'latest'`.

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,44 @@
...
+SKIP_STALECHECK_FOR_METHODS = set([
+    'eth_getBlockByNumber',
+])
+
+
+def _isfresh(block, allowable_delay):
+    return block and time.time() - block['timestamp'] <= allowable_delay
+
+
+def make_stalecheck_middleware(allowable_delay=0):
```
pipermerriam: I think that this should not have a default value to ensure it's required that they specify a delay in their code.  (I get this is the current behavior but I think the validation should be separate).

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,44 @@
...
+    Use to require that a function will run only of the blockchain is recently updated.
+
+    This middleware takes an argument, so unlike other middleware, you must make the middleware
+    with a method call.
+    For example: `make_stalecheck_middleware(allowable_delay=60*5)`
+
+    If the latest block in the chain is older than 5 minutes in this example, then the
+    middleware will raise a StaleBlockchain exception.
+    '''
+    if not allowable_delay:
```
pipermerriam: This seems like the check should be `if delay <= 0:` since the delay should also be strictly positive as well as non-zero.

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,44 @@
...
+        raise ValueError("You must set a non-zero allowable delay for this middleware")
+
+    def stalecheck_middleware(make_request, web3):
+        cache = {'latest': None}
+
+        def middleware(method, params):
+            if method not in SKIP_STALECHECK_FOR_METHODS:
```
pipermerriam: Can you add a new keyword argument to the factory, something like `skip_stalecheck_for_methods=SKIP_STALECHECK_FOR_METHODS` so that users can pass in their own list for this?

web3/middleware/stalecheck.py
```diff
@@ -0,0 +1,44 @@
...
+    For example: `make_stalecheck_middleware(allowable_delay=60*5)`
+
+    If the latest block in the chain is older than 5 minutes in this example, then the
+    middleware will raise a StaleBlockchain exception.
+    '''
+    if not allowable_delay:
+        raise ValueError("You must set a non-zero allowable delay for this middleware")
+
+    def stalecheck_middleware(make_request, web3):
+        cache = {'latest': None}
```
pipermerriam: Since we only ever access the key `'latest'` why not just store the block itself.  `latest_block = None;  latest_block = web3.eth.getBlock('latest')`

tests/core/middleware/test_stalecheck.py
```diff
@@ -0,0 +1,108 @@
...
+        'eth_getTransactionByBlockHashAndIndex',
+        'eth_getTransactionByBlockNumberAndIndex',
+        'eth_getUncleCountByBlockHash',
+        'eth_getUncleCountByBlockNumber',
+    ]
+)
+def test_stalecheck_ignores_get_by_block_methods(request_middleware, rpc_method):
+    # This is especially critical for getBlock('latest') which would cause infinite recursion
+    with patch('web3.middleware.stalecheck._isfresh', return_value=True):
+        params = Mock()
```
pipermerriam: why is `params` being mocked.  Can you not just pass in an empty list?

web3/middleware/stalecheck.py
carver: I don't know how to reassign the variable that the inner function accesses in the outer scope, without using a mutable object or the `nonlocal` keyword, which is unavailable in Python 2.

From the the commit:

> Using a dict for closure around the cache variable, but once we move
to python 3, it can change to storing the block itself, and the inner
function can use the `nonlocal` keyword.

pipermerriam: Makes complete sense.

web3/middleware/stalecheck.py
carver: Yeah, you mentioned this on the first review. I still don't understand when someone would use it. I'll add for it for now, but I'm curious what case you have in mind.

pipermerriam: It seems like someone might want to either add more methods, or remove methods from that list.  Given the uncertainty about which methods are most appropriate to skip or not skip this seems like a good feature to have (even if you don't document it) to allow work arounds for special cases were they to arise...

web3/middleware/stalecheck.py
carver: :man_facepalming: yeah, this was a remnant of the old arguments style.

PR #281, opened 2017-09-01
Piper/ethereum tester library integration

pipermerriam:
### What was wrong?

Get started integrating the `ethereum-tester` library into the integration tests.

### How was it fixed?

* Implementing a new provider that is dependent on the `ethereum-tester` library being installed.
* Implement a few middlewares for that provider to get everything normalized to web3 expectations.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,11 @@ env:
     - TOX_POSARGS="-e py27-core-gevent"
     - TOX_POSARGS="-e py34-core-gevent"
     - TOX_POSARGS="-e py35-core-gevent"
+    # eth-testrpc
     - TOX_POSARGS="-e py27-integration-ethtestrpc -e py34-integration-ethtestrpc -e py35-integration-ethtestrpc"
+    # ethereum-tester
+    - TOX_POSARGS="-e py27-integration-ethtester -e py34-integration-ethtester -e py35-integration-ethtester"
+    # go-ethereum
     - TOX_POSARGS="-e py27-integration-goethereum -e py34-integration-goethereum -e py35-integration-goethereum" GETH_VERSION=v1.6.0
     - TOX_POSARGS="-e py27-integration-goethereum -e py34-integration-goethereum -e py35-integration-goethereum" GETH_VERSION=v1.6.1
     - TOX_POSARGS="-e py27-integration-goethereum -e py34-integration-goethereum -e py35-integration-goethereum" GETH_VERSION=v1.6.4
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -68,11 +68,36 @@ IPCProvider
         >>> web3 = Web3(Web3.IPCProvider("~/Library/Ethereum/geth.ipc")
 
 
-.. py:currentmodule:: web3.providers.tester
-
+.. py:currentmodule:: web3.providers.eth_tester
 
 EthereumTesterProvider
 ~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning:: Experimental:  This provider is experimental and should be used with caution.
+
+.. py:class:: EthereumTesterProvider(eth_tester)
+
+    This provider integrates with the ``ethereum-tester`` library.  The
+    ``eth_tester`` constructor argument should be an instance of the
+    ``eth_tester.EthereumTester`` class provided by the ``ethereum-tester``
+    library.  See the ``ethereum-tester`` library documentation for
+    instructions on how to use the library.
+
+    .. code-block:: python
+
+        >>> from web3 import Web3
+        >>> from web3.providers.eth_tester import EthereumTesterProvider
+        >>> from eth_tester import EthereumTester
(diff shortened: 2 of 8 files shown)


pipermerriam:
@carver ready for really preliminary (high level) feedback if you find time.  No rush.

carver:
First thoughts:

* The `eth_tester/middleware.py` feels heavier than I expected. Shouldn't we be able to avoid a lot of remappings by designing eth_tester to match with the expected key values? Also, aren't formatters like these more general than just eth_tester? `'eth_getUncleCountByBlockNumber': apply_formatter_if(to_integer_if_hex, is_not_named_block)`
* `web3/providers/eth_tester/provider.py` looks solid
* Mostly fixtures in here `tests/integration/test_ethereum_tester.py` nothing really to comment on yet, but I'm surprised you built all the fixtures without the tests that use them :)

Overall, looks good

pipermerriam:
> Mostly fixtures in here tests/integration/test_ethereum_tester.py nothing really to comment on yet, but I'm surprised you built all the fixtures without the tests that use them :)

Take a look at `web3.utils.module_testing`.  It's probably worth me documenting the *required* fixtures that need to be available for the module tests to work.  You can see the same large amount of fixtures in the `test_goethereum` and `test_ethtestrpc` tests.

pipermerriam:
> The eth_tester/middleware.py feels heavier than I expected. Shouldn't we be able to avoid a lot of remappings by designing eth_tester to match with the expected key values? Also, aren't formatters like these more general than just eth_tester? 'eth_getUncleCountByBlockNumber': apply_formatter_if(to_integer_if_hex, is_not_named_block)

I tried to write `ethereum-tester` so that it wasn't web3 which meant python naming conventions and normal python types (like integer filter ids rather than hexidecimal ones).  It does certainly making glueing them together quite heavy. 

 One thing that makes me more *ok* with this is that `ethereum-tester` is very strict about inputs and outputs.  I wouldn't mind some strong typing...

I read a topical quote today: *"Python is the second best language for everything"*

carver:
> Take a look at `web3.utils.module_testing`. It's probably worth me documenting the _required_ fixtures

Ah, right, makes more sense now. Yeah, "How to add testing for a new provider" is probably a thing that should exist. 

> I tried to write ethereum-tester so that it wasn't web3 which meant python naming conventions and normal python types

I thought you might say something like that. I agree with the premise, but still want to _grumble_ about it. Crazy idea: could it be run as a "convention converter", like:

```python
def snakesToCamel(snakes):
    first, _, *rest = snakes.split('_')
    return first+''.join(word[0].upper() + word[1:] for word in rest if word)
```

> I read a topical quote today: _"Python is the second best language for everything"_

Haha, a little too true.

pipermerriam:
@carver super messy commit history but this is green.  I don't think it should really be merged yet as I want to write some docs and clean it up.

web3/providers/eth_tester/provider.py
```diff
@@ -115,6 +126,30 @@ def create_log_filter(eth_tester, params):
     return filter_id
 
 
+def generate_random_private_key():
+    return encode_hex(bytes(bytearray((
+        random.randint(0, 255)
```
carver: Maybe a little note here in case anyone stumbles in, saying that this shouldn't be used for generating private keys in production, because `random` is not meant for generating secure secrets.

web3/providers/eth_tester/provider.py
```diff
@@ -0,0 +1,379 @@
...
+
+def create_log_filter(eth_tester, params):
+    filter_params = params[0]
+    filter_id = eth_tester.create_log_filter(**filter_params)
+    return filter_id
+
+
+def generate_random_private_key():
+    return encode_hex(bytes(bytearray((
+        random.randint(0, 255)
```
carver: Maybe a little note here in case anyone stumbles in, saying that this shouldn't be used for generating private keys in production, because `random` is not meant for generating secure secrets.

carver:
Sounds good, I'll hold off. Not sure how you feel about it, but I'm in favor of rebasing to squash the commits on a PR like this.

web3/providers/eth_tester/provider.py
pipermerriam: Gonna add a `_` prefix and the comment.

pipermerriam:
@carver see #290, #291, #292, and #293

pipermerriam:
@carver now has been rebased to have a clean commit history and only the actual changes specific to adding the provider.

carver:
> now has been rebased to have a clean commit history and only the actual changes specific to adding the provider. Thanks for keeping me honest.

Great! Glad to :)

> I don't think it should really be merged yet as I want to write some docs and clean it up.

I'm happy to merge with a quick and dirty version of the docs, we can tweak it going forward.

pipermerriam:
@carver wrote some very basic docs.  Looking to merge if you're 👍

PR #295, opened 2017-09-09
Adding toBytes, with same functionality as toAscii

carver:
### What was wrong?

`toAscii` is not quite semantically right: it outputs bytes that may not be valid ascii.

### Still To-do

- [x] mark `toAscii` deprecated
- [x] use a converter that doesn't `force_text` or `force_bytes`
- [x] move `toBytes` and `toText` to web3.utils.encoding
- [ ] update docs, from to/fromAscii and to/fromUtf8 to `toBytes` or `toText`, as appropriate
- [x] pass all tests
- [ ] py2 deprecation warning (probably a separate PR, though, for discussion)

### How was it fixed?

Added `toBytes` and `toText`, with `toAscii` remaining for backwards compatibility.

Moving away from `fromAscii` concept, specifying the from type in the argument instead. Like `toBytes(hexstr='0x...')`

With bonus import alphabetizing.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -50,10 +50,12 @@ from web3.utils.decorators import (
     deprecated_for,
 )
 from web3.utils.encoding import (
-    hex_encode_abi_type,
-    to_hex,
-    to_decimal,
     from_decimal,
+    hex_encode_abi_type,
+    to_bytes,
+    to_decimal,
+    to_hex,
+    to_text,
 )
 
 
@@ -88,8 +90,10 @@ class Web3(object):
     Iban = Iban
 
     # Encoding and Decoding
-    toHex = staticmethod(to_hex)
+    toBytes = staticmethod(to_bytes)
     toDecimal = staticmethod(to_decimal)
+    toHex = staticmethod(to_hex)
+    toText = staticmethod(to_text)
     fromDecimal = staticmethod(from_decimal)
 
     # Currency Utility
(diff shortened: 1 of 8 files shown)


web3/main.py
```diff
@@ -89,9 +89,11 @@ class Web3(object):
 
     # Encoding and Decoding
     toHex = staticmethod(to_hex)
-    toAscii = staticmethod(decode_hex)
+    toBytes = staticmethod(decode_hex)
+    toAscii = toBytes
```
pipermerriam: Thoughts on going ahead and marking this deprecated?

web3/main.py
```diff
@@ -89,9 +89,11 @@ class Web3(object):
 
     # Encoding and Decoding
     toHex = staticmethod(to_hex)
-    toAscii = staticmethod(decode_hex)
+    toBytes = staticmethod(decode_hex)
```
pipermerriam: Something to think about.  

The `force_text` and `force_bytes` functions from `eth_utils` are *wrong* in certain ways (since they use `latin1` instead of `utf8`).  I can't recall if we've had this discussion but, this came about because you can't encode all of the 256 possible bytes using utf8.

Both the `decode_hex` and `encode_hex` functions use these `force_X` functions under the hood.  

If continuing to use these functions in new apis we are increasing the number of places that these are used and they will almost definitely need to be fixed at some point in the near-to-medium-term future.

This has been on my mind for a while.  I don't know what the right fix is for this, but I'm inclined to at least start talking about it because `eth_utils` uses those functions **everywhere**.

web3/main.py
carver: I'm a fan, will do

web3/main.py
carver: > I can't recall if we've had this discussion but, this came about because you can't encode all of the 256 possible bytes using utf8.

Yeah, we did. If it's an arbitrary bytestring, we shouldn't try to encode it, it should just stay `bytes` (or second-best, a hexstr). Staying as `bytes` won't be possible until py3 is required, of course.

> Both the decode_hex and encode_hex functions use these force_X functions under the hood... If continuing to use these functions in new apis we are increasing the number of places that these are used

Ok, you convinced me. I'll do this the less-lazy way. The new `toBytes` and `fromBytes` won't have exactly the same functionality then.

Between this, and `sha3`, it's about time to start including a py2 deprecation warning when initializing Web3.

carver:
The changes were pretty big, so I want to get another look before merging. All that's left is documentation.

Issue #288, opened 2017-09-07
Changes to `web3.sha3`


pipermerriam:
### What was wrong?

Currently, the implementation of `web3.sha3` has some unexpected behavior and non ideal behavior.

1. It uses the `web3_sha3` RPC endpoint which allows for different providers to return different values.
2. By default it expects the value to be hashed to be passed in as a hex encoded string.
3. If you specify *any* encoding besides hex, it treats it as if it was bytes.

#### 1. Use `eth_utils.keccak` to compute the hash.  

This is more reliably going to produce the result that users expect as well as reducing RPC request overhead.

#### 2. Change the default encoding to `bytes`.

This would be a breaking change so figuring out how to deprecate the current behavior while giving users the appropriate time to upgrade would be ideal.  This could be done in a two step process.

We should have an explicit list of encodings that are supported and have logic for dealing with those appropriately.  I suggest `bytes`, `utf8`, and `hex`.

carver:
Encoding Confusion
---

There is a caveat to # 2 above, which is that if someone passes `bytes`, then the encoding is meaningless and should be ignored. Also, "encoding" as the keyword argument is confusing here, because it actually means a different thing for each of these options:

1. bytes (it's already encoded)
2. utf8 (take this string of characters, and encode them using utf-8)
3. hex (take this hex representation of bytes, and decode them into their underlying bytes)

```
"txt"   == encode utf-8 ==>   b'\x74\x78\x74'   == encode hex str ==>   "0x747874"
"txt"   <== decode utf-8 ==   b'\x74\x78\x74'   <== decode hex str ==   "0x747874"
```

Also confusing is that python supports hex literals, which seems reasonable to support too, like:
`web3.sha3(0x74657874)`. The primitive here is `int`. Again, like `bytes`, if an int is passed, then "encoding" should be ignored.

The "encoding" here is only ever referring to the type of string we're looking at. We can (and should) auto-detect all the rest. We'd be done, if it wasn't for those meddling kids and their hex-string encodings.

New API:
---

So what I'm ramping up to is a new API (the breaking change):
```
def sha3(primitive=None, text=None, hex=None)
```
My least favorite part is overwriting the builtin `hex`, but I really want a short syntax here. Even core python does stuff like this (see `file` kwarg in `print`).

So you could call any of:

```
web3.sha3(b'\x74\x78\x74')
web3.sha3(0x747874)
web3.sha3(hex="0x747874")
web3.sha3(text="txt")
```

[...]

PR #296, opened 2017-09-09
New web3.sha3

carver:
Seeking high-level feedback. I'll figure out the failing tests on Monday.

### What was wrong?

See #288 - new sha3 API, calculated without hitting RPC.

### How was it fixed?

It's not yet. Just sharing the checkpoint where all old tests pass, but not the new ones yet.

Also accidentally included alphabetizing imports in this and another pull, but the merge will get ugly if I try to strip it out. I'll rebase off the other if it gets merged in.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -162,13 +162,12 @@ class Web3(object):
                 "code to use sha3(text='txt'), sha3(hexstr='0x747874'), "
                 "sha3(b'\\x74\\x78\\x74'), or sha3(0x747874)."
             ))
-        else:
-            if not isinstance(primitive, (bytes, int, type(None))):
-                warnings.warn(DeprecationWarning(
-                    "The first argument as a string has been deprecated. Please update your "
-                    "code to use sha3(text='txt'), sha3(hexstr='0x747874'), "
-                    "sha3(b'\\x74\\x78\\x74'), or sha3(0x747874)."
-                ))
+        elif not isinstance(primitive, (bytes, int, type(None))):
+            warnings.warn(DeprecationWarning(
+                "The first argument as a string has been deprecated. Please update your "
+                "code to use sha3(text='txt'), sha3(hexstr='0x747874'), "
+                "sha3(b'\\x74\\x78\\x74'), or sha3(0x747874)."
+            ))
 
         if isinstance(primitive, bytes):
             if bytes == str:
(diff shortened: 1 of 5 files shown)


web3/main.py
```diff
@@ -152,13 +153,53 @@ def currentProvider(self):
         ))
         return self.manager.providers[0]
 
-    @coerce_return_to_text
-    def sha3(self, value, encoding="hex"):
-        if encoding == 'hex':
-            hex_string = value
+    @staticmethod
+    @apply_to_return_value(encode_hex)
+    def sha3(primitive=None, text=None, hexstr=None, encoding=None):
```
pipermerriam: I think we should either:

* implement validation that ensures that you can't call this function *wrongly* such as `web3.sha3(b'arst', text='bar', hexstr='baz')`
* do some shenanigans with the function signature to make it impossible to call in that manner.

web3/main.py
```diff
@@ -152,13 +153,53 @@ def currentProvider(self):
...
+        if encoding is not None:
+            warnings.warn(DeprecationWarning(
+                "The encoding keyword has been deprecated.  Please update your "
+                "code to use sha3(text='txt'), sha3(hexstr='0x747874'), "
+                "sha3(b'\\x74\\x78\\x74'), or sha3(0x747874)."
+            ))
         else:
-            hex_string = encode_hex(value)
-        return self.manager.request_blocking('web3_sha3', [hex_string])
+            if not isinstance(primitive, (bytes, int, type(None))):
```
pipermerriam: `elif not ....`

Looks like you have one level of unnecessary indentation here.

web3/main.py
```diff
@@ -23,16 +24,18 @@
 from web3.admin import Admin
 from web3.db import Db
 from web3.eth import Eth
+from web3.iban import Iban
```
pipermerriam: Side Note:  I'd like to extract the iban functionality into a stand-alone library and get it out of web3.  It's a format that isn't used or supported very broadly in the ecosystem and while I'm still fine supporting it, I'd rather it live in a different repository.

web3/main.py
carver: I'm anti-shenanigan (at least in this context), will validate.

PR #298, opened 2017-09-12
fix JSONBaseProvide.isConnected

chenfengyuan:
### What was wrong?
After https://github.com/pipermerriam/web3.py/commit/699a775411062bfd748a38565c2de3703d757b30, JSONBaseProvide.isConnected is broken because of doing json.loads on dict object

### How was it fixed?
remove the superfluous json.loads in isConnected

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/providers/base.py
+++ b/web3/providers/base.py
@@ -54,8 +54,7 @@ class JSONBaseProvider(BaseProvider):
 
     def isConnected(self):
         try:
-            response_raw = self.make_request('web3_clientVersion', [])
-            response = json.loads(force_text(response_raw))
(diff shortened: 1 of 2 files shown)


chenfengyuan:
#299

carver:
Thanks for finding and fixing!

Could you also help make sure this doesn't revert in the future by adding a test?

Something very simple could be good enough for merging, like:
```
    def test_is_connected(self, web3):
        assert web3.isConnected()
```
in `web3/utils/module_testing/web3_module.py`.

chenfengyuan:
@carver 
Hi, test id added and passed 😄

PR #301, opened 2017-09-13
New eth.sign and recovery hash generator

carver:
### What was wrong?

`eth.sign` was ambiguous about the kind of data it received. It was also difficult to verify the message; users were completely on their own.

### How was it fixed?

`eth.sign` now follows the pattern of the other converters, by passing one of these: `(bytesdata, hexstr='0x..', text='my msg')` 

There is also a new hash generator `eth._recoveryMessageHash` which adds the preamble and generates the hash that `ecrecover` takes as a first argument.  The preamble is the one defined by geth, a loose standard until [EIP 191](https://github.com/ethereum/EIPs/issues/191) is solidified.

It looks like we won't get `ecrecover` functionality until we upgrade pyethereum.

Also, I'll add docs when I get a :ship: 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/shh-module/test_shh_filter.py
+++ b/tests/core/shh-module/test_shh_filter.py
@@ -4,24 +4,25 @@ from web3.utils.compat import sleep
 def test_shh_filter(web3, skip_if_testrpc):
     skip_if_testrpc(web3)
     recieved_messages = []
-    shh_filter = web3.shh.filter({"topics": [web3.fromAscii("test")]})
+    topic = web3.toHex(text="test")
+    shh_filter = web3.shh.filter({"topics": [topic]})
     shh_filter.watch(recieved_messages.append)
 
     payloads = []
     payloads.append(str.encode("payload1"))
     web3.shh.post({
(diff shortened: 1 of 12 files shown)


tests/core/shh-module/test_shh_filter.py
```diff
@@ -4,24 +4,25 @@
...
+    shh_filter = web3.shh.filter({"topics": [topic]})
     shh_filter.watch(recieved_messages.append)
 
     payloads = []
     payloads.append(str.encode("payload1"))
     web3.shh.post({
-        "topics": [web3.fromAscii("test")],
-        "payload": web3.fromAscii(payloads[len(payloads) - 1]),
+        "topics": [topic],
+        "payload": web3.toHex(text=payloads[-1]),
```
pipermerriam: Doesn't need to be cleaned up here, but it might be wise for us to remove the use of `web3.toHex` here in favor of using the `eth_utils.encode_hex` function so as to not make our tests dependent on this API.

PR #310, opened 2017-09-18
named middleware: remove and replace

### LINKED ISSUE #284: Name middleware, for removing/replacing existing middleware
### What was wrong?

It is difficult to modify or replace existing middleware after you add it.

### How can it be fixed?

Optionally, name the middleware at add time, like:

```
web3.add_middlware('stale-check', make_stalecheck_middleware(hours=2))
# or
web3 = Web3(..., middlewares=[caching_middleware, ('stale-check', stalecheck_middleware)])
```

Then, you can remove them individually, like:

```
web3.remove_middleware('stale-check')
# or 
web3.remove_middleware(caching_middleware)
```

Or replace with a new one:
```
web3.replace_middleware('stale-check', make_stalecheck_middleware(hours=12))
```

@pipermerriam referenced some difficulty during [our chat](https://gitter.im/pipermerriam/web3.py?at=59a9e60c210ac26920e721d5) about removing/replacing a middleware in place, because of how they're chained together. I don't understand something there, because It seems like it should be something straightforward like:

manager.py:
```
def remove(self, middleware):
    remove_idx = self._middlewares.indox(middleware)
    self._middlewares = self._middlewares[:remove_idx] + self._middewares[remove_idx + 1:]
    self._generate_request_functions()
```

(at least for the case where you are removing by the function. The name takes a bit more work)

Anyway, it will all come out in the implementation.

carver:
Also, a NamedElementStack, which might be useful for runtime
modification of the providers list, too.

Closes #284 (see for details)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -40,18 +40,78 @@ Pythonic
     where appropriate. For example, it converts the raw hex string returned by the RPC call
     ``eth_blockNumber`` into an ``int``.
 
+.. _Modifying_Middleware:
+
+Modifying Middleware
+------------------
+
+Middleware can be added, removed, replaced, and cleared at runtime. To make that easier, you
+can name the middleware for later reference. Alternatively, you can use a reference to the
+middleware itself.
+
+.. py:method:: Web3.middleware_stack.add(middleware, name=None)
+
+    Middleware will be added to the top of the stack. That means the new middleware will modify the
+    request first, and the response last. You can optionally name it with any hashable object,
+    typically a string.
+
+    .. code-block:: python
+
+        >>> w3 = Web3(...)
+        >>> w3.middleware_stack.add(web3.middleware.pythonic_middleware)
+        # or
+        >>> w3.middleware_stack.add(web3.middleware.pythonic_middleware, 'pythonic')
+
+.. py:method:: Web3.middleware_stack.remove(middleware)
+
+    Middleware will be removed from wherever it sat in the stack. If you added the middleware with
+    a name, use the name to remove it. If you added the middleware as an object, use the object to
+    remove it.
+
+    .. code-block:: python
+
+        >>> w3 = Web3(...)
+        >>> w3.middleware_stack.remove(web3.middleware.pythonic_middleware)
+        # or
+        >>> w3.middleware_stack.remove('pythonic')
+
+.. py:method:: Web3.middleware_stack.replace(old_middleware, new_middleware)
+
+    Middleware will be replaced wherever it sat in the stack. If the middleware was named, it will
+    continue to have the same name. If it was un-named, then you will now reference it with the new
+    middleware object.
+
+    .. code-block:: python
+
+        >>> from web3.middleware import pythonic_middleware, attrdict_middleware
+        >>> w3 = Web3(...)
+
+        >>> w3.middleware_stack.replace(pythonic_middleware, attrdict_middleware)
+        # this is now referenced by the new middleware object, so to remove it:
+        >>> w3.middleware_stack.remove(attrdict_middleware)
+
+        # or, if it was named
+
+        >>> w3.middleware_stack.replace('pythonic', attrdict_middleware)
+        # this is still referenced by the original name, so to remove it:
+        >>> w3.middleware_stack.remove('pythonic')
+
+.. py:method:: Web3.middleware_stack.clear()
+
+    Empty all the middlewares, including the default ones.
+
+    .. code-block:: python
+
+        >>> w3 = Web3(...)
+        >>> w3.middleware_stack.clear()
+        >>> assert len(w3.middlewares) == 0
+
+
 Built-in Middleware
 ------------------
 
-Web3 ships with middleware for custom use, as desired. Middleware can be added after creating
-your Web3 object, like:
-
-.. code-block:: python
-
-    w3 = Web3(...)
-    w3.add_middleware(my_middleware)
-
-Alternatively, you can pass in middlewares to the Web3 constructor.
+Web3 ships with non-default middleware, for your custom use. In addition to the other ways of
+:ref:`Modifying_Middleware`, you can specify a list of middleware when initializing Web3, with:
 
 .. code-block:: python
 
@@ -60,8 +120,10 @@ Alternatively, you can pass in middlewares to the Web3 constructor.
 .. warning::
   This will
   *replace* the default middlewares. To keep the default functionality,
-  either use ``add_middleware()`` from above, or add the default middlewares to your list of
+  either use ``middleware_stack.add()`` from above, or add the default middlewares to your list of
   new middlewares.
+
+Below is a (maybe partial) list of built-in middleware.
 
 Stalecheck
 ~~~~~~~~~~~~
@@ -80,7 +142,7 @@ Stalecheck
     .. code-block:: python
 
         two_day_stalecheck = make_stalecheck_middleware(60 * 60 * 24 * 2)
-        web3.add_middleware(two_day_stalecheck)
+        web3.middleware_stack.add(two_day_stalecheck)
 
     If the latest block in the blockchain is older than 2 days in this example, then the
     middleware will raise a ``StaleBlockchain`` exception on every call except
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -104,7 +104,7 @@ middleware itself.
 
         >>> w3 = Web3(...)
         >>> w3.middleware_stack.clear()
-        >>> assert len(w3.middlewares) == 0
+        >>> assert len(w3.middleware_stack) == 0
 
 
 Built-in Middleware
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -123,7 +123,7 @@ Web3 ships with non-default middleware, for your custom use. In addition to the 
   either use ``middleware_stack.add()`` from above, or add the default middlewares to your list of
   new middlewares.
 
-Below is a (maybe partial) list of built-in middleware.
+Below is a list of built-in middleware.
 
 Stalecheck
 ~~~~~~~~~~~~
--- a/tests/core/manager/conftest.py
+++ b/tests/core/manager/conftest.py
@@ -1,4 +1,3 @@
-
 import itertools
 import pytest
 
--- a/tests/core/manager/conftest.py
+++ b/tests/core/manager/conftest.py
@@ -1,0 +1,28 @@
+
+import itertools
+import pytest
+
+
+@pytest.fixture
+def middleware_factory():
+    unique_ids = itertools.count()
+
+    def factory(key=None):
+        if key is None:
+            key = next(unique_ids)
+        key = str(key)
+
+        class Wrapper:
+            def __repr__(self):
+                return 'middleware-' + key
+
+            def __call__(self, make_request, web3):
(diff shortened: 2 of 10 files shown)


tests/core/manager/test_middleware_add_and_clear_api.py
```diff
@@ -6,62 +9,158 @@
...
+    manager.middleware_stack.add(middleware_factory(), 'a')
+
+    id_add = id(manager._wrapped_provider_request_functions)
+
+    assert id_add != id_start
+
+    # replace
+    manager.middleware_stack.replace('a', middleware_factory())
+
+    id_replace = id(manager._wrapped_provider_request_functions)
```
pipermerriam: we don't necessarily need to fix this now, but I'm wondering if using mocks and just asserting that the `_generate_request_functions` method was called after replacing it with a mock.

web3/utils/datastructures.py
```diff
@@ -87,3 +89,81 @@ def __eq__(self, other):
             return self.__dict__ == dict(other)
         else:
             return False
+
+
+class NamedElementStack:
```
pipermerriam: This feels like it *should* inherit from at least one of the `collections.abc` classes.  Maybe any or all of:

* `Iterable`
* `Sequence`
* `Mapping`

pipermerriam: Oh, and for python2, this needs to inherit from `object` at a minimum.

carver: Hm, I started this is a subclass of an `OrderedDict`, and now I forget exactly why that was a bad idea. I think it was that adding via `stack[key] = val` seemed like an unnatural way to say that val is now at the top of the stack.

I'm on board with `Sequence`, which itself inherits from `Iterable`.

carver: Actually, even `OrderedDict` isn't a sequence, so I'll stick with `Iterable` for now. If we have a good reason to add `Mapping` later, we can add it.

carver: Actually, non-mutable `Mapping` looks good.

web3/manager.py
```diff
@@ -33,37 +38,36 @@ def __init__(self, web3, providers, middlewares=None):
         self.providers = providers
 
     web3 = None
-    _middlewares = None
     _providers = None
 
     @property
     def middlewares(self):
-        return self._middlewares or tuple()
+        return tuple(self.middleware_stack)
```
pipermerriam: I figured that with the introduction of `RequestManager.middleware_stack` we'd remove this property.  Is there a reason to keep it that I'm not seeing?

web3/manager.py
```diff
@@ -33,37 +38,36 @@ def __init__(self, web3, providers, middlewares=None):
...
 
     @middlewares.setter
-    def middlewares(self, value):
-        self._middlewares = tuple(value)
+    def middlewares(self, new_middlewares):
+        self.middleware_stack = NamedElementStack(new_middlewares)
         self._generate_request_functions()
+        self.middleware_stack.on_change(self._generate_request_functions)
```
pipermerriam: This makes me feel funny.  It's totally pragmatic and I could almost call it *elegant* but it feels like cramming javascript into python.  I don't love the alternative of just ensuring that `_generate_request_functions()` is called at all points where middlewares are modified, but.... I also don't love that this intertwines these two objects or that it feels a bit *magic*.

Proposed 3rd way to do this?

* Change `NamedElementStack` to be `MiddlewareStack`.
* Initialize it with the `provider` instances as well.
* Move the request function generation into the `MiddlewareStack` object itself, maybe as a  computed property which is cached based on the providers and middlewares in the stack.

Thoughts?

web3/utils/datastructures.py
```diff
@@ -87,3 +89,96 @@ def __eq__(self, other):
...
+            self._replace_with_new_name(old, new)
+        else:
+            self._queue[old] = new
+        self._changed()
+
+    def remove(self, old):
+        if old not in self._queue:
+            raise ValueError("You can only remove something that has been added")
+        del self._queue[old]
+        self._changed()
```
pipermerriam: Just an idea: this could be done with a decorator.  I think the following should work if you like the pattern (I'm not sure it's actually better)

```python
def update_when_called(method):
    def inner(self, *args, **kwargs):
        return_value = method(self, *args, **kwargs)
        self._changed()
        return return_value
    return inner
```

web3/providers/base.py
```diff
@@ -11,24 +11,23 @@
 
 
 class BaseProvider(object):
-    _middlewares = None
+    _middlewares = []
```
pipermerriam: This will/can leak across instances of the `Provider`.  Since lists are mutable, every instance will by default share this instance of this list.  Current implementation won't actual fail, but it's still bad mojo.

docs/middleware.rst
```diff
@@ -40,18 +40,78 @@ Pythonic
...
-.. code-block:: python
+.. py:method:: Web3.middleware_stack.add(middleware, name=None)
+
+    Middleware will be added to the top of the stack. That means the new middleware will modify the
+    request first, and the response last. You can optionally name it with any hashable object,
+    typically a string.
+
+    .. code-block:: python
+
+        >>> w3 = Web3(...)
```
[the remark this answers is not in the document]
carver: I don't love `web3` being the instance name actually, since the package is also called web3. But consistency is more important here, so I agree in the context of this PR and will update. Later, I'll see if I can convince myself (and then you) that we should change the standard practice everywhere in the docs and tests. It's a big change now, but will only get more massive over time.

For now, I'll go with something like:

```
from web3.middleware import pythonic_middleware
web3 = Web3(...)
web3.middleware_stack.add(web3.middleware.pythonic_middleware)
```

(A little preview of my raw thoughts: the current standard means you can never use `import web3`, which is a valid style choice in some repositories. You end up losing access to the module as soon as you initialize `web3 = Web3()`. I was using the module that way in the little snippets, with `web3.middleware.pythonic_middleware`, for example. The standard also probably leads to confusion for beginners who might expect to `import web3` and use it statically without initializing.)

tests/core/manager/conftest.py
```diff
@@ -0,0 +1,28 @@
+
```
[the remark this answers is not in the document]
carver: Interesting, I think I've been doing that in lots of places. I like the space at the top of the editor, but don't mind forgoing it in the future for consistency. If it's worth mentioning in a comment, it's probably worth adding to the style guide.

tests/core/manager/test_middleware_add_and_clear_api.py
```diff
@@ -1,67 +1,179 @@
...
 
-
-def middleware_factory():
-    def generated_middleware(*args, **kwargs):
-        def middleware(*args, **kwargs):
-            pass
-        return middleware
-    return generated_middleware
+if sys.version_info.major >= 3:
+    from unittest.mock import Mock
```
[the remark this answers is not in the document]
carver: Eh, I'll move it inside for now. No reason to add the backport with py3-only close enough on the horizon.

web3/manager.py
carver: Adds some convenience, but more on the setter side than the get. I guess whether this makes sense depends on your next comment.

web3/providers/base.py
carver: Yeah, it's super defensive, since no one should be messing with it directly, but I'm on board. Easy enough to change to `()`

web3/manager.py
carver: > feels like cramming javascript into python

I'd say it's not so much a javascript pattern as a general object-oriented pattern. But if we can get rid of it, it's still a complexity win, so let's look at alternatives.

> 3rd way
> * Change NamedElementStack to be MiddlewareStack.
> * Initialize it with the provider instances as well.
> * Move the request function generation into the MiddlewareStack object itself, maybe as a computed property which is cached based on the providers and middlewares in the stack.

I'm interpreting this idea like:

```
ms = MiddlewareStack(middleware, providers)
...
class RequestManager:
    def _make_request(self, method, params):
        for provider in self.providers:
            request_func_for_provider = ms.request_func(self.web3, provider)
            try:
                return request_func_for_provider(method, params)
            except CannotHandleRequest:
                continue
```

It has the advantage of moving the change observation out of RequestManager, making it simpler. But MiddlewareStack gets a lot more responsibility of managing all the providers, rather than mimicking a basic collection.

That extra responsibility would add significant new lines of code, just pushing the problem to MiddlewareStack.

So option # 4:

Instead of pushing the request generation into MiddlewareStack, push it into the provider. Why is that any better? Each Provider only cares about its own middlewares, so no one is keeping a list of request functions or lists of MiddlewareStacks around, or regenerating caches for all providers at once.

A new `BaseProvider.request_func(web3, outer_middlewares)` would implement `combine_middlewares` so RequestManager can make a request with:

```
class RequestManager:
    def _make_request(self, method, params):
        for provider in self.providers:
            make_request_fn = provider.request_func(self.web3, self.middleware_stack)
            try:
                return make_request_fn(method, params)
            except CannotHandleRequest:
                continue
```

Then:

1. `RequestManager` never messes with provider middleware internals at all (encapsulation win)
2. Caching logic moves from `RequestManager` to `BaseProvider`. RequestManager was managing a list of cached request functions, BaseProvider is only caching a single request function (complexity win)
3. `on_change` becomes unnecessary (complexity win)
3. Don't have to generate the request function until the provider needs it (tiny performance win)
4.  Pushes a little extra logic into Providers, making testing harder for any provider that doesn't extend BaseProvider (testing loss)
5. Misbehaving Providers could theoretically ignore RequestManager middlewares... but why? (encapsulation loss)

docs/middleware.rst
```diff
@@ -40,18 +40,78 @@ Pythonic
...
+
+        >>> w3.middleware_stack.replace(pythonic_middleware, attrdict_middleware)
+        # this is now referenced by the new middleware object, so to remove it:
+        >>> w3.middleware_stack.remove(attrdict_middleware)
+
+        # or, if it was named
+
+        >>> w3.middleware_stack.replace('pythonic', attrdict_middleware)
+        # this is still referenced by the original name, so to remove it:
+        >>> w3.middleware_stack.remove('pythonic')
```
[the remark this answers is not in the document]
carver: `default_normalization` seems reasonable (although maybe applies to returned vals better than request vals). Maybe `builtin_types` captures both directions? Either way, it probably belongs in a separate PR. Although conceptually simple, it will add a lot of line change noise to this already-long PR.

web3/utils/datastructures.py
carver: Slightly positive on it. Moot if we get rid of change listener. Punting, until that's cleared up.

docs/middleware.rst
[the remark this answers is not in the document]
pipermerriam: I'm convinced.  `w3` works for me.

web3/manager.py
pipermerriam: I'm +1 on moving it into the providers.  Seems the best option on the table that I've seen.

tests/core/manager/test_middleware_add_and_clear_api.py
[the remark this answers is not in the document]
pipermerriam: Your assuming I won't pull a guido on you and extend python 2 support for 10 more years.

Issue #279, opened 2017-08-31
When a contract call results in an empty address, return None


carver:
The next step is to flesh out the specs, and building a design for review. It may or may not require py3.

### What was wrong?

If a contract returns a uint, I want it to get the return value as an `int`. If it is an address, then a value of `0x000...000` should return `None` instead, and so on. (TODO: flesh out)

One motivator is to extract the `web3utils` library from `ens.py` so ENS support can be cleanly integrated.

### How can it be fixed?

It probably means building a layer in the contract that does ABI-aware modification of the eth_call return value. At first, I was thinking middleware, but because it relies on the contract ABI, and fact that it only affects eth_call, probably means a contract-layer implementation makes the most sense.

carver:
Ok, so for some reason, I thought this was a bigger job than it was. Really the only thing I want to change is to make 0x0 addresses return `None`.

This should happen at the end of a contract call, so a new kwarg on contract like `eth.contract(return_none=True)` seems natural. This should be a straightforward change to `normalize_return_type()`

I would just make it happen by default, but it's not backwards compatible, so that will have to wait for v4.

pipermerriam:
I don't think it's a good idea to return `None` in place of an empty address.  Reasoning is:

* The empty address is a valid address 
* This will introduced a second semantic for the empty address meaning code will now have to both check whether its `0x0000...000` as well as `None`.  

Also gets screwy if you want to use the result of one call as the input to another call.

```python
token_A.balanceOf(contract_B.call().owner())
```

This would result in a failure if the owner of `contract_B` was the empty address....

carver:
Ok, I'll pause advocating for this until I see how painful it is to extract out of the ENS lib. I'm not sure how many places will have this transformation required:

From
```
def resolve():
  resolver_addr = ens.resolver(namehash)
  if resolver_addr:
    resolver = Resolver(resolver_addr)
    return resolver.addr(namehash)
  else:
    return None
```

To something like:
```
def resolve():
  resolver_addr = ens.resolver(namehash)
  if Web3.toDecimal(resolver_addr):
    resolver = Resolver(resolver_addr)
    return resolver.addr(namehash)  
  else:
    return None
```
For now, it just makes me a little sad that the code won't be as elegant.

----
_But, while we're talking about it..._

I don't agree with all the implications of this:

> The empty address is a valid address

While technically valid, 0x0 _is_ a special case. Specifically, uninitialized values in the EVM start as `0x0`. Here's an example:

[...]

PR #312, opened 2017-09-19
Pluggable normalization of return values

carver:
### What was wrong?

I want to have my contract methods return `None` instead of 0x0, mostly so I can do cleaner tests on my lookups, like:

```
resolver = ens_contract.resolver(namehash)
if resolver:
  # triggers some logic, like setting up the existing resolver
else:
  # triggers other logic, like setting up and pointing to the resolver
```

Background #279 

### How was it fixed?

Fiddling with the idea of plugging in custom normalization mechanisms.

This started to open up other things I'm excited about, so I feel like it's a good direction.  For example, it made it trivial to decode only `string` abi types as utf-8.

#### Cute Animal Picture

Cute animal picture


pipermerriam:
I have a suggestion for re-architecting this.  

Currently, web3 uses the `normalize_return_type` function by default to normalize return data.  This PR adds an additional layer of normalization via a custom normalizer passedinto the `normalize_return_type` function.

I propose that we instead keep the normalization a single layer deep.  This can be done by adding a new method to the contract object `Contract.normalize_return_data(function_abi, output_data)`.

* `function_abi` is the full ABI for the function that was called
* `output_data` is the result of decoding the output data from the function ABI.

The default implementation will call the `normalize_return_type` function in the same manner that it is currently called.  Users may choose to override this method on the contract object to provide their own normalization.

## Why?

Benefits I see are:

1. Keeps the normalization simple and a single layer deep.
2. Allows normalization based on both the data type as well as the actual function which was called.

web3/utils/abi.py
```diff
@@ -388,15 +393,13 @@ def abi_to_signature(abi):
     return function_signature
 
 
+@curry
```
pipermerriam: I ❤️ this function soooo much.  Makes things that you end up `partial`-ing so much cleaner and elegant.

web3/utils/abi.py
```diff
@@ -438,11 +442,15 @@ def return_val_and_type(elements):
 
 @coerce_return_to_text
 def data_tree_vals(data_tree):
-    return recursive_map(lambda els: els[1] if _is_two_tuple(els) else els, data_tree)
+    values = recursive_map(lambda els: els[1] if _is_two_tuple(els) else els, data_tree)
```
pipermerriam: *nitpick* lambda is bugging me a bit.  Seems like a simple thing to

```
def _first_if_two_tuple(value):
    if is_to_tuple(value):
        return value[0]
    else:
        return value
```

which then makes the line in question:

```python
values = recursive_map(_first_if_two_tuple, data_tree)
```

which *might* be easier to read.... your call.

web3/utils/abi.py
```diff
@@ -461,3 +469,18 @@ def normalize_return_type(data_type, data_value):
...
         return data_value
+
+
+def map_abi_data(normalizers, types, data):
+    all_normalizers = itertools.chain(
+        DEFAULT_RETURN_NORMALIZERS,
+        normalizers,
+    )
+
+    pipeline = itertools.chain(
```
pipermerriam: I really like this variable name.

web3/utils/abi.py
carver: Ok, attacked this, plus the issue of ambiguity (in case we end up with tuples of two values at some point, I don't want to accidentally trigger this mapping). Added a minimal class for type-checking. Next, to the docstrings...

tests/core/utilities/test_abi.py
```diff
@@ -0,0 +1,79 @@
...
+    collapse_type,
+    map_abi_data,
+)
+
+
+@pytest.mark.parametrize(
+    'original, expected',
+    [
+        ('address', 'address'),
+        # ('uint[2][]', 'uint256[2][]'),  # pending an eth_abi pull request
```
pipermerriam: Just made a release with this included.  `ethereum-abi-utils>=0.4.2`

web3/utils/abi.py
```diff
@@ -1,23 +1,42 @@
...
 from eth_abi.abi import (
     process_type,
 )
 
+from web3.utils.formatters import (
+    recursive_map
+)
+
+BASE_RETURN_NORMALIZERS = [
+    lambda typ, data: (typ, to_checksum_address(data)) if typ == 'address' else (typ, data),
```
pipermerriam: use of a lambda here is bugging me.  lambda + long line + turnary-statement = 😢

web3/utils/abi.py
```diff
@@ -1,23 +1,42 @@
...
+    # lambda typ, data: data.decode(errors='ignore') if typ == 'string' else data,
```
pipermerriam: leftover comment?

web3/utils/abi.py
```diff
@@ -378,8 +397,151 @@ def abi_to_signature(abi):
...
+            [
+                abi_sub_tree(sub_type, sub_value)
+                for sub_value in data_value
+            ],
+        ])
+    else:
+        return ABITypedData([collapsed, data_value])
+
+
+# TODO if this makes any sense at all, it probably belongs in eth_abi
```
pipermerriam: 👍

web3/utils/abi.py
```diff
@@ -378,8 +397,151 @@ def abi_to_signature(abi):
...
+def strip_abi_type(elements):
+    if isinstance(elements, ABITypedData):
+        return elements.data
+    else:
+        return elements
+
+
 @coerce_return_to_text
 def normalize_return_type(data_type, data_value):
+    # TODO replace all uses with map_abi_data
```
pipermerriam: Is this still a `TODO`?

web3/utils/abi.py
carver: The simplest alternative is to define a new function just above, but it bugs me to have the constant declared beneath other functions. Which leads me to creating a new module to hold the new formatter, but the name formatters is already taken, and doesn't mean the same thing. I guess web3/util/normalizers.py unless you have other suggestions, then I'll do something like:

```
from web3.util.normalizers import (
  address_to_checksum,
)

BASE_RETURN_NORMALIZERS = [
  address_to_checksum,
]
```

web3/utils/abi.py
carver: Oops, yeah, that led me down the #318 path. I'll remove it.

web3/utils/abi.py
carver: Yeah, I was going to punt it, but I'll go add it now.

carver: Glad I did. I realized the obvious, when applying the same pattern to events (that popping single elements out of their list is a special case). I removed the special case from `map_abi_data` and into the contract code.

web3/utils/abi.py
pipermerriam: I'd be fine with the function definition above.  There's plenty of precedence of that format in my other code....  but you're call.

PR #317, opened 2017-09-19
Remove all positional eth.contract args, except address

carver:
### What was wrong?

There was ambiguity about specifying the ABI in a contract. This ambiguity made it more challenging to do simple things like auto-convert the ABI from a json string, as needed.

### How was it fixed?

* Remove all the deprecated positional arguments.
* Add automatic json encoding of ABI.
* Address must be specified when instantiating contract

#### Cute Animal Picture

Cute animal picture


web3/contract.py
```diff
@@ -120,85 +118,36 @@ class Contract(object):
...
-        if code_runtime is not empty:
-            self.bytecode_runtime = code_runtime
-        if source is not empty:
-            self._source = source
-
-        if address is not empty:
-            validate_address(address)
-            self.address = to_normalized_address(address)
+        if address:
+            self.address = self.normalize_property('address', address)
```
pipermerriam: Thoughts on doing this via a `property` with a `setter` which performs this normalization?  Could do the same with `abi`

web3/contract.py
```diff
@@ -207,13 +156,16 @@ def factory(cls, web3, contract_name=None, **kwargs):
 
         kwargs['web3'] = web3
 
-        for key in kwargs:
+        for key in dict(kwargs):
```
pipermerriam: why is this wraped in `dict`?

carver: I suppose it was a kind of knee-jerk reaction to modifying `kwargs` while looping through it. I don't have a specific reason that I know the assignment mid-loop would break without the change.

pipermerriam: Ah, I have typically allowed myself to modify `kwargs` since it's not going to leak the same way modifying a normal dictionary would.  I'd suggest at least leaving a comment here, or using one of the `toolz.dicttoolz` functions for the assignment below such as `assoc`

web3/contract.py
carver: Runs into some issues with class property `abi` and pseudo-class property `address`. (the property setter confuses `@combomethod`s that have `self.address`)

web3/contract.py
carver: I was abstractly worried about modifying the dict mid-iteration. But, I just [RTFA in Python](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects):

> Iterating views while adding or deleting entries in the dictionary may raise a RuntimeError or fail to iterate over all entries.

I take that to mean that replacing existing entries while iterating won't cause any problems.

tests/core/contracts/conftest.py
```diff
@@ -344,3 +344,8 @@ class LogTopics(object):
 @pytest.fixture()
 def emitter_log_topics():
     return LogTopics
+
+
+@pytest.fixture()
+def some_address():
+    return '0x5B2063246F2191f18F2675ceDB8b28102e957458'
```
pipermerriam: In other news, I've enjoyed using the original DAO address for this type of thing :)

web3/contract.py
```diff
@@ -120,85 +118,36 @@ class Contract(object):
...
-            self.abi = abi
-        if code is not empty:
-            self.bytecode = code
-        if code_runtime is not empty:
-            self.bytecode_runtime = code_runtime
-        if source is not empty:
-            self._source = source
-
-        if address is not empty:
-            validate_address(address)
```
pipermerriam: Did address validation get dropped?  Not seeing this added back in anywhere...

web3/eth.py
```diff
@@ -308,37 +307,18 @@ def uninstallFilter(self, filter_id):
...
-        if has_address:
-            if 'address' in kwargs:
-                address = kwargs.pop('address')
-            elif is_address(args[0]):
-                address = args[0]
-            elif is_address(args[1]):
-                address = args[1]
-                kwargs['abi'] = args[0]
+
+        NewContract = ContractFactoryClass.factory(self.web3, **kwargs)
```
pipermerriam: I'd like to figure out clear terminology for the *class* and *instance*.  I've used the term *contract* to mean *instance* and *factory* to mean *class*.  This name conflicts a bit with that terminology so I'd either like to come up with terminology that we both like or use the existing terminology and keep things consistently named.

web3/contract.py
carver: Nope, it's still there at: https://github.com/pipermerriam/web3.py/pull/317/files#diff-2171ba22a13b6793b04c4a1287049a67R147

web3/eth.py
carver: :+1: for switching to `ContractFactory`

tests/core/eth-module/test_eth_contract.py
```diff
@@ -15,10 +15,9 @@
     'args,kwargs,expected',
     (
         ((ADDRESS,), {}, None),
-        ((ADDRESS, ABI), {}, None),
         ((INVALID_CHECKSUM_ADDRESS,), {}, ValueError),
         ((), {'address': INVALID_CHECKSUM_ADDRESS}, ValueError),
-        ((ABI), {'address': INVALID_CHECKSUM_ADDRESS}, ValueError),
+        ((), {'address': INVALID_CHECKSUM_ADDRESS}, ValueError),
```
pipermerriam: This looks identical to the line above it?

PR #318, opened 2017-09-20
Better string encoding

carver:
### What was wrong?

Strings are utf-8 encoded in contracts, but latin-1 encoded in web3 v3

### How was it fixed?

It's not completely, but it's a little test case for using pluggable normalization to decode only strings as utf-8.

This PR includes the code for pluggable return value normalization. Only commits 2671951...6bb4f39 are relevant to this specific goal of string encoding.

Python 2 support is not a goal. This will block until web3 v4.

#### Cute Animal Picture

Cute animal picture


tests/core/contracts/test_normalization_of_return_types.py
```diff
@@ -2,6 +2,15 @@
 
 from web3.utils.abi import normalize_return_type
 
+EMPTY_ADDR = '00' * 20
```
pipermerriam: looks unused.

web3/contract.py
```diff
@@ -589,6 +591,21 @@ def __getattr__(self, function_name):
...
     # Private Helpers
     #
+    _return_data_normalizers = []
+
+    @classmethod
+    def _normalize_return_data(cls, output_types, output_data):
+        data_tree = abi_data_tree(output_types, output_data)
+
+        for normalizer in cls._return_data_normalizers:
+            data_tree = data_tree_map(normalizer, data_tree)
```
pipermerriam: to remove mutation...

```python
from cytoolz import pipe
data_tree = pipe(
    output_data, 
    partial(abi_tree_data, output_types),
    *(partial(data_tree_map, normalizer) for normalizer in cls._return_data_normalizers)
)
```

Readability could be cleaned up by decomposing this into multiple lines with intermediate variables.

web3/utils/compat/compat_codecs.py
```diff
@@ -0,0 +1,24 @@
...
+
+def backslashreplace(ex):
+    # The error handler receives the UnicodeDecodeError, which contains arguments of the
+    # string and start/end indexes of the bad portion.
+    bstr, start, end = ex.args[1:4]
+
+    # The return value is a tuple of Unicode string and the index to continue conversion.
+    # Note: iterating byte strings returns int on 3.x but str on 2.x
+    return (
+        u''.join(
```
pipermerriam: You should be able to remove the `u` by adding a `from __future__ import unicode_literals` to the top of this module.

web3/utils/compat/compat_codecs.py
```diff
@@ -0,0 +1,24 @@
...
+            '\\x{:02x}'.format(c if isinstance(c, int) else ord(c))
+            for c in bstr[start:end]
+        ),
+        end,
+    )
+
+
+def setup_backslashreplace():
+    if sys.version_info[:2] < (3, 5):
+        codecs.register_error('backslashreplace', backslashreplace)
```
pipermerriam: Does anything bad happen if this happens to be called twice?

carver: Probably just a performance penalty. I would guess that the error handling in the default implementation is pretty optimized.

web3/contract.py
carver: Ok, I took another swing at it in the functional style. I'm roughly familiar with the principles, but it doesn't come naturally to me yet. LMK what you think.

carver: (updated in #312)

carver:
I'm going to close this one to avoid confusion and leave #312 open. I'll reopen this one, rebased, when 312 gets merged.

PR #321, opened 2017-09-20
Changes to web.eth.estimateGas

### LINKED ISSUE #320: Changes to web3.eth.estimateGas
Check whether 'from' parameter is already in transaction and use it. Otherwise use defaultAccount.
Basically to implement the same behavior as for 'sendTransaction'.

bashalex:
Closes #320

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -255,7 +255,7 @@ class Eth(Module):
 
     def estimateGas(self, transaction):
         # TODO: move to middleware
-        if is_address(self.defaultAccount):
+        if 'from' not in transaction and is_address(self.defaultAccount):
             transaction = assoc(transaction, 'from', self.defaultAccount)
 
         return self.web3.manager.request_blocking(


carver:
Welcome to the repo!

The functionality here makes sense. The preferred fix IMO would be to add a middleware that does this for all requests that contain a transaction (as mentioned in the `#TODO`).

Is that something you'd be interested in doing? The contribution would definitely be appreciated.

If so, that would involve adding a new middleware in `web3/middlewares`, like `web3/middleware/formatting.py` and then adding it to [the default list](https://github.com/pipermerriam/web3.py/blob/master/web3/manager.py#L30). It would probably need to have a whitelist of methods where it tries to apply the default `from` address. Some [middleware docs](http://web3py.readthedocs.io/en/latest/internals.html#middlewares) could be useful.

bashalex:
@carver Thanks for the suggestion. I think I will try to implement it in the described way as soon as I have some free time.

pipermerriam:
In the meantime @carver I think we should merge this since it's the expected behavior.

Issue #323, opened 2017-09-22
Implement the upcoming web3.js `web3.eth.accounts` API


carver:
### What was wrong?

There is no good API for local key manipulation and signing.

### How can it be fixed?

The new [eth_keys](https://github.com/pipermerriam/ethereum-keys) package should be able to handle most of the implementation of actually manipulating keys and signing with them.

The proposed v1 API of [`web3.eth.accounts`](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html) in web3.js seems reasonable. Let's implement it.

Out of scope for this particular issue (but probably reasonable to consider in the future):

* [`accounts.wallet.*`](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#wallet) - "Contains an in memory wallet with multiple accounts"

pipermerriam:
The `accounts.encrypt` and `accounts.decrypt` code is already written in [`eth_keyfile`](https://pypi.python.org/pypi/ethereum-keyfile).  You just have to hook in to the existing functions.

carver:
Nice, I'll put those in scope then.

carver:
Merged into master

PR #324, opened 2017-09-25
eth.accounts API

carver:
### What was wrong?

See #323 

### How was it fixed?

Added a new Account sub-module to Eth. Using the eth_keys and eth_keyfile packages for underlying signing and keyfile generation.

Some hacks to get py2 working (for examlpe, py3 can use bytes or hex str, but py2 can only use hexstr)

This checklist is crossed off when there are sufficient first-pass tests, and they are green:

- [x] pass transaction signature from eth_test
- [x] use the curried hexstr_if_str & text_if_str
- [x] create()
- [x] create().address
- [x] create().privateKey
- [x] create().signTransaction
- [x] create().sign
- [x] create().privateKey
- [x] create().encrypt
- [x] privateKeyToAccount
- [x] signTransaction
- [x] recoverTransaction
- [x] hashMessage
- [x] sign
- [x] recover
- [x] encrypt
- [x] decrypt
- [x] accept hexstr for private key arguments (depends on #332 )
- [x] docs

#### Cute Animal Picture

Cute animal picture


web3/account.py
```diff
@@ -0,0 +1,75 @@
...
+        key.address = key.public_key.to_checksum_address()
+        key.privateKey = key
+        key.sign = compose(sign, signature_wrapper, to_bytes)
+        key.signTransaction = compose(sign, to_bytes)
+        return key
+    return wrapper
+
+
+class Account(Module):
+    _keys = KeyAPI()
```
pipermerriam: This has been updated so that you can just do `from eth_keys import keys` and access all the apis from the `keys` instance.

web3/account.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def create(self, extra_entropy=''):
+        extra_key_bytes = to_bytes(text=extra_entropy)
+        key_bytes = keccak(os.urandom(32) + extra_key_bytes)
+        return self.privateKeyToAccount(key_bytes)
+
+    @return_key_api
+    def privateKeyToAccount(self, primitive=None, hexstr=None):
+        key_bytes = to_bytes(primitive, hexstr=hexstr)
+        if len(key_bytes) != 32:
```
pipermerriam: Instead of this you can just do:

```python
from eth_keys.exceptions import ValidationError

try:
    return self._keys.PrivateKey(key_bytes)
except ValidationError:
    raise ...
```

web3/utils/encoding.py
```diff
@@ -175,7 +175,10 @@ def to_decimal(value=None, hexstr=None, text=None):
             ))
             return to_decimal(hexstr=value)
         else:
-            return int(value)
+            try:
+                return int(value)
+            except ValueError:
+                return to_decimal(hexstr=to_hex(value))
```
pipermerriam: These strike me as something that should have a corresponding test case for whatever failure they are handling?

setup.py
```diff
@@ -17,6 +17,7 @@
 install_requires = [
     "cytoolz>=0.8.2",
     "ethereum-abi-utils>=0.4.1",
+    "ethereum-keys>=0.1.0a5",
```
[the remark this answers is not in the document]
carver: Ahh, the bleeding edge. :)

web3/utils/encoding.py
carver: done

web3/account.py
```diff
@@ -0,0 +1,189 @@
...
+                sys.version_info.major < 3 and isinstance(private_key, unicode)  # noqa: F821
+            ):
+            key_bytes = to_bytes(hexstr=private_key)
+        else:
+            key_bytes = to_bytes(private_key)
+        account = self.privateKeyToAccount(key_bytes)
+
+        # detect nonce for this account
+        if 'nonce' not in transaction_dict:
+            transaction_dict['nonce'] = self.web3.eth.getTransactionCount(account.address)
```
pipermerriam: This is an acceptable solution for initial implementation but it will fail anytime two transactions are signed before the 1st transaction has been mined since `getTransactionCount` will return the same value.

Eventually we will need to perform local tracking on the nonce value, taking into account any transactions we've signed locally and broadcast.

web3/module.py
```diff
@@ -3,3 +3,24 @@ class Module(object):
 
     def __init__(self, web3):
         self.web3 = web3
+
+    @classmethod
+    def attach(cls, target, module_name=None):
```
pipermerriam: This looks like it's extractable to be merged independently.

web3/utils/signing.py
```diff
@@ -9,3 +16,38 @@ def signature_wrapper(message, version=b'E'):
...
+    if above_id_offset < 0:
+        return (None, raw_v)
+    else:
+        (chain_id, v_bit) = divmod(above_id_offset, 2)
+        return (chain_id, v_bit + V_OFFSET)
+
+
+def sign_transaction_hash(account, transaction_hash, chain_id):
+    signature = account.sign_msg_hash(transaction_hash)
+    (v_raw, r, s) = (getattr(signature, part) for part in 'vrs')
```
pipermerriam: I believe you can just access `signature.vrs` making this:

```python
(v_raw, r, s) = signature.vrs
```

carver: :man_facepalming: thanks, I missed that attribute

web3/account.py
carver: Yup, agreed. It looks like web3.js is punting on internal nonce management too... at least according to their docs.

docs/web3.eth.account.rst
```diff
@@ -0,0 +1,362 @@
...
+
+        >>> Web3.toHex(text=msg)
+        '0x49e299a55346'
+        >>> w3.eth.account.sign(message_hexstr='0x49e299a55346', private_key=key)
+        >>> w3.eth.account.sign(0x49e299a55346, key)
+
+
+.. py:method:: Account.recoverMessage(data=None, hexstr=None, text=None, vrs=None, signature=None)
+
+    Get the address of the account that signed the message with the given hash.
```
pipermerriam: copy/paste

Looks like this should be *"Get the address of the account that signed the given message"*.

docs/web3.eth.account.rst
```diff
@@ -0,0 +1,362 @@
...
+
+        >>> msg = "I♥SF"
+        >>> w3.eth.account.hashMessage(text=msg)
+        '0x1476abb745d423bf09273f1afd887d951181d25adc66c4834a70491911b7f750'
+
+
+.. py:method:: Account.signTransaction(transaction_dict, private_key)
+
+    Sign a transaction using a local private key. Produces signature details
+    and a rawTransaction that you can broadcast.
```
pipermerriam: Maybe better? 

*"and the hex encoded transaction suitable for broadcast using `web3.eth.sendRawTransaction`."*

setup.py
```diff
@@ -17,6 +17,8 @@
 install_requires = [
     "cytoolz>=0.8.2",
     "ethereum-abi-utils>=0.4.3",
+    "ethereum-keyfile>=0.3.0",
+    "ethereum-keys>=0.1.0-alpha.7",
```
pipermerriam: This should get moved to a beta....  will try to carve time to do this tomorrow.  We shouldn't release features depending on alpha libraries right (even if they are our own)?

tests/core/eth-module/test_accounts.py
```diff
@@ -0,0 +1,276 @@
...
+    msg_hash = to_hex_if_py2(msg_hash)
+    signature = to_hex_if_py2(signature_bytes)
+    from_account = web3.eth.account.recover(msg_hash, signature=signature)
+    assert from_account == '0xFeC2079e80465cc8C687fFF9EE6386ca447aFec4'
+
+
+def test_eth_account_recover_vrs(web3):
+    v, r, s = (
+        27,
+        5634810156301565519126305729385531885322755941350706789683031279718535704513,
```
pipermerriam: Wondering if it would be good to source these test fixtures as well as maybe a small sample script for generating them from a known good implementation (*cough* `coincurve` *cough*).

This is more about demonstrating due-diligence than anything else.

web3/account.py
```diff
@@ -0,0 +1,194 @@
...
+    def create(self, extra_entropy=''):
+        extra_key_bytes = text_if_str(to_bytes, extra_entropy)
+        key_bytes = keccak(os.urandom(32) + extra_key_bytes)
+        if sys.version_info.major < 3:
+            key_bytes = to_hex(key_bytes)
+        return self.privateKeyToAccount(key_bytes)
+
+    def decrypt(self, keyfile_json, password):
+        if is_string(keyfile_json):
+            keyfile = json.loads(keyfile_json)
```
pipermerriam: Feels like this shouldn't throw a decode error if I pass in either a random string or bad JSON.  It'd be nice to issue a better error if decoding fails.

pipermerriam: Other note.  `is_string` will return `True` for a `bytes` value but I believe `json.loads` will fail in python 3 if you pass it a byte string.

web3/account.py
```diff
@@ -0,0 +1,194 @@
...
+            In Python 2, a bytes, unicode or str object will be interpreted as hexstr
+            In Python 3, only a str object will be interpreted as hexstr
+        '''
+        msg_bytes = to_bytes(message, hexstr=message_hexstr, text=message_text)
+        msg_hash = self.hashMessage(msg_bytes)
+        key_bytes = hexstr_if_str(to_bytes, private_key)
+        key = self._keys.PrivateKey(key_bytes)
+        (v, r, s, eth_signature_bytes) = sign_message_hash(key, msg_hash)
+        (r_hex, s_hex, eth_signature_hex) = map(to_hex, (r, s, eth_signature_bytes))
+        return AttributeDict({
```
pipermerriam: This makes me think about whether the accounts API should be done using a middleware that listens for non-standard JSON-RPC calls (`accounts_sign`, `accounts_recover`).

Do you have thoughts on this?  Is it weird to you that all of the other API's can be mucked with by middlewares, but this one stands separate?

web3/utils/signing.py
```diff
@@ -9,3 +43,101 @@ def signature_wrapper(message, version=b'E'):
         return preamble + size + message
     else:
         raise NotImplementedError("Only the 'Ethereum Signed Message' preamble is supported")
+
+
+def annotate_transaction_with_chain_id(txn_obj):
```
pipermerriam: This function name is confusing because it doesn't match what it seems like the function is doing.  I'd suggest renaming to something like:

* `get_transaction_obj_for_signing`
* `get_unsigned_transaction`

web3/utils/signing.py
```diff
@@ -9,3 +43,101 @@ def signature_wrapper(message, version=b'E'):
...
+        return self._publicapi.sign(*args, **kwargs)
+
+    def signTransaction(self, transaction_dict):
+        return self._publicapi.signTransaction(transaction_dict, self.privateKey)
+
+    def __bytes__(self):
+        return self.privateKey
+
+    # Python 2 support
+    def __str__(self):
```
pipermerriam: How does python3 feel about this?  In python3 the expected return type of `str(priv_key_obj)` would be a text string, but this would return bytes right?

web3/account.py
carver: If I passed an invalid json string into the parameter `keyfile_json`, I would expect a JSONDecodeError describing where in the string it failed to decode the json. What do you think would be more helpful to say here?

Ah, thanks. Passing in `bytes` would be an odd thing to try, but I'll check for it.

web3/account.py
carver: The attrdict middleware is a cute way to handle this line, but I don't think other middlewares buy us anything (`pythonic` -- neutral, `stalecheck` -- negative, others ??). It also introduces a risk if the user removes the default middlewares: in some cases, I think web3 would send their private key to the provider(s) over http. *yikes*

`web3.eth.account` does stand out as different, since it doesn't call the client. I think it **should** stand out as different. We want to be absolutely clear with users that this module won't send your private key anywhere.

In fact, I'm starting to think we should move the default prefills out of `Account.signTransaction()`. That is the only method that is currently making provider calls. Then we can simply say the `eth.account` module has absolutely no calls to the provider.

*Downside:*

 * It becomes slightly less convenient, when signing a transaction without all the details
 * It becomes slightly less API-equivalent with web3.js
   * We would invent some new function like `w3.eth.fillTransactionDefaults({....})`
   * `Account.signTransaction()` would no longer accept a transaction missing any fields

*Upside:*

 * It's much clearer to say "`web3.eth.account` never makes client calls"
 * We could even strip out the `web3` attribute from `Account` to be as clear as possible, and avoid future changes that accidentally re-introduce client calls

Thoughts?

web3/utils/signing.py
carver: I fought with the name for long enough that I just decided to change the abstraction boundary.

It now returns the hash of the signed transaction. It's called `hash_of_signed_transaction()`.

web3/utils/signing.py
carver: Probably pretty grumpy. Fixed.

tests/core/eth-module/test_accounts.py
carver: Fair enough. Off to learn me some `coincurve`...

web3/account.py
pipermerriam: > What do you think would be more helpful to say here?

Maybe just the text of the `JSONDecodeError` with a *header* Something like:

```
Error parsing keyfile JSON:

<JSONDecodeError-text-here>
```

web3/account.py
pipermerriam: > I think web3 would send their private key to the provider(s) over http. yikes

This line sold me. I'm now firmly in the camp that this shouldn't be mock RPC'd

pipermerriam: > In fact, I'm starting to think we should move the default prefills out of `Account.signTransaction()`

👍

pipermerriam: > `Account.signTransaction()` would no longer accept a transaction missing any fields

I think that rather than a global `w3.eth.fillTransactionDefaults({....})` we should instead implement localized APIs on `web3.eth`, `web3.eth.accounts`, The `account` object that comes back from `web3.eth.accounts`, and `contract`.  These would all be the same API.  Something like

`web3.eth.setTransactionDefaults({...})`.  This would replace `web3.eth.defaultAccount` (or that could just be a shortcut for `web3.eth.setTransactionDefaults({'from': '0x1234...'})`.

So *everything* that sends transactions or creates transactions get's the same API and we avoid a global set of transaction defaults because I don't actually think that makes sense.

pipermerriam:
:shipit:

web3/account.py
carver: > we should instead implement localized APIs on `web3.eth`, `web3.eth.accounts`, The `account` object that comes back from `web3.eth.accounts`, and `contract`.

I don't fully understand the proposal. You're saying we'd have a `web3.eth.accounts.create().setTransactionDefaults()` too?

One thing that might be obscuring the issue is that many of the defaults aren't static, but dynamic from the client. To highlight the "defaults":
```
TRANSACTION_DEFAULTS = {
    'gasPrice': lambda web3: web3.eth.gasPrice,
    'value': lambda web3: 0,
    'data': lambda web3: b'',
    'chainId': lambda web3: int(web3.net.version),
}
# implemented elsewhere:
nonce = web3.eth.getTransactionCount(account.address)
```

So having those get set anywhere in the `web3.eth.account` module (including the `LocalAccount` object returned by a `web3.eth.account.create()`) would involve a client call, which I'd love to prohibit. Although, `value` and `data` could still get defaulted in `web3.eth.account`.

I *think* the rest are globally default-able. `nonce` is the biggest question mark.

carver:
I added [two transaction signing tests from ethereum/tests](https://github.com/ethereum/tests/blob/develop/BasicTests/txtest.json), which inspired me to support chain-naive transactions. It passes the first now, but still fails the second. The only apparently relevant difference in the second is that it has no `to` field and has a `data` field. I'm stalled on debugging it, at the moment.

In something that seems very weird to me: `r` matches exactly, but `s` and `v` differ. The unsigned parts of the transaction seem to be equal.

```
>       assert signed.rawTransaction == add_0x_prefix(expected_raw_txn)
E       AssertionError: assert '0xf87f8085e8...dc6e341c11426' == '0xf87f8085e8d...497a98e752d1b'
E         - 0xf87f8085e8d4a510008227108080af6025515b525b600a37f260003556601b596020356000355760015b525b54602052f260255860005b525b54602052f21ca05afed0244d0da90b67cf8979b0f246432a5112c0d31e8d5eedd2bc17b171c694a044efca37cb9883d1ee7a47236f3592df152931a930566933de2dc6e341c11426
E         + 0xf87f8085e8d4a510008227108080af6025515b525b600a37f260003556601b596020356000355760015b525b54602052f260255860005b525b54602052f21ba05afed0244d0da90b67cf8979b0f246432a5112c0d31e8d5eedd2bc17b171c694a0bb1035c834677c2e1185b8dc90ca6d1fa585ab3d7ef23707e1a497a98e752d1b

# web3's signed r, s, v:
r          = '0x5afed0244d0da90b67cf8979b0f246432a5112c0d31e8d5eedd2bc17b171c694'
s          = '0x44efca37cb9883d1ee7a47236f3592df152931a930566933de2dc6e341c11426'
v          = 28

# expected r, s, v:
r          = '0x5afed0244d0da90b67cf8979b0f246432a5112c0d31e8d5eedd2bc17b171c694'
s          = '0xbb1035c834677c2e1185b8dc90ca6d1fa585ab3d7ef23707e1a497a98e752d1b'
v          = 27
```

carver:
I read these lines as saying that the msg_hash and the private key are the same in the web3 implementation and the test data:
https://github.com/pipermerriam/ethereum-keys/blob/master/eth_keys/backends/native/ecdsa.py#L78-L80

carver:
eth_keys with native and coincurve backends both produce the same `vrs`, which doesn't match the test result. (except for `r`, which still seems to suggest that both the message hash and private key are being provided the same value.)

carver:
Decided w/ @pipermerriam that the signature is non-deterministic. They both recover to the right sender. Now it only evaluates recovery, and the `r` value. Will merge to master once the tests pass.

PR #331, opened 2017-09-29
Piper/integration test improvements

pipermerriam:
### What was wrong?

Deficiencies in the integration tests:

* Needed to test against go-ethereum `v1.7.0`
* Needed to test log entry serialization
* Writing integration tests required institutional knowledge about what fixtures were required.
* Lots of fields from ethereum tester were not being properly remapped to their web3 equivalents (`gas_price > gasPrice`)

### How was it fixed?

* Added tests to the integration suite for log entries.
* Wrote a script that can be used to generate the go-ethereum datadir fixture.
* generated a `v1.7.0` geth datadir fixture.
* added `tests/integration/README.md` to document how to write a new integration test.
* expanded the `ethereum-tester` middleware to include remappers for all of the missing fields.

#### Cute Animal Picture

TODO


pipermerriam:
depends on https://github.com/pipermerriam/py-geth/pull/36

.gitignore
```diff
@@ -106,3 +106,4 @@ crashlytics.properties
 crashlytics-build.properties
 fabric.properties
 
+!tests/integration/**.log
```
pipermerriam: I don't think this actually works.  removing.

setup.py
```diff
@@ -23,6 +23,7 @@
     "requests>=2.12.4",
     "rlp>=0.4.7",
     "toolz>=0.8.2",
+    "ethereum-tester==0.1.0-beta.1",
```
pipermerriam: Thoughts on this?  It doesn't add any system dependencies and is pure python.  The `web3.providers.eth_tester.EthereumTesterProvider` requires it.

Issue #329, opened 2017-09-28
Unified approach to string formatting.


pipermerriam:
* Version: 3.15.0
* Python: 2.7/3.4/3.5

### What was wrong?

We need a well though through string handling solution for `bytesXX`, `bytes` and `string` types being passed into functions as arguments.

Currently, users expect to be able to provide hex encoded strings to some of these methods which either fails due to these methods expecting bytestrings, *or* it appears to succeed when in fact it has sent the provided hex-string as-is.

### How can it be fixed?

For `web3.sha3` and `web3.eth.sign` we have a workable solution.  See https://github.com/pipermerriam/web3.py/issues/289

For `bytesXX` we should be able to be *smart* and detect hex encoded values.

* If the value is a byte string, validate `len(v) == XX` where `XX` is the `bytesXX` length.
* If the value is a text string, assume hex encoding.  validate `len(v_without_0x_prefix) == 2 * XX` where `XX` is the `bytesXX` length. 

For `bytes` types.

* If the value is a byte string, pass it through unmodified.
* If the value is a text string, validate that it is a valid hexidecimal encoded value.

For `string` types.  

* If the value is a byte string, pass it through unmodified.
* If the value is a text string, decode it as `utf8` into it's bytes representation.

carver:
> 1. if the type of the value is bytes, pass it through as-is. This will allow certain silent failures in python2 but we're about to remove support so I'm ok with sweeping that under the rug.

Absolutely on board

> 2. if the type of the value is text, it is required to be valid hexidecimal.

Mostly on board.

So lets look at all these cases for Python 3: (am I missing any critical ones for this discussion?)

* python type of an argument to a contract function call -->
  * ABI type for the matching function call
* `bytes`
  * `bytes`: pass through
  * `bytesNN`: pass through, after validating the length
  * `string`: pass through, after validating that the bytes can be UTF-8 decoded, else raise a validation error.
* `str`
  * `bytes`: `to_bytes(hexstr=arg)`
  * `bytesNN`: `to_bytes(hexstr=arg)`, and validate length
  * `string`: `to_bytes(hexstr=arg)`

Only the last bullet is problematic.

I would be very surprised for my `str` in python to be treated like hex when being sent to a `string` ABI type. Instead, I would expect:

* `str`
  * `string`: `to_bytes(text=arg)` aka `arg.encode('utf-8')`

[...]

PR #332, opened 2017-09-30
non_hexstr() & non_text(): inferred string conversions

carver:
### What was wrong?

In eth.accounts PR #324 -- I wanted to have the user pass `bytes` or a hex`str` for a private key. Sending text (a unicode `str`) here would be non-sensical: you can't cover the whole private key space with utf-8 encoded strings.

There was no way to do this with a single clean function call (esp after dealing with py2).

Also, #329 is a proposal to do something like this with contract arguments. So a general utility function seems useful.

### How was it fixed?

I added two methods: `text_if_str()` and `hexstr_if_str()`. They work like so:

**In Python 3**

`text_if_str(to_bytes, 'ö')` will automatically infer utf8, giving `b'm\xc3\xb6'`
`text_if_str(to_bytes, '0xab')` will still infer utf8, giving `b'0xab'` (NOT `b'\xab'`)
`text_if_str(to_bytes, b'\xff')` will pass through `b'\xff'`

Any other type passes through as a primitive, like `text_if_str(to_bytes, 3) == to_bytes(3)`

`hexstr_if_str(to_bytes, 'ö')` will raise a TypeError
`hexstr_if_str(to_bytes, '0xab')` will decode the hex to b'\xab'
`hexstr_if_str(to_bytes, b'0xab')` will pass through the bytes `b'0xab'`
`hexstr_if_str(to_bytes, b'\xab')` will pass through `b'\xab'`

Any other type passes through as a primitive, like `hexstr_if_str(to_bytes, 3) == to_bytes(3)`

**In Python 2** - which gets much more complicated (and will only last till Q1 :tada: )

`text_if_str(to_bytes, u'ö')` will automatically infer utf8, giving b'm\xc3\xb6'
`text_if_str(to_bytes, u'0xab')` will still infer utf8, giving b'0xab'
`text_if_str(to_bytes, b'\xff')` will pass through `b'\xff'`
`text_if_str(to_bytes, '\xff')` will pass through `b'\xff'`

_If you do something really unfortunate, like ask for unicode point 00ff when you think you're asking for byte b'\xff', you'll get a surprise:_
`text_if_str(to_bytes, u'\xff')` will decode into `b'\xc3\xbf'`
_^ Just don't do this, and switch to py3 to avoid the issue altogether. We can't fix this, because the user might actually want to encode the character `'ÿ'` into bytes using utf-8_

Any other type passes through as a primitive, like `text_if_str(to_bytes, 3) == to_bytes(3)`

Non-text will greedily assume hex, for unicode, str, and bytes:
`hexstr_if_str(to_bytes, 'ö')` will raise a TypeError
`hexstr_if_str(to_bytes, 'g')` will raise a TypeError
`hexstr_if_str(to_bytes, b'g')` will raise a TypeError
`hexstr_if_str(to_bytes, u'0xab')` will generate `b'\xab'`
`hexstr_if_str(to_bytes, '0xab')` will generate `b'\xab'`
`hexstr_if_str(to_bytes, b'0xab')` will generate `b'\xab'`

Any other type passes through as a primitive, like `hexstr_if_str(to_bytes, 3) == to_bytes(3)`

#### Cute Animal Picture

Cute animal picture


pipermerriam:
After a very brief glance, this looks like the type of function I *might* test with hypothesis (if you happen to be looking for a reason to learn your way around the library).  Seems like there are some pretty clear *properties* that can be expressed about the inputs and outputs of these functions.

tests/core/utilities/test_encoding.py
```diff
@@ -80,3 +97,122 @@ def test_hex_encode_abi_type(abi_type, value, expected):
...
+
+@pytest.mark.skipif(sys.version_info.major > 2, reason="these test values only valid for py2")
+@given(
+    st.one_of(st.text(), st.binary()),
+    st.sampled_from((to_bytes, to_hex, to_decimal)),
+)
+def test_hexstr_if_str_invalid_hex_py2(val, converter):
+    try:
+        is_hexstr = (is_hex(val) or val == '')
+    except ValueError:
```
pipermerriam: *curious*

What's throwing the `ValueError` here?  Something inside `is_hex`?  If so, I think that should be fixed upstream.

tests/core/utilities/test_encoding.py
```diff
@@ -80,3 +97,122 @@ def test_hex_encode_abi_type(abi_type, value, expected):
...
+    assert to_type.call_args == ((val, ), {'hexstr': None})
+
+
+@pytest.mark.skipif(sys.version_info.major < 3, reason="these test values only valid for py3")
+@given(st.from_regex(HEX_REGEX))
+@example('0x')
+@example('0')
+def test_hexstr_if_str_on_valid_hex(val):
+    to_type = Mock(return_value='zoot')
+    assert hexstr_if_str(to_type, val) == 'zoot'
```
pipermerriam: Are you sure this is true even for small values of `'zoot'`?  What about as `'zoot'` approaches infinity?  Can you link me to the `'zoot'` documentation so that I can read more about this `'zoot'` thing?

🤡

tests/core/utilities/test_encoding.py
```diff
@@ -80,3 +97,122 @@ def test_hex_encode_abi_type(abi_type, value, expected):
...
+@pytest.mark.skipif(sys.version_info.major < 3, reason="these test values only valid for py3")
+@given(st.one_of(st.integers(), st.booleans(), st.binary()))
+@example(b'')
+def test_hexstr_if_str_passthrough(val):
+    to_type = Mock(return_value='zoot')
+    assert hexstr_if_str(to_type, val) == 'zoot'
+    assert to_type.call_args == ((val, ), {'hexstr': None})
+
+
+@pytest.mark.skipif(sys.version_info.major < 3, reason="these test values only valid for py3")
```
pipermerriam: I think you should declare this at the top and then reuse the marker in each test.

```python
only_python3 = pytest.mark.skipif(sys.version_info.major < 3, reason="these test values only valid for py3")

@only_python3
def test_thing():
    ...
```

tests/core/utilities/test_encoding.py
carver: https://github.com/pipermerriam/ethereum-utils/pull/22

tests/core/utilities/test_encoding.py
carver: [Docs](https://getyarn.io/yarn-clip/c99f2aa3-9ec4-46fb-8488-b8d6f67bba01)

Issue #336, opened 2017-10-09
Logistics for v4 Development


carver:
Let's create a plan for developing and releasing version 4.

Proposal: we create a v3 branch for backporting important bugfixes, then implement the planned breaking changes on master for release as v4.

Pull Requests:

 * v4-only changes belong in a PR to master
 * v3 & v4 changes go in a PR to master for review. After green-lit, open another PR to `v3` branch, for insta-merge after tests pass.

Timeline:

 * ~October:
   * Create v3 branch
   * Deprecation warning in v3
   * Start merging in breaking changes to master
 * ~November:
   * Finish breaking change merges to master, then release as v4.0.0-beta
   * Backport any important bugfixes to v3
 * ~December:
   * Normal development progresses on master, delete cruft for backwards-compatible support
   * Release master as v4.0.0-stable
 * ~January: Stop backporting anything but critical security fixes to v3 ([no earlier](https://medium.com/@pipermerriam/dropping-python-2-support-d781e7b48160))
 * ~April: Stop all backporting and support for v3

Prefer to get all breaking changes in before v4-beta, but they _must_ all go in before v4-stable.

To address a couple in-flight projects:

 * steps 1-2 of #294 can go in v3, and step 3 can go in master (v4)
 * #324 is a bit trickier: it would be nice to release in v3, but it depends on an alpha package. Maybe it's best to just go in master, to avoid maintaining a v3-beta branch.

pipermerriam:
This looks solid to me.  Might be good to create some labels for categorizing issues and pull requests.

carver:
v4 launched Apr 2.

Issue #300, opened 2017-09-13
Deprecation warning for Py2


carver:
* Version: 3.*
* Python: 2.7
* OS: osx/linux/win

### What was wrong?

Many functions would greatly benefit from being able to differentiate `bytes` and `str`. Hacks and workarounds are all around the system. We should start warning users now so they can prepare their Py3 stacks.

### How can it be fixed?

Add a `DeprecationWarning` if `sys.version_info.major < 3`.

djrtwo:
Excellent.

PR #337, opened 2017-10-09
Warn that py2 is deprecated, plus guidance in docs

carver:
### What was wrong?

#300 Needed to warn about py2 deprecation.

### How was it fixed?

Added deprecation warning, plus an environment note in the docs

Also, trying out merging to the v3 branch, a la #336 

For backported items that belong in both versions, we can manually merge to master after doing a PR to v3. Any better idea?

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -4,6 +4,22 @@ Quickstart
 .. contents:: :local:
 
 
+Environment
+------------
+
+Web3.py prefers Python 3, and will soon require it. Often, the
+best way to guarantee a clean Python 3 environment is with ``virtualenv``, like:
+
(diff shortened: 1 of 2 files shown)


pipermerriam:
> For backported items that belong in both versions, we can manually merge to master after doing a PR to v3. Any better idea?

* PR into master.
* Once merged and happy create v3 PR that only contains the new commits (merge when green).

I think we need to PR both so that we can see that tests pass on both branches since they will slowly diverge.

carver:
Sounds good, added that process to #336 .

A dependency install failed on one of the tests. Just waiting for it to re-run then I'll merge.

PR #338, opened 2017-10-09
[v4] Warn that py2 is deprecated, plus guidance in docs

carver:
Dup #337


carver:
Something weird is going on. I remember this showing green earlier (also, Travis seems to only be running one environment at a time, so **very** slowly). Restarting the build...

Issue #166, opened 2017-02-27
Windows CI


pipermerriam:
Add CI for windows environments via https://www.appveyor.com/

carver:
> There was an error while creating new AppVeyor account. Please contact AppVeyor support.

:( waiting to hear back...

carver:
Status update:

Build is kicking off now, but failing to install requirements-dev: https://ci.appveyor.com/project/JasonCarver/web3-py

pipermerriam:
This *might* help: https://stackoverflow.com/questions/1710922/how-to-install-pkg-config-in-windows

pipermerriam:
Alternatively, it might be time to drop `eth-testrpc` (and the underlying `ethereum` requirement).  Requires building the JSON-RPC webserver component for eth-tester... but probably worth it.

carver:
The `pyethereum` gone, the next roadblock is `pyethash` (used by `py-evm`).

https://ci.appveyor.com/project/JasonCarver/web3-py/build/1.0.4

Others have faced this issue, esp on Windows. See https://github.com/ConsenSys/ethjsonrpc/issues/5

carver:
It doesn't look plausible to run the test suite on windows anytime soon. Removing from the v4 stable milestone.

voith:
Travis now supports windows: https://blog.travis-ci.com/2018-10-11-windows-early-release?utm_source=intercom&utm_medium=web&utm_campaign=windows_early_release

fselmo:
closed by #1907

PR #344, opened 2017-10-10
Auto web3

Sebohe:
### What was wrong?

This request provides a solution for: https://github.com/pipermerriam/web3.py/issues/302

### How was it fixed?

The auto module configures a web3 instance instance for you. If you don't specify which provider you want it will default on IPC first, then HTTP.

#### Cute Animal Picture
Cute animal picture
(I'm starting to hoard cute animal pictures on my phone)


web3/auto/__init__.py
```diff
@@ -0,0 +1,7 @@
+from web3.auto import ipc
+from web3.auto import http
+
+w3 = ipc.ipc()
+
+if not w3:
+    w3 = http.http()
```
carver: Right now, this is opening more connections than it needs to:

1. It connects to http even if ipc is available (on module import of http)
2. It re-attempts the connection that was already tested during module import

You could get around both of those issues by switching to a dynamic import and using the w3 already defined, looking something like:

```
for connector in ('ipc', 'http'):
    connection = importlib.import_module('web3.auto.%s' % connector)
    if connection.w3:
        ...
```

---

It could be nice to give these modules a consistent interface, something like `ipc.connect()` and `http.connect()`. Not critical, though.

web3/auto/ipc.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+def ipc():
+
+    home = path.expanduser('~')
+
+    # The following 2 lines are default paths
+    parity_path = home + '/.local/share/io.parity.ethereum/jsonrpc.ipc'
+    geth_path = home + '/.ethereum/geth.ipc'
```
carver: It doesn't look like this will cover the major operating systems, for geth. Take a look at how the IPC module does it: https://github.com/pipermerriam/web3.py/blob/master/web3/providers/ipc.py#L38-L65

In fact, it might be better to just initialize `IPCProvider()` without any args, which will automatically look for the geth file.

---

Nice that it looks for the parity path! Can you add in support for parity on other OS's, too?

web3/auto/http.py
```diff
@@ -0,0 +1,16 @@
+from web3 import (
+    HTTPProvider,
+    Web3
+)
+
+
+def http():
+
+    w3 = Web3(HTTPProvider('http://localhost:8545'))
```
pipermerriam: It's nice to be able to configure this type of thing using environment variables.  This can be done something like the following.

```python
w3 = Web3(HTTPProvider(
    os.environ.get('WEB3_HTTP_PROVIDER_URI', 'http://localhost:8545')
))
```

This allows for this functionality to be used beyond just `localhost`.

web3/auto/ipc.py
pipermerriam: I would be supportive of just adding the parity paths to the `web3.providers.ipc.get_default_ipc_path` function.  In fact, I think it would be good to modify that function so that rather than just returning a path, it enumerates a number of different paths and checks to see if there is an IPC socket in place that it can connect to and returns the first one.

Order of operations I would like to see on that is:

1. go-ethereum (mainnet)
2. go-ethereum (rinkeby)
3. go-ethereum (ropsten)
4. parity (mainnet)
5. parity (ropsten)
6. parity (kovan)

web3/auto/http.py
```diff
@@ -0,0 +1,16 @@
...
+    if w3.isConnected():
+        return w3
+    else:
+        return False
+
+
+w3 = http()
```
pipermerriam: I'm thinking that it may not be good to instantiate this here since it's done in `web3.auto`

web3/auto/ipc.py
```diff
@@ -0,0 +1,28 @@
...
+        return w3
+
+    w3 = Web3(IPCProvider(geth_path))
+    if w3.isConnected():
+        return w3
+
+    return False
+
+
+w3 = ipc()
```
pipermerriam: Same here.  Maybe remove this instantiation since it's done independently in `web3.auto`?

web3/auto/http.py
carver: This supports the API from the issue:
```
from web3.auto.http import w3
```
Which would give you an autocreated http connection even if an IPC one exists. I suppose we could revisit whether that's necessary. I don't have a strong use case for it, it just seemed like a good idea in the abstract.

carver: Also, this comment should address the double-instantiation issue: https://github.com/pipermerriam/web3.py/pull/344#discussion_r143795348

Sebohe: Thanks for the feedback. I will work on the recommendations.

^.^

web3/auto/__init__.py
Sebohe: I like the `ipc.connect()` and `http.connect()`, makes it clearer.

web3/auto/http.py
Sebohe: My two cents on the issue of whether to use from `web3.auto.http import w3` and  `web3.auto.ipc import w3`.

I don't think people will be running three different ethereum clients, only I was to test this implementation. They will likely be running a single client. I might be wrong on this but it is usually testrpc or parity or geth.

carver: I'm okay with this direction: either you're using a simple, single-client setup and can use `from web3.auto import w3` or you have multiple clients and you need to specify the provider explicitly. That doesn't seem like an unreasonable burden.

web3/auto/__init__.py
```diff
@@ -0,0 +1,9 @@
+import importlib
+
+for connector in ('ipc', 'http'):
+    connection = importlib.import_module('web3.auto.' + connector)
+    if connection.w3:
+        w3 = connection.w3
+
+        if w3.isConnected():
```
carver: This check looks redundant: if w3 is set, then `isConnected()` is True

web3/providers/ipc.py
```diff
@@ -42,22 +42,52 @@ def get_default_ipc_path(testnet=False):
...
+        ))
+        if os.path.exists(ipc_path):
+            return ipc_path
+
     elif sys.platform == 'win32':
-        return "\\\\.\\pipe\\geth.ipc"
+        if os.path.exists("\\\\.\\pipe\\geth.ipc"):
+            return "\\\\.\\pipe\\geth.ipc"
+        elif os.path.exists("\\\\.\\pipe\\jsonrpc.ipc"):
+            return "\\\\.\\pipe\\jsonrpc.ipc"
```
carver: A utility function like `first_available_path(paths)` could maybe remove some of the repetition here.

web3/providers/ipc.py
```diff
@@ -42,22 +42,52 @@ def get_default_ipc_path(testnet=False):
...
+            "share",
+            "io.parity.ethereum",
+            "jsonrpc.ipc"
+        ))
+        if os.path.exists(ipc_path):
+            return ipc_path
+
     elif sys.platform == 'win32':
-        return "\\\\.\\pipe\\geth.ipc"
+        if os.path.exists("\\\\.\\pipe\\geth.ipc"):
```
pipermerriam: I think these should still use `os.path.join` to construct these paths as it will use the proper path separators for the os.

carver: Heh, I've been out of windows for so long that I don't remember what `\\.\pipe\geth.ipc` even means.

@Sebohe because of the '\\' start, it wasn't intuitive to me how to use join here. This was the first way I could find to do the join: `os.path.join('\\\\', '.', 'pipe', 'geth')`

It looks like this is the last change needed to get merged.

web3/providers/ipc.py
```diff
@@ -42,22 +42,65 @@ def get_default_ipc_path(testnet=False):
...
+            "io.parity.ethereum",
+            "jsonrpc.ipc"
         ))
+        if os.path.exists(ipc_path):
+            return ipc_path
+
     elif sys.platform == 'win32':
-        return "\\\\.\\pipe\\geth.ipc"
+        ipc_path = os.path.join(
+            "\\\\",
```
pipermerriam: I so wish we had windows CI to know that this works.... 😢

pipermerriam:
:shipit:

web3/providers/ipc.py
carver: Tagging #166 for motivation.

Sebohe: I tested it on my windows machine if it is some consolation to you. I couldn' get seperated `"\\", "\\"` to work. It would merge them to one `\`. That is why I used 4 backslashes.

Sebohe: I also had geth and parity running while doing my in house testing, but I agree a more thourough test would be comforting.

PR #347, opened 2017-10-10
ethereum-tester requirement fix

Aviah:
### What was wrong?
requirement pinned to a version

### How was it fixed?
== to >=

#### Cute Animal Picture

chic

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ install_requires = [
     "requests>=2.12.4",
     "rlp>=0.4.7",
     "toolz>=0.8.2",
-    "ethereum-tester>=0.1.0-beta.1",
+    "ethereum-tester==0.1.*",
 ]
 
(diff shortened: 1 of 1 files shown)


carver:
Do we need a newer version for something? It is sometimes a pain to track down bugs caused by a newer dependency that has breaking changes, especially when that dependency gets pulled into testing environments silently.

I know we're not consistent about it right now, but especially with a 0.* (and beta) release, my instinct is to pin the version and re-pin it as we upgrade and test locally.

If that doesn't work for some reason, an alternative could be to match against ethereum-tester==0.1.* and promise not to break backwards compatibility without at least a minor version bump.

pipermerriam:
@carver newer version is needed to un-pin `rlp` from an exact version which is causing breakages downstream with populus.

PR #349, opened 2017-10-11
remove python tests for py2 and py3.4

carver:
### What was wrong?

We are unnecessarily testing against old Python versions on v4 (`master` branch), slowing the test process.

### How was it fixed?

Removed the tox environments and travis runs

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,49 +10,33 @@ matrix:
   include:
     # go-ethereum
     - python: "3.5"
-      env: TOX_POSARGS="-e py27-integration-goethereum -e py34-integration-goethereum -e py35-integration-goethereum" GETH_VERSION=v1.6.0
+      env: TOX_POSARGS="-e py35-integration-goethereum" GETH_VERSION=v1.6.0
     - python: "3.5"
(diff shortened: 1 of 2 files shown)


pipermerriam:
To be sure I understand, our plan is to merge stuff like this into `master`.  Along-side that we'll have a `v3` branch that we backport any bugfixes and changes to which will remain py27/34 compatible.

carver:
Right, this would merge only into master, the py27/34 tests will remain on the `v3` branch for backporting.

carver:
Good to go?

pipermerriam:
@carver is there a reason this doesn't also remove the python2 support code?  Are you planning to do that separately?  Good to go either way.

pipermerriam:
Going forward can you assign me as a reviewer.  It sends me a notification which helps me know what needs my attention.

carver:
Right, removing all the py2 support code is still on the to-do list. I just wanted to separate the test removal out from the ENS PR #343 , since it didn't really seem to belong in there (but ENS depends on removing py2 tests).

PR #356, opened 2017-10-19
Reuse IPC Socket

carver:
### What was wrong?

When making many consecutive requests over IPC, they were slow, used a lot of RAM, and eventually crashed Parity. See https://github.com/paritytech/parity/issues/6516

### How was it fixed?

Reuse IPC connection across requests.

Thanks @ifA88 for testing out that the approach has a significant performance boost.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 
 import sys
 import os
-import contextlib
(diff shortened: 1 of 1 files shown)


pipermerriam:
Just realized you might not be done with this... Either way, looks good.

PR #357, opened 2017-10-19
notify gitter on travis build success (1st time)

carver:
Experimenting with posting successful build notifications into gitter. Curious what the signal/noise will look like.

#### Cute Animal Picture

Cute animal picture


carver:
Notification spam seems pretty harmless in gitter. Merging, and will remove again later, if we're unhappy with it.

Issue #340, opened 2017-10-09
Return `bytes` type from methods where appropriate


carver:
* Version: 4
* Python: 3.5+

### What was wrong?

We now have cleaner separation in many method arguments between hex strings and text, but most methods that result in arbitrary bytestrings still return a hex string.

Returning `bytes` instead gives several benefits:

 * As an argument to the next call, `bytes` is unambiguous
 * It's the natural representation
 * Comparing two `bytes` objects is much simpler to do correctly. Hex equality tests can go wrong due to case difference, or `0x` prefix difference.

### How can it be fixed?

Convert these methods (maybe more?) to return `bytes`:

- [x] `Web3.sha3()`
- [x] `Web3.soliditySha3()`
- [x] `Web3.eth.getBlock()` attributes
  - [x] `hash`
  - [x] `extraData`
  - [x] `logsBloom`
  - [x] `mixHash`
  - [x] `parentHash`
  - [x] `sha3Uncles`
  - [x] `stateRoot`
  - [x] `transactions`
  - [x] `transactionsRoot`
  - [x] `uncles`
- [x] `Web3.eth.getTransaction()` attributes
  - [x] `blockHash`
  - [x] `hash`
  - [x] `publicKey`
  - [x] `r`
  - [x] `raw`
  - [x] `s`
  - [x] `standardV` (should be `int`, not `bytes`)
  - [x] `v` (should be `int`, not `bytes`)
- [x] `Web3.eth.getTransactionReceipt()` attributes
  - [x] `blockHash`
  - [x] `logsBloom`
  - [x] `logs[].topics`
  - [x] ~`root`~
  - [x] `transactionHash`
- [x] `Web3.eth.getStorageAt()`
- [x] `Web3.eth.getCode()`
- [x] `Web3.eth.sendTransaction()`
- [x] `Web3.eth.sendRawTransaction()`
- [x] `Web3.eth.sign()`
- [x] Filter changes & logs that result in hex attributes (except addresses)
- [x] Contracts that return ABI type `bytes` or `bytesNN` (Related to #329 ) -- in PR #363 
- [x] `Web3.eth.account.sign()` attributes
  - [x] `messageHash`
  - [x] `r`
  - [x] `s`
  - [x] `signature`
- [x] `Web3.eth.account.signTransaction()` attributes
  - [x] `hash`
  - [x] `r`
  - [x] `rawTransaction`
  - [x] `s`
- [x] `Web3.eth.account.hashMessage()`
[...]

PR #364, opened 2017-10-20
Return bytes in web3.eth.account

carver:
### What was wrong?

`web3.eth.account` is returning hex strings where a bytes-like value would be more appropriate (see #340 for further rationale)

This is a test-run for changing all the values in #340 -- if it looks good, I'll apply this approach to the other methods.

### How was it fixed?

Return bytes instead of hex strings for `sign()`, `signTransaction()`, and `hashMessage()`. Unfortunately, the output was so ugly that I was embarrassed to put it in the docs.

So I also created a new `HexBytes` class that is just like `bytes` (and tests == to `bytes` natively), but has a prettier display.

Key Features:

```py
>>> assert HexBytes('0xaf') == b'\xaf'
>>> HexBytes('0xaf')
HexBytes('0xaf')
```

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -7,6 +7,11 @@ from collections import (
     Sequence,
 )
 
+from web3.utils.encoding import (
+    hexstr_if_str,
+    to_bytes,
+)
+
 from web3.utils.formatters import recursive_map
 
 # Hashable must be immutable:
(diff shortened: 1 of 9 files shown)


web3/utils/datastructures.py
```diff
@@ -168,3 +173,18 @@ def __reversed__(self):
...
         return iter(elements)
+
+
+class HexBytes(bytes):
+    def __new__(cls, val):
+        bytesval = hexstr_if_str(to_bytes, val)
+        return super().__new__(cls, bytesval)
+
+    def __init__(self, val):
+        super().__init__()
```
pipermerriam: *curious*

This *looks* like a passthrough except that it doesn't actually pass `val` through.  Since tests are passing, I suspect this actually works, but I'm curious why (which indicates this could maybe use a small comment explaining why `val` isn't passed through).

carver: Yeah, that was me trying to figure out how to subclass `bytes` and forgetting to remove the cruft. I just removed it.

X) I really like this one: `BytesThatLookLikeHexidecimalButThatYouCanTreatLikeBytes`

PR #366, opened 2017-10-23
add Parity trace functions

jennyailin:
add functions from https://github.com/paritytech/parity/wiki/JSONRPC-trace-module


web3/trace.py
```diff
@@ -0,0 +1,77 @@
+from web3 import formatters
+from web3.utils.empty import (
+    empty,
+)
+
+class Trace(object):
+    def __init__(self, web3):
+        self.web3 = web3
+
```
carver: Using `class Trace(Module):` means that the `__init__` method can be deleted. Yay for less code! (requires importing `from web3.module import Module`)

web3/trace.py
```diff
@@ -0,0 +1,77 @@
...
+    defaultAccount = empty
```
carver: Looks like this isn't used and can be deleted? (which means the import of `empty` can be, too)

web3/trace.py
```diff
@@ -0,0 +1,77 @@
+from web3 import formatters
```
carver: This module is gone.

web3/trace.py
```diff
@@ -0,0 +1,77 @@
...
+
+class Trace(object):
+    def __init__(self, web3):
+        self.web3 = web3
+
+    defaultAccount = empty
+    defaultBlock = "latest"
+
+    def call(self, transaction, types, block_identifier=None):
+        formatted_transaction = formatters.input_transaction_formatter(self, transaction)
```
carver: This function has been refactored away. It looks like adding a two new request formatters for `'trace_call'` to `web3/middleware/pythonic.py` is the right path forward here.

web3/trace.py
```diff
@@ -0,0 +1,77 @@
...
+        if block_identifier is None:
+            block_identifier = self.defaultBlock
+        return self.web3._requestManager.request_blocking(
```
carver: All of the `_requestManager` properties are now just `manager`

allquantor:
Hey @jennyailin do you still think to merge this? Is is an valuable addition IMO.

jennyailin:
@allquantor I do plan on it, but unsure when I can get to it. Hopefully in a few weeks :)

carver:
Looks like #850 took care of most/all of this. Feel free to comment or reopen if any functionality is still missing.  Several of the patterns from this PR made their way into #850. Thank you for the contribution!

Issue #341, opened 2017-10-09
[v4] Remove all deprecated methods


carver:
### What was wrong?

Deprecated methods are kept around for backward compatibility, which isn't necessary in v4.

### How can it be fixed?

All methods that have a deprecation warning should be removed as part of v4.

carver:
All known deprecated usages are removed

PR #369, opened 2017-10-23
Remove deprecated contract attributes and function

carver:
### What was wrong?

#341 `Contract` had a deprecated function and several attributes.

### How was it fixed?

Removed them

#### Cute Animal Picture

The killer bunny is coming for you, deprecated code!

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -4,7 +4,6 @@
 import functools
 import itertools
 import json
-import warnings
 
 from eth_utils import (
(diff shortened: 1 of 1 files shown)


pipermerriam:
I think there are test cases that need to be removed as well.

carver:
I wasn't able to find any mention of the method, `source`, `code`, or `code_runtime`. Plus the tests pass. I think we're good.

PR #372, opened 2017-10-23
Remove deprecated providers and .currentProvider

carver:
### What was wrong?

Some deprecated providers and access mechanisms were floating around #341 

### How was it fixed?

Removed them

#### Cute Animal Picture

That's the most foul, cruel, and bad-tempered rodent you ever set eyes on! 

Cute animal picture


carver:
The removed methods weren't there, but interestingly there was a reference to `Web3.setProvider` which doesn't exist anymore. Updated that bit.

PR #380, opened 2017-10-24
return bytes in 'hash' of getBlock

carver:
### What was wrong?

#340 - 'hash' in getBlock was returning a hex string, let's return a bytes value instead

### How was it fixed?

A small change to `pythonic_middleware` gets this done. The other ramifications are a bit larger.

In many tests, the returned hash gets passed back in as an argument to other json-rpc calls. They were all expecting a hex string value. So this commit also allows bytes as input when retrieving block by hash.

It accomplishes this with an "abi-like" mechanism in `pythonic_middleware` to handle bytes values passed in where a block hash goes (treating it as bytes32).

Finally, had to move `BASE_RETURN_NORMALIZERS` to web3/utils/normalizers.py to avoid an import cycle. (makes more sense there anyway)

#### Cute Animal Picture

Cute animal picture

^ This should totally be our Devcon mascot

### DIFF
--- a/web3/middleware/pythonic.py
+++ b/web3/middleware/pythonic.py
@@ -3,6 +3,9 @@ from __future__ import absolute_import
 import codecs
 import operator
 
+from cytoolz import (
+    curry,
+)
 from cytoolz.curried import (
     keymap,
     valmap,
@@ -25,6 +28,12 @@ from eth_utils import (
     encode_hex,
 )
 
+from web3.utils.abi import (
+    map_abi_data,
+)
+from web3.utils.datastructures import (
+    HexBytes,
+)
 from web3.utils.formatters import (
     apply_formatter_if,
     apply_formatters_to_dict,
@@ -33,6 +42,10 @@ from web3.utils.formatters import (
     apply_one_of_formatters,
     hex_to_integer,
     integer_to_hex,
+)
+from web3.utils.normalizers import (
+    abi_int_to_hex,
+    abi_bytes_to_hex,
 )
 
 from .formatting import (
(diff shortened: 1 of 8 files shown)


web3/middleware/pythonic.py
```diff
@@ -79,6 +92,20 @@ def is_array_of_dicts(value):
...
+@curry
+def to_hexbytes(num_bytes, val):
+    if isinstance(val, str):
+        return HexBytes(val)
+    elif isinstance(val, bytes):
+        if len(val) == num_bytes:
+            return HexBytes(val)
+        else:
+            # some providers return values with hex as bytes, like b'0xEFFF' :(
+            return HexBytes(val.decode('utf-8'))
```
pipermerriam: should this validate that `len(val.decode('utf-8')) == num_bytes`?

web3/middleware/pythonic.py
```diff
@@ -128,7 +155,7 @@ def is_array_of_dicts(value):
     'gasUsed': to_integer_if_hex,
     'size': to_integer_if_hex,
     'timestamp': to_integer_if_hex,
-    'hash': to_ascii_if_bytes,
+    'hash': apply_formatter_if(to_hexbytes(32), is_not_null),
```
pipermerriam: this is surprisingly legible.  I was skeptical when I wrote it.  I think the argument order could be reversed at some point for better legibility

```python
apply_formatter_if(is_not_null, to_hexbytes(32))
```

carver: Yeah, it's grown on me.  I agree about swapping the arguments. When I was first getting familiar with `apply_formatter_if`, my expectation was that the test would come first. (it curries a bit more nicely, too, I think)

PR #395, opened 2017-10-27
Contract argument string & byte handling

carver:
### What was wrong?

The way strings and bytes are handled when passed as arguments to contract functions was surprising, and inconsistent with the rest of web3. #329 

### How was it fixed?

If provided a string, ABI strings are utf-8 encoded, ABI bytes are hex-decoded. Either will accept bytes natively.

The only variation away from the #329 plan is that bytesXX will accept a "too-short" bytes value and pad it appropriately.

#### Cute Animal Picture

Cute animal picture


carver:
I've never found myself happier in practice with the new-style `.format()`. The way I usually use it, it doesn't really seem any more concise or readable (although not necessarily worse).

The [new, new string formatting](https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals) is pretty cool, but I don't think we can require python 3.6 since it's less than a year old.

pipermerriam:
I only really use `.format()` for consistency at this point.  At some point I thought it was *better* but now it's just apples/oranges.  Agreed on both points for new new string formatting.

PR #390, opened 2017-10-26
first argument in apply_formatter_if now boolean, updated codebase

wschwab:
#382 
### What was wrong?

Order of arguments in `apply_formatter_if` needed to be flipped around to put the boolean first

### How was it fixed?

`grep`ped for instances and changed  'em

#### Cute Animal Picture

Cute animal picture

PR #399, opened 2017-10-29
Updated apply_formatter_if

wschwab:
#390 
### What was wrong?

Argument order in `apply_formatter_if` had boolean condition after the formatter.

### How was it fixed?

The `apply_formatter_if` function was rearranged, and the codebase updated. `flake8` tests were run.

#### Cute Animal Picture

Cute animal picture


carver:
I tried locally, and it can be fast-forward merged, so passing the push test should be sufficient (Travis has had problems all day).

Thanks @wschwab for the contribution!

PR #404, opened 2017-10-30
Really trigger travis on wschwab apply formatter

carver:
Travis won't trigger when I commit on top of a remote fork from inside github :(

This PR is just to trigger the push and PR test for #399


carver:
Nope, travis was just being suuuper slow today

PR #411, opened 2017-11-05
Adding implementation for eth_getLogs

jstoxrocky:
### What was wrong?
web3.eth.getLogs had a placeholder NotImplementedError raised when called.

### How was it fixed?
Followed the same pattern as similar requests, calling "eth_getLogs" from the provider.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -289,7 +289,9 @@ class Eth(Module):
         )
 
     def getLogs(self, filter_params):
-        raise NotImplementedError("Not yet implemented")
+        return self.web3.manager.request_blocking(
+            "eth_getLogs", [filter_params],
+        )
 
     def uninstallFilter(self, filter_id):
         return self.web3.manager.request_blocking(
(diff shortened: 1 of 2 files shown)


carver:
I suspect the reason that this wasn't implemented yet is that we would want to expose a better format for filter parameters than the json-rpc does. There are other places in the code that do filter parameter generation, so it might just be a matter of hooking into one of those. It's a part of the library I'm not as familiar with so I don't have better advice than that, right now. @pipermerriam probably has a better idea of the background.

Excellent work with the pug.

jstoxrocky:
Cool! I was working with the [Infura API](https://infura.io/docs/#supported-json-rpc-methods) and found that the getLogs method was the only way that they allow you to query events. They only offer a limited amount of JSON-RPC methods, so I was a little bummed that web3.py had not yet implemented this method. Would be interested in helping get this method added in whatever way you guys think is best.

PR #413, opened 2017-11-09
Resolve HexBytes incompatibility issue on Python 2

wip727:
### What was wrong?
HexBytes would error when calling web3.eth.account.create(), this is an incompatibility issue on Python 2 since "super()" only works for Python 3.

### How was it fixed?
Expand super() to super(HexBytes, cls)

#### Cute Animal Picture
My cookie is missing

### DIFF
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -178,10 +178,10 @@ class NamedElementStack(Mapping):
 class HexBytes(bytes):
     def __new__(cls, val):
         bytesval = hexstr_if_str(to_bytes, val)
-        return super().__new__(cls, bytesval)
+        return super(HexBytes, cls).__new__(cls, bytesval)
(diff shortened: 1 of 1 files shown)


carver:
Hi @wip727 thanks for your interest in the project!

Python 2 is explicitly not supported in the master branch, which is nearly released as v4-beta now. You'll notice that web3 just refuses to start in Py2 after #408 was merged to master.

The options for resolving your issue are to either upgrade to Python 3 (the preferred route), or pin to version 3 (the `v3` branch).

carver:
Since py2 support for web3 v4 is not on the table, I'll close this PR. Let me know if there's anything else I can do to help.

wip727:
Thanks a lot for the clarification @carver, we will upgrade Python for the v4-beta.

Issue #342, opened 2017-10-09
ENS Integration


carver:
### What was wrong?

We should enable sending transactions from/to an ENS address, and create a contract by referencing an ENS name.

### How can it be fixed?

Integrate [ens.py](https://github.com/carver/ens.py) inside `web3` as a side-loaded module, then add default web3 middleware that converts from values ending in '.eth' to their address as resolved in ENS. ie~ `ens.address('to-addr-name.eth')`

For security, the name lookup should be relatively fresh, so stalecheck will become a default middleware.

Additionally, an optional middleware `reverse_ens` could do [reverse name lookups](https://github.com/carver/ens.py#get-name-from-address) when an address would be returned. You might see something like `miner = getBlock('latest')` return block with: `block.miner == 'f2pool.eth'`

- [x] ~add stalecheck middleware as default (maybe 72 hours?)~ (This logic can stay inside ENS)
- [x] add `ens` module #343 

These are in limbo, waiting to be rebased and reopened after #343 is merged:
- [x] create `name_to_address_middleware` as a default middleware (auto-populates from/to in `sendTransaction`)
- [x] ENS-aware Contracts ~(wait for #317 )~
  - [x] ABI address inputs could look up `.eth` names
  - [x] Initialize contracts with ENS addresses: eg~ `TokenContract('totally-legit-ico.eth')`
  - [x] Deploy contracts with args of ENS addresses: eg~ `TokenContract.deploy(args=['withdrawal-wallet.eth'])`
- [x] Documentation
  - [x] make it very clear that the address can change mid-stream when using ENS names.
  - [x] how to do `token = TokenContract(resolve('name.eth'))` for when a user only wants to resolve up-front and only use the address

Reminder to open these issues:
- [x] create `address_to_name_middleware` as an optional middleware which would replace returned addresses with ENS names
- [x] support name lookups on networks other than mainnet (like testnets), probably using a chain ID lookup in `ENS.fromWeb3()`

Punt until user demand builds:
- ~optionally trigger contract failures if name resolution changes (see part # 3 of _Documentation_ above)~
- ~solicit feedback for dynamic contract name lookup, but with strictness around address changes (maybe changes in a recent period, or any changes)~

carver:
When creating a contract, should we save the resolved address or save the name and resolve it on demand? (either way, we should resolve it once to validate that the name points to an address)

For example, if I have a long-running process that sets up a link to some token contract on init:

```
[...]

PR #414, opened 2017-11-10
Resolve ENS names to addresses

carver:
### What was wrong?

It was cumbersome to resolve names manually before using them throughout the web3 module. See #342 

### How was it fixed?

Detect .eth names and resolve them anywhere that addresses are used.

Three main entry points:
 * middleware that intercepts calls like `eth_getBalance` and replaces ENS name arguments with the address that they resolve to. Another example: the `to` & `from` fields in `eth_sendTransaction`
 * contract initialization - store the name and resolve it when you interact with the contract
 * accept ENS names in address-typed arguments to contract calls/transactions, and resolve those names before invoking

#### Cute Animal Picture

Cute animal picture


pipermerriam:
If you'll start using the "Ready for Review" label it'll help me do drive-by checks on things that need review.

carver:
Sounds good, will do. It's ready for a look now.

carver:
Okay, TODO:

- [ ] explore backend pattern for ENS
- [ ] accept any TLD (using `addr.startswith('0x') and '.' not in addr` to decide if address, otherwise test if valid ENS name)

carver:
I didn't add an extra interface layer, but you can easily swap out ENS implementations in the web3 instance now. All usages (including the middleware and contracts) use the ens instance in the web3 object, so it's pretty painless to swap it out, even temporarily. Any objections to that approach before I merge @pipermerriam ?

If no comment by tomorrow afternoon, I'll assume we're good to go. Then we can cut the first v4-beta release!

pipermerriam:
:shipit:

Issue #400, opened 2017-10-30
Deprecation of Contract.on and Contract.pastEvents


pipermerriam:
* Version: 4.x

### What was wrong?

With the recent deprecation of the old filter API, we forgot to address `Contract.on` and `Contract.pastEvents`.  These still hold some of the old semantics and I think they would benefit from being replaced by a new API.

### How can it be fixed?

I suggest deprecating both of the old contract filter APIs in favor of a new `Contract.eventFilter` which doesn't do any implicit *magic* with the block numbers and simply creates a filter that the user can then decide how they want to interact with it.

PR #416, opened 2017-11-10
New contract.eventFilter() to replace on() and pastEvents()

### LINKED ISSUE #251: Drop gevent support
### What was wrong?

Gevent support is a leftover feature from a long time ago when all of the threading/async code in these libraries was gevent based.  Given that gevent users can use the monkeypatching approach, I'm inclined to remove support.

### LINKED ISSUE #415: Implement eventFilter
* Version: 4

### What was wrong?

We removed the old asynchronous filtering methods on contract. The new filter mechanism has to be added. See #400 

### How can it be fixed?

If the name changes from `eventFilter` during implementation, be sure to grep for `eventFilter` to find all the tests that were disabled and need to be reimplemented. (If the name stays the same, the tests will start failing, as a reminder to rewrite them)

carver:
### What was wrong?

* Needed to remove features deprecated in v3:
  * gevent
  * contract.on - related to #400
  * contract.pastEvents - related to #400
* Needed new way to filter events on contracts, with synchronous-only access

### How was it fixed?

* Removed gevent library & tests - closes #251
* Removed contract.on and contract.pastEvents.
* Added `eventFilter()` - closes #415 
* Updated docs

#### Cute Animal Picture

Cute animal picture


web3/utils/compat/__init__.py
```diff
@@ -14,30 +14,17 @@ def get_threading_backend():
...
-        make_server,
-        GreenletThread,
-        spawn,
-        subprocess,
+    from .compat_requests import (  # noqa: F401
         make_post_request,
     )
 else:
-    raise ValueError("Unsupported threading backend.  Must be one of 'gevent' or 'stdlib'")
+    raise ValueError("Unsupported threading backend.  Must be 'stdlib'")
```
pipermerriam: Thoughts on just removing this alltogether? and making the imports direct?  too much?

carver: Yeah, I think that's reasonable. I'll take a crack at it. If it starts to get messy, I'll merge and create an issue. If it's clean enough, I'll upload it to this one before merging.

PR #423, opened 2017-11-14
Replace is_address checks with is_checksum_address

carver:
### What was wrong?

Some non-checksum addresses might have slipped through.

### How was it fixed?

Grepped for all `is_address` usages, and replaced (nearly) all of them.

Note that this changes the meaning of `Web3.isAddress`. Since we now require checksummed addresses everywhere, and return them everywhere, `isAddress` is now semantically the same as `isChecksumAddress`.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -1,6 +1,5 @@
 
 from eth_utils import (
-    is_address,
     is_checksum_address,
     to_checksum_address,
 )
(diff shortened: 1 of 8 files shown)


docs/overview.rst
```diff
@@ -193,30 +195,22 @@ Addresses
 
 .. py:method:: Web3.isAddress(value)
 
-    Returns ``True`` if the value is one of the recognized address formats.
+    Returns ``True`` if the value is a valid `EIP55`_ checksummed address.
+    Note that EIP55 requires the '0x' prefix.
```
pipermerriam: My initial reaction is that the functionality of `Web3.isAddress` should not change.

It's still a useful thing to be able to check if something is an address in the basic definition of address as a hex encoded 20 byte value.  With this change users will have to resort to digging into `eth_utils` if they want this functionality and I don't think that's a good thing.

carver:
In gitter, @ pipermerriam gave :+1: with the isAddress change reverted.

PR #427, opened 2017-11-14
 Do not guess .eth ending for ENS names in web3

### LINKED ISSUE #426: Stop inferring .eth on name lookups without a TLD
### What was wrong?

The `ens.address(name)` method appends '.eth' if no recognized TLD is present. This is handy at the command line, but problematic in web3, where ambiguity is created, because we want to support all possible TLDs out of the gate.

### How can it be fixed?

* Change signature to `ens.address(name, guess_tld=True)`
* Set `guess_tld` to `False` in all web3 usages
* Confirm that ENS.py can resolve any TLD when `guess_tld=False`

carver:
### What was wrong?

If someone tried to use the name `xyz`, web3 would look up `xyz.eth` instead of the top level domain `xyz`. This leads to bad ambiguities when more & longer TLDs are added.

Closes #426 

### How was it fixed?

Never append .eth to names

* Added a strict version of ENS address lookup
* Exclusively use strict lookups in middleware & contracts

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -27,6 +27,7 @@ from ens.utils import (
     init_web3,
     is_valid_name,
     label_to_hash,
+    name_to_hash,
     normalize_name,
 )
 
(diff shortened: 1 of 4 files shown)


ens/main.py
```diff
@@ -75,14 +76,19 @@ def fromWeb3(cls, web3, addr=None):
         '''
         return cls(web3.manager.providers, addr=addr)
 
-    def address(self, name):
+    def address(self, name, guess_tld=True):
```
pipermerriam: I'm beginning to be of the opinion that we should just **require** the full name and not support guessing the TLD.  This is like a browser letting you type `google` and it taking you magically to `google.com`.

carver: >  This is like a browser letting you type google and it taking you magically to google.com

I actually use ctrl-enter a lot in browsers, so I might have pitched doing that in a browser. heh.

But yeah, I can see the argument for it. Autofilling .eth made a little more sense when it was essentially the only TLD. That's still basically true, but not for much longer.

Issue #339, opened 2017-10-09
Python Modernization


carver:
### What was wrong?

A lot of code cruft exists to support versions of Python before 3.5. It complicates the implementation, and requires more time to test.

### How can it be fixed?

In v4, as part of dropping support for Python 2 and Python <=3.4, we can remove this cruft.  Also, let's drop test environments for python 2.7 and 3.4, and add one for 3.6.

- [x] Remove Python 2.7 and 3.4 tests #349 
- [x] Add Python 3.6 test environment
- [x] Remove all `sys.version_info` checks, presuming Python 3.5+
- [x] Remove all `bytes == str` checks, presuming Python 3.5+
- [x] Remove the `web3.utils.compat` module, and replace all usages with native calls #251 
- [ ] **Bonus**: Convert logic, where appropriate, to use native methods in 3.5+ (like `b'\xbe\xef'.hex()`, etc.)

gitcoinbot:
__This issue now has a funding of 0.95 ETH (313.05 USDT) attached to it.__  To view or claim this funding, [click here](https://gitcoin.co/funding/details?url=https://github.com/pipermerriam/web3.py/issues/339).

jasonrhaas:
Just submitted a claim on this bounty through gitcoin.co!  I did it through the Google Form since I was having a bit of trouble with MetaMask.

owocki:
@jasonrhaas i got your claim.  i'm confused though.  it looks like @techtonik is the one who submitted the actual code?

jasonrhaas:
@owocki I was under the impression that there was still work that needed to be done since not all of the checkmarks were marked complete and the claim was still open.

owocki:
ah yes, you are correct.

jasonrhaas:
@owocki Am I good to start working on this issue for the bounty?  I'm new to the whole gitcoin.co thing, first time trying it out.

owocki:
please do @jasonrhaas !

pipermerriam:
@jasonrhaas assuming @techtonik is also interested in this (since he has already submitted one pull request on the subject) it's likely we'll look at dividing the bounty between the two of you (also assuming that @techtonik is interested in collecting the bounty).  Exact division TBD based on something like the portion of the work done by the two of you.

Wanted to be upfront about expectations.

jasonrhaas:
@pipermerriam Fine with me, I'll leave it up to you to decide how the bounty gets split based on work put in.

[...]

PR #438, opened 2017-11-17
Python modernization

jasonrhaas:
### What was wrong?

Need to remove Python 2 code since we are dropping support for it in the next release.  Also - take advantage of native Python 3 methods where it makes sense.

Reference issue #339 

### How was it fixed?

- Remove support for Python2
- Add tests for Python 3.6
- Add docker based test environment

#### Cute Animal Picture

```
   __
 .^o ~\
Y /'~) }      _____
l/  / /    ,-~     ~~--.,_
   ( (    /  ~-._         ^.
    \ "--'--.    "-._       \
     "-.________     ~--.,__ ^.
               "~r-.,___.-'-. ^.
                YI    \      ~-.\
                ||     \        `\
                ||     //
                ||    //
                ()   //
                ||  //     -Row
                || ( c
   ___._ __  ___I|__`--__._ __  _
 "~     ~  "~   ::  ~~"    ~  ~~
                ::
                .:
                 .
```


jasonrhaas:
Tests are passing when I run `pytest -n 4 -f -k "not goethereum"`.  I am having issues getting go-etheruem installed in my docker container, but I don't think any of my changes should affect this part of the test suite.

`============= 852 passed, 29 skipped, 2 warnings in 68.35 seconds ==============`

jasonrhaas:
OK, I think this one is ready for review.  The only travis test that is failing is the `flake8` part for some reason, not sure what is going on there.

carver:
@jasonrhaas If you click into the flake8 test, it shows you log output, which includes the flake8 failure info. (flake8 will also run quickly in your local dev environment)

tox.ini
```diff
@@ -4,6 +4,10 @@ envlist=
     py35-core-stdlib
     py35-integration-{ethtestrpc,goethereum,ethtester}
     flake8
+    py36-ens
```
pipermerriam: Prefered format would be to inline these with the definitions above using the following syntax

```
py{35,36}-ens
py{35,36}-core  # you can drop the `-stdlib` suffix
py{35,36}-integration-{...}
```

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY setup.py .
+COPY README.md .
+COPY requirements-dev.txt .
+COPY requirements-docs.txt .
+
+# Install python dependencies
+RUN pip install -r requirements-dev.txt
+RUN pip install -e .
+
+WORKDIR /code
```
pipermerriam: I think I could get onboard with the inclusion of a dockerfile but would mind extracting it into a separate pull request and writing up some basic documentation in the `README.md` explaining how one might go about using this.

.travis.yml
```diff
@@ -42,6 +40,41 @@ matrix:
...
+      env: TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.6.0
+    - python: "3.6"
+      env: TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.6.1
+    # 1.6.2 and 1.6.3 are failing due to go-ethereum internal errors.
+    - python: "3.6"
+      env: TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.6.4
+    - python: "3.6"
+      env: TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.6.5
+    - python: "3.6"
+      env: TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.6.6
```
pipermerriam: I believe @carver was looking to drop testing against everything but `v1.7.2` since it's the *only* version of go-ethereum that is currently viable to be run on the mainnet.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /usr/src/app
+
+# Install Linux dependencies
+RUN apt-get update && apt-get install -y libssl-dev
+
+# Copy over requirements
+COPY setup.py .
+COPY README.md .
+COPY requirements-dev.txt .
+COPY requirements-docs.txt .
```
carver: Probably don't need to copy in requirements-docs.txt here, unless you're planning on building the documentation inside this container.

Dockerfile
carver: *nitpick* please add trailing newline.

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+services:
+  sandbox:
+    build:
+      context: .
+    volumes:
+      - .:/code
+    command: tail -f /dev/null
```
carver: *nitpick* - trailing newline, again

tests/core/utilities/test_encoding.py
```diff
@@ -1,7 +1,8 @@
 # encoding: utf-8
 
-from __future__ import unicode_literals
+# from __future__ import unicode_literals
```
carver: Looks like this line should just be deleted.

tests/core/utilities/test_encoding.py
```diff
@@ -1,7 +1,8 @@
...
 
+from unittest.mock import Mock
 import pytest
 import sys
```
carver: This is the nitpickiest of all, but I've been alphabetizing the imports whenever I touch them. (only inside the [PEP8 subgroups](https://www.python.org/dev/peps/pep-0008/#imports), so `unittest` would go after `sys` here).

.travis.yml
carver: Correct. #428 will push that change for the 3.5 python environments, so you can leave them out of the 3.6 ones.

jasonrhaas:
All review comments have been implemented.  I added one other commit 6edd103 that removes `__future__` imports across the project.  Tests are passing 💯

jasonrhaas:
@carver @pipermerriam this one should be ready to merge.

techtonik:
Removed tests test Python 2 code, but I don't see Python 2 code removed from main codebase. Without tests it will be harder to locate it.

carver:
@techtonik I understand the sentiment, and it would have been reasonable to make those changes in this PR, but it's also okay to put it in a separate one. It won't be too hard to find, because they use `sys.version_info`.

zhuyuecai:
I don't understand why dropped support for python2, actually a lot dev favour python2 over python3, including me.

techtonik:
@zhuyuecai if you can list reasons why you favour Python 2, then there is a chance to make Python great again in Python 4 and add support for that.

pipermerriam:
@zhuyuecai I'd be interested in hearing your reasons for preferring python2 as well.

Issue #440, opened 2017-11-18
sendRawTransaction doesn't produce valid RPC request


monokh:
* Version: 4.0.0-beta.1
* Python: 3.5
* OS: osx

### What was wrong?

Trying to sign and send a raw transaction produces an error indicating the RPC request is invalid. 

* The code which produced the error ([straight from the docs](http://web3py.readthedocs.io/en/latest/web3.eth.account.html#web3.eth.account.Account.signTransaction)) :
```
from web3 import Web3, HTTPProvider

w3 = Web3(HTTPProvider('https://mainnet.infura.io'))
transaction = {
    'to': '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
    'value': 1000000000,
    'gas': 2000000,
    'gasPrice': 234567897654321,
    'nonce': 0,
    'chainId': 1
}
key = '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
signed = w3.eth.account.signTransaction(transaction, key)
print(Web3.toHex(signed.rawTransaction))
tx = w3.eth.sendRawTransaction(signed.rawTransaction)
```
* The full output of the error
```
Traceback (most recent call last):
  File "justtesting.py", line 14, in <module>
    tx = w3.eth.sendRawTransaction(signed.rawTransaction)
  File "/Users/monokh/Code/web3.py/web3/eth.py", line 222, in sendRawTransaction
    [raw_transaction],
  File "/Users/monokh/Code/web3.py/web3/manager.py", line 92, in request_blocking
    raise ValueError(response["error"])
ValueError: {'message': 'invalid argument 0: json: cannot unmarshal invalid hex string into Go value of type hexutil.Bytes', 'code': -32602}
```
* What type of node you were connecting to.
https://mainnet.infura.io
`testrpc` produces similar error: `ValueError: {'code': -32000, 'message': 'Error: Invalid Signature`

### How can it be fixed?

The web3py version produces the signed message: `0xf86a8086d55698372431831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a009ebb6ca057a0535d6186462bc0b465b561c94a295bdb0621fc19208ab149a9ca0440ffd775ce91a833ab410777204d5341a6f9fa91216a6f3ee2c051fea6a0428`
The web3.js version produces an identical signed message.

Here's the web3.js version: ``

```
var Web3 = require('web3');
[...]

PR #442, opened 2017-11-19
accept raw bytes to sendRawTransaction

carver:
### What was wrong?

#440 

### How was it fixed?

Add an ABI definition for the ABI middleware, so the method will accept bytes or hex input.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/module_testing/eth_module.py
+++ b/web3/utils/module_testing/eth_module.py
@@ -223,8 +223,9 @@ class EthModuleTest(object):
                 '0x1f80f8ab5f12a45be218f76404bda64d37270a6f4f86ededd0eb599f80548c13',
             ),
             (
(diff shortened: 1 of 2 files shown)


carver:
The code is fixed, but the test is failing because geth refuses the second transaction submission with the same hash. Need to generate a new valid second transaction.

pipermerriam:
@carver did you see https://github.com/pipermerriam/web3.py/blob/master/tests/generate_go_ethereum_fixture.py

pipermerriam:
Also, it looks like I never saved the private key for `0x39eeed73fb1d3855e90cbd42f348b3d7b340aaa6` so you'll need to just generate a new private key and re-generate the fixtures.

PR #445, opened 2017-11-19
Convert 'status' field from hex to int in getTransactionReceipt

jstoxrocky:
### What was wrong?
web3.eth.getTransactionReceipt returned a hex value for the status field. The status field was newly introduced following the Byzantium update. This was inconsistent with other returned values that were converted from hex to int.

### How was it fixed?
status was added to to the list of RECEIPT_FORMATTERS with a formatter of "to_integer_if_hex"

#### Cute Animal Picture

Cute animal picture


carver:
Thanks @jstoxrocky ! Would you mind opening a second PR against the `v3` branch too? This should probably be backported.

PR #448, opened 2017-11-20
thread-aware check for recursive repeats

carver:
Porting #447 to v4

### DIFF
--- a/tests/core/utilities/test_decorators.py
+++ b/tests/core/utilities/test_decorators.py
@@ -1,0 +1,22 @@
+import time
+
+from web3.utils.threads import (
+    spawn,
+)
+from web3.utils.decorators import (
+    reject_recursive_repeats,
+)
+
+
+def test_reject_recursive_repeats_multithreaded():
+    @reject_recursive_repeats
+    def recurse(sleep_now):
+        time.sleep(sleep_now)
+        try:
+            recurse(0.05)
+            return True
+        except ValueError:
+            return False
+    thd1 = spawn(recurse, 0)
+    thd2 = spawn(recurse, 0.02)
+    assert thd2.get() and thd1.get()
--- a/web3/utils/decorators.py
+++ b/web3/utils/decorators.py
@@ -1,4 +1,5 @@
 import functools
+import threading
 import warnings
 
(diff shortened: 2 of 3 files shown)


web3/utils/decorators.py
```diff
@@ -24,14 +25,15 @@ def reject_recursive_repeats(to_wrap):
 
     @functools.wraps(to_wrap)
     def wrapped(*args):
-        instances = tuple(map(id, args))
-        if instances in to_wrap.__already_called:
+        arg_instances = tuple(map(id, args))
+        unique_call = (threading.get_ident(),) + arg_instances
```
pipermerriam: I *think* it may be slighly easier to parse if you used a `threading.local` to store the state rather than keying based on the current thread.

tests/core/utilities/test_decorators.py
```diff
@@ -0,0 +1,22 @@
...
+    def recurse(sleep_now):
+        time.sleep(sleep_now)
+        try:
+            recurse(0.05)
+            return True
+        except ValueError:
+            return False
+    thd1 = spawn(recurse, 0)
+    thd2 = spawn(recurse, 0.02)
+    assert thd2.get() and thd1.get()
```
pipermerriam: *curious* 

When you tested this, did it trigger the error 100% of the time?

carver: I tested it 3 or 4 times, and it triggered every time. I'll go run a quick test to make sure it triggers at least 200 times in a row.

carver: On my laptop, I can trigger this failure 500 times in a row without the thread-aware decorator.

web3/utils/decorators.py
carver: So I briefly considered that, but the problem that I ran into is I had to add my own namespace mechanism rather than just add the dict to the wrapped method, and check for/create a new dict if you're inside a new thread.

Maybe I missed an easier way to do it, but this seemed a lot cleaner. I can play with the variable names a bit to make it more readable, though.

pipermerriam:
:shipit:

PR #455, opened 2017-11-22
Remove all sys.version_info checks, presuming Python 3.5+

miguel550:
### What was wrong?
It have python 2 and python 3.3 checks all over the places.

### How was it fixed?
I remove them carefully.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_init.py
+++ b/tests/core/contracts/test_contract_init.py
@@ -1,3 +1,4 @@
+
 import pytest
 
 from web3.exceptions import (


carver:
I'm not sure how you accomplished this, but this commit seems to be removing and re-adding every file? Maybe it's changing the newlines from `\n` to `\r\n`? The diff doesn't make any sense to me.

miguel550:
After investigating it seems like Windows is changing all the files' permissions to 777 and github is taking it as a change. It does not seems like a problem but if it is I can make the changes again in a clean way.
 
More info: https://github.com/Microsoft/WSL/issues/184

carver:
@miguel550 yes, this definitely needs to be fixed. It is impossible to code review this way. Try again after the suggestion from the thread you linked: 

```
# For the current repository
git config core.filemode false   

# Globally
git config --global core.filemode false
```

(force pushing the re-issued commits to your branch is fine here)

PR #456, opened 2017-11-22
Remove all sys.version_info checks, presuming Python 3.5+ (Second try)

miguel550:
### What was wrong?
It have python 2 and python 3.3 checks all over the places.

### How was it fixed?
I remove them carefully.

#### Cute Animal Picture

Cute animal picture


miguel550:
@carver can give me a hint of why this is failing?

carver:
Are you familiar with looking at the Travis logs? Click on Details next to "The Travis CI build failed", then you can click on a particular failing build to get the full logs.

I didn't look at all the failures, but one of them is because `transaction_hash` was added back into the return tuple in `sign_transaction_dict`.  That suggests to me that you rebased onto master, which is a good idea. But then something went wrong during the conflict resolution. You can fix this particular issue by removing `transaction_hash`, but please review the full diff of this PR to make sure other conflicts got resolved correctly.

Also, all the version_info checks should be able to be completely deleted.

So if the original is:
```py
if sys.version_info < (3, 3):
  do_something()
else:
  do_something_py2()
```

Then it should all be replaced with only:
```py
do_something()
```

miguel550:
Thanks, I'll take a look!

miguel550:
@carver I removed the remaining sys.version_info from the tests. I let only one sys.version_info and is in web3/__init__.py to let the user know that they need use python 3.5 or above to use web3.py without problems.

miguel550:
Does this PR have something else to be done? @carver

web3/__init__.py
```diff
@@ -1,8 +1,8 @@
 import pkg_resources
 import sys
 
-if sys.version_info.major < 3:
-    raise EnvironmentError("Python 3 is required")
+if sys.version_info < (3, 5):
+    raise EnvironmentError("Python 3.5 or above is required")
```
carver: Ah, thanks for catching this.

PR #460, opened 2017-11-22
Re-raise formatting errors with field name

### LINKED ISSUE #459: Add field name of formatter failure to apply_formatters_to_dict
* Version: 3.x & 4.x
* Python: 2.7/3.4/3.5
* OS: osx/linux/win

### What was wrong?

When applying formatters to a dict, if one of the values fails, it's impossible to tell from the trace which one did.

### How can it be fixed?

Catch ValueError exceptions in `apply_formatters_to_dict` and rethrow with the name of the field that failed.

carver:
### What was wrong?

Very difficult to debug which formatter is failing when applied to a dict.

### How was it fixed?

Re-raise the formatting exception with the field name.

Closes #459 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/formatters.py
+++ b/web3/utils/formatters.py
@@ -67,7 +67,10 @@ def apply_formatter_if(condition, formatter, value):
 def apply_formatters_to_dict(formatters, value):
     for key, item in value.items():
         if key in formatters:
-            yield key, formatters[key](item)
+            try:
+                yield key, formatters[key](item)
+            except (TypeError, ValueError) as exc:
+                raise type(exc)("Could not format value %r as field %r" % (item, key)) from exc
         else:
             yield key, item
 
--- a/tests/core/utilities/test_formatters.py
+++ b/tests/core/utilities/test_formatters.py
@@ -2,6 +2,7 @@
 import pytest
 
 from web3.utils.formatters import (
+    apply_formatters_to_dict,
     map_collection,
     recursive_map,
 )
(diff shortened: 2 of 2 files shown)


web3/utils/formatters.py
```diff
@@ -67,7 +67,10 @@ def apply_formatter_if(condition, formatter, value):
 def apply_formatters_to_dict(formatters, value):
     for key, item in value.items():
         if key in formatters:
-            yield key, formatters[key](item)
+            try:
+                yield key, formatters[key](item)
+            except (TypeError, ValueError) as exc:
+                raise type(exc)("Could not format value %r as field %r" % (item, key)) from exc
```
pipermerriam: I think this should be changed to include the original exception message (it doesn't appear it currently does)

```python
raise type(exc)("Could not format value %r as field %r: %r" % (item, key, str(exc))) from exc
```

carver: Since it is a `raise from` the old trace and message all still show up though. If you missed the `raise from`, do you think this is easier to read?

```
except (TypeError, ValueError) as old_exception:
    new_exception = type(exc)("Could not format value %r as field %r: %r" % (item, key, str(exc)))
    raise new_exception from old_exception
```

pipermerriam: I didn't miss it, I just wasn't aware that it showed both messages.  Current format is good with me.

Issue #454, opened 2017-11-21
When trying to use web3 on Windows 10 in Powershell or CMD, get "ModuleNotFoundError: No module named 'win32file'"


Mikerah:
* Version: 3.16.2
* Python: 3.6
* OS: win10

### What was wrong?

When trying to call `web3.eth.filter('latest')` by following the example in the documentation, I get the following error:
~~~~
Python 3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 17:26:49) [MSC v.1900 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from web3 import Web3, IPCProvider
>>> web3 = Web3(IPCProvider())
>>> web3.eth.filter('latest')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\eth.py", line 270, in filter
    "eth_newBlockFilter", [],
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\manager.py", line 93, in request_blocking
    response = self._make_request(method, params)
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\manager.py", line 76, in _make_request
    return request_func(method, params)
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\middleware\attrdict.py", line 20, in middleware
    response = make_request(method, params)
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\middleware\formatting.py", line 25, in middleware
    response = make_request(method, params)
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\providers\ipc.py", line 84, in make_request
    with get_ipc_socket(self.ipc_path) as sock:
  File "c:\python36\Lib\contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\providers\ipc.py", line 25, in get_ipc_socket
    from web3.utils.windows import NamedPipe
  File "C:\Users\Mikerah\Documents\GitHub\py-eth-parser\env\lib\site-packages\web3\utils\windows.py", line 8, in <module>
[...]

PR #470, opened 2017-11-27
Added windows specific dependencies to setup.py

Mikerah:
### What was wrong?
In issue #454 , I noted that `pypiwin32` didn't get installed when I installed web3 through pip in a virtualenv on Windows 10.

### How was it fixed?
By following the suggestions in this [stackoverflow](https://stackoverflow.com/questions/16055403/setuptools-platform-specific-dependencies/32955538#32955538) post

So, I simply removed the line adding the `pypiwin32` package from the `installed_packages` list and instead added a line to the `extra_requires` parameter of `setup`.

#### Cute Animal Picture

Cute animal picture


carver:
Also note the [comment below the answer](https://stackoverflow.com/questions/16055403/setuptools-platform-specific-dependencies/32955538#comment79776267_32955538), which suggests this syntax:

`'platform_system=="Windows"'`

I'll fix the conflict and update for that syntax. Thanks for tracking this down @Mikerah !

PR #471, opened 2017-11-27
Python modernization 2

jasonrhaas:
### What was wrong?
Finishing remaining items on #339.

### How was it fixed?
Deleted old code.

#### Cute Animal Picture
[hippo](https://static.boredpanda.com/blog/wp-content/uploads/2017/05/59083ecd5984d_5iXoAQJr__700.jpg)


web3/account.py
```diff
@@ -60,14 +58,11 @@ class Account(object):
...
         key_bytes = keccak(os.urandom(32) + extra_key_bytes)
-        if sys.version_info.major < 3:
-            key_bytes = to_hex(key_bytes)
         return self.privateKeyToAccount(key_bytes)
 
     @staticmethod
     def decrypt(keyfile_json, password):
-        if isinstance(keyfile_json, str) or (
-                sys.version_info.major < 3 and isinstance(keyfile_json, unicode)):  # noqa: 821
+        if isinstance(keyfile_json, str):  # noqa: 821
```
techtonik: Why `noqa` here?

web3/account.py
```diff
@@ -36,8 +35,7 @@
     hexstr_if_str,
     text_if_str,
     to_bytes,
-    to_int,
-    to_hex,
```
techtonik: :+1:

web3/utils/encoding.py
```diff
@@ -231,16 +229,10 @@ def text_if_str(to_type, text_or_primitive):
 
     @param to_type is a function that takes the arguments (primitive, hexstr=hexstr, text=text),
         eg~ to_bytes, to_text, to_hex, to_int, etc
-    @param hexstr_or_primitive in bytes, str, or int. (or unicode in py2)
-        In Python 2, only a unicode object will be interpreted as unicode text
-        In Python 3, only a str object will be interpreted as interpreted as unicode text
+    @param hexstr_or_primitive in bytes, str, or int.
+        In Python 3, only a str object will be interpreted as unicode text
```
techtonik: It is all Python 3 now.

web3/utils/encoding.py
```diff
@@ -194,7 +193,6 @@ def to_bytes(primitive=None, hexstr=None, text=None):
     elif is_integer(primitive):
         return to_bytes(hexstr=to_hex(primitive))
     elif hexstr is not None:
-        hexstr = hexstr.rstrip('L')  # handle longs in Python 2
```
techtonik: See also line 152 https://github.com/ethereum/web3.py/pull/471/files#diff-0260bdf984abdd4b8ab8987ea75d032cR152

web3/account.py
jasonrhaas: @techtonik that was already in the code so I just left it there.  I'm not sure what the comment is referencing.

web3/utils/encoding.py
jasonrhaas: I guess its safe to delete all the "python 3" comments?

web3/utils/encoding.py
jasonrhaas: fixed

jasonrhaas:
Its failing in Travis on some weird `ValueError: embedded null byte` error.  It doesn't happen on my local docker image, I'm assuming its something with the Linux Travis environment?  Not sure exactly.

```
self = environ({'LANG': 'en_US.UTF-8', 'GOPATH': '/home/travis/gopath', 'GOROOT': '/home/travis/.gimme/versions/go1.7.4.linux...:/opt/pyenv/bin', 'PYTHONHASHSEED': '3431167186', 'VIRTUAL_ENV': '/home/travis/build/ethereum/web3.py/.tox/py36-core'})

key = b'PYTEST_CURRENT_TEST'

value = b'tests/core/eth-module/test_accounts.py::test_eth_account_hash_message_text[instance-Hello World-\\xa1\\xde\\x98\\x86\x00\\xa4,KJ\\xb0\\x89\\xb6\x19)|\x17\\xd5<\\xff\\xae]Q \\xd8-\\x8a\\x92\\xd0\\xbb;x\\xf2] (setup)'

    def __setitem__(self, key, value):

        key = self.encodekey(key)

        value = self.encodevalue(value)

>       self.putenv(key, value)

E       ValueError: embedded null byte
```

pipermerriam:
Looks like a pytest bug.  The test name is being derived from the fixture values and the fixture has null bytes in it.   Looks similar to https://github.com/pytest-dev/pytest/issues/2644

pipermerriam:
@jasonrhaas work-around would be to to explicitely name the tests using one of these methods:

https://docs.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids

pipermerriam:
https://github.com/pytest-dev/pytest/issues/2649 looks related.

jasonrhaas:
@pipermerriam I worked around this issue by assigning `ids` to all of the `@paramatrize()` statements.  In most cases I used something like `ids=['test_' + str(x) for x in range(10]` to give it a list of strings.  Slightly hacky, but it fixes the problem.

web3/account.py
carver: Just FYI, these noqa comments are a way of telling `flake8` not to issue the specified warning. Most of the time it's best to fix the thing causing the warning, but occasionally the best thing to do is suppress it.

The list of warnings are here: http://flake8.pycqa.org/en/3.4.1/user/error-codes.html#error-violation-codes

821 is usage of an undefined name, which is usually pretty bad, but in this case it was just because `unicode` isn't defined in py3, and we know a py3 codepath to that `unicode` in this line is impossible.

Now that `unicode` is removed, we can and should remove the warning supressor.

carver:
Hey, I've been reviewing and merging from oldest submissions to newest, and it looks like #456 did a lot of these same changes (hence the conflicts). I just took a pass at the conflicts, and will do a final review before merging...

web3/utils/encoding.py
carver: :+1:

carver:
Pinning pytest to 3.2.5 also does the trick, but specifying ids also make the test results easier to read so I'm happy to see those in place too.

tests/core/utilities/test_abi_is_encodable.py
```diff
@@ -56,6 +56,7 @@
         ('0xname.eth', 'address', True),  # 0x in name is fine, if it is not a TLD
         ('0xff', 'address', False),  # but any valid hex starting with 0x should be rejected
     ),
+    ids=['test_' + str(x) for x in range(42)]
```
carver: Since these `ids` actually obscure the test results, let's just remove them now that pytest is pinned to 3.2.5.

tests/core/web3-module/test_conversions.py
```diff
@@ -18,7 +18,8 @@
         (256, b'\x01\x00'),
         (True, b'\x01'),
         (False, b'\x00'),
-    )
+    ),
+    ids=['test_' + str(x) for x in range(10)]
```
carver: The `ids` in this module can go too (although the trailing comma on the list of args can stay).

tests/core/web3-module/test_sha3.py
```diff
@@ -11,7 +11,8 @@
     [
         ('cowmö', HexBytes('0x0f355f04c0a06eebac1d219b34c598f85a1169badee164be8a30345944885fe8')),
         ('', HexBytes('0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470')),
-    ]
+    ],
+    ids=['test_' + str(x) for x in range(2)]
```
carver: These `ids` can go too (although the trailing comma on the list of args can stay).

jasonrhaas:
@carver this should be ready to merge once the Travis Tests finish running.

PR #473, opened 2017-11-28
Fix ABI encoding of arrays of strings and bytes

carver:
### What was wrong?

Only in v4: `contract.encodeABI` fails to encode a byte-array with:

> `TypeError: expected an int in first arg, or keyword of hexstr or text`

It was trying to convert an array of hex strings into bytes all at once.

This conversion to bytes only happens in contract ABI encoding, because `encode_abi` expects byte arguments, and the other places that do this encoding (like middleware before sending to a client) expect hex-encoded arguments.

### How was it fixed?

Check that conversion from hex to bytes only happens at the basic type level, not the array level.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,7 +5,7 @@ flaky>=3.3.0
 flake8==3.4.1
 hypothesis>=3.31.2
 py-geth==1.10.2
-pytest>=3.2.1
+pytest==3.2.5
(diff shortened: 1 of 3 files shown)


carver:
pytest 3.3.0 was breaking our tests, due to https://github.com/pytest-dev/pytest/issues/2957

Issue #466, opened 2017-11-23
Account.sign() fails 343 out of 25600 times


miohtama:
* Version: master
* Python: 3.5
* OS: osx

### What was wrong?

I am trying to use Account.sign() to sign orders against 0x Exchange.sol contract. Account.sign() has been grabbed from the master for this purpose. Account.sign() seems to be produce a signature that Exchange.sol does not accept 343 out of 25600 times. Account.sign() input is seemingly random (SHA3 hash of 0x order parameters).

The private key used to sign is 0x5e95384d8050109aab08c1922d3c230739bc16976553c317e5d0b87b59371f2a and does not change between samples.

I am using Populus 1.11.x testrpc for the EVM bit.

Here is the full log of 343 failed hashes (to detect a pattern): https://gist.github.com/miohtama/93b87a963dcc7c944b5d5e6142815347

Below is my test case

```
def test_order_signature(web3, exchange, parties_ready_to_trade, bought_token, sold_token, fee_recipient, private_key_hex, private_key_account):
    """Python and Solidity agree how to sign orders."""

    buyer, seller = parties_ready_to_trade

    failure_count = 0
    for i in range(0, 25600):
        to_buy = 300
        to_sell = 200
        maker_fee = random.randint(0, 100)
        taker_fee = random.randint(0, 100)

        fee_recipient = "0x0000000000000000000000000000000000000000"

        o = create_test_order(
            web3,
            exchange_address=exchange.address,
            fee_recipient=fee_recipient,
            maker_address=seller,
            taker_address=buyer,
            maker_token=sold_token.address,
            taker_token=bought_token.address,
            maker_token_amount=to_sell,
            taker_token_amount=to_buy,
            maker_fee=maker_fee,
            taker_fee=taker_fee,
        )

        # Solidity takes our signature as granted
        hash = o.calculate_hash()
        hash_bytes = decode_hex(hash)

        data = o.sign_private_key(private_key_hex)

# is same as
#        hash = self.calculate_hash()
#        hash_bytes = decode_hex(hash)
#
  #      a = Account()
 #       sign_data = a.sign(hash_bytes, private_key=decode_hex(private_key_hex))
   #     signature_params = {
     #       "hash": hash,
      #      "hash_bytes": hash_bytes,
       #     "v": sign_data["v"],
        #    "r": sign_data["r"],
         #   "s": sign_data["s"],
        #}
        #return signature_params


        try:
[...]

PR #477, opened 2017-11-30
zero-pad signature bytes of messages, return r & s as integers

carver:
### What was wrong?

#466 signature bytes were not 0-padded on message signatures

### How was it fixed?

- 0-pad them r and s in signature bytes
- return `r` and `s` as integers
- `r` and `s` in transactions remain not 0-padded

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/signing.py
+++ b/web3/utils/signing.py
@@ -120,9 +120,6 @@ def to_eth_v(v_raw, chain_id=None):
     return v
 
 
-to_bytes32 = compose(zpad_bytes(32), to_bytes)
-
-
 def sign_transaction_hash(account, transaction_hash, chain_id):
     signature = account.sign_msg_hash(transaction_hash)
     (v_raw, r, s) = signature.vrs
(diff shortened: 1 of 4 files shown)


web3/utils/signing.py
```diff
@@ -107,21 +112,30 @@ def to_standard_v(enhanced_v):
...
 
 
 def sign_message_hash(key, msg_hash):
     signature = key.sign_msg_hash(msg_hash)
-    (v_standard, r, s) = signature.vrs
-    v = v_standard + V_OFFSET
-    eth_signature_bytes = b''.join(map(to_bytes, (r, s, v)))
+    (v_raw, r, s) = signature.vrs
+    v = to_eth_v(v_raw)
+    (r_bytes, s_bytes) = map(to_bytes32, (r, s))
```
pipermerriam:
```python
r_bytes, s_bytes = to_bytes32(r), to_bytes32(s)
# or
r_bytes = to_bytes32(r)
s_bytes = to_bytes32(s)
```

Minor readability improvements?

PR #479, opened 2017-12-01
bumpversion and release in one-liner

carver:
### What was wrong?

I'm too lazy to type or remember three lines for releasing new versions.

### How was it fixed?

Pushed all the version bump and release logic into the makefile

*Sorry I didn't come up with this before you copied it everywhere @pipermerriam X)*

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,12 @@ linux-docs: build-docs
 	xdg-open docs/_build/html/index.html
 
 release: clean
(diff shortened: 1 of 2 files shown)


pipermerriam:
I have an alias that works for my specific setup since `origin` is now always my fork of the main repo and `upstream` is always the main repository.  I suspect it *could* still be good to federate these across *all-the-repos* but currently you and I are the only ones cutting releases..

> *kicks the can down the road*

Issue #419, opened 2017-11-13
Flip the contract function calls


carver:
* Version: 4

### What was wrong?

The order of calls (`contract.call().doSomething()`) could be more intuitive. 

#### Basic Approach

web3.js seems to have a solid solution: `contract.functions.method().call()`

Plus:

 * `..method().transact()`
 * `..method().estimateGas()`
 * etc

This API can coexist alongside the old way of calling contract methods.

#### API Naming 

`functions` is a little long, here are some other options:

- `contract.api.doSomething().call()`
- `contract.public.doSomething().call()`
- `contract.build.doSomething().call()`
- `contract.make.doSomething().call()`
- `contract.prep.doSomething().call()`
- `contract.useabi.doSomething().call()`
- `contract.go.doSomething().call()`
- `contract.methods.doSomething().call()` _Added_

Also, think about how the name will sound when used with other execution options, like `transact`, `estimateGas`, and `buildTransaction`

#### Requirements

- [ ] add new API
- [ ] document new API in `docs/`
- [ ] add deprecation warning to old call/transact/etc methods
- [ ] **Bonus**: Update `ConciseContract` to use the new Contract API under the hood (with no change to the external API)

pipermerriam:
I would add `contract.methods.doSomething().call()` to the list to be considered.

I also like `api` but I'm not a fan of how broad a concept API is.  I think some users might assume that events are available under `api` as well when I think they belong under `contract.events.MyEvent()`.

carver:
Good point about the events disambiguation. That invokes another alternative, which is to mirror the word in [the `type` field from the abi json spec](https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI#json) exactly. By that, I mean something like:

- `myContract.function.myMethod().transact()`
- `myContract.event.myMethod().transact()`
- `myContract.constructor.args().transact()`
- `myContract.fallback.transact()`

But as is apparent above, it does lead to some non-parallel APIs in the constructor and fallback, which makes me sad.

pipermerriam:
I don't exactly follow your meaning here:

> But as is apparent above, it does lead to some non-parallel APIs in the constructor and fallback, which makes me sad.

I like the api you list above (`function/event/constructor/fallback`) though my spider sense says that special casing fallback and constructor *might* be wrong though that is a very weak *might* and on the surface I can find no reason not to do it.

carver:
> I don't exactly follow your meaning here:
[...]

Issue #430, opened 2017-11-15
New contract method execution option: generating the transaction dictionary


carver:
* Version: 4

### What was wrong?

There is no convenient way to prepare a contract transaction for offline signing.

Specifically, the prepared transaction should work smoothly with [`w3.eth.account.signTransaction()`](http://web3py.readthedocs.io/en/latest/web3.eth.account.html#web3.eth.account.Account.signTransaction).

#### New contract method preparation API

Add a new way to prepare a contract function transaction. It would look something like:

```
# new API
transaction_dict = myContract.functions.myMethod().asdict()

# existing APIs
signed = w3.eth.account.signTransaction(transaction_dict, private_key)
w3.eth.sendRawTransaction(signed.rawTransaction)
```

`asdict` should accept a dictionary that defines transaction fields, just like call/transact/etc, and return an `AttributeDict`.

#### Fields included

Besides the natural fields, like `to` and `data`, this must also generate other fields like:

- nonce
- chainId
- gas
- gasPrice

If they are not supplied as arguments to `asdict()`, then they should be inferred before returning. `signTransaction` will not attempt to infer these values, so they must be explicitly added.

#### API Naming

`asdict()` is okay, but isn't necessarily the obvious winner.

Some alternatives:

- `myContract.functions.myMethod().todict()`
- `myContract.functions.myMethod().tofields()`
- `myContract.functions.myMethod().prepare()`
- `myContract.functions.myMethod().generate()`

#### Crossover issue

Depending on whether #419 is already implemented, and how it's implemented, this might require a bit of extra work to expose `asdict` in both ways.

#### TODOs

- [ ] choose naming
- [ ] implement new API
- [ ] document in `docs/`
- [ ] reference this method in signTransaction docs `docs/web3.eth.account.rst`
- [ ] Support for #419 - if implementation available

pipermerriam:
suggest names with the word transaction in them

* `myContract.myMethod().prepareTransaction()`
* `myContract.myMethod().asTransactionDict()`
* `myContract.myMethod().generateTransaction()`

carver:
Yeah, I was feeling grumbly about how long it got. Especially when you add in the prep word (which I just realized I forgot to do. (Now updated in the top)

With prep word:

* `myContract.functions.myMethod().prepareTransaction()`
* `myContract.functions.myMethod().asTransactionDict()`
* `myContract.functions.myMethod().generateTransaction()`

pipermerriam:
[...]

PR #481, opened 2017-12-01
Flip contract function calls

dylanjw:
### What was wrong?
Contract function calls not intuitive #419 

### How was it fixed?
Now support more intuitive syntax:
contract.[abitype].[contractmethod].call|transact|estimateGas()

#### Cute Animal Picture

hi


dylanjw:
This is what I have done so far for the function contract method type.  I still need to do events + contructor, fallback.

I have a couple questions.

Im have worked on implementing this form (so far just function abi type):
<contract>.<abi_type>.<abi_name>(<args>).transact()

For example:
`contract.function.add(3,9).call()`

where `contract.function.add(3,9)` returns a class built from class a factory for contract methods.  

Question 1:
Is a class factory for contract methods a good way to go?

Question 2/3:
What should happen when calling the contract method instance? Do a .call() by default?
Should we support changing the parameters on calls to instances of the contract method object?  I currently do, but this can change what is returned from the abi lookup.  Im not sure if this is good or bad.

For example, this test:
```
def test_default_action_call_with_argument_change(math_contract):
    math_contract_add = math_contract.function.add(9,7)
    result = math_contract_add(8,9)
    assert result == 17
```

Question 4:

contract.function returns an object containing the contract method classes.  Im using a descriptor.  Initially my thought was to support the different syntaxes, e.g.
```
math_contract.function['add']().call()
math_contract.function.add().call()
math_contract.function().add().call
```
Does the descriptor seem seem weird?  Do you see a simpler way?

I have several TODO items, but I wanted to put this up in case you want me to take it in a different direction.

pipermerriam:
I think that having dedicated `ContractFunctions` and `ContractEvents` objects that implement the appropriate `__getattribute__` lookups would result in easier to understand/read code.  Descriptors are inherently a bit confusing, especially in this case where the descriptor returns a dynamically created class.

pipermerriam:
Oh, and this will probably be easier to test in isolation with dedicated `ContractFunctions` and `ContractEvents` objects since you won't have to go through the overhead of setting up a contract and then digging into it's `contract.functions`.

pipermerriam:
btw, when I say `ContractFunctions` object, I merely mean implementing the class that you're dynamically constructing in the descriptor as a dedicated class that is passed all of the necessary values from the contract.

web3/contract.py
```diff
@@ -89,6 +89,56 @@
...
 
 
+class ContractFunctions(object):
+    """Class containing contract function objects
+    """
+
+    _method_names = []
+
+    def __init__(self, contract):
+        self.contract = contract
```
pipermerriam: I don't know if it's feasible but I'm curious if we can do this without passing in the whole contract.  In theory we *should* be able to get by with only passing the address, ABI and the web3 instance.  This should make this **significantly** easier to test in isolation.

web3/contract.py
```diff
@@ -89,6 +89,56 @@
...
+                    ContractMethod.factory(contract=self.contract,
+                                           web3=self.contract.web3,
+                                           method_name=method['name']))
+
+    def __getitem__(self, item):
+        if item in self._method_names:
+            return getattr(self, item)
+
+    def __call__(self):
+        return self
```
pipermerriam: This is odd.  It suggests that you could do `contract.functions().myMethod()` *and* `contract.functions.myMethod()`.  Unless there is a reason for the former, thoughts on removing the `__call__` and just sticking with property based access?

web3/contract.py
```diff
@@ -122,6 +172,8 @@ class Contract(object):
     clone_bin = None
 
     dev_doc = None
+    function = None
```
pipermerriam: *naming nitpick*

I think this an `events` should be pluralized.

web3/contract.py
```diff
@@ -722,6 +776,161 @@ def __prepared_function(self, **kwargs):
         return getattr(contract_modifier_func(modifier_dict), self.__function)
 
 
+class ContractMethod(object):
+    """
+    """
+    contract = None
+    method_name = None
+    web3 = None
+    transaction = None
```
pipermerriam: This doesn't appear to be used anywhere.

web3/contract.py
carver: I like that these are all exactly the same as the ABI type field right now (which is why the written issue had them singular as well). At the same time I get the model. I proposed one more option for naming back on #419. It would require a significant change, but hopefully address this naming issue.

web3/contract.py
```diff
@@ -146,6 +198,8 @@ def __init__(self, address=None):
 
         if not self.address:
             raise TypeError("The address argument is required to instantiate a contract.")
+        setattr(self, 'function', ContractFunctions(self))
+        setattr(self, 'event', ContractEvents(self))
```
carver: Why not `self.event = ContractEvents(self)`?

web3/contract.py
```diff
@@ -722,6 +776,161 @@ def __prepared_function(self, **kwargs):
...
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def __call__(self, *args, **kwargs):
+        if args:
+            self.args = args
+        if kwargs:
+            self.kwargs = kwargs
+        return self.call()
```
carver: As I mentioned in the tests, I think this `__call__` should be dropped altogether.

web3/contract.py
```diff
@@ -722,6 +776,161 @@ def __prepared_function(self, **kwargs):
...
+    def factory(cls, **kwargs):
+        if "method_name" not in kwargs:
+            kwargs["method_name"] = cls.__name__
+
+        for key in kwargs:
+            if not hasattr(cls, key):
+                raise AttributeError(
+                    "Property {0} not found on ContractMethod class. "
+                    "`ContractMethod.factory` only accepts keyword arguments which are "
+                    "present on the contract class".format(key)
```
carver: present on the ~contract~ ContractMethod class

web3/contract.py
dylanjw: Its used in the factory method on [line 924](https://github.com/ethereum/web3.py/pull/481/files#diff-2171ba22a13b6793b04c4a1287049a67R924) where the kwargs are checked against the ContractMethod attributes before they are passed to type().

dylanjw:
I rebased to get the buildTransaction() changes, and did a forced push.  Sorry if that makes the review confusing.

Im failing a test introduced in #430:

```
def test_build_transaction_with_contract_class_method(web3, MathContract, math_contract):
    txn = MathContract.functions.increment().buildTransaction({'to': math_contract.address})
    assert txn == {
        'to': math_contract.address,
        'data': '0xd09de08a',
        'value': 0,
        'gas': 43120,
        'gasPrice': 1,
        'chainId': 1
    }   
```

Do I need to support `functions` and `events` on the contract type invocation?  If that is the case, I have a bit of a chicken and egg problem.  How I have it now, where functions and events are instatiated in __init__ they are only available in contract instances.  I could include them in the contract factory attributes like so:

```
    @classmethod
    def factory(cls, web3, contract_name=None, **kwargs):
        if contract_name is None:
            contract_name = cls.__name__

        kwargs['web3'] = web3
        kwargs['functions'] = ContractFunctions(cls)
        kwargs['events'] = ContractEvents(cls)

        for key in kwargs:
            if not hasattr(cls, key):
                raise AttributeError(
                    "Property {0} not found on contract class. "
                    "`Contract.factory` only accepts keyword arguments which are "
                    "present on the contract class".format(key)
                )   
            else:
                kwargs[key] = cls.normalize_property(key, kwargs[key])

        return type(contract_name, (cls,), kwargs)
```

...but then im not sure how to get all of the parameters I need, since at this point abi, address, web3 are absent from `cls`.  .  @pipermerriam Suggested removing the contract parameter from ContractMethod(), which I think is feasible, by moving `contract._find_matching_fn_abi` and `contract._prepare_transaction` to ContractMethod.  Another problem I see is wouldnt know what to do if 'address' is absent from contract.factory kwargs.

Originally, ContractMethod() was a descriptor, which delayed its invocation until I had all the parameters I needed in the Contract type.  Is there a different way to do this?

carver:
> I rebased to get the buildTransaction() changes, and did a forced push. Sorry if that makes the review confusing.

No worries, rebase seems like the best option you had.

> Do I need to support `functions` and `events` on the contract type invocation?

Yes.

> but then im not sure how to get all of the parameters I need, since at this point abi, address, web3 are absent from cls. . @pipermerriam Suggested removing the contract parameter from `ContractMethod()`, which I think is feasible, by moving `contract._find_matching_fn_abi` and `contract._prepare_transaction` to `ContractMethod`. Then I could just pass in abi, web3 and address from the factory kwargs.

Passing in abi and web3 here seems promising.

> Another problem I see is wouldnt know what to do if 'address' is absent from contract.factory kwargs.

Note that `address` is not available in the factory in any normal usage, because it's only supplied on initialization in `w3.eth.contract`: https://github.com/ethereum/web3.py/blob/master/web3/eth.py#L296-L301

On a static call to a function the `to` field will always have to be supplied in the action, like `.transact({'to': '0x....'})`. (It should also handle ENS names)

So the question is how to initialize `functions` in `factory()` but then have `functions` discover the address of the initialized contract later on an instance-call. One way to go is to initialize it once in the factory without an address (for static function calls), and then re-initialize it in `__init__` once `address` becomes available.

dylanjw:
Ive gotten past the issue with failing function calls to the contract factory types by taking @carvers suggestion and instatiating functions/events in both the contract factory and __init__.  This made it obligatory to no longer pass the contract into `ContractFunctions` and `ContractEvents` to avoid a circular reference.  To remove the dependency I needed to copy some methods from Contract to ContractMethods.  These included: `_encode_abi`, `_find_matching_fn_abi`, `_get_function_info`, ` _encode_transaction_data`, `_prepare_transaction`.  I will next work on some reworking of these methods.  Do you have any comments on these particular methods, as far as migrating them from Contract to ContractMethods?  Are they part of a separate functionality, or do they need to be shared?

carver:
One issue is that `contract.encodeABI` is a public interface that depends on `_encode_abi`, so there will have to be some way to share the logic, rather than just remove it. Some of the others will be similar, I'm sure. That would be the first thing I'd look for: what can be removed without changing public APIs.

web3/contract.py
pipermerriam: Maybe but given that it looks like a more complex change than I predicted I'll just open up an issue noting this as an area of minor debt that we can try to clean up at some point.  Skip it for now.

dylanjw: Ive already removed the need to pass in the whole contract.  It was necessary to avoid a circular reference.  Ive left the helper functions as they are by copying some contract methods to the ContractMethod class.  I tried to make the required change minimal.  There is more clean up that can be done, so maybe that can be dealt with in a new issue.

dylanjw:
Ive submitted more changes and am ready for another review.
I am going to submit some documentation later this week, and I can submit any requested changes at the same time.

carver:
The crash happening in `ens` implies that the `ConciseContract` API got broken somehow. `ConciseContract._none_addr` was expecting a call to return an address, and got `None` instead.

carver:
Also still has some flake8 errors. Running `flake8` locally is really fast, much quicker than waiting for Travis  to do it. :)

dylanjw:
@carver Ive committed changes to resolve the failing tests and flake8 errors.

carver:
Cool, nearly there. I'm concerned about the code duplication in `ContractMethod` and the maintenance burden it will add. One solution would be to extract the duplicated classmethods from `Contract` into pure utility functions, and call the same functions from both `Contract` and `ContractMethod`. 

For example: `_find_matching_fn_abi()` can be converted pretty easily like so:

- add an `abi` argument
- in the method body, replace `cls.abi` with `abi`
- move it to a standalone method `find_matching_fn_abi()`, in a new module `web3/utils/contracts.py`.

web3/contract.py
```diff
@@ -589,38 +592,27 @@ def __getattr__(self, function_name):
...
-            elif not function_candidates:
-                break
+        return prepare_transaction(cls.abi,
+                                   cls.address,
+                                   cls.web3,
+                                   fn_name=fn_name,
+                                   fn_args=fn_args,
+                                   fn_kwargs=fn_kwargs,
+                                   transaction=transaction)
+
```
carver: This line is causing the `flake8` errors. Deleting it will fix both errors. Once travis-ci passes, and we should be good to merge.

dylanjw:
@carver Ready!

carver:
Thanks @dylanjw !

PR #482, opened 2017-12-02
Fix from pull request #470 didn't work

Mikerah:
### What was wrong?
After submitting #470, I thought I would be able to use web3.py without having to manually install `pypiwin32`. However, this was not the case. The suggestion from the stackoverflow post in #470 didn't work. 

### How was it fixed?
I looked up the [documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies) for pip and there are 2 options:
..* Add an optional key, so that if the user is on Windows, install web3.py like `pip install web3['win']`
..* Add an extra value to the install requires list

The second option is what was implemented before #470. Thus, I decided to do the first option. Also, I think the first option is better because in the future, there might be more windows specific dependencies. To add any new dependencies, it simply requires adding the module to the list and the user installs as usual.

I also added how to install web3 if on a windows machine to the `Quickstart` documentation page.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -36,7 +36,11 @@ Or to install with support for the ``TestRPCProvider`` and
 
    $ pip install web3[tester]
 
-
+Installing Web3.py on Windows using ``pip`` can be done as follows.
+.. code-block:: shell
+   
+   $ pip install['win']
+   
 Installation from source can be done from the root of the project with the
 following command.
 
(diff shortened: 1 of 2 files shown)


pipermerriam:
I think this warrants going ahead and getting windows based CI setup.

Mikerah:
Any updates on this pull request?

carver:
I'd like to see this pass a Windows CI before merging, so it's blocking on #166

carver:
This can close after #789 is merged.

carver:
As of #789 pypiwin32 is installed by default on windows systems. See: https://github.com/ethereum/web3.py/blob/4db93fa8cca8bd4bcb6e375a625af5bbc378faa4/setup.py#L30

PR #487, opened 2017-12-04
New AutoProvider

### LINKED ISSUE #485: Auto-detect providers if none provided to Web3
* Version: 4

### What was wrong?

More and more functions are valuable even if not connected to a node. Initializing with no providers available would be helpful to access those functions.

Additionally, it would be natural to try to auto-detect providers if none are provided, equivalent to how `from web.auto import w3` works.

#### Todo

- [x] Review/improve error message when trying to connect to one of the providers and it is unavailable.
- [x] add default value to `providers` in `Web3` and `Manager`, to put the request manager in a special auto-detect mode
- [x] move autodetection logic from `web3.auto` into `web3.providers.auto`
- [x] auto-detect mode: save last-working provider to reattempt first
- [x] auto-detect mode: if no previously-working provider, or last-working is unavailable, re-attempt detection
~- [ ] auto-detect mode: if failure to detect, raise an exception describing it different from the first case~
- [x] auto-detect mode: look up environment variable first
- [x] replace `web3.auto.w3` implementation with `w3 = Web3()`
- [x] map environment URI to provider

#### Specifying No Providers

To prevent connection to any node, allow initialization of `Web3` with `None` provider specified. This is already technically supported, it just fails with unhelpful error messages at the moment.

```
from web3 import Web3
w3 = Web3(providers=None)
# or simply
w3 = Web3(None)
```

If any call attempt is made, it will always fail with an error saying that `Web3` explicitly disallows node interaction when started with `Web3(providers=None)`.

#### Auto-detecting Providers

To auto-detect providers, allow initialization of `Web3` with no provider specified:

```
from web3 import Web3
w3 = Web3()
```

A `Web3` instance is returned immediately, and can be used for anything that does not interact with the node.

The first time a call to a node is attempted, it tries to auto-detect a node. If one is found, save it for later use and automatically reuse it. In this mode, any time the connection fails, it will try to reconnect to any node that it can find.

#### Auto-detection in `web3.auto.w3`

```
from web3.auto import w3
```

Will now be exactly equivalent to:

```
from web3 import Web3
w3 = Web3()
```

That means two changes:

1. If no nodes are found, the import still succeeds and works for functions that don't need node interaction
2. If a node is not found at import time, `w3` will attempt to re-detect it on every attempt to call to the node

#### Explicitly Setting Providers

*No change in functionality*

For the normal use case (previously, the only explicitly supported use case), `Web3` would only attempt to use the providers specified. If none of them are available, it will **not** attempt to auto-detect another one. (If you try to connect to a testnet node and fail, we don't want to automatically connect you to a production node without you realizing)

carver:
### What was wrong?

Auto-providers were still a little fiddly.

### How was it fixed?

New `AutoProvider` that proxies to whatever provider it can find.

Closes #485 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -1,0 +1,40 @@
+from contextlib import contextmanager
+import os
+import pytest
+
+from web3.providers.auto import (
+    load_provider_from_environment,
+)
+from web3.providers import (
+    IPCProvider,
+    HTTPProvider,
+)
+
+
+@contextmanager
+def env(name, tempval):
+    oldval = os.environ.get(name, None)
+    os.environ[name] = tempval
+    try:
+        yield
+    finally:
+        if oldval is None:
+            del os.environ[name]
+        else:
+            os.environ[name] = oldval
+
+
+@pytest.mark.parametrize(
+    'uri, expected_type, expected_attrs',
+    (
+        ('', type(None), {}),
+        ('http://1.2.3.4:5678', HTTPProvider, {'endpoint_uri': 'http://1.2.3.4:5678'}),
+        ('file:///root/path/to/file.ipc', IPCProvider, {'ipc_path': '/root/path/to/file.ipc'}),
+    ),
+)
+def test_load_provider_from_env(uri, expected_type, expected_attrs):
+    with env('WEB3_PROVIDER_URI', uri):
+        provider = load_provider_from_environment()
+        assert isinstance(provider, expected_type)
+        for attr, val in expected_attrs.items():
+            assert getattr(provider, attr) == val
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -1,5 +1,3 @@
-from contextlib import contextmanager
-import os
 import pytest
 
 from web3.providers.auto import (
@@ -11,19 +9,6 @@ from web3.providers import (
 )
 
 
-@contextmanager
-def env(name, tempval):
-    oldval = os.environ.get(name, None)
-    os.environ[name] = tempval
-    try:
-        yield
-    finally:
-        if oldval is None:
-            del os.environ[name]
-        else:
-            os.environ[name] = oldval
-
-
 @pytest.mark.parametrize(
     'uri, expected_type, expected_attrs',
     (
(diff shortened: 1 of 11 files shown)


web3/providers/auto.py
```diff
@@ -0,0 +1,53 @@
...
+
+def load_provider_from_environment():
+    uri_string = os.environ.get('WEB3_PROVIDER_URI', '')
+    uri = urlparse(uri_string)
+    if uri.scheme == 'file':
+        return IPCProvider(uri.path)
+    elif uri.scheme == 'http':
+        return HTTPProvider(uri_string)
+    else:
+        return None
```
pipermerriam: maybe `raise NotImplementedError`? or something other than a silent return of `None`?

web3/providers/auto.py
```diff
@@ -0,0 +1,53 @@
...
+            provider = Provider()
+            if provider is not None and provider.isConnected():
+                self._active_provider = provider
+                return provider
+
+        return None
+
+
+def load_provider_from_environment():
+    uri_string = os.environ.get('WEB3_PROVIDER_URI', '')
```
pipermerriam: Maybe do an early return here if `WEB3_PROVIDER_URI` is not set?

pipermerriam:
Tests?

carver:
I had trouble coming up with a meaningful way to test this, since it's looking in the real locations for clients. I don't want to accidentally pick up real clients when testing on a local box, and I don't want to create fake client endpoints at the real locations. So for now I just tested the empty (no provider found) example.

carver:
Ok, took almost that approach, but instead of subclasses, it optionally initializes the list of providers in `AutoProvider`.

carver:
@pipermerriam look good to merge?

tests/core/providers/test_auto_provider.py
```diff
@@ -0,0 +1,40 @@
...
+    load_provider_from_environment,
+)
+from web3.providers import (
+    IPCProvider,
+    HTTPProvider,
+)
+
+
+@contextmanager
+def env(name, tempval):
```
pipermerriam: You're looking for the pytest `monkeypatch` fixture which does this natively:

https://docs.pytest.org/en/latest/monkeypatch.html

tests/core/providers/test_auto_provider.py
```diff
@@ -32,9 +17,9 @@ def env(name, tempval):
...
     ),
 )
-def test_load_provider_from_env(uri, expected_type, expected_attrs):
-    with env('WEB3_PROVIDER_URI', uri):
-        provider = load_provider_from_environment()
-        assert isinstance(provider, expected_type)
-        for attr, val in expected_attrs.items():
-            assert getattr(provider, attr) == val
+def test_load_provider_from_env(monkeypatch, uri, expected_type, expected_attrs):
+    monkeypatch.setenv('WEB3_PROVIDER_URI', uri)
```
pipermerriam: I ❤️ this fixture so much.  Also look at `tmpdir` if you haven't already.

carver: Yeah, these are great. Such a concise way to set up the test environment.

PR #490, opened 2017-12-05
 enable block filter test, and bugfix BlockFilter

carver:
### What was wrong?

BlockFilter (ie~ `eth.filter('latest')`) was broken.

### How was it fixed?

Depends on #489 

- removed 'skip' in the block test
- set up EthereumTester to work with `wait_for_block()`
- bugfix `BlockFilter` (move several required functions from `LogFilter` up to superclass `Filter`)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/filtering/test_filter_against_latest_blocks.py
+++ b/tests/core/filtering/test_filter_against_latest_blocks.py
@@ -1,54 +1,28 @@
-from flaky import flaky
-
+from web3.providers.eth_tester import EthereumTesterProvider
 from web3.utils.threads import (
     Timeout,
 )
 
 
-@flaky(max_runs=3)
 def test_sync_filter_against_latest_blocks(web3, sleep_interval, wait_for_block, skip_if_testrpc):
-    skip_if_testrpc(web3)
+    if EthereumTesterProvider not in map(type, web3.providers):
+        web3.providers = EthereumTesterProvider()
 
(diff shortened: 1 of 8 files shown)


tests/core/filtering/test_filter_against_latest_blocks.py
```diff
@@ -1,54 +1,28 @@
-from flaky import flaky
-
+from web3.providers.eth_tester import EthereumTesterProvider
 from web3.utils.threads import (
     Timeout,
 )
 
 
-@flaky(max_runs=3)
 def test_sync_filter_against_latest_blocks(web3, sleep_interval, wait_for_block, skip_if_testrpc):
```
pipermerriam: looks like `skip_if_testrpc` is no longer used?

web3/utils/filters.py
```diff
@@ -100,17 +99,22 @@ def _filter_valid_entries(self, entries):
...
 
     def get_all_entries(self):
-        self._ensure_not_running("get_all_entries")
-
         log_entries = self._filter_valid_entries(self.web3.eth.getFilterLogs(self.filter_id))
         return self._format_log_entries(log_entries)
 
+    def _format_log_entries(self, log_entries=None):
+        if log_entries is None:
+            log_entries = []
```
pipermerriam: Should this just `return []`?

Issue #409, opened 2017-11-05
Remove pylru as a dependency


carver:
It is GPL-licensed, and it is barely used. Not worth the licensing uncertainty it brings.

If an lru cache is really the best way to manage session state, maybe an alternative like: https://github.com/amitdev/lru-dict

But probably something like adding the lru cache from `functools` on generate_cache_key. (Which is now an option, since we are on Python 3.)

gitcoinbot:
__This issue now has a funding of 0.15 ETH (49.43 USDT) attached to it.__  To view or claim this funding, [click here](https://gitcoin.co/funding/details?url=https://github.com/pipermerriam/web3.py/issues/409).

Sebohe:
I would like to take this on. :+1:

pipermerriam:
@Sebohe all yours.

owocki:
@Sebohe great to have you.  want to claim the issue at https://gitcoin.co/funding/details?url=https://github.com/pipermerriam/web3.py/issues/409 ?

owocki:
@Sebohe you still workin on this?

jasonrhaas:
If @Sebohe isn't around, I can take this one on.

Sebohe:
Yeah I'm still working on it, did I miss something??

owocki:
@Sebohe dont thinik so.  we were just checking in since it's been 13 days since we heard from ya on this thread.

owocki:
hey @Sebohe -- just checking in to see if there's any progress.  no pressure, but just making sure you're still interested bc if not id give it to someone else.  thanks :)

Sebohe:
@owocki ^ I couldn't get @functools.lru_cache to work... We will see how the PR goes.

pipermerriam:
I think the only thing left is one more PR against the v3 branch and this can be closed.

carver:
Ready to pay out @owocki !

owocki:
submitting the bounty payment now.  while you wait for the tx to clear, you are welcome to stare at this kitty: 

![https_ 2f 2fblueprint-api-production s3 amazonaws com 2fuploads 2fstory 2fthumbnail 2f64318 2f433830cd-6b06-4d7d-b704-48f66ba8ab11](https://user-images.githubusercontent.com/513929/33689126-7dd7d656-da9b-11e7-9648-50eda191201c.png)

owocki:
[...]

PR #491, opened 2017-12-06
Lru issue 409

Sebohe:
### What was wrong?

This PR is for issue https://github.com/ethereum/web3.py/issues/409

I tried to implement functools.lru_cache but this would occur: `TypeError: unhashable type: 'dict'` on the wrapped generating_cache_key function.

### How was it fixed?

Changed requirement from pylru to lru-dict

#### Cute Animal Picture

Cute animal picture


carver:
Looks good. @Sebohe can you also submit this back to the `v3` branch?

Issue #503, opened 2017-12-09
Transaction receipt events not getting decoded properly


gakonst:
* Version: 4.0.0b4
* Python: 3.5 and 3.6
* OS: Ubuntu 17.04

### What was wrong?
Expected for the get_new_entries() method to catch the event. Instead it seems to not be able to decode the transaction receipt. I have tested the error with ethereumjs-testrpc (ganache-cli), geth --dev mode, full node running on private chain (custom genesis.json) and with eth-tester

### Installed python packages (output of pip freeze after installing pyethereum/web3 in fresh venv)
```
bitcoin==1.1.42
certifi==2017.11.5
cffi==1.11.2
chardet==3.0.4
cytoolz==0.8.2
eth-abi==0.5.0
eth-keyfile==0.4.0
eth-keys==0.1.0b3
eth-tester==0.1.0b5
eth-utils==0.7.1
ethereum==1.6.1
idna==2.6
lru-dict==1.1.6
pbkdf2==1.3
py-solc==2.1.0
pycparser==2.18
pycryptodome==3.4.7
pyethash==0.1.27
pysha3==1.0.2
PyYAML==3.12
repoze.lru==0.7
requests==2.18.4
rlp==0.6.0
scrypt==0.8.0
secp256k1==0.13.2
semantic-version==2.6.0
toolz==0.8.2
typing==3.6.2
urllib3==1.22
web3==4.0.0b4
```

## Script Output with error
I print the ABI first, just to make sure it's getting encoded properly. Then, the get_all_entries() function returns an empty list, as expected. After making a call that should fire the event, I get a proper transaction receipt, however the get_new_entries() function is unable to decode it and return a populated list.

```py
[{'constant': False, 'inputs': [], 'name': 'test', 'outputs': [], 'payable': False, 'stateMutability': 'nonpayable', 'type': 'function'}, {'anonymous': False, 'inputs': [{'indexed': True, 'name': 'x', 'type': 'uint256'}], 'name': 'Test', 'type': 'event'}]
[]
[...]

PR #504, opened 2017-12-10
Fixes event logging bug in utils/events.py

gakonst:
Fixes: https://github.com/ethereum/web3.py/issues/503

### What was wrong?
Events were not getting properly decoded due to use of camelCase instead of snake_case

### How was it fixed?
Changed all occurences of camelCase to snake_case, e.g. logIndex to log_index

#### Cute Animal Picture
le-poodle

### DIFF
--- a/web3/providers/eth_tester/middleware.py
+++ b/web3/providers/eth_tester/middleware.py
@@ -197,6 +197,9 @@ ethereum_tester_middleware = construct_formatting_middleware(
         'eth_newFilter': integer_to_hex,
         'eth_newBlockFilter': integer_to_hex,
         'eth_newPendingTransactionFilter': integer_to_hex,
+        'eth_getLogs': apply_formatter_to_array(log_key_remapper),
+        'eth_getFilterChanges': apply_formatter_to_array(log_key_remapper),
+        'eth_getFilterLogs': apply_formatter_to_array(log_key_remapper)
     },
 )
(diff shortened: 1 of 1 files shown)


gakonst:
The tests in travis are failing because they are setup to work with TestRPCProvider instead of EthereumTesterProvider, should this go to a different issue?

pipermerriam:
This is fixing the wrong thing.  We need to fix the middleware that the `eth-tester` provider uses to translate to the web3.py version of the data.

In this file: https://github.com/ethereum/web3.py/blob/master/web3/providers/eth_tester/middleware.py

We need to add three new entries to the `results_formatters` for `eth_getLogs`, `eth_getFilterChanges`, and `eth_getFilterLogs` that apply the `log_key_remapper` function.  

Looks like there are some other deficiencies that can be addressed at the same time.  We probably will need a new `log_entry_formatter` to normalize any of the fields in log entries.   Also, that same formatter and remapping needs to be applied to the `logs` field of the receipt which doesn't currently appear to be happening.

pipermerriam:
I just pushed f0f8282 which accounts for the fact that the return from a filter is *sometimes* an event log, but for `pending` and `latest` filters they will be transaction or block hashes.

pipermerriam:
@carver this is good to merge/release  (should also be backported to v3)

Issue #496, opened 2017-12-06
Better gas price engine algorithm


pipermerriam:
Depends on #494 

### What needs to be done?

The algorithm used by https://ethgasstation.info/ is pretty solid for determining gas prices.  Lets implement something similar

### How can it be done?

Implement a new gas price engine (see #494 ) that uses the following algorithm to compute gas prices.

The algorithm will be configured with the following parameters:

- Let `N` be the number of historical blocks that should be sampled.
- Let `P` be a number between [0-1] which represents the probability for gas price estimation.  0 meaning 0% probability and 1 meaning 100% probability.
- Let `D` be the desired maximum wait time in seconds for the transaction to be mined.

We then compute the following information by fetching data from the blockchain.

- Let `A` be the *average* block time for the last `N` blocks.
- Let `M` be the set of miners from the last `N` blocks.

For each miner in `M` compute:

- Let `m.totalBlock` be the number of blocks mined by a given miner.
- Let `m.minPrice` be the minimum gas price accepted for all transactions in all blocks mined by the given miner.
- Let `g_max` be the maximum gas price for all `m.minPrice` from the set of miners `M`.
- Let `g_min` be the minimum gas price for all `m.minPrice` from the set of miners `M`.

Now we compute `B`, the maximum wait time in blocks using the formula `int(ceiling(D / A))`.

Now we can use our computed data from the miners to compute the set of probabilities that a transaction will be included at each miner's min price.  For a miner `x` let `h` be the total number of blocks mined by all miners where `m.minPrice <= x.minPrice`.  The probability is computed using the formula `(1 - (N - h)/N) ** B`.  Using this formula we can compute the set of probabilities `p0, p1, ...` for each miner.  Note that each probability `p0` corresponds to a miner `m0`.  We also assume that these probabilities are ordered such that `p0 < p1, p1 < p2, ...`.

Now compute the following values.

- Let `p_max` be the maximum probability from the set of probabilities `p0, p1, ...`.
- Let `p_min` be the minimum probability from the set of probabilities `p0, p1, ...`.

Now, to compute the gas price for a transaction to be included within `B` blocks with probability `P` as follows.

- If `P` > `p_max`, the gas price is `g_max`
- If `P` < `p_min`, the gas price is `g_min`

Otherwise, we find the location within `p0, p1, ...` such that `pn <= P <= pn+1`.  The desired gas price can then be computed as follows.

If `pn == pn+1` then the desired gas price is mn.minPrice.  Otherwise...

[...]

PR #514, opened 2017-12-14
Gas price engine API

### LINKED ISSUE #494: Gas price engine API
### What was wrong?

Currently, web3.py just uses the `eth_gasPrice` JSON-RPC API to determine the gas price to use for transactions.  Most nodes do not currently implement very good algorithms for determining gas prices.

### How can it be fixed?

Lets implement a new API for determining gas prices.

* Add a new API: `Web3.set_gas_price_engine(engine)`
* Add new API: `Web3.get_gas_price(transaction_params)` which delegates to whatever *engine* is set.

An `engine` must be a callable with the function signature:

```
engine(w3, transaction_params)
```

* `w3` will be the `Web3` instance
* `transaction_params` will be a dictionary of the transaction params being sent.

The callable **must** return a positive integer which represents the `gas_price` in the *wei* denomination.

The **default** engine that a web3 instance should use is one that simply returns `web3.eth.gasPrice`

Documentation should be added covering the following topics.

* Overview of the `Web3.set_gas_price_engine()` API (maybe in `docs/overview.rst`)
* Writing your own gas price engine (maybe in new document `docs/gas_pricing_engine.rst` or `docs/web3.main.rst`)
* List of available engines, where to import them and how to use them.

monokh:
### What was wrong?

Currently, web3.py just uses the eth_gasPrice JSON-RPC API to determine the gas price to use for transactions. Most nodes do not currently implement very good algorithms for determining gas prices.

### How was it fixed?

Add an API that allows adding of custom gas price "engines" to allow various implementations of calculating gas prices.

#### TODOs

- [x] Add skeleton
- [x] Confirm API and package structure
- [x] Implement the default gas price engine
- [x] Extend tests
- [x] Add a "gas_price" middleware that includes the gas price when a gas pricing strategy has been set.
- [x] Add documentation about new API in `docs/web3.eth.rst`
- [x] Add documentation about creating your own API with simple example `gas_price.rst`. Should also include list of available engines

#### Cute Animal Picture

Cute animal picture

Fixes #494


web3/utils/gas_price_engines.py
```diff
@@ -0,0 +1,4 @@
+
+
+def default_gas_price_engine(web3, transaction_params):  # TODO: call eth.gasPrice
```
pipermerriam: What do you think about moving this to `web3.gas_engine.rpc` and renaming it to `rpc_gas_price_engine`.

web3/main.py
```diff
@@ -182,3 +186,15 @@ def ens(self):
     @ens.setter
     def ens(self, new_ens):
         self._ens = new_ens
+
+    @property
+    def gasPriceEngine(self):
```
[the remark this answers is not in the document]
monokh: Sure, just couldn't decide what the convention was 😄 
Will make this a plain property. If we do need to add extra functionality on get/set then we could always add the getter/setter back in without impacting consumers.

web3/main.py
```diff
@@ -182,3 +186,15 @@ def ens(self):
...
+        return self._gas_price_engine
+
+    @gasPriceEngine.setter
+    def gasPriceEngine(self, new_gas_price_engine):
+        self._gas_price_engine = new_gas_price_engine
+
+    def getGasPrice(self, transaction_params={}):
```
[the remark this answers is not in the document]
monokh: Wow I did not know that 😅  will do that

web3/utils/gas_price_engines.py
monokh: I'm not sure I understand what the `rpc` part of the package and module names really mean?  I'm happy to give this it's own package though

On the subject of naming, how do you feel about an alternative name "gas pricing strategy"?

pipermerriam: I'm not in love with the name `rpc` but the idea was that this engine get's the price via the JSON-RPC connection.  `default` is a tempting name but I was hoping to come up with a name that was more descriptive of what it actually does.  Maybe `rpc` is worse.

I'm good going forward with `default` as the namespace.

> On the subject of naming, how do you feel about an alternative name "gas pricing strategy"?

I like it.  Make it so!

monokh: Oh i see what you mean. I've changed it to `rpc_gas_pricing_strategy`. I'll have a think if there's a better name that talks more accurately but this should do for now

monokh:
@pipermerriam what shall we do in terms of any extra gas pricing strategies? Is #496 in a state where it can be used?

web3/gas_strategies/rpc.py
```diff
@@ -0,0 +1,11 @@
...
+TODO: Provide explanation of gas price engines
+"""
+
+
+def rpc_gas_pricing_strategy(web3, transaction_params=None):  # TODO: call eth.gasPrice
+    """
+    This is the default gas pricing strategy. It derives it's value from the eth_gasPrice
+    JSON-RPC call.
+    """
+    return web3.eth.gasPrice
```
pipermerriam: I'd like to make a small change.

* Change this function body to call directly out to the JSON-RPC endpoint

```python
def rpc_gas_pricing_strategy(...):
    web3.manager.request_blocking("eth_gasPrice", [])
```

Then, we can replace the current code in `web3.eth.gasPrice` with code that delegates to `self.web3.getGasPrice()`.  This gives us one definitive place to query gas pricing.

monokh: Good thinking 😃  changed.

monokh:
@pipermerriam i've mentioned in the docs:

> With Web3, the gas price included in a transaction (if not explicitly set) is 
determined by the selected "Gas Pricing Strategy".

I initially thought that this would happen when calling `Web3.eth.sendTransaction` but it doesn't. It happens with `contract.buildTransaction` and with `signTransaction` but not in this case. Only the gas is calculated in `Web3.eth.sendTransaction`. Do you agree that gasPrice should be too? If so, should I include in this MR or a follow up?

pipermerriam:
> I initially thought that this would happen when calling `Web3.eth.sendTransaction` but it doesn't. It happens with `contract.buildTransaction` and with `signTransaction` but not in this case. Only the gas is calculated in `Web3.eth.sendTransaction`. Do you agree that gasPrice should be too? If so, should I include in this MR or a follow up?

@carver can you weigh in on this as well?

Here is what seems most *correct* to me.  Curious to hear what others think.

- Change API so that:
  - returning `None` from `web3.get_gas_price()` indicates *nothing* should happen.
  - otherwise the `gas_price` portion of the transaction dictionary should be filled in with the returned value.
- Change `sendTransaction` to populate the `gas_price` of transactions using this API.
- Make sure *other* transaction sending APIs do this as well (e.g. `web3.personal.sendTransaction`, `web3.eth.contract.buildTransaction`).

I also think this implies that the current *default* strategy `rpc_gas_price_strategy` should just be removed.  Thoughts?

pipermerriam:
@monokh I just want to apologize for the churn on this issue.  I hope you aren't annoyed.  This has turned out to be one of those APIs that has built-in subtlety that I failed to see up-front.  I'll see about upping the bounty on this to account for the un-expected extra difficulty.

web3/main.py
```diff
@@ -187,6 +187,5 @@ def ens(self):
     def ens(self, new_ens):
         self._ens = new_ens
 
-    def getGasPrice(self, transaction_params=None):
-        # TODO: sanity checks?
+    def get_gas_price(self, transaction_params=None):
```
carver: Is gas price a concept that belongs on the global `Web3` class? I don't think it has anything to do with swarm or whisper or other protocols that Web3 interacts with, it seems like an `eth`-only concept, and belongs in that box.  (I'm really asking: not sure if gas applies to anything outside the core ethereum protocol).

I don't think we're perfectly consistent on that distinction, but would rather avoid making it less consistent than it currently is.

monokh: I agree, I felt it was out of place as well but decided to go with it. Happy to move to `Web3.eth.` if you are too @pipermerriam

monokh:
@pipermerriam Oh no problem at all. The bounty is just an eye catcher, i'm here to do something interesting and learn so the more twists and turns the better 😄 

About the API, i'm slightly confused.
> returning None from web3.get_gas_price() indicates nothing should happen.
What would be the function of this method then? Do you mean that it would return None as a default and the result of the gas pricing strategy if set?

> Change sendTransaction to populate the gas_price of transactions using this API.
Would it do this using get_gas_price()? It would seem the logical path to me so to centralise the defaulting of the gas price.

> the current default strategy rpc_gas_price_strategy should just be removed
Not sure i understand why. Do you mean that if get_gas_price() returned none, we would just use `eth_gasPrice` directly in `sendTransaction` (e.g.)?

Also another question I had was if you think sendTransaction should *not* add a gasPrice by default. Is the idea to keep expectations the same since it wasn't doing this before?

I might be more than slightly confused 😄

web3/main.py
pipermerriam: My bad, happy to have it moved to `web3.eth` as well.

pipermerriam:
> Also another question I had was if you think sendTransaction should not add a gasPrice by default. Is the idea to keep expectations the same since it wasn't doing this before?

I think this captures what I was thinking.  Let me lay it out in my own words and we can see if we're in agreement of what should happen.

- The *default* behavior of `Web3` remains to **not** populate `gasPrice` for transactions.
- *if* a strategy has been set, `Web3` should use the strategy to populate the `gasPrice` field for a transaction.

I'm indifferent to exactly how this gets done.  You can ignore my previous attempt to *specify* this API with the return `None` stuff.

monokh:
@pipermerriam thanks, that makes sense :slightly_smiling_face:. I'll make the changes when I get some time tomorrow.

carver:
Sorry I didn't get a chance to weigh in until now. It looks like you got what you need for the next steps.

> Also another question I had was if you think sendTransaction should not add a gasPrice by default. Is the idea to keep expectations the same since it wasn't doing this before?

> The default behavior of Web3 remains to not populate gasPrice for transactions.

I strongly agree with the conclusion you came to: the default behavior should not change. The default behavior is to leave the gasPrice empty (and let the node decide what gas price to use). 

Which brings us to the next question, how do we change the gas price engine? Right now it looks like the idea is to set `w3.gas_price_engine = new_engine_func`. That's okay, if a little different style from the rest of Web3 (and should probably be on `w3.eth.gas_price_engine`).

pipermerriam:
> and should probably be on `w3.eth.gas_price_engine`

@monokh I'd be good with either of the following APIs or a 3rd option if you have ideas on the subject

- `web3.eth.gas_price_strategy = my_gas_price_strategy`
- `web3.eth.set_gas_price_strategy(my_gas_price_strategy)`

monokh:
I do like `set_gas_price_strategy()` ( is it price? is it pricing? :joy: ). The only thing to note then is that then  `gas_pricing_strategy` should be a private property, to avoid multiple ways of setting it (?). The consequence of that: no clear way to "get" the price strategy but I can't think of a use-case for this anyway. Maybe also a `get_gas_pricing_strategy()` ?

pipermerriam:
Then I'm inclined to go with

- `eth.set_gas_price_strategy(...)`
- sets to `eth._gas_price_strategy`
- externally exposed with `eth.get_gas_price(transaction_params)`
- *if* there is no strategy set, `eth.get_gas_price(...)` should either throw an exception or return `None` (I don't know which is better, see what you think is best).

monokh:
One more thing... I've been moving this into the `web3.eth` package and feel the difference in naming convention really pops out here... I feel from a usability perspective, it would be best to follow the camelCase convention. Thoughts?

pipermerriam:
Yeah, I'm fine with `web3.eth.getGasPrice`.

@carver any thoughts on alternate names for this API since it somewhat conflicts with what would be the non-property based accessor for `web3.eth.gasPrice`.

pipermerriam:
@monokh given the simplicity of renaming the API, go with what you think is best for now.

carver:
> I do like set_gas_price_strategy() ( is it price? is it pricing? :joy: )

It could make sense either way. I think of "price strategy" as a little more static: how to choose a price. "Pricing strategy" would be a bit more meta: choosing how to choose a price. With that intepretation, I guess "pricing strategy" would be a discussion of which "price strategy" to use, so the natural representation would be `setGasPriceStrategy`.

(The first time I wrote the above paragraph, I came to the opposite conclusion and deleted it because this one makes more sense... But since it's highly subtle, they're both good enough that I have no objection to whichever you choose @monokh . :) )

> @carver any thoughts on alternate names for this API since it somewhat conflicts with what would be the non-property based accessor for web3.eth.gasPrice.

Maybe:

- `w3.eth.estimateGasPrice`
- `w3.eth.guessGasPrice`
- `w3.eth.generateGasPrice`

monokh:
@carver I'll use `w3.eth.setGasPriceStrategy` 😃 

In terms of names for getting gas price, i see `w3.eth.generateGasPrice` being nicer.

I'm feeling uncomfortable in finding and going through every part of the code that sends a transaction and adding gas price strategy execution. Wasn't this the problem that middlewares are solving?

i.e. couldn't the gas price be set with a middleware that adds a gas price if a strategy is set and a gas price has not? Maybe I have a different idea about what the middleware is.

pipermerriam:
> I'm feeling uncomfortable in finding and going through every part of the code that sends a transaction and adding gas price strategy execution. Wasn't this the problem that middlewares are solving?

@carver thoughts?  New default middleware?

pipermerriam:
> In terms of names for getting gas price, i see `w3.eth.generateGasPrice` being nicer.

+1

web3/eth.py
```diff
@@ -313,3 +314,11 @@ def getCompilers(self):
 
     def getWork(self):
         return self.web3.manager.request_blocking("eth_getWork", [])
+
+    def getGasPrice(self, transaction_params=None):
+        if self.gasPriceStrategy:
+            return self.gasPriceStrategy(self, transaction_params)
+
+    def setGasPriceStrategy(self, gas_price_strategy):
+        # TODO: check passed method has 2 parameters and show handy error
```
pipermerriam: Given that this is a low level API I think it's ok to skip this check.  Also, it should be allowed for someone to call `web3.eth.setGasPriceStrategy(None)` to remove the strategy unless there is already an alternate way to unset the strategy that I'm not seeing.

monokh: ok :)  makes sense

monokh:
Added the gas price middleware, feels quite neat now. What do you think? 😃

web3/middleware/gas_price.py
```diff
@@ -0,0 +1,18 @@
...
+
+def gas_price_middleware(make_request, web3):
+    """
+    Includes a gas price using the gas price strategy
+    """
+    def middleware(method, params):
+        if method == 'eth_sendTransaction':
+            transaction = params[0]
+            generated_gas_price = web3.eth.generateGasPrice(transaction)
+            if 'gasPrice' not in transaction and generated_gas_price:
```
pipermerriam: I think the check should be explicitly `... and generated_gas_price is not None` to allow the subtle edge case of returning `0` as the gas price which while strange, should be a valid response.

pipermerriam: I think that the check for `if 'gasPrice' not in transaction` should happen prior to the call to `web3.eth.generateGasPrice(...)` so that we can avoid the effort of computing the gas price if one was explicitly specified.

web3/middleware/gas_price.py
```diff
@@ -0,0 +1,18 @@
+from cytoolz.dicttoolz import (
+    assoc,
+)
+
+
+def gas_price_middleware(make_request, web3):
```
pipermerriam: naming nitpick

can we change this to `gas_price_strategy_middleware`

monokh:
Good catches 😃  updated

pipermerriam:
@carver LGTM, I'll leave it up to you on whether to click merge.

benediamond:
I'm a very late comer to this game, but have any of you noticed that `generateGasPrice()`, even under `fast_gas_price_strategy`, generates gas prices that are typically well below the SafeLow gas price at https://ethgasstation.info/ (which appears to be reliable)?

pipermerriam:
No, but I suspect this API is not currently being very heavily used so not surprising if it has some bugs built into it.

carver:
Let's move discussion to #894

Issue #493, opened 2017-12-06
Middleware to sign transactions locally


carver:
* Version: 4

## What was wrong?

It would be convenient to use the `Web3.eth` interface as if you were using your node's account management features, even if the key is only available locally.

Basically, we want this to work:

```
from web3.auto import w3

# some middleware setup on w3, with private key 0x000...0001

w3.eth.sendTransaction({ ... normal transaction ... , 'from': '0xAccountForPrivateKey1'})
```

## How can it be fixed?

**New Local Middleware**

A reasonable approach is to add a new middleware that is configured with a private key. It would look for transactions from the account with the associated private key, sign it locally, and send it to the node already signed.

pipermerriam:
@jasonrhaas this is the issue I mentioned.

jasonrhaas:
I will take a look at this as soon as I can get my head above water on my other work :)

pipermerriam:
> I will take a look at this as soon as I can get my head above water on my other work :)

Sounds good.  I'm going to hold it for you for two weeks.

@jasonrhaas should be considered to have exclusive rights to this issue through December 20th, 2017.  Others should feel free to express interest prior to that date.

gitcoinbot:
__This issue now has a funding of 0.42 ETH (283.5 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/493).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/493)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $10040.76 more Funded OSS Work Available at: https://gitcoin.co/explorer

monokh:
Happy to get busy with this if Jason doesn't get a chance. :slightly_smiling_face:

pipermerriam:
@monokh consider yourself *second in line*.

jasonrhaas:
Started looking at this.. first step is getting everything running.  Having some issues running the quickstart, something to do with `py-solc` or just maybe solidity isn't installed right in my docker container.  https://github.com/ethereum/py-solc/issues/40.

Anyway I'll take the nitty gritty stuff off this thread but want to let you know that I'm looking at this and will keep you all up to date on progress.

jasonrhaas:
[...]

PR #517, opened 2017-12-18
[WIP] Middleware to sign transactions locally

jasonrhaas:
### What was wrong?
https://github.com/ethereum/web3.py/issues/493

### How was it fixed?
Work in progress.  I have the test script `issue_493.py` to demonstrate the problem and the checklist to keep track of progress.  The middleware and test stubs are there, but still need to be tested.

### To do list
For `web3.middleware.signing.construct_transaction_signing_middleware.py`

#### Accept Private Key as:
- [x] raw bytes
- [ ] hex encoded string
- [x] PrivateKey object from eth-keys
- [x] LocalAccount object as returned by the web3.eth.account API

#### Do nothing if
- [x] if the method is not "eth_sendTransaction"
- [x] if the from key in the transaction object is not the same address as the one generated from  our private key

#### Checks and conversions
- [x] Check that address for transaction is the same as private key
- [x] Sign transaction using private key
- [x] Convert to raw Transaction
- [x] Add `nonce` field if it doesn't exist
- [x] Add `gas` and `gas_price` is it doesn't exist in the transaction

#### Testing
For `tests.core.middleware.test_signing_middleware.py`
- [x] Low level middle ware tests (similar to fixtures)
- [ ] Integration testing with real transactions?
- [x] Use TestRPC or just dictionaries?
- [x] Need to use a web3 instance for testing?

#### Cute Animal Picture

```
                  (      )
                  ~(^^^^)~
                   ) @@ \~_          |\
                  /     | \        \~ /
                 ( 0  0  ) \        | |
                  ---___/~  \       | |
                   /'__/ |   ~-_____/ |
    o          _   ~----~      ___---~
      O       //     |         |
             ((~\  _|         -|
       o  O //-_ \/ |        ~  |
            ^   \_ /         ~  |
                   |          ~ |
                   |     /     ~ |
                   |     (       |
                    \     \      /\
                   / -_____-\   \ ~~-*
                   |  /       \  \
                   / /         / /
                 /~  |       /~  |
```


jasonrhaas:
@pipermerriam @carver I pushed up a first cut of the new middleware.  Haven't tested it yet, but wanted to keep y'all updated on the progress.  Next steps will be adding tests and debugging.

issue_493.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Send the raw transaction
+receipt = w3.eth.sendRawTransaction(signed.rawTransaction)
+
+print(f'Transaction receipt: {receipt}')
+
+# Balances after transaction
+print('Balances after raw transaction from local account to TestRpc:')
+print(f'Local Account Balance: {w3.eth.getBalance(acct.address)}')
+print(f'TestRpc Account[0] Balance: {w3.eth.getBalance(w3.eth.accounts[0])}')
```
pipermerriam: Ooh, fancy new python3 formatting.

web3/middleware/signing.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns:
+        TYPE: Description
+    """
+
+    # Check if the private_key is either
+    # - eth_key.keys.PrivateKey
+    # - web3.eth.account.create().privateKey
+    # - bytes
+    valid_key_types = (PrivateKey, HexBytes, bytes)
```
carver: `HexBytes` inherits from `bytes`, so an explicit `HexBytes` check is not necessary here. (Although a check against `LocalAccount` would be nice)

web3/middleware/signing.py
```diff
@@ -0,0 +1,62 @@
...
+            _private_key = private_key
+            break
+
+    def transaction_signing_middleware(make_request, web3):
+        def middleware(method, params):
+            # Only operate on the `eth.sendTransaction` method
+            # Only operate on if the private key matches the public key in the transaction
+            # Note: params == transaction in this case.
+            if method is not 'eth_sendTransaction':
+                return None
```
carver: The way to "pass through" a middleware without any effect is to `return make_request(method, params)`. Returning `None` would skip the call and probably crash anything upstream looking for either a `result` or `error` key in the return `dict`.

web3/middleware/signing.py
```diff
@@ -0,0 +1,62 @@
...
+
+            transaction = params
+
+            signed = web3.eth.account.signTransaction(transaction, _private_key)
+            raw_transaction = signed.rawTransaction
+
+            if 'gas' not in raw_transaction:
+                raw_transaction['gas'] = web3.eth.estimateGas(transaction)
+            if 'gas_price' not in raw_transaction:
+                raw_transaction['gas_price'] = web3.eth.gasPrice
```
carver: The gas price key name follows [json-rpc](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sendtransaction) convention, so is named `gasPrice`.

issue_493.py
carver: Yeah, I like it! I'm sad this isn't in py3.5 so we can't use it everywhere yet :/

It's the first python string formatting approach that feels right to me.

web3/middleware/signing.py
jasonrhaas: Right -- I thought that was the case, had `None` in there as a placeholder until I verified that 👍

web3/middleware/signing.py
jasonrhaas: roger that.  Need to add `LocalAccount` in there.

jasonrhaas: OK, I remember the issue I was having.  In the transaction, just the regular address is passed, right?  The `acct.address` returns a "HexBytes" object, which looks similar to a plain string.  It seems to lose its class information.

```
In [34]: acct = w3.eth.account.create()
In [35]: transaction = {'from': acct.address, 'to': 'testrpc'}
In [37]: isinstance(acct.privateKey, web3.utils.signing.LocalAccount)
Out[37]: False
In [38]: isinstance(acct.address, web3.utils.signing.LocalAccount)
Out[38]: False
In [39]: isinstance(acct, web3.utils.signing.LocalAccount)
Out[39]: True
```

Any suggestions for an elegant way to check that the `acct.address` was in fact generated by `web3.eth.account.create()`?

jasonrhaas: Github marked this as outdated -- not true.  Mentioning @carver so it doesn't get lost in the noise :)

web3/middleware/signing.py
```diff
@@ -0,0 +1,64 @@
...
+            # Might also want to check that bytes length == 32
+            _private_key = private_key
+            break
+
+    def transaction_signing_middleware(make_request, web3):
+        def middleware(method, params):
+            # Only operate on the `eth.sendTransaction` method
+            # Only operate on if the private key matches the public key in the transaction
+            # Note: params == transaction in this case.
+            if method is not 'eth_sendTransaction':
```
pipermerriam: My python says that `is` is the wrong operator here and that this should be `!=` or `if not thing == other` since the `is` operator is supposed to operate on whether two things are the *same* which is subtly different from equality.

It appears that this does in fact work as expected but I still think it's semantically incorrect.

web3/middleware/signing.py
```diff
@@ -0,0 +1,64 @@
...
+                return make_request(method, params)
+            if params.get('from') != web3.eth.account.privateKeyToAccount(_private_key):
```
pipermerriam: since `privateKeyToAccount` involves some ECDSA operations, we should probably pre-compute the address and do a comparison against the pre-computed address.

Also, while we expect these two values to both be checksummed, it is probably a good idea to use `eth_utils.is_same_address` here since it gracefully handles comparisons of addresses which are *equal* but who's strings are not equal.

web3/middleware/signing.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Check if the private_key is either
+    # - eth_key.keys.PrivateKey
+    # - web3.eth.account.create().privateKey
+    # - bytes
+    valid_key_types = (PrivateKey, HexBytes, bytes)
+    _private_key = None
+
+    for t in valid_key_types:
+        if isinstance(private_key, t):
```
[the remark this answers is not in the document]
jasonrhaas: yup -- good call.  It accepts a tuple and returns `True` if any of the objects are instances of the class

web3/middleware/signing.py
jasonrhaas: fixed, along with the `nonce` value.

web3/middleware/signing.py
jasonrhaas: fixed 👍 .  `eth_utils` is definitely the way to go.

web3/middleware/signing.py
carver: > The acct.address returns a "HexBytes" object

That would definitely be a bug, `acct.address` should be a checksummed hex string with the account address. That's what I get at the command line in a quick test right now. Can you show how to reproduce in an issue if you are seeing `HexBytes` in `acct.address`?

> Any suggestions for an elegant way to check that the acct.address was in fact generated by web3.eth.account.create()?

`transaction['from'] == acct.address` will tell you if the `acct` has the private key for the sender of `transaction`. But I feel like I might be missing something about what you're asking, so can you rephrase if I misunderstood?

web3/middleware/signing.py
carver: Just to point out that this isn't only a hypothetical or style issue, it is pretty easy to bust the string interning* that cpython does:

```
In [80]: 'ab' is 'ab'
Out[80]: True

In [81]: 'ab' is ''.join('ab')
Out[81]: False

In [82]: 'ab' == ''.join('ab')
Out[82]: True
```

*Reuse of instances of string literals.

web3/middleware/signing.py
carver: Or `eth_utils.is_same_address` as Piper mentioned later. (Although I think there are no code paths where addresses without a checksum here is possible, it can't hurt)

jasonrhaas: @carver I misspoke.  Meant to say that the `acct.privateKey` returns the HexBytes.

```
In [7]: acct.privateKey
Out[7]: HexBytes('0xab6f1f3e4327d2cf3aac02abd898ded798a57f5deab603ac16597686eddaeeda')
In [8]: type(acct.privateKey)
Out[8]: web3.utils.datastructures.HexBytes
```

So what I was asking was the best way to check that the privateKey was generated from the `web3.eth.account.create()` method.

jasonrhaas:
Hey guys, made some significant changes on this today.  I think most of the logic in is there, although I think more test cases need to be added.  Also, I'm still getting some weird errors in some cases, see the `# Raise error?` comment.  Also, I have a few `TODO's` in there on the testing code.  Let me know what you think.

web3/middleware/signing.py
carver: Maybe the simplest interface is to do `Account.privateKeyToAccount(key).address`

issue_493.py
```diff
@@ -0,0 +1,75 @@
+"""
```
pipermerriam: Not sure if this is still in "work in progress" mode, but if not, can this script get removed?

requirements-dev.txt
```diff
@@ -5,6 +5,7 @@ flaky>=3.3.0
 flake8==3.4.1
 hypothesis>=3.31.2
 py-geth==1.10.2
+py-solc==2.1.0
```
pipermerriam: Not clear why this dependency was added?

tests/core/middleware/test_signing_middleware.py
```diff
@@ -0,0 +1,125 @@
...
+bytes_private_key = b'\x01' * 32
+hex_str_private_key = '0xd65215905d39210e8c1aba8449d0ed3a95f2bd4b8d5f1f253135986a247d3aec'
+
+
+@pytest.fixture(scope='module', params=[
+                                        # TestRPCProvider,
+                                        EthereumTesterProvider,
+                                        ])
+def w3(request):
+    # Might want to use EthereumTesterProvider instead
```
pipermerriam: Looks like this comment can be removed.

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Returns:
+        TYPE: Description
+    """
+
+    # Check if the private_key is either
+    # - eth_keys.keys.PrivateKey
+    # - web3.eth.account.create().privateKey -- can just check that its HexBytes
+    # - bytes
+    # TODO: - hexbytes
```
pipermerriam: Looks like this comment may be able to go away.

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+    valid_key_types = (PrivateKey, HexBytes, bytes)
+
+    _private_key = private_key if isinstance(private_key, valid_key_types) else None
+    if not _private_key:
```
pipermerriam: These three lines could be replaced by the following which I believe is easier to read.

```
if not isinstance(private_key, valid_key_types):
    raise ...
```

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def transaction_signing_middleware(make_request, web3):
+        def middleware(method, params):
+            # Only operate on the `eth.sendTransaction` method
+            # Only operate on if the private key matches the public key in the transaction
+            # Note: params == transaction in this case.
+            if method != 'eth_sendTransaction':
+                return make_request(method, params)
+
+            transaction = params
```
pipermerriam: I don't believe this is correct.  `params` should be a `list` with a single element which is the transaction dictionary.  I think this should be changed to `transaction = params[0]` (though it would be good to do a pre-check to ensure that `params` and `params[0]` are both well formed:

- list with a single element
- single element is a dictionary

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+            private_key_address = web3.eth.account.privateKeyToAccount(_private_key).address
+
+            if not is_same_address(transaction_from_address, private_key_address):
+                return make_request(method, params)
+
+            if 'gas' not in transaction:
+                try:
+                    transaction['gas'] = web3.eth.estimateGas(transaction)
+                except ValueError:
+                    # Raise Error?
```
pipermerriam: This error should probably bubble.  If no gas was provided and `estimateGas` fails then we have to fail hard.

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+                    transaction['gas'] = 21000
+            if 'gas_price' not in transaction:
+                transaction['gasPrice'] = web3.eth.gasPrice
+            if 'chainId' not in transaction:
+                transaction['chainId'] = 1
```
pipermerriam: This should be populated from `web3.net.version`

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+            if 'nonce' not in transaction:
+                try:
+                    transaction['nonce'] = web3.eth.getTransactionCount(transaction_from_address)
+                except InvalidAddress:
+                    # Raise error?
```
pipermerriam: This error should bubble.

web3/middleware/signing.py
```diff
@@ -0,0 +1,78 @@
...
+                    transaction['nonce'] = 1
+
+            signed = web3.eth.account.signTransaction(transaction, _private_key)
+            raw_transaction = signed.rawTransaction
```
pipermerriam: While this works, I don't think this middleware should assume property access since the `attrdict_middleware` isn't guaranteed to be present.  Should be changed to use dictionary lookup style.

issue_493.py
jasonrhaas: removing..

requirements-dev.txt
jasonrhaas: I added this so that the "quickstart" example in the README.md could be run.  It's purely a developer convenience, so they don't have to install this package to work with solidity code in python.  Can remove it if you don't want it in there.

web3/middleware/signing.py
jasonrhaas: Yea, the `estimateGas` method does bomb in my test suite.

`transaction['gas'] = web3.eth.estimateGas(transaction)` fails with a long traceback.  The end of it reads:

```
E           ValueError: b'\x1d\xa8\xbd\xea\xe6\xa5\xf5.\xa2\xca\x1fR\x0c\xc9i\xa5\xb0\xe3\x14M' is not in list

_from      = b'\x1d\xa8\xbd\xea\xe6\xa5\xf5.\xa2\xca\x1fR\x0c\xc9i\xa5\xb0\xe3\x14M'
data       = b''
gas        = None
gas_price  = None
self       = <testrpc.client.client.EthTesterClient object at 0x7f0eb873a208>
to         = b'0x82A978B3f5962A5b0957d9ee9eEf472EE55B42F1'
value      = 123
```

Looks like there is a formatting different between the `from` and the `to` fields.  Could that be the issue?

jasonrhaas:
Implemented and/or reviewed all comments.  Thanks @pipermerriam.  I think this thing is at the 80 - 90% point -- there are still some quirks that I haven't been able to work around.

dylanjw:
Work is continued in https://github.com/ethereum/web3.py/pull/649

PR #519, opened 2017-12-20
Implicit contract

fubuloubu:
### What was wrong?
Wanted a contract type that could magically figure out (from the ABI) when a method should be a call by default, or a transaction by default

### How was it fixed?
I added a subclass of `ConciseContract` that checked the ABI whether or not the method was `constant`/`pure` ('constant': True in ABI), and passed that info to a subclass of `ConciseMethod`. If it was determined to make a transaction by default, the transaction is made using the defaultAccount and zero value.

#### Cute Animal Picture
Cute animal picture

### DIFF
--- a/tests/core/contracts/test_implicit_contract.py
+++ b/tests/core/contracts/test_implicit_contract.py
@@ -61,7 +61,7 @@ def test_implicitcontract_call_default(math_contract, get_transaction_count):
 def test_implicitcontract_transact_default(math_contract, get_transaction_count):
     # Use to verify correct operation later on
     start_count = math_contract.counter()
-    assert is_integer(start_count) # Verify correct type
+    assert is_integer(start_count)  # Verify correct type
     # When a function is called that defaults to transact
     blocknum, starting_txns = get_transaction_count("pending")
     math_contract.increment()
@@ -87,7 +87,7 @@ def test_implicitcontract_call_override(math_contract, get_transaction_count):
 def test_implicitcontract_transact_override(math_contract, get_transaction_count):
     # Use to verify correct operation later on
     start_count = math_contract.counter()
-    assert is_integer(start_count) # Verify correct type
+    assert is_integer(start_count)  # Verify correct type
     # When a function is called with call override that defaults to transact
     blocknum, starting_txns = get_transaction_count("pending")
     math_contract.increment(call={})
--- a/tests/core/contracts/test_implicit_contract.py
+++ b/tests/core/contracts/test_implicit_contract.py
@@ -1,0 +1,89 @@
+import pytest
+
+from web3.contract import (
+    ImplicitContract,
+)
+
+
+@pytest.fixture()
+def math_contract(web3, MATH_ABI, MATH_CODE, MATH_RUNTIME):
+    # Deploy math contract
+    # NOTE Must use non-specialized contract factory or else deploy() doesn't work
+    MathContract = web3.eth.contract(
+        abi=MATH_ABI,
+        bytecode=MATH_CODE,
+        bytecode_runtime=MATH_RUNTIME,
+    )
+    tx_hash = MathContract.deploy()
+    tx_receipt = web3.eth.getTransactionReceipt(tx_hash)
+    math_address = tx_receipt['contractAddress']
+    # Return interactive contract instance at deployed address
+    # TODO Does parent class not implement 'deploy()' for a reason?
+    MathContract = web3.eth.contract(
+        abi=MATH_ABI,
+        bytecode=MATH_CODE,
+        bytecode_runtime=MATH_RUNTIME,
+        ContractFactoryClass=ImplicitContract,
+    )
+    return MathContract(math_address)
+
+
+@pytest.fixture()
+def get_transaction_count(web3):
+    def get_transaction_count(blocknum_or_label):
+        block = web3.eth.getBlock(blocknum_or_label)
+        # Return the blocknum if we requested this via labels
+        # so we can directly query the block next time (using the same API call)
+        # Either way, return the number of transactions in the given block
+        if blocknum_or_label in ["pending", "latest", "earliest"]:
+            return block.number, len(block.transactions)
+        else:
+            return len(block.transactions)
+    return get_transaction_count
+
+
+def test_implicitcontract_internals(math_contract):
+    # counter() is constant, so it calls by default
+    assert math_contract.counter.call_by_default
+    # return13() does not since it is not constant
+    assert not math_contract.return13.call_by_default
+
+
+def test_implicitcontract_call_default(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called that defaults to call
+    # (Verify something was returned)
+    assert math_contract.counter() is not None
+    # Check that a call was made and not a transact
+    # (Auto-mining is enabled, so query by block number)
+    assert get_transaction_count(blocknum) == starting_txns
+
+
+def test_implicitcontract_transact_default(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called that defaults to transact
+    # (Verify something was returned)
+    assert math_contract.return13() is not None
+    # Check that a transaction was made and not a call
+    # (Auto-mining is enabled, so query by block number)
+    assert get_transaction_count(blocknum) == starting_txns + 1
+
+
+def test_implicitcontract_call_override(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called with transact override that defaults to call
+    # (Verify something was returned)
+    assert math_contract.counter(transact={}) is not None
+    # Check that a transaction was made and not a call
+    # (Auto-mining is enabled, so query by block number)
+    assert get_transaction_count(blocknum) == starting_txns + 1
+
+
+def test_implicitcontract_transact_override(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called with call override that defaults to transact
+    # (Verify something was returned)
+    assert math_contract.return13(call={}) is not None
+    # Check that a call was made and not a transact
+    # (Auto-mining is enabled, so query by block number)
+    assert get_transaction_count(blocknum) == starting_txns
--- a/tests/core/contracts/test_implicit_contract.py
+++ b/tests/core/contracts/test_implicit_contract.py
@@ -4,6 +4,10 @@ from web3.contract import (
     ImplicitContract,
 )
 
+from eth_utils import (
+    is_integer,
+)
+
 
 @pytest.fixture()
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -4,6 +4,7 @@ ethereum>=1.6.1,<2.0
 flaky>=3.3.0
 flake8==3.4.1
 hypothesis>=3.31.2
+lru-dict>=1.1.6
```
pipermerriam: this requirement is already present in `setup.py`.  You just need to re-install the package locally to pickup the new requirement.

web3/contract.py
```diff
@@ -764,9 +765,48 @@ def __prepared_function(self, **kwargs):
...
+
+    > contract.withdraw(amount)
+
+    is equivalent to this call in the classic contract:
+
+    > contract.transact({'from': eth.defaultAccount, 'value': 0, ...}).withdraw(amount)
+    '''
+    def __getattr__(self, attr):
+        # If method_name.constant is true in the abi, then call by default
+        call_by_default = \
```
pipermerriam: This is too difficult to parse/read.  I would recommend breaking this out into a small utility function which **should** also make it easier to test the abi inspection logic directly.

Alternatively you could break this up some across multiple lines, but I think this would benefit from being a stand-alone easy-to-test function.

web3/contract.py
```diff
@@ -764,9 +765,48 @@ def __prepared_function(self, **kwargs):
...
+        self.call_by_default = call_by_default
+        super().__init__(contract, function)
+
+    def _prepared_function(self, **kwargs):
+        # Modifier is not provided and method is not constant/pure
+        # then do a transaction instead assuming default conditions
+        # Defaults are: from coinbase and no value sent
+        if not kwargs and not self.call_by_default:
+            default_account = self._contract.web3.eth.defaultAccount
+            return super()._prepared_function(transact={"from": default_account, "value": 0})
```
pipermerriam: I think that `value` defaults to zero if not included in the transaction dict which suggests it shouldn't be included here.

web3/contract.py
```diff
@@ -764,9 +765,48 @@ def __prepared_function(self, **kwargs):
...
+    is equivalent to this call in the classic contract:
+
+    > contract.transact({'from': eth.defaultAccount, 'value': 0, ...}).withdraw(amount)
+    '''
+    def __getattr__(self, attr):
+        # If method_name.constant is true in the abi, then call by default
+        call_by_default = \
+                [m['constant'] if 'constant' in m.keys() else False \
+                    for m in self._classic_contract.abi if \
+                        m['type'] != 'constructor' and m['name'] == attr][0]
```
pipermerriam: There is an api on the contract object `_find_matching_fn_abi` that will retrieve the ABI for a named function for you.

web3/contract.py
fubuloubu: :+1: so just `from` is needed?

web3/contract.py
fubuloubu: :+1: will update, also helps simplify

requirements-dev.txt
fubuloubu: I'll remove this commit, glad I made it separately.

web3/contract.py
fubuloubu: `_find_matching_fn_abi` makes this cleaner. I separated it out

fubuloubu:
As a part of #419 will the tests for `ConciseContract` be updated?

I was also just thinking I only added two tests that are unique to this subclass, perhaps I can forgo the original `ConciseContract` tests, keeping `test_implicitcontract_transact_default` and `test_implicitcontract_abi_inspection` only. The first fully tests (IMO) the functionality behind `ImplicitMethod` even though `Mock` is used, and the second fully tests that the correct calls are made on a real contract. I basically set up the flag to check whether or not the method should be called by default when the call is made with the given context, I don't see how the `Mock` really plays into this.

fubuloubu:
Should also keep `test_implicitcontract_call_default` as this shows the other side of the coin for `ImplicitMethod`

pipermerriam:
I'm still not ok with the mock based tests.  They represent an anti-pattern and I can't accept the maintenance cost they represent.  A few end-to-end tests which use real contracts and a real EVM should be sufficient to test the code.  I don't think we necessarily need to directly test the `ImplicitMethod` object.

fubuloubu:
Alright then, will do!

Does this mean `ConciseContract`'s should get updated as well for the same concern?

fubuloubu:
Changed all the test cases to check the specific functionality I added in my subclass. I am no longer testing any functionality in the `ConciseContract`/`ConciseMethod` classes.

I couldn't think of a way to test it by asking if the contract had made a `transact`/`call` (like how `ConciseContract` is tested when using `unittest.Mock`), so instead I simply checked if the transaction was added (and automined) to the latest block.

If this is not sufficient, please let me know.

fubuloubu:
Noticed there was a pretty large rewrite of `web3/contracts.py` so I had to completely redo my commits here. Squashed down to one commit.

web3/contract.py
```diff
@@ -698,17 +698,17 @@ def _none_addr(datatype, data):
 )
 
 
-class ConciseMethod:
+class ConciseMethod(object):
```
carver: This shouldn't be necessary, web3 is py3-only.

web3/contract.py
```diff
@@ -698,17 +698,17 @@ def _none_addr(datatype, data):
...
     def __init__(self, function, normalizers=None):
         self.__function = function
         self.__function._return_data_normalizers = normalizers
 
     def __call__(self, *args, **kwargs):
-        return self.__prepared_function(*args, **kwargs)
+        return self._prepared_function(*args, **kwargs)
 
-    def __prepared_function(self, *args, **kwargs):
+    def _prepared_function(self, *args, **kwargs):
```
carver: I'm kind of paranoid about causing name conflicts, and this exposes us to another one. Some options are:

1. Rename to `_web3py_prepared_function` which would be a diabolical thing to call a Solidity function :P
1. Extract to a utility function, removing it from the class entirely, and pass in `self.__function` as an argument.
2. Rename to a Solidity keyword that makes sense, although the only one I can think of is `function` which is confusingly close to `__function` and the classic contract's `functions`.

I have a slight preference for 1, mostly because 2 might be a big rewrite. I have a strong preference for 2 over 3 (but could be convinced if the perfect keyword comes along).

web3/contract.py
fubuloubu: Does this happen under the hood for py3? I always thought `class MyObj` was py2-preferred and `class MyObj(object)` was py3

tests/core/contracts/test_implicit_contract.py
```diff
@@ -0,0 +1,89 @@
...
+    assert math_contract.counter.call_by_default
+    # return13() does not since it is not constant
+    assert not math_contract.return13.call_by_default
+
+
+def test_implicitcontract_call_default(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called that defaults to call
+    # (Verify something was returned)
+    assert math_contract.counter() is not None
```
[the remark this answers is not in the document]
fubuloubu: Can do

tests/core/contracts/test_implicit_contract.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test_implicitcontract_transact_default(math_contract, get_transaction_count):
+    blocknum, starting_txns = get_transaction_count("pending")
+    # When a function is called that defaults to transact
+    # (Verify something was returned)
+    assert math_contract.return13() is not None
+    # Check that a transaction was made and not a call
+    # (Auto-mining is enabled, so query by block number)
+    assert get_transaction_count(blocknum) == starting_txns + 1
```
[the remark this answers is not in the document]
fubuloubu: I like it, will do

web3/contract.py
carver: > I always thought class MyObj was py2-preferred and class MyObj(object) was py3

It is the reverse, actually. `class MyObj(object)` was how you got some Py3-style object features into Py2. It is [(nearly) a no-op in Py3](https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x).

fubuloubu: Ah, you learn something new everyday. Or more accurately, unlearn a bad habit lol.

web3/contract.py
fubuloubu: I could create *another* method in this class called something like `_get_prepared_function(*args, **kwargs)`, that calls this directly. Trying to figure out why this function needs to be totally private versus protected. Or I can just completely go `super()._ConciseContract__prepared_function(*args, **kwargs)` to call it in the subclass. Obviously not preferred, but if the shoe fits...

web3/contract.py
fubuloubu: P.S. There are 26 instances of this in `web3/`:
```bash
$ grep -r "(object)" web3/ | wc -l
26
```

tests/core/contracts/test_implicit_contract.py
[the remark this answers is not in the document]
fubuloubu: I don't know `return13()` returns, but it certainly isn't the number 13 lol. I don't know where the source code is for this, what return value do you want me to check of this pre-compiled contract?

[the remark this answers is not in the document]
fubuloubu: Ah, I see how this is different. Disregard previous comment

web3/contract.py
fubuloubu: Figured out a workaround

tests/core/contracts/test_implicit_contract.py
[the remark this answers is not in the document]
fubuloubu: I covered all of these concerns, but not in every single test since 3/4 tests do transactions, which returns a txn_hash instead of the actual return value.

tests/core/contracts/test_implicit_contract.py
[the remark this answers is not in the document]
fubuloubu: 1/4 tests covers this concern. If it works for one call, it works for any call to the same method I'd say.

fubuloubu:
@carver See my individual comments on your prior review. I rewrote what I had initially, reducing the amount of state being stored in my subclass and optimizing around the `_function` member of `ConciseMethod` as a work around in order to find whether or not the method being called is `constant` in the abi.

web3/contract.py
carver: Yeah, not a surprise that some are leftover from our py2 support days. Do you want to do the honors of deleting them? :)

If not, I'll take care of it.

carver:
Yup, that all looks good to me!

It looks like you addressed @pipermerriam 's comments, too, but I'll wait for the rest of the day for him to review again if he likes. If he's too busy, I'll just merge it at end of day.

pipermerriam:
Blindly approving.  Blaming it on @carver if anything is wrong :P

fubuloubu:
Yay blind approvals!

PR #520, opened 2017-12-21
Add support for sendRawTransaction in EthereumTesterProvider

mhchia:
### What was wrong?
Recently `sendRawTransaction` is supported by `EthereumTester`, and therefore we can support it in `web3.providers.eth_tester.EthereumTesterProvider`

### How was it fixed?
1. Modify the config in `web3/providers/eth_tester/main.py`
2. Modify the version of `eth-tester` in the deps in `setup.py` to the latest one
3. Enable the according test
4. Fix the problem that the original `test_eth_sendRawTransaction` would fail because the sender of the raw transaction doesn't have enough funds in the case1.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/module_testing/eth_module.py
+++ b/web3/utils/module_testing/eth_module.py
@@ -232,17 +232,9 @@ class EthModuleTest(object):
     )
     def test_eth_sendRawTransaction(self,
                                     web3,
-                                    unlocked_account,
                                     raw_transaction,
+                                    funded_account_for_raw_txn,
                                     expected_hash):
-        txn_params = {
-            'from': unlocked_account,
-            'to': "0x39EEed73fb1D3855E90Cbd42f348b3D7b340aAA6",
-            'value': (21000 * 100000000000 + 1),  # tx.gas * tx.gasprice + tx.value
-            'gas': 21000,
-            'gas_price': web3.eth.gasPrice,
-        }
-        txn_hash = web3.eth.sendTransaction(txn_params)
         txn_hash = web3.eth.sendRawTransaction(raw_transaction)
         assert txn_hash == web3.toBytes(hexstr=expected_hash)
 
--- a/web3/utils/module_testing/eth_module.py
+++ b/web3/utils/module_testing/eth_module.py
@@ -230,16 +230,19 @@ class EthModuleTest(object):
             ),
         ]
     )
-    def test_eth_sendRawTransaction(self, web3, raw_transaction, expected_hash):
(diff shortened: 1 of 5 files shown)


mhchia:
One concern, I don't know if it is okay to modify the test in that way.

web3/utils/module_testing/eth_module.py
```diff
@@ -229,7 +230,19 @@ def test_eth_sendTransaction(self, web3, unlocked_account):
...
+                                    raw_transaction,
+                                    expected_hash):
+        txn_params = {
+            'from': unlocked_account,
+            'to': "0x39EEed73fb1D3855E90Cbd42f348b3D7b340aAA6",
+            'value': (21000 * 100000000000 + 1),  # tx.gas * tx.gasprice + tx.value
+            'gas': 21000,
+            'gas_price': web3.eth.gasPrice,
+        }
+        txn_hash = web3.eth.sendTransaction(txn_params)
```
pipermerriam: This shouldn't be required.  Part of the necessary test setup for these tests *should* be pre-loading the raw transaction sender with the appropriate ether.

pipermerriam:
Lets see if 63e200a fixes it

pipermerriam:
@carver good to merge if you're 👍

web3/utils/module_testing/eth_module.py
mhchia: Cool! Thank you for the solution

web3/utils/module_testing/eth_module.py
```diff
@@ -229,7 +230,11 @@ def test_eth_sendTransaction(self, web3, unlocked_account):
             ),
         ]
     )
-    def test_eth_sendRawTransaction(self, web3, raw_transaction, expected_hash):
+    def test_eth_sendRawTransaction(self,
+                                    web3,
+                                    raw_transaction,
+                                    funded_account_for_raw_txn,
```
carver: I think it's cleaner to set the fixture to `autouse=True` than to add a variable that goes unused.

carver:
Just saw this was holding up https://github.com/ethereum/py-evm/pull/206
I'll merge now

web3/utils/module_testing/eth_module.py
mhchia: @carver sorry for the late reply. Should I still modify this part? or are you going to handle this:)?
Thank you for the merge anyway!

carver: No worries, it was pretty minor. No need to submit a PR for it. :) @mhchia

PR #523, opened 2017-12-22
Stricter specification of dependencies

pipermerriam:
### What was wrong?

Most versions in `setup.py` were declared with a simple `>=`.  This allows for major version bumps of dependencies which is problematic given that major version bumps tend to correspond to breaking changes.

### How was it fixed?

Tightened down the version requirements by adding a strict `<` the next major version.

#### Cute Animal Picture

a51fdc9ae54d783f252e1a842890d6e8


carver:
Interesting, the `keccak('')` assertion in eth-utils is failing. I modified it to spit out the actual result and got: 

>       assert keccak('') == b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p", keccak('')  # noqa: E501
>     AssertionError: b'\xa7\xff\xc6\xf8\xbf\x1e\xd7fQ\xc1GV\xa0a\xd6b\xf5\x80\xffM\xe4;I\xfa\x82\xd8\nK\x80\xf8CJ'

carver:
Looks like pysha3 at v0.3 is causing the failure, but it works at v1+, so I'll bump the major version constraints to `>=1,<2`

PR #530, opened 2017-12-31
Replace eth-testrpc with eth-tester in core tests

carver:
### What was wrong?

eth-testrpc is deprecated, but used all over the tests

### How was it fixed?

Swap eth-tester in for eth-testrpc

Still work to do to pass all tests

#### Cute Animal Picture

Cute animal picture


web3/providers/eth_tester/middleware.py
```diff
@@ -218,7 +220,30 @@ def is_named_block(value):
...
+            transaction = params[0]
+            if 'from' not in transaction:
+                if web3.eth.coinbase:
+                    default_from = web3.eth.coinbase
+                elif web3.eth.accounts:
+                    default_from = web3.eth.accounts[0]
+                else:
+                    default_from = None
+                if default_from:
+                    transaction['from'] = web3.eth.coinbase or web3.eth.accounts
```
pipermerriam: This has the potential to mutate the dictionary from another context.  Can you use `cytoolz.assoc` to remove the mutation?

PR #533, opened 2018-01-05
HTTP Retry Request Middleware

bweick:
### What was wrong?
It's not uncommon for requests sent over the network to fail due to network interruptions, and in most cases retrying the request will resolve the issue.

### How was it fixed?
A default middleware for HTTPProvider was created that retried requests up to 5 times if the errors returned were due to network connectivity issues (ConnectionError, HTTPError, Timeout, TooManyRedirects). However not all methods can be allowed to do this (we don't want to send multiple `eth_sendTransaction` requests in case it sends multiple transactions) so a whitelist was created against which the request method is checked to see if any failed requests should be retried.

#### Cute Animal Picture

Cute Animal Picture


web3/providers/rpc.py
```diff
@@ -24,6 +25,7 @@ class HTTPProvider(JSONBaseProvider):
     endpoint_uri = None
     _request_args = None
     _request_kwargs = None
+    _middlewares = (http_retry_request_middleware, 'http_retry_request')
```
carver: I would expect this to need to be something like: `[(http_retry_request_middleware, 'http_retry_request')]`, with the named entry pair as the first element of an iterable. This is an example of when the Mock testing isn't as effective as we might hope.

tests/core/middleware/test_http_request_retry.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture
+def http_retry_request_setup():
+    make_request, web3 = Mock(), Mock()
+    setup = http_retry_request_middleware(make_request, web3)
+    setup.web3 = web3
+    setup.make_request = make_request
+    setup.check_method = check_method
+    return setup
```
carver: This mock testing doesn't cover all our needs (and might be too brittle as the internal implementations change over time).

If you must mock, maybe mock out something much deeper, like `make_post_request`, so that all the HTTPProvider and its new built-in middlewares get tested.

web3/providers/rpc.py
bweick: Fixed this, ended up using the NamedElementStack data structure to be able to merge with the outer_middlewares.

tests/core/middleware/test_http_request_retry.py
bweick: Changed this and added some tests to make sure the whole process works.

web3/middleware/http_retry_request.py
```diff
@@ -0,0 +1,84 @@
...
+    root = method.split('_')[0]
+    if root in whitelist:
+        return True
+    elif method in whitelist:
+        return True
+    else:
+        return False
+
+
+def http_retry_request_middleware(make_request, web3):
```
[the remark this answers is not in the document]
bweick: Cool did this and renamed the file. Any other exception/retry we want to try can just reside in this script if that makes sense to you.

tests/core/middleware/test_http_request_retry.py
```diff
@@ -0,0 +1,72 @@
...
+    assert 'http_retry_request' not in web3.middlewares
+
+
+@patch('web3.providers.rpc.make_post_request', side_effect=ConnectionError)
+def test_check_with_all_middlewares(make_post_request_mock):
+    provider = HTTPProvider()
+    w3 = web3.Web3(provider)
+    with pytest.raises(ConnectionError):
+        w3.eth.blockNumber()
+    assert make_post_request_mock.call_count == 5
```
carver: This looks great. Were you able to confirm that this test failed, before you implemented the fix in `web3/providers/rpc.py`?

It looks right to me, eyeballing, just wanted to make sure you saw the test go from red to green.

bweick: Yep it sure did! Glad this all worked out.

pipermerriam:
@bweick are you willing to submit another follow up PR with basic documentation for the `http_` middleware?  Should be easy to following the existing format in `docs/middleware.rst`

bweick:
@pipermerriam Sure no problem. I'll write it up and then submit the PR.

PR #534, opened 2018-01-05
Caching middlewares

pipermerriam:
### What was wrong?

Many RPC requests can be cached reliably using various caching strategies.

### How was it fixed?

Implemented three new middleware factories.

- `construct_simple_cache_middleware`: *dumb* caching which returns any values which are present in the cache.  Works well for methods like `eth_protocolVersion` and `web3_clientVersion`.  Anything which should *always* have the same response given the same input parameters.
- `construct_time_based_cache_middleware`: caches certain responses for a specified amount of time.  Good for things that *may* change but are not expected to change often such as `eth_accounts` and `eth_coinbase`.
- `construct_block_number_based_cache_middleware`: caching for which all values are keyed by the current latest block number as well.  Works well for endpoints like `eth_call` or `eth_getBalance` which only change at block boundaries.

All middlewares intelligently refuse to cache `null` responses, error responses, and responses without a `'result'` key in the returned JSON.  This prevents things like.

- caching a `null` response for a transaction/block/receipt hash which isn't yet known by the node.
- caching error responses
- caching weird responses

TODO:

- [x] documentation
- [x] tests

#### Cute Animal Picture

englishspringerspanielunderthechair

### DIFF
--- a/web3/middleware/cache.py
+++ b/web3/middleware/cache.py
@@ -73,6 +73,14 @@ def should_cache(method, params, response):
 def construct_simple_cache_middleware(cache,
                                       rpc_whitelist=SIMPLE_CACHE_RPC_WHITELIST,
                                       should_cache_fn=should_cache):
+    """
+    Constructs a middleware which caches responses based on the request
+    ``method`` and ``params``
+
+    :param cache: Any dictionary-like object
+    :param rpc_whitelist: A set of RPC methods which may have their responses cached.
+    :param should_cache_fn: A callable which accepts ``method`` ``params`` and ``response`` and returns a boolean as to whether the response should be cached.
+    """
     def simple_cache_middleware(make_request, web3):
         def middleware(method, params):
             if method in rpc_whitelist:
@@ -148,6 +156,15 @@ def construct_time_based_cache_middleware(cache,
                                           cache_expire_seconds=15,
                                           rpc_whitelist=TIME_BASED_CACHE_RPC_WHITELIST,
                                           should_cache_fn=should_cache):
+    """
+    Constructs a middleware which caches responses based on the request
+    ``method`` and ``params`` for a maximum amount of time as specified
+
+    :param cache: Any dictionary-like object
+    :param cache_expire_seconds: The number of seconds an item may be cached before it should expire.
+    :param rpc_whitelist: A set of RPC methods which may have their responses cached.
+    :param should_cache_fn: A callable which accepts ``method`` ``params`` and ``response`` and returns a boolean as to whether the response should be cached.
+    """
     def simple_cache_middleware(make_request, web3):
         def middleware(method, params):
             if method in rpc_whitelist:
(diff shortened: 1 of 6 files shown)


web3/middleware/cache.py
```diff
@@ -0,0 +1,328 @@
...
+
+import lru
+
+from web3.utils.caching import (
+    generate_cache_key,
+)
+
+
+SIMPLE_CACHE_RPC_WHITELIST = {
+    'web3_clientVersion',
```
carver: What happens if someone connects to a new provider? I don't think we can do a simple cache on `web3_clientVersion`.

web3/middleware/cache.py
```diff
@@ -0,0 +1,359 @@
...
+                return response
+            else:
+                return make_request(method, params)
+        return middleware
+    return simple_cache_middleware
+
+
+time_based_cache_middleware = construct_time_based_cache_middleware(
+    cache=lru.LRU(256),
+)
```
carver: I'm a compulsive double-checker. So I'll almost always go and request `eth_accounts` immediately after creating an account. Not seeing the new account would feel broken here, and 15s is kind of a long time in that context. Short enough to be reasonable in that use case also means too short to be of much good, I would think.

As long as this cache doesn't become a default, I suppose it's not a problem, though.

web3/middleware/cache.py
```diff
@@ -0,0 +1,359 @@
...
+AVG_BLOCK_SAMPLE_SIZE_KEY = 'avg_block_sample_size'
+AVG_BLOCK_TIME_UPDATED_AT_KEY = 'avg_block_time_updated_at'
+
+
+def should_cache_by_block_number(method, params, response):
+    if method == 'eth_getBlockByNumber':
+        if params == ['latest'] or params == ['pending']:
+            return False
+
+    if 'error' in response:
```
carver: maybe `return should_cache(method, params, response)` here?

web3/middleware/cache.py
```diff
@@ -0,0 +1,359 @@
...
+            else:
+                avg_block_time_age_in_blocks = (
+                    (time.time() - avg_block_time_updated_at) / avg_block_time
+                )
+
+            if avg_block_time_age_in_blocks >= avg_block_sample_size:
+                # If the length of time since the average block time as
+                # measured by blocks is greater than or equal to the number of
+                # blocks sampled then we need to recompute the average block
+                # time.
```
carver: To me, this was harder to read to have everything in blocks instead of time.

For example, `cache_expiration_seconds` seems more natural than `avg_block_sample_size`, and leads to a test that probably doesn't require explanation, like:

```py
if time.time() - avg_block_time_updated_at >= cache_expiration_seconds
```

It seems worth the cost of having to do a bit more calculation to get the ancestor block number for the next sample.

web3/middleware/cache.py
pipermerriam: I don't think any of these should ever be default so I think we're ok there.

web3/middleware/cache.py
pipermerriam: I wrote it this way to cover the case where there is not enough chain history to provide an accurate average block time.

With your formula, the avg time would get set to `block_1_timestamp - genesis_timestamp` if queried on the 1st block on the chain.  Then it wouldn't get refreshed until `cache_expiration_seconds` later.

The current code would result in it getting updated at approximately blocks 1, 2, 4, 8, 16, ... until the chain history was at least `cache_expiration_seconds` long after which if would update at regular intervals.

Thoughts on whether the complexity is worth covering this admittedly minority use case?

web3/middleware/cache.py
pipermerriam: I'm going to make some changes that will cause the cache to get busted anytime the middleware stack is rebuilt (which I believe happens anytime providers change).

pipermerriam:
> At 0.7 ABT expiration, then you won't see it until 1.4 ABT

I don't think this is correct.  Once the time is `> 0.7 ABT` it will check for a new block on *every* request.

pipermerriam:
@carver tests added and addressed the cache-busting issue.

web3/middleware/cache.py
carver: Doesn't seem worth it to me. I'd rather see a hard-coded corner case that is easy to read in code (and easy to explain in documentation), than to complicate the main body.

**But** with the included commentary, and the fact that it's a non-default feature, I'm happy to merge now and come back to it later.

carver:
> I don't think this is correct. Once the time is > 0.7 ABT it will check for a new block on every request.

Ah, yes, I was mixing up the cache with the average block time calculation itself.

> @carver tests added and addressed the cache-busting issue.

:+1:

pipermerriam:
blarg.  `latest_block_based_cache_middleware` throws infinite recursion error.... will fix on monday.

carver:
Whoops, how did the tests pass?

pipermerriam:
I think it's due to me faking the `make_request` function

PR #535, opened 2018-01-05
Piper/gas station api

### LINKED ISSUE #496: Better gas price engine algorithm

pipermerriam:
fixes #496 

### What was wrong?

With added network congestion, users need better tools to choose gas prices..info

### How was it fixed?

Implements a gas pricing strategy that is akin to the one used in EthGasStation

### DIFF
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -1,0 +1,132 @@
+import pytest
+
+from web3 import Web3
+from web3.providers.base import BaseProvider
+from web3.gas_strategies.time_based import (
+    construct_time_based_gas_price_strategy,
+)
+from web3.middleware import (
+    construct_fixture_middleware,
+)
+
+
+def wrap_result(fn):
+    def inner(*args, **kwargs):
+        return {
+            'result': fn(*args, **kwargs),
+        }
+    return inner
+
+
+@wrap_result
+def _get_block_by_something(method, params):
+    block_identifier = params[0]
+    if block_identifier == 'latest' or block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000005':
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000005',
+            'number': 5,
+            'parentHash': '0x0000000000000000000000000000000000000000000000000000000000000004',
+            'transactions': [
+                {'gasPrice': 70},
+                {'gasPrice': 60},
+            ],
+            'miner': '0xA',
+            'timestamp': 100,
+        }
+    elif block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000004':
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000004',
+            'number': 4,
+            'parentHash': '0x0000000000000000000000000000000000000000000000000000000000000003',
+            'transactions': [
+                {'gasPrice': 100},
+                {'gasPrice': 80},
+                {'gasPrice': 60},
+            ],
+            'miner': '0xB',
+            'timestamp': 80,
+        }
+    elif block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000003':
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000003',
+            'number': 3,
+            'parentHash': '0x0000000000000000000000000000000000000000000000000000000000000002',
+            'transactions': [
+                {'gasPrice': 10},
+                {'gasPrice': 50},
+                {'gasPrice': 100},
+            ],
+            'miner': '0xC',
+            'timestamp': 60,
+        }
+    elif block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000002':
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000002',
+            'number': 2,
+            'parentHash': '0x0000000000000000000000000000000000000000000000000000000000000001',
+            'transactions': [
+                {'gasPrice': 50},
+                {'gasPrice': 70},
+                {'gasPrice': 100},
+            ],
+            'miner': '0xB',
+            'timestamp': 40,
+        }
+    elif block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000001':
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000001',
+            'number': 1,
+            'parentHash': '0x0000000000000000000000000000000000000000000000000000000000000000',
+            'transactions': [
+                {'gasPrice': 30},
+                {'gasPrice': 35},
+                {'gasPrice': 65},
+            ],
+            'miner': '0xA',
+            'timestamp': 20,
+        }
+    elif block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000000' or block_identifier == 0:
+        return {
+            'hash': '0x0000000000000000000000000000000000000000000000000000000000000000',
+            'number': 0,
+            'parentHash': None,
+            'transactions': [
+                {'gasPrice': 30},
+                {'gasPrice': 50},
+                {'gasPrice': 60},
+            ],
+            'miner': '0xA',
+            'timestamp': 0,
+        }
+    else:
+        assert False
+
+
+@pytest.mark.parametrize(
+    'strategy_params,expected',
+    (
+        (dict(max_wait_seconds=60, sample_size=5, probability=98), 58),
+        (dict(max_wait_seconds=60, sample_size=5, probability=90), 46),
+        (dict(max_wait_seconds=60, sample_size=5, probability=50), 11),
+        (dict(max_wait_seconds=40, sample_size=5, probability=90), 56),
+        (dict(max_wait_seconds=20, sample_size=5, probability=90), 60),
+        (dict(max_wait_seconds=80, sample_size=5, probability=90), 36),
+    ),
+)
+def test_time_based_gas_price_strategy(strategy_params, expected):
+    fixture_middleware = construct_fixture_middleware({
+        'eth_getBlockByHash': _get_block_by_something,
+        'eth_getBlockByNumber': _get_block_by_something,
+    })
+
+    w3 = Web3(
+        providers=[BaseProvider()],
+        middlewares=[fixture_middleware],
+    )
+
+    time_based_gas_price_strategy = construct_time_based_gas_price_strategy(
+        **strategy_params,
+    )
+    w3.eth.setGasPriceStrategy(time_based_gas_price_strategy)
+    actual = w3.eth.generateGasPrice()
+    assert actual == expected
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -105,12 +105,22 @@ def _get_block_by_something(method, params):
 @pytest.mark.parametrize(
     'strategy_params,expected',
     (
-        (dict(max_wait_seconds=60, sample_size=5, probability=98), 58),
-        (dict(max_wait_seconds=60, sample_size=5, probability=90), 46),
+        # 120 second wait times
+        (dict(max_wait_seconds=80, sample_size=5, probability=98), 35),
+        (dict(max_wait_seconds=80, sample_size=5, probability=90), 27),
+        (dict(max_wait_seconds=80, sample_size=5, probability=50), 10),
+        # 60 second wait times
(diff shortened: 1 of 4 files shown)


pipermerriam:
I don't want to merge this until I get the caching stuff cleaned up because I want to see how they play together.

web3/gas_strategies/time_based.py
```diff
@@ -0,0 +1,152 @@
...
+            break
+        # we intentionally trace backwards using parent hashes rather than
+        # block numbers to make caching the data easier to implement.
+        block = w3.eth.getBlock(block['parentHash'], full_transactions=True)
+        for transaction in block['transactions']:
+            yield (block['miner'], block['hash'], transaction['gasPrice'])
+
+
+def aggregate_miner_data(raw_data):
+    data_by_miner = groupby(1, raw_data)
```
carver: Shouldn't this be `groupby(0, raw_data)`?

web3/gas_strategies/time_based.py
```diff
@@ -0,0 +1,152 @@
...
+        miner_data,
+        key=operator.attrgetter('min_gas_price'),
+        reverse=True,
+    ))
+    for idx in range(len(miner_data_by_price)):
+        min_gas_price = miner_data_by_price[idx].min_gas_price
+        num_blocks_accepting_price = sum(m.num_blocks for m in miner_data_by_price[idx:])
+        inv_prob_per_block = (sample_size - num_blocks_accepting_price) / sample_size
+        probability_accepted = 1 - inv_prob_per_block ** wait_blocks
+        yield Probability(min_gas_price, probability_accepted)
```
carver: This looks like an optimistic probability calculation. If a miner mines 20 blocks out of a sample of 100, and includes one transaction at a 1 gwei price, then this treats it as if a 1 gwei transaction is just as likely to get included in all 20 blocks.

Am I reading that right?

web3/gas_strategies/time_based.py
```diff
@@ -0,0 +1,152 @@
...
+            raise Exception('Invariant')
+
+        adj_prob = desired_probability - right.prob
+        window_size = left.prob - right.prob
+        position = adj_prob / window_size
+        gas_window_size = left.gas_price - right.gas_price
+        gas_price = int(math.ceil(right.gas_price + gas_window_size * position))
+        return gas_price
+    else:
+        raise Exception('Invariant')
```
carver: Notes about why this is invariant will help if something breaks later, and maybe even add clarity to the rest of the loop.

web3/gas_strategies/time_based.py
```diff
@@ -0,0 +1,152 @@
...
+    ))
+    for idx in range(len(miner_data_by_price)):
+        min_gas_price = miner_data_by_price[idx].min_gas_price
+        num_blocks_accepting_price = sum(m.num_blocks for m in miner_data_by_price[idx:])
+        inv_prob_per_block = (sample_size - num_blocks_accepting_price) / sample_size
+        probability_accepted = 1 - inv_prob_per_block ** wait_blocks
+        yield Probability(min_gas_price, probability_accepted)
+
+
+def compute_gas_price(probabilities, desired_probability):
```
carver: Maybe a docstring about probabilities being sorted by increasing price.

web3/middleware/fixture.py
```diff
@@ -6,9 +6,11 @@ def construct_fixture_middleware(fixtures):
...
         def middleware(method, params):
             if method in fixtures:
-                return {
-                    'result': fixtures[method],
-                }
+                response = fixtures[method]
+                if callable(response):
+                    return response(method, params)
+                else:
+                    return response
```
carver: Why change the fixture API here? The old one that automatically inserts the value with a `{'result': val}` seemed fine.

Leaving the old API even seems to make  `tests/core/gas-strategies/test_time_based_gas_price_strategy.py` simpler by removing the need for the `wrap_result` decorator.

web3/gas_strategies/time_based.py
pipermerriam: Yes, I'm open to suggestions for how to account for this.

Also, here's a list of other things I know that aren't currently being accounted for.

- transaction `gas` limit.
- competing transactions in the transaction pool.

I'm sure there are others.  The rabbit hole is deep.

web3/middleware/fixture.py
pipermerriam: Originally this was a "theoretically I bet it would be useful to be able to return error responses".

Now with the bug in the caching branch, I'm planning to actually use that functionality so the ability to return any type of json is desired.  I'm open to suggestions for cleaning up the API for the common use case.

pipermerriam:
Now based on and depends on #536 to be merged first.

web3/gas_strategies/time_based.py
carver: > suggestions for how to account for this

Maybe track the lowest gwei per block, ignoring the miner for this calculation? That may be a bad idea for the increase in required data. Frankly, I am still a little pessimistic about the approach of estimating gas price at all, so I suppose I'm in the camp of limit effort and I will just say "it's fine as is."

(The alternative to estimating gas price I am thinking is a daemon that watches your pending transactions and repeatedly bumps the price if they haven't been picked up for a while)

> The rabbit hole is deep.

For real

pipermerriam:
Ready for review but I still want to wait on #542 before merging.

pipermerriam:
Feel free to mock me for 9fdc0bf  - Turns out the recursion problems were *still* present.  Fixed with what I think is a bulletproof solution using a lock which only allows the cache portion of each of the caching middleware functions to be entered once per request.

pipermerriam:
@carver ready for new 👀

carver:
> Feel free to mock me for 9fdc0bf - Turns out the recursion problems were still present.

lol @ that commit message

> using a lock which only allows the cache portion of each of the caching middleware functions to be entered once per request

Seems like a good idea anyway, for people using a web3 instance across multiple threads.

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
```diff
@@ -0,0 +1,142 @@
...
+        (dict(max_wait_seconds=40, sample_size=5, probability=90), 38),
+        (dict(max_wait_seconds=40, sample_size=5, probability=50), 16),
+        # 20 second wait times
+        (dict(max_wait_seconds=20, sample_size=5, probability=98), 49),
+        (dict(max_wait_seconds=20, sample_size=5, probability=90), 45),
+        (dict(max_wait_seconds=20, sample_size=5, probability=50), 25),
+    ),
+)
+def test_time_based_gas_price_strategy(strategy_params, expected):
+    fixture_middleware = construct_fixture_middleware({
```
carver: here

PR #542, opened 2018-01-08
Piper/fix recursion error in caching middleware

pipermerriam:
Based on #536 which is likely to change in structure.

### What was wrong?

Caching middleware is broken.  It causes infinite recursion

### How was it fixed?

Each of the middlewares now do a pre-check using the `should_cache_fn` prior to fetching a response, and then again with the response to see if it should be cached.  All tests now use *legit* web3 instances and requests that are passed through the middleware stack just as they would be normally.

#### Cute Animal Picture

hqdefault

### DIFF
--- a/tests/core/middleware/test_latest_block_based_cache_middleware.py
+++ b/tests/core/middleware/test_latest_block_based_cache_middleware.py
@@ -1,11 +1,20 @@
+import codecs
+import itertools
 import time
 import uuid
 
 import pytest
 
-from cytoolz import assoc
+from eth_utils import (
+    is_integer,
+    to_tuple,
+)
 
-from web3.middleware.cache import (  # noqa: F401
+from web3 import Web3
+from web3.providers.base import BaseProvider
+from web3.middleware import (  # noqa: F401
+    construct_result_generator_middleware,
+    construct_error_generator_middleware,
     construct_latest_block_based_cache_middleware,
 )
 from web3.utils.caching import (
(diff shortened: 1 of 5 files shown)


carver:
@pipermerriam #536 is merged, so this is ready to rebase on master

tests/core/middleware/test_time_based_cache_middleware.py
```diff
@@ -1,125 +1,150 @@
...
         {'result': None},
-        {'error': 'some error message'},
     )
 )
-def test_time_based_cache_middleware_does_not_cache_bad_responses(response):
-    def make_request(method, params):
-        return assoc(response, 'id', str(uuid.uuid4()))
+def test_time_based_cache_middleware_does_not_cache_bad_responses(w3_base,
+                                                                  response,
+                                                                  time_cache_middleware):
```
carver: *style nitpick* guide says w3_base belongs on its own line with one indent

tests/core/middleware/test_time_based_cache_middleware.py
```diff
@@ -1,125 +1,150 @@
...
+    with pytest.raises(ValueError) as err:
+        w3.manager.request_blocking('fake_endpoint', [])
+    assert 'error-number-1' in str(err)
 
-    response_a = middleware('not_whitelisted', [])
-    assert 'result' in response_a
 
-    response_b = middleware('not_whitelisted', [])
-    assert 'result' in response_b
+def test_time_based_cache_middleware_does_not_endpoints_not_in_whitelist(w3):
```
carver: *naming nitpick* `does_not_*cache*_endpoints`

web3/middleware/cache.py
```diff
@@ -60,20 +60,21 @@
...
-def should_cache(method, params, response):
-    if 'error' in response:
-        return False
-    elif 'result' not in response:
-        return False
-
-    if response['result'] is None:
-        return False
+def _should_cache(method, params, response=None):
+    if response is not None:
```
carver: If the whole response is None, we should cache it? Not sure how that would happen, but it doesn't feel right.

web3/middleware/cache.py
```diff
@@ -60,20 +60,21 @@
...
+
+        if response['result'] is None:
+            return False
     return True
 
 
 def construct_simple_cache_middleware(cache_class,
                                       rpc_whitelist=SIMPLE_CACHE_RPC_WHITELIST,
-                                      should_cache_fn=should_cache):
+                                      should_cache_fn=_should_cache):
```
carver: *style nitpick* `cache_class` on its own line

web3/middleware/cache.py
```diff
@@ -162,7 +163,7 @@ def middleware(method, params):
 def construct_time_based_cache_middleware(cache_class,
                                           cache_expire_seconds=15,
                                           rpc_whitelist=TIME_BASED_CACHE_RPC_WHITELIST,
-                                          should_cache_fn=should_cache):
+                                          should_cache_fn=_should_cache):
```
carver: *style nitpick* `cache_class` on its own line

tests/core/middleware/test_latest_block_based_cache_middleware.py
```diff
@@ -1,108 +1,239 @@
...
+    w3.manager.request_blocking('fake_endpoint', [])
+
+    assert next(counter) == 2
+
 
-    # prime the internal latest block cache.
-    middleware('fake_endpoint', [])
+def test_latest_block_cache_middleware_does_not_cache_error_response(w3_base,
+                                                                     block_data_middleware,
+                                                                     latest_block_based_cache_middleware):
```
carver: *style nitpick* `w3_base` to its own line (will also fix `flake8` failure)

tests/core/middleware/test_latest_block_based_cache_middleware.py
```diff
@@ -1,108 +1,239 @@
...
 
-    middleware = construct_latest_block_based_cache_middleware(
-        cache_class=dict,
-        average_block_time_sample_size=10,
-        default_average_block_time=10,
-        rpc_whitelist={'fake_endpoint'},
-    )(make_request, web3)
+def test_latest_block_cache_middleware_does_not_cache_bad_responses(w3_base,
+                                                                    block_data_middleware,
+                                                                    latest_block_based_cache_middleware):
```
carver: *style nitpick* `w3_base` to its own line (will also fix `flake8` failure)

tests/core/middleware/test_latest_block_based_cache_middleware.py
```diff
@@ -1,108 +1,239 @@
...
-        {},
-        {'result': None},
-        {'error': 'some error message'},
-    )
-)
-def test_latest_block_cache_middleware_does_not_cache_bad_responses(web3, response):
-    def make_request(method, params):
-        return assoc(response, 'id', str(uuid.uuid4()))
+    assert w3.manager.request_blocking('fake_endpoint', []) == result
+    assert not w3.manager.request_blocking('fake_endpoint', [1]) == result
```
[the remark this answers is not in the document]
pipermerriam: I've been having an internal battle on this one.  Your order of operations take on it is a good enough reason for me to settle on `!=`.

tests/core/middleware/test_latest_block_based_cache_middleware.py
```diff
@@ -1,108 +1,239 @@
...
+    w3 = w3_base
 
-    response = middleware('fake_endpoint', [])
-    assert 'result' in response
+    def error_cb(method, params):
+        next(counter)
+        return "the error message"
 
-    assert middleware('fake_endpoint', []) == response
+    w3 = w3_base
```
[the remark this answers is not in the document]
pipermerriam: 👍 copy/paste

tests/core/middleware/test_time_based_cache_middleware.py
pipermerriam: This may take a few more reminders for me to break this habit.

web3/middleware/cache.py
pipermerriam: Ah, this is the pre request/post-request should I cache this check to prevent recursion.

- `response` is `None` when it's being called prior to making the request
- `response` is not `None` when checking if the response should be cached.

Having written this out, I'm now annoyed with the *messyness* of this API and would be happy for suggestions on cleaning it up.  Basically, there needs to be a way to say, don't even generate a response for requests which match this pattern.

pipermerriam: Alright, I actually chose to remove this and special case the check for whether the request is for the 'latest' block since that request should **never** be cached by this middleware.

pipermerriam:
I think I cleaned it all up.

web3/middleware/cache.py
```diff
@@ -267,19 +268,20 @@ def middleware(method, params):
...
-    if method == 'eth_getBlockByNumber':
-        if params == ['latest'] or params == ['pending']:
-            return False
-
-    return _should_cache(method, params, response)
+def _is_latest_or_pending_block_request(method, params):
+    if method != 'eth_getBlockByNumber':
+        return False
+    elif params == ['latest']:
+        return True
```
carver: It's called `is_latest_or_pending` but caches only `latest`?

pipermerriam: Last minute change that didn't make it into the rename of the function.  fixing.

carver:
Cool. I don't see how it could cause a new test failure, so I'll optimistically merge and eat my :tophat: if it fails with a true positive.

PR #543, opened 2018-01-10
Adding tests for byte arrays as arguments passed to functions

jstoxrocky:
### What was wrong?
Referencing #538.  Adding tests to make sure byte arrays can be passed to contract functions. 

### How was it fixed?
A new contract added to `/tests/core/contracts/conftest.py` with several functions that accept and return `bytes32[]` and `byte[]`. Tests for these functions are also added.

#### Cute Animal Picture

Cute animal picture


tests/core/contracts/conftest.py
```diff
@@ -305,6 +305,44 @@ def emitter(web3_empty, Emitter, wait_for_transaction, wait_for_block):
...
...000010492830192600103820291508084116106f45790505b5050505050905090565b60038181548110151561078257fe5b9060005260206000209060209182820401919006915054906101000a90047f01000000000000000000000000000000000000000000000000000000000000000281565b80600090805190602001906107db9291906108ce565b5050565b602060405190810160405280600081525090565b602060405190810160405280600081525090565b82805482825590600052602060002090601f016020900481019282156108bd5791602002820160005b8382111561088e57835183826101000a81548160ff02191690837f0100000000000000000000000000000000000000000000000000000000000000900402179055509260200192600101602081600001049283019260010302610830565b80156108bb5782816101000a81549060ff021916905560010160208160000104928301926001030261088e565b505b5090506108ca9190610921565b5090565b828054828255906000526020600020908101928215610910579160200282015b8281111561090f5782518290600019169055916020019190600101906108ee565b5b50905061091d9190610951565b5090565b61094e91905b8082111561094a57600081816101000a81549060ff021916905550600101610927565b5090565b90565b61097391905b8082111561096f576000816000905550600101610957565b5090565b905600a165627a7a72305820d0caf89bd0d39b343a907ac9d0f0b9bcddb1b8dc910706160d9d26fdc552afa40029"  # noqa: E501
```
pipermerriam: Will you provide the source code used to generate this bytecode?

jstoxrocky: Yes, is it enough to post it here in the PR? I could not find any contracts source code in the repo itself.

jstoxrocky:
Contract that was added:
```
contract ArraysContract {

    bytes32[] public bytes32Value;
    bytes32[] public bytes32ConstValue;
    byte[] public byteValue;
    byte[] public byteConstValue;
    
    
    function ArraysContract(bytes32[] _bytes32Value, byte[] _byteValue) {
        bytes32Value = _bytes32Value;
        byteValue = _byteValue;
        bytes32ConstValue = [keccak256('A'), keccak256('B')];
        byteConstValue = [bytes1(0), bytes1(1)];
    }
    
    function setBytes32Value(bytes32[] _bytes32Value) {
        bytes32Value = _bytes32Value;
    }

    function getBytes32Value() returns (bytes32[]) {
        return bytes32Value;
    }

    function getBytes32ConstValue() returns (bytes32[]) {
        return bytes32ConstValue;
    }

    function setByteValue(byte[] _byteValue) {
        byteValue = _byteValue;
    }

    function getByteValue() returns (byte[]) {
        return byteValue;
    }
    
    function getByteConstValue() returns (byte[]) {
        return byteConstValue;
    }
}
```

tests/core/contracts/conftest.py
pipermerriam: All the previous ones were added in the early days by me.  With the addition of *untrusted* code, I think it's prudent that we start including source code so that:

1. we can verify that the contract does what we think it does in the case of a confusing test failure.
2. we can re-generate bytecode for future changes to the VM (sharding will introduce some VM changes for which some mainnet VM code will no longer be runnable on a shard)

Can you add it to the module just as a multiline string

```python
CONTACT_SOURCE = """
contract ArraysContract {
    ...
}
"""
```

carver:
Piper's request was the source code, which is now added. Merging...

PR #546, opened 2018-01-10
adding contract address to concise contract

jstoxrocky:
### What was wrong?
I love the ConciseContract class but its annoying to have to call `concise_contract._classic_contract.address` to get the address. This commit allows it to be directly callable with `concise_contract.address`

### How was it fixed?
Added `address` as an attribute to the ConciseContract object when its instantiated.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -676,6 +676,7 @@ class ConciseContract(object):
     def __init__(self, classic_contract):
         classic_contract._return_data_normalizers += CONCISE_NORMALIZERS
         self._classic_contract = classic_contract
+        self.address = self._classic_contract.address
 
     @classmethod
     def factory(cls, *args, **kwargs):


carver:
It's not unreasonable for a contract to have a function called `address`. I don't see a great way to minimize name conflicts besides hanging all the metadata off a single attribute, like we're doing with `_classic_contract`.

Are the underscores the most annoying part? Unfortunately, anything short and easy is at high risk for name conflicts.

jstoxrocky:
haha yes the underscores :(

Hmmm, I was under the impression that "address" is a protected term in Solidity since its also a datatype. In which case there can't be a name conflict. I can't seem to create a contract in Remix that has a function or variable named "address".  Is this your concern, that someone might call a variable or function "address", or is it something specific to Web3.py?

carver:
> haha yes the underscores :(

Heh, sometimes I forget that underscores are annoyingly far on a QWERTY layout, so I might use them more than otherwise. (my dash/underscore is where the quote key is on QWERTY)

> I was under the impression that "address" is a protected term in Solidity since its also a datatype

Ah, you are correct. I just confirmed that Solidity won't compile with a function named `address`. AFAIK, it's not technically prohibited by the ABI (if someone were to compile to EVM bytecode using an alternative language), but being impossible in solidity is good enough for me, since it's so pervasive.

PR #547, opened 2018-01-11
Update readme for docs requirements

monokh:
### What was wrong?

Instructions for setting up docs env had a slight error

### How was it fixed?

Correct instructions to point to docs requirement file

#### Cute Animal Picture

Cute animal picture


carver:
Thanks!

PR #552, opened 2018-01-13
Block id contract call

bweick:
### What was wrong?
Users could not directly specify a block to query contract state from when using the contract.functions.methodName.call() API. 

### How was it fixed?
A `block_identifier` parameter is now passed to the call method that allows users to query a specific block based on predefined blocks (`latest`, `earliest`), block number indexed from front or back of chain, and block hash.

Note: The tests for this feature feel partially incomplete because the eth-tester package does not allow one to query past contract state (outlined here ethereum/eth-tester#47). Instead the web3.eth.call method is assumed to be working as expected regarding querying past blocks (probably a safe assumption) and is mocked out.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -1,13 +1,9 @@
 import pytest
 
-<<<<<<< HEAD
 from hexbytes import (
     HexBytes,
 )
 
-=======
-from unittest.mock import patch
->>>>>>> Added ability to query a specific block when using the contract.functions.methodName.call() method. Blocks can be queried by 'latest', 'earliest', block number (indexed from back and front of chain) and block hash.
 from web3.exceptions import (
     BadFunctionCallOutput,
     InvalidAddress,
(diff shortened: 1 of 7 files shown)


carver:
@bweick can you rebase on latest so that the tests run?

Thanks for opening the issue on eth-tester.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -314,3 +324,44 @@ def test_call_undeployed_contract(undeployed_math_contract):
...
+def test_throws_error_if_block_out_of_range(web3, math_contract):
+    web3.providers[0].make_request(method='evm_mine', params=[20])
+    with pytest.raises(BlockNumberOutofRange):
+        math_contract.functions.counter().call(block_identifier=50)
+
+    with pytest.raises(BlockNumberOutofRange):
+        math_contract.functions.counter().call(block_identifier=-50)
+
+
+@patch('web3.eth.Eth.call', return_value='0x' + '0' * 64)
```
pipermerriam: I'd like to try and remove the mocks.

Since we're using the `Math` contract which has the `increment` method, it should be pretty trivial for us to execute a few transactions incrementing the counter and then execute the calls against those block numbers checking that the `counter` value is what we expect.

bweick: Cool yeah 100% agree, I built it that way originally until I realized `eth-tester` didn't keep past blocks. Let me know when that's merged and I'll update the code here.

bweick: @pipermerriam I don't want this to fall through the cracks. Did the eth-tester updates ever get merged to be able to finish this? If not I'd be happy to take a look at tackling that issue.

pipermerriam: There have been a lot of updates merged and released into eth-tester.  I'd suggest rebasing your PR and seeing if the issue you ran into is fixed.

Otherwise, I'm not sure I follow your statement:

> until I realized `eth-tester` didn't keep past blocks.

I don't understand what you're referring to here.  Can you expand?

bweick: Ok I can rebase. The problem I was having with the `eth-tester` backends was that they only allowed me to query the state of the last block, so I couldn't increment the counter from one block to the next and then query the previous block's state, which caused me to have to do the Mock's in my tests on `web3.eth.call` (thus making the implicit assumption that the `block_identifier` parameter works within that method) instead of directly querying the previous state. Does that make sense?

carver: Can you try rebasing and running again? It looks to me like eth-tester supports it: https://github.com/ethereum/eth-tester/blame/master/eth_tester/main.py#L406

Although I'm a little confused about:
> The tests for this feature feel partially incomplete because the eth-tester package does not allow one to query past contract state (outlined here ethereum/eth-tester#47)

because it seems like that feature has been there a while. Anyway, once this is rebased, we can take another look at where it's failing.

carver:
@bweick https://github.com/ethereum/eth-tester/pull/61 is released in beta 17 of eth-tester, so this should be unblocked.

carver:
I posted a PR (#663 ) with some small overlap with this PR (since this wasn't linked to #576 yet). I think I'll be able to just close mine if you can rebase and add some small pieces:

1. convert the block number to hex, for json-rpc compliance with other back ends, like: https://github.com/ethereum/web3.py/pull/663/files#diff-06d2058561b889dd1e40e63c2496b111R264
2. convert block numbers back to int for eth-tester only, like: https://github.com/ethereum/web3.py/pull/663/files#diff-ef15c20827fc5f198f0ee6f0c4bcc561R191
3. an integration test in `utils/module_testing/eth_module.py` (can be very simple, single call to an old block)

The integration test would have caught that we missed step 1 above.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -366,48 +362,6 @@ def test_call_undeployed_contract(undeployed_math_contract, call):
...
-    web3.providers[0].make_request(method='evm_mine', params=[15])
-    math_contract.functions.counter().call(block_identifier=-6)
-    eth_call_mock.assert_called_with(call_transaction, block_identifier=11)
-
-
-@patch('web3.eth.Eth.call', return_value='0x' + '0' * 64)
-def test_returns_data_from_specified_block(eth_call_mock, web3, math_contract, call_transaction):
-    web3.providers[0].make_request(method='evm_mine', params=[20])
-    math_contract.functions.counter().call(block_identifier=14)
-    eth_call_mock.assert_called_with(call_transaction, block_identifier=14)
```
carver: Did you mean to delete all these tests you wrote?

bweick: Yeah it's fine those were all the mock based tests I needed to replace that led me down the whole `eth-tester` rabbit hole.

carver:
FYI, `isort --recursive web3 ens tests` will do all the import sorting for you.

carver:
Still working on this @bweick @brian-biometrix ? I'm trying to get v4-stable out the door by Monday.

If you can't work on it, I'm happy to pick it up where you left off.

bweick:
Yeah I'm a little busy at the moment if you want to take it to the finish line. Still a little unsure why the integration test is breaking...Parity for example doesn't seem to be recognizing 'latest' as a block_identifier, that or for whatever reason the counter isn't incrementing. The goethereum break seems to have something to do with the account not being authorized to send the incrementing transaction see [here](https://ethereum.stackexchange.com/questions/30439/error-unknown-account)

bweick:
Then the linting issue is that super quick fix you outlined above!

carver:
Merged these changes into #663 -- thanks for getting this kicked off @bweick !

PR #553, opened 2018-01-13
adding missing apt packages

ycdk:
**Problem**: Some packages missing for compiling in clean virtualenv

**Solution**: Add packages to `apt` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ print('Contract value: {}'.format(contract_instance.greet()))
 If you would like to hack on web3.py, set up your dev environment with:
 
 ```sh
-sudo apt-get install libssl-dev
+sudo apt-get install libssl-dev libffi-dev autoconf automake libtool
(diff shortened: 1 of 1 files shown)


carver:
Cool, thanks for fleshing this out a bit.

Why do you want to install `libffi-dev`? I don't have it locally.

carver:
I'm not certain all of these are required, but it's relatively low cost to overinstall a few standard dev libraries, so I'm merging.

ycdk:
sorry @carver, did not see note while travelling.   if i remember, the `apt` packages were required to build `pyethereum` and its `cffi` dependency. 

maybe we really copy the full list from that webpage:
https://github.com/ethereum/pyethereum/#installation

carver:
Since we're in the process of removing the pyethereum dependency, I probably won't go add the full list, but if you wanted to, I'd merge it. (maybe separating the dependencies into a separate line, so they are easy to remove again when pyethereum is fully removed)

PR #558, opened 2018-01-15
Clean up contract tests covering contract functions api

dylanjw:
### What was wrong?

With #481, I duplicated the contract test files to create variants for the new contract method api: Contract.functions._myMethod_.call|transact().  While having all the deprecated api tests in one file makes them easy to remove, it also makes writing tests annoying because two files must be edited for every new contract test.  I would like to improve this, but haven't decided how.  Im putting up this example for how tests are written to cover both api's, while I think about how to improve this.   Suggestions are welcome.

### How was it fixed?

TBD

#### Cute Animal Picture

Cute animal picture


carver:
> it also makes writing tests annoying because two files must be edited for every new contract test

Not only annoying, but error prone. My biggest concern is the risk of not detecting new regressions because someone forgot to put the change in both places.

> I would like to improve this, but haven't decided how... Suggestions are welcome.

One sort of crazy idea is to parametrize the contract fixture with a value like `call_style='func_first'` or `call_style='func_last'` and then write a meta function that takes a contract and calls it in the old way or the new way, like:

```py
def invoke_contract(contract_obj, function_name, interaction, txn_dict, style, *func_args):
    ...

# implement such that this call
invoke_contract(arrays_contract, 'getBytes32Value', 'call', {'gas': 90000}, 'func_first')
# invokes this:
arrays_contract.functions.getBytes32Value().call({'gas': 90000})
# and this
invoke_contract(arrays_contract, 'getBytes32Value', 'call', {'gas': 90000}, 'func_last')
# invokes this:
arrays_contract.call({'gas': 90000}).getBytes32Value()
```

pipermerriam:
+1 t jason's parametrization idea.  Should be a good way to keep logic separate but still enforce that we test things both ways.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -127,26 +133,30 @@ def test_call_with_multiple_arguments(math_contract, call_args, call_kwargs):
     ),
 )
 def test_saved_method_call_with_multiple_arguments(math_contract, call_args, call_kwargs):
+    # Is it Ok to limit some test cases to new api style?
```
carver: Yes, seems totally reasonable. It's already tested both ways in other tests.

carver:
Looks like you might need to rebase on master because of #563 - should be straightforward, since you can probably just overwrite whatever changes conflict in there.

dylanjw:
@carver Can you retrigger the travis build?  It looks like the build failed due to an error getting the pypandoc package.  The package is retrievable for me, so maybe there was connection issue.

pipermerriam:
@dylanjw it's messy but for now, if you're blocked by needing build to re-trigger (though it does it as a whole), you can push an empty commit, and then rebase it away and force push.  It's dirty but it works.

carver:
> @ dylanjw it's messy but for now, if you're blocked by needing build to re-trigger (though it does it as a whole), you can push an empty commit, and then rebase it away and force push. It's dirty but it works and doesn't polute history.

This is a totally reasonable short-term fix if you're blocked and not sure how long till I can get to it... But since I am here, I just restarted that one job that had a dependency download error.

carver:
There are some style tweaks I'd prefer, but I don't want to wait through another tests cycle, so I won't hold it up. For future PRs:

- https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#first-thing-on-its-own-line
- https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#function-signatures

Also, this isn't yet applied globally in the repo. The style guide came after a lot of this code. So don't be surprised if you find other ones. Please "fix the broken windows" if you are working near the non-stylized code. :)

PR #565, opened 2018-01-19
Ability to serialize the values within an AttributeDict

jstoxrocky:
### What was wrong?
`web3.Account.sign` returns an `AttributeDict` containing `HexByte` objects. Since state-channel-like applications need to sign and then send values to another user, it is necessary to serialize these values first. I found it time consuming and full of bloat code to have to iterate over the key/values within an `AttributeDict` and serialize them manually. I thought it would be useful to have a way to simply serialize the values within an `AttributeDict`

### How was it fixed?
I added two methods to the `AttributeDict` class (I'm not sure if this is the right spot, maybe a different location is better). The first `serializeHexBytes` accepts a value, checks whether it is a `HexBytes` instance, and if so, reduces it to its string representation - otherwise it does nothing. The second method composes an assortment of serializer functions (currently only `serializeHexBytes`) and recursively runs the values of the `AttributeDict` through them.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -12,10 +12,6 @@ from web3.utils.encoding import (
     to_bytes,
 )
 
-from toolz.functoolz import (
-    compose,
-)
-
 from web3.utils.formatters import recursive_map
 
 # Hashable must be immutable:
@@ -98,15 +94,6 @@ class AttributeDict(ReadableAttributeDict, Hashable):
             return self.__dict__ == dict(other)
         else:
             return False
-
-    @staticmethod
-    def serializeHexBytes(value):
-        return value.hex() if isinstance(value, HexBytes) else value
-
-    @property
-    def serialized(self):
-        serializers = compose(AttributeDict.serializeHexBytes)
-        return recursive_map(serializers, self.__dict__)
 
 
 class NamedElementStack(Mapping):
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -12,10 +12,6 @@ from web3.utils.encoding import (
     to_bytes,
 )
 
-from toolz.dicttoolz import (
-    valmap,
-)
-
 from toolz.functoolz import (
     compose,
 )
@@ -110,7 +106,7 @@ class AttributeDict(ReadableAttributeDict, Hashable):
     @property
     def serialized(self):
         serializers = compose(AttributeDict.serializeHexBytes)
-        return valmap(serializers, self.__dict__)
+        return recursive_map(serializers, self.__dict__)
 
 
 class NamedElementStack(Mapping):
--- a/web3/utils/datastructures.py
+++ b/web3/utils/datastructures.py
@@ -12,6 +12,14 @@ from web3.utils.encoding import (
     to_bytes,
 )
 
+from toolz.dicttoolz import (
+    valmap,
+)
+
+from toolz.functoolz import (
+    compose,
+)
+
 from web3.utils.formatters import recursive_map
 
 # Hashable must be immutable:
@@ -94,6 +102,15 @@ class AttributeDict(ReadableAttributeDict, Hashable):
             return self.__dict__ == dict(other)
         else:
             return False
+
+    @staticmethod
+    def serializeHexBytes(value):
+        return value.hex() if isinstance(value, HexBytes) else value
+
+    @property
+    def serialized(self):
+        serializers = compose(AttributeDict.serializeHexBytes)
+        return valmap(serializers, self.__dict__)
 
 
 class NamedElementStack(Mapping):
--- a/tests/core/utilities/test_attributedict.py
+++ b/tests/core/utilities/test_attributedict.py
@@ -110,15 +110,32 @@ def test_attributedict_set_in_recursive_dict():
 
 
 def test_serialize_hex_bytes():
-    _input = HexBytes(b'\x11')
-    expected_output = _input.hex()
-    output = AttributeDict.serializeHexBytes(_input)
+    data = HexBytes(b'\x11')
+    expected_output = data.hex()
+    output = AttributeDict.serializeHexBytes(data)
     assert output == expected_output
     assert isinstance(output, str)
 
 
 def test_serialized():
-    _input = AttributeDict({'h': HexBytes(b'\x11')}).serialized
-    expected_output = '{"h": "0x11"}'
-    output = json.dumps(_input)
+    data = {'b': HexBytes(b'\x11')}
+    data = AttributeDict(data).serialized
+    expected_output = '{"b": "0x11"}'
+    output = json.dumps(data)
+    assert output == expected_output
+
+
+def test_recursive_serialized():
+    data = {'a': {'b': HexBytes(b'\x11')}}
+    data = AttributeDict(data).serialized
+    expected_output = '{"a": {"b": "0x11"}}'
+    output = json.dumps(data)
+    assert output == expected_output
+
+
+def test_recursive_serialized_array():
+    data = {'a': [{'b': HexBytes(b'\x11')}]}
+    data = AttributeDict(data).serialized
+    expected_output = '{"a": [{"b": "0x11"}]}'
+    output = json.dumps(data)
     assert output == expected_output
--- a/tests/core/utilities/test_attributedict.py
+++ b/tests/core/utilities/test_attributedict.py
@@ -2,7 +2,9 @@ import pytest
 
 from web3.utils.datastructures import (
     AttributeDict,
+    HexBytes,
 )
+import json
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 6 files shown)


jstoxrocky:
@pipermerriam Good call! Added.

web3/utils/datastructures.py
```diff
@@ -95,6 +99,15 @@ def __eq__(self, other):
         else:
             return False
 
+    @staticmethod
+    def serializeHexBytes(value):
```
pipermerriam: Rather than this being a staticmethod, I think I'd rather it be implemented as a stand-alone function.  I prefer keeping our classes as small as possible and I don't think this will be regularly used directly.  Probably belongs in `web3.utils.encoding` as something like `to_hex_if_bytes`.

web3/utils/datastructures.py
```diff
@@ -95,6 +99,15 @@ def __eq__(self, other):
...
+        return value.hex() if isinstance(value, HexBytes) else value
```
pipermerriam: This should probably be `isinstance(value, bytes)` to catch both `bytes` and `HexBytes` types.

web3/utils/datastructures.py
```diff
@@ -95,6 +99,15 @@ def __eq__(self, other):
...
+
+    @property
+    def serialized(self):
```
pipermerriam: @carver curious to hear your opinion on this.  My inclination is to *not* add a new propery/method to the `AttributeDict` class, but to write this as a stand-alone function that people can use, maybe even as something available off of the base `Web3` object.  

No good names immediately come to mind

- `Web3.prepareForJSON`
- `Web3.normalizeForJSON`
- `Web3.makeJSONEncodable`

Thoughts?

carver: Since it seems awkward to describe the preparation, and the only thing it's good for is when you're about to encode to json, how about a `Web.toJSON(val)` that does (roughly) this:

```py
# in web3/utils/encoding.py
def to_json(obj):
    prepared = recursive_map(_encode_for_json, obj)
    return json.dumps(prepared)

def _encode_for_json(val):
    if isinstance(val, bytes): return val.hex() #or if we need the prefix, to_hex(val)
    elif isinstance(val, int) and val > JS_MAX_INT: return to_hex(val)
    else: return val

# in web3/main.py
class Web3:
    toJSON = to_json
```

web3/utils/datastructures.py
carver: I agree it's better to have this standalone.

My instinct is that this should be internally handled as part of a `to_json()` method that is parallel to the other `to_bytes`, `to_hex`, etc. Example code in another comment.

web3/utils/datastructures.py
```diff
@@ -95,6 +99,15 @@ def __eq__(self, other):
...
+        serializers = compose(AttributeDict.serializeHexBytes)
```
carver: Compose with just one function is a noop.

jstoxrocky:
Thanks @carver I agree all your points except returning the raw JSON string. The serialized data is much more valuable to me than the raw JSON string. Here are two reasons why:

Motivation behind this PR is basically to facilitate signing data to send to a user to send to a state-channel contract. So for example, with a simplified state-channel contract's functionality, the first 4 arguments come from `web3.Account.sign` but the additional 2 are the raw values that combine to make up the signed message. These raw values are needed to verify a piece of data contains what  the sender says it should.

```
function verifyData(
        bytes32 messageHash, uint8 v, bytes32 r, bytes32 s, 
        address _addr, uint256 _value) public {
        // Verify signer
        address signer = ecrecover(msgHash, v, r, s);
        require(signer == stateChannelSender);

        // Verify msgHash
        bytes memory preamble = "\x19Ethereum Signed Message:\n32";
        bytes memory message = keccak256(this, _addr, _value);
        bytes32 proof = keccak256(preamble, message);
        require(proof == msgHash);

        // Do something with _addr and _value
        // ...
    }
```

So I would like some functionality in `web3.py` that allows me to sign some values, then send a user those values along with the signature of those values. Returning the raw JSON would add extra steps to this process. Below is what I would like to do:

```python
msg = Web3.soliditySha3(['address', 'uint256'], [addr, value])
signed = Account.sign(msg, PRIVATE_KEY)
combined_data = merge(signed.serialized, {'addr', addr, 'value': value})
```

Second, common practice with web frameworks like `Django` or `Flask` is to wrap a python dictionary in a `jsonify` function which returns some kind of `Response` object that handles all the HTTP headers for you along with the raw JSON data. Again, returning the raw JSON in `web3.py` would add the extra step of converting back to a python dictionary, and then passing to this `jsonify` function.

Thanks for your help!

web3/utils/datastructures.py
jstoxrocky: Ah right, haha i thought this would be helpful for adding future serializers but I realize now its just garbage hahaha

carver:
This background really helps in understanding where you're headed. We're getting closer. I still want a fuller picture though. Can you continue from here, though:

```py
combined_data = merge(signed.serialized, {'addr': addr, 'value': value})
```

What's the next step that you want to take? Show me all the way till the end.

BTW, the json thing was based on an older understanding of what you were trying to accomplish, so if you don't think it helps, then drop it.

jstoxrocky:
Sure, @carver thanks for your patience!

Here is a scenario. I want to send someone a signed message along with a pre-image that will perform some action in a contract IFF (1) the message is signed by me, and (2) the hashed pre-image matches the message that was signed. As a state-channel example, the recipient can only redeem her coins by providing the amount we've agreed upon (pre-image) and her permission slip to take them (signed message).

```python
    # Calculate two values for the pre-image
    # This could be any number of values. I wanted to show that by hashing them together
    # we get a nice fixed length byte-array that interacts well with the Solidity contract 
    addr = '0x' + '1'*40 
    value = 1
    # Hash these values to get fixed length of 32-bytes
    msg = Web3.soliditySha3(['address', 'uint256'], [addr, value])
    # Sign with a private key
    PRIVATE_KEY = b'\x11' * 32
    signed = Account.sign(msg, PRIVATE_KEY) #  Hashes "\x19Ethereum Signed Message:\n" + len(message) + msg
    # Create a dict containing the signed message and the pre-image
    # This is the important part I was hoping to save time with. These values all need to be
    # sent to the recipient. It would be nice to have a serialized version of the results from
    # `Account.sign`
    combined_data = merge(signed.serialized, {'addr': addr, 'value', value})
```

I like developing webapps with Flask or Django so here is a scenario where the server will sign the message:

```python
@app.route(methods=['GET'])
def my_endpoint():
    addr = '0x' + '1'*40 
    value = 1
    msg = Web3.soliditySha3(['address', 'uint256'], [addr, value])
    PRIVATE_KEY = b'\x11' * 32
    signed = Account.sign(msg, PRIVATE_KEY)
    combined_data = merge(signed.serialized, {'addr': addr, 'value', value})
    # Flask's `jsonify` and Django's `JsonResponse` accept a Python dictionary
    # Returns some kind of helpful server response object complete with correct json headers etc.
    return jsonify(combined_data)
```

What I am trying to achieve in this PR is (1) remove a bunch of boilerplate serialization code I have been including in my projects and reduce it to something like `signed.serialized` or `Web3.toSerial(signed)`, and (2) be able to merge those results with the primitive Python pre-image values. Keeping this step as a Python dictionary helps everything interact well with Python web-frameworks. Let me know if there is anything you're not clear on!

jstoxrocky:
I guess a 2-liner thats not terrible boilerplate to include could be:
```python
from toolz.dictoolz import valmap
valmap(lambda x: x.hex() if isinstance(x, bytes) else x, signed)
```
But it won't recursively serialize the dict. Let me know what you think. I think this PR could be helpful to others since the point of signing something seems to be to send it to another machine/user. But I'm also fine with using a two liner solution like this.

carver:
We aren't exactly "serializing" the dict here, it's just preparing it for json encoding. I can't imagine this being used for anything other than calling something that converts the dict to a json object next.

[Another option](https://stackoverflow.com/a/26961568/8412986) is to make an alternative to `jsonify` that accepts a json str:

```py
def json_response(json_str):
    return app.response_class(
        response=json_str,
        status=200,
        mimetype='application/json'
    )
    
def my_endpoint():
    combined_str = Web3.toJSON(combined_data)
    return json_response(combined_str)
```

This seems more generally usable across frameworks, namable, and easy to document as a public Web3 API. Otherwise, it is "Web3.convert_to_a_dict_that_is_json_encodable(obj)"...

Worst case, people can `json.loads()` from the str, if they *need* a dict that is json encodable, but not json encoded (but I think those cases are few).

---

If you feel like doing a `toJSON()` implementation, I'm for it. Or if you would rather do the hex mapping, then let's just close this PR.

jstoxrocky:
@carver I don't really need the JSON implementation at the moment, so I'm just going to close this PR. I initially thought that it was the `HexBytes` objects that were causing problems but the real issue is that Python's `bytes` aren't serializable. So I don't think there is anything to change to Web3.py anymore.

PR #568, opened 2018-01-19
Add Emitter contract test fixture source code

dylanjw:
First of several contract fixtures being brought in from populous, with
modifications extrapolated from the abi and test cases,
for the purpose of review and extention of contract test fixtures.

### What was wrong?

It would be nice to have the contract test fixture source code checked in for review and additions.

### How was it fixed?

Original source code taken from populous.  There are modifications made that had to be extrapolated from the abi and tests.  I can continue to make pull requests with these as I become acquainted with the codebase.

#### Cute Animal Picture

Cute animal picture


dylanjw:
Can we run the build again?  The geth integration tests all have a socket not found error:

```
>           sock.connect(ipc_path)
E           FileNotFoundError: [Errno 2] No such file or directory
```

Im not sure if this is a real issue or something with the build.  Maybe the new contract code crashed geth?

carver:
> Maybe the new contract code crashed geth?

I doubt it, especially because the integration test on py3.6 succeeded. Restarted the test.

PR #569, opened 2018-01-20
Use py-evm in place of pyethereum as a backend for eth-tester, in all tests

### LINKED ISSUE #505: Deprecate and/or remove the `eth-testrpc` based providers.
With no plans to support `eth-testrpc` and the plan to replace it with `eth-tester` I think it's time to add some deprecations warnings to the providers that use `eth-testrpc` if they aren't already present and then to remove them.

We should also update the test suite to not depend on `eth-testrpc` anymore.

carver:
### What was wrong?

`pyethereum` is deprecated, in favor of py-evm.

### How was it fixed?

Moved eth-tester onto a py-evm backend so that all ongoing development is aligned in the same ecosystem. The latest eth-tester & py-evm is now good enough to replace pyethereum. :tada: 

(Although some incoming improvements to `call` and gas estimation will continue to simplify the web3 implementation)

closes #505 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -21,6 +21,7 @@ def deploy(web3, Contract, args=None):
     deploy_receipt = web3.eth.getTransactionReceipt(deploy_txn)
     assert deploy_receipt is not None
     contract = Contract(address=deploy_receipt['contractAddress'])
+    assert len(web3.eth.getCode(contract.address)) > 0
     return contract
 
 
--- a/web3/providers/eth_tester/middleware.py
+++ b/web3/providers/eth_tester/middleware.py
@@ -254,12 +254,17 @@ ethereum_tester_fixture_middleware = construct_fixture_middleware({
 
 
 def guess_from(web3, transaction):
-    if web3.eth.coinbase:
-        return web3.eth.coinbase
-    elif web3.eth.accounts:
+    coinbase = web3.eth.coinbase
+    if coinbase is not None:
+        return coinbase
+
+    try:
         return web3.eth.accounts[0]
-    else:
-        return None
+    except KeyError as e:
+        # no accounts available to pre-fill, carry on
+        pass
+
+    return None
 
 
 def guess_gas(web3, transaction):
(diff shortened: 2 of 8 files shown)


tests/core/contracts/test_contract_call_interface.py
```diff
@@ -21,6 +21,7 @@ def deploy(web3, Contract, args=None):
     deploy_receipt = web3.eth.getTransactionReceipt(deploy_txn)
     assert deploy_receipt is not None
     contract = Contract(address=deploy_receipt['contractAddress'])
+    assert len(web3.eth.getCode(contract.address)) > 0
```
pipermerriam: This my not be accurate anymore because of `HexBytes` but this was always `> 2` to account for the `0x` prefix.

web3/providers/eth_tester/middleware.py
```diff
@@ -251,23 +253,51 @@ def is_named_block(value):
 })
 
 
+def guess_from(web3, transaction):
+    if web3.eth.coinbase:
+        return web3.eth.coinbase
```
pipermerriam: Each of these if/return statements causes a double RPC call.  It would be better to write this as

```
try:
    return web3.eth.coinbase
except Whatever:
    pass

try:
    return web3.eth.accounts[0]
except Whatever:
    pass

return None
```

web3/providers/eth_tester/middleware.py
```diff
@@ -251,23 +253,51 @@ def is_named_block(value):
...
+
 def default_transaction_fields_middleware(make_request, web3):
     def middleware(method, params):
-        if method in (
-            'eth_call',
+        # TODO send call to eth-tester without gas, and remove guess_gas entirely
+        if method == 'eth_call':
+            filled_transaction = pipe(
+                params[0],
+                fill_default_from(web3),
```
pipermerriam: instead of currying these for each call, you could curry them in the outer scope of this function once during middleware construction.

tests/core/contracts/test_contract_call_interface.py
carver: Right, this changed with `HexBytes`. I confirmed it while debugging another issue that caused the eth-tester/py-evm deployment to fail (but still return a `contractAddress`!).

web3/providers/eth_tester/middleware.py
carver: In the first case I'm checking if `coinbase` is not None, which wouldn't raise anything, but I can still do:

```py
coinbase = web3.eth.coinbase
if coinbase is not None:
    return coinbase
```

pipermerriam: Consider my code above to be psuedo code.  Change it out for whatever matches your desired logic (while still reducing API calls).

Issue #429, opened 2017-11-15
(feature request?) parse transaction receipts for event log data output?


drandreaskrueger:
# How to extract event logged output?

Please have a look first how it is working in javascript/truffle:  
[logArgs.contractId](https://github.com/chatch/hashed-timelock-contract-ethereum/blob/1adc535ac5d63e74c64a8deabf8eb26afb774e42/test/htlc.js#L65-L67)  
[txLoggedArgs](https://github.com/chatch/hashed-timelock-contract-ethereum/blob/1adc535ac5d63e74c64a8deabf8eb26afb774e42/test/htlc.js#L35)

I *really* want to work with Python and not JS - and apart from some minor hickups it is looking good.

But now, I need to get a specific value `logArgs.contractId` from an event log after a contract call, and I am stuck.  
Because when I try the same as above with the py transaction receipts that I get via

    rec = web3py.eth.getTransactionReceipt(tx)

then this of course fails:

    logargs = rec.logs[0].args

with

     AttributeError: 'AttributeDict' object has no attribute 'args'

because it simply does not exist - and that seems different from javascript/truffle?

Is it because the event output `args` are still all mixed up in the `rec.logs [0].topics` field (or in the `rec.logs [0].data` field) ?

Example (complete receipt below).

```
'data': '0x00000000000000000000000000000000000000000000000000038d7ea4c68000ec099eeb7c6385269c410da8eb2e028bf832c8418d731bf99f987e9d9357abe9000000000000000000000000000000000000000000000000000000005a0c6ae0'
'topics': ['0xa0085ce7b3f5d49ff7ac7b4f8caaf2f75147d0228c3811a6d3d41eb2029fe122', '0xa4162351c79a5ed77bfc705394785c93842b6fc2e55d2b302c86ee55540c5b24', '0x000000000000000000000000c62d9f5092737fa7689579e279a3a1dd5543fb9e', '0x00000000000000000000000025f579697ce3fe1cf122fc5ffae9f1d83f16c676'],
```

Wouldn't the ABI now help, to take those outputs apart?   

Is there a helper function for that in web3py?   which one?

If not, wouldn't it be useful to have?    Or ... which external tool can I use now?

Or am I thinking this in a completely twisted way?  
[...]

Issue #469, opened 2017-11-27
`web3.utils.events.get_event_data` is rather slow


konradkonrad:
* Version: 3.16.3
* Python: 3.5
* OS: linux

### What was wrong?

Processing events with `web3.utils.events.get_event_data` takes a rather long time, in the order of 1ms per event. This becomes quickly noticeable, when trying to decode event data for thousands of events. 

Please see my benchmarking script & output at https://gist.github.com/konradkonrad/87d76dd2526f39e6f5bad19b85374948

### How can it be fixed?

I don't want to rule out any user error here -- maybe there is a more efficient way to parse events data?

pipermerriam:
At a glance, there may be *some* low hanging fruit in there:

I know there is in-efficiency in the address utilities that could probably be optimized.

```
   110000    0.079    0.000    2.413    0.000 /usr/local/lib/python3.5/site-packages/web3/utils/abi.py:464(map_to_typed_data)
    30000    0.021    0.000    2.259    0.000 /usr/local/lib/python3.5/site-packages/web3/utils/normalizers.py:7(addresses_checksummed)
    20000    0.052    0.000    2.115    0.000 /usr/local/lib/python3.5/site-packages/eth_utils/address.py:124(to_normalized_address)
10000 0.065 0.000 2.007 0.000 /usr/local/lib/python3.5/site-packages/eth_utils/address.py:176(to_checksum_address)
```

Also the use of `coerce_return_to_text` and other related utils are both slow and recently discovered to be *wrong* and capable of returning string data encoded as latin-1 which is very confusing.

PR #575, opened 2018-01-23
parse transaction receipts for event log data

dylanjw:
### What was wrong?

See https://github.com/ethereum/web3.py/issues/429

### How was it fixed?

Im putting up this is a quick implementation, while Im still exploring the feature.  It simply passes the logs through the already existing function: utils.events.get_event_data.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -966,9 +966,6 @@ class ContractEvent(object):
             decoded_log = get_event_data(self.abi, log)
             yield decoded_log
 
-    def _decode_log(self, log):
-        return get_event_data(self.abi, log)
-
     @classmethod
     def factory(cls, class_name, **kwargs):
         return PropertyCheckingFactory(class_name, (cls,), kwargs)
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -19,6 +19,7 @@ from eth_abi.exceptions import (
 
 from toolz.functoolz import (
     compose,
+    partial,
 )
 
 from web3.exceptions import (
@@ -61,6 +62,8 @@ from web3.utils.filters import (
 from web3.utils.normalizers import (
     BASE_RETURN_NORMALIZERS,
     normalize_address,
+    normalize_abi,
+    normalize_bytecode,
 )
 from web3.utils.transactions import (
     fill_transaction_defaults,
@@ -93,10 +96,11 @@ class ContractFunctions(object):
                     self,
                     function['name'],
                     ContractFunction.factory(
+                        function['name'],
                         web3=web3,
                         contract_abi=self.abi,
                         address=address,
-                        method_name=function['name']))
+                        function_name=function['name']))
 
 
 class ContractEvents(object):
@@ -115,6 +119,7 @@ class ContractEvents(object):
                     self,
                     event['name'],
                     ContractEvent.factory(
+                        event['name'],
                         web3=web3,
                         contract_abi=self.abi,
                         address=address,
@@ -187,13 +192,22 @@ class Contract(object):
         self.events = ContractEvents(self.abi, self.web3, self.address)
 
     @classmethod
-    def factory(cls, web3, contract_name=None, **kwargs):
-        if contract_name is None:
-            contract_name = cls.__name__
+    def factory(cls, web3, class_name=None, **kwargs):
 
         kwargs['web3'] = web3
 
-        contract = PropertyCheckingFactory(contract_name, (cls,), kwargs)
+        normalizers = {
+            'abi': normalize_abi,
+            'address': partial(normalize_address, kwargs['web3'].ens),
+            'bytecode': normalize_bytecode,
+            'bytecode_runtime': normalize_bytecode,
+        }
+
+        contract = PropertyCheckingFactory(
+            class_name or cls.__name__,
+            (cls,),
+            kwargs,
+            normalizers=normalizers)
         setattr(contract, 'functions', ContractFunctions(contract.abi, contract.web3))
         setattr(contract, 'events', ContractEvents(contract.abi, contract.web3))
 
@@ -696,7 +710,7 @@ class ImplicitMethod(ConciseMethod):
     def __call_by_default(self, args):
         # If function is constant in ABI, then call by default, else transact
         function_abi = find_matching_fn_abi(self._function.contract_abi,
-                                            fn_name=self._function.method_name,
+                                            fn_name=self._function.function_name,
                                             args=args)
         return function_abi['constant'] if 'constant' in function_abi.keys() else False
 
@@ -715,7 +729,7 @@ class ContractFunction(object):
     is a subclass of this class.
     """
     address = None
-    method_name = None
+    function_name = None
     web3 = None
     contract_abi = None
     abi = None
@@ -796,7 +810,7 @@ class ContractFunction(object):
                                       self.web3,
                                       self.address,
                                       self._return_data_normalizers,
-                                      self.method_name,
+                                      self.function_name,
                                       call_transaction,
                                       *self.args,
                                       **self.kwargs)
@@ -829,7 +843,7 @@ class ContractFunction(object):
         return transact_with_contract_function(self.contract_abi,
                                                self.address,
                                                self.web3,
-                                               self.method_name,
+                                               self.function_name,
                                                transact_transaction,
                                                *self.args,
                                                **self.kwargs)
@@ -864,7 +878,7 @@ class ContractFunction(object):
         return estimate_gas_for_function(self.contract_abi,
                                          self.address,
                                          self.web3,
-                                         function_name=self.method_name,
+                                         function_name=self.function_name,
                                          transaction=estimate_transaction,
                                          *self.args,
                                          **self.kwargs)
@@ -900,7 +914,7 @@ class ContractFunction(object):
         return build_transaction_for_function(self.contract_abi,
                                               self.address,
                                               self.web3,
-                                              self.method_name,
+                                              self.function_name,
                                               built_transaction,
                                               *self.args,
                                               **self.kwargs)
@@ -912,11 +926,8 @@ class ContractFunction(object):
     _return_data_normalizers = tuple()
 
     @classmethod
-    def factory(cls, **kwargs):
-        if "method_name" not in kwargs:
-            kwargs["method_name"] = cls.__name__
-
-        return PropertyCheckingFactory(kwargs["method_name"], (cls,), kwargs)
+    def factory(cls, class_name, **kwargs):
+        return PropertyCheckingFactory(class_name, (cls,), kwargs)
 
 
 class ContractEvent(object):
@@ -959,10 +970,8 @@ class ContractEvent(object):
         return get_event_data(self.abi, log)
 
     @classmethod
-    def factory(cls, **kwargs):
-        if "event_name" not in kwargs:
-            kwargs["event_name"] = cls.__name__
-        return PropertyCheckingFactory(kwargs["event_name"], (cls,), kwargs)
+    def factory(cls, class_name, **kwargs):
+        return PropertyCheckingFactory(class_name, (cls,), kwargs)
 
 
 def call_contract_function(abi,
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -615,28 +615,10 @@ class Contract(object):
 
     @classmethod
     def _find_matching_event_abi(cls, event_name=None, argument_names=None):
-        filters = [
-            functools.partial(filter_by_type, 'event'),
-        ]
-
-        if event_name is not None:
-            filters.append(functools.partial(filter_by_name, event_name))
-
-        if argument_names is not None:
-            filters.append(
-                functools.partial(filter_by_argument_name, argument_names)
-            )
-
-        filter_fn = compose(*filters)
-
-        event_abi_candidates = filter_fn(cls.abi)
-
-        if len(event_abi_candidates) == 1:
-            return event_abi_candidates[0]
-        elif not event_abi_candidates:
-            raise ValueError("No matching functions found")
-        else:
-            raise ValueError("Multiple functions found")
+        return find_matching_fn_abi(
+            contract_abi=cls.abi,
+            event_name=event_name,
+            argument_names=argument_names)
 
     @combomethod
     @coerce_return_to_text
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -29,8 +29,6 @@ from web3.exceptions import (
 
 from web3.utils.abi import (
     filter_by_type,
-    filter_by_name,
-    filter_by_argument_name,
     get_abi_output_types,
     get_constructor_abi,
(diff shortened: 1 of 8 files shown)


web3/contract.py
```diff
@@ -968,14 +953,66 @@ def factory(cls, **kwargs):
...
+
+    @classmethod
+    def factory(cls, **kwargs):
+        if "event_name" not in kwargs:
+            kwargs["event_name"] = cls.__name__
+
+        for key in kwargs:
+            if not hasattr(cls, key):
+                raise AttributeError(
+                    "Property {0} not found on ContractEvent class. "
```
pipermerriam: Since this logic is repeated a few times, what do you think about abstracting it away into a `Factory` class in `web3.utils.datatypes`

web3/utils/contracts.py
```diff
@@ -35,6 +42,32 @@
...
+        filters.append(functools.partial(filter_by_name, event_name))
+
+    if argument_names is not None:
+        filters.append(
+            functools.partial(filter_by_argument_name, argument_names)
+        )
+
+    filter_fn = compose(*filters)
+
+    event_abi_candidates = filter_fn(abi)
```
pipermerriam: We could probably make this more concise using `cytoolz.pipe`

```python
event_abi_candidates = pipe(abi, *filters)
```

web3/utils/events.py
```diff
@@ -98,10 +98,10 @@ def construct_event_data_set(event_abi, arguments=None):
         for key, value in arguments.items()
     }
 
-    indexed_args = exclude_indexed_event_inputs(event_abi)
+    non_indexed_args = exclude_indexed_event_inputs(event_abi)
```
pipermerriam: Thanks for cleaning this up.

pipermerriam:
Oh, can you add documentation for this as well.

web3/contract.py
```diff
@@ -181,57 +178,26 @@ def __init__(self, address=None):
...
-                    "present on the contract class".format(key)
-                )
-            else:
-                kwargs[key] = cls.normalize_property(key, kwargs[key])
-
-        kwargs['functions'] = ContractFunctions(kwargs.get('abi'), kwargs['web3'])
-        kwargs['events'] = ContractEvents(kwargs.get('abi'), kwargs['web3'])
+        contract = PropertyCheckingFactory(contract_name, (cls,), kwargs)
+        setattr(contract, 'functions', ContractFunctions(contract.abi, contract.web3))
+        setattr(contract, 'events', ContractEvents(contract.abi, contract.web3))
```
carver: Is there an advantage here, over the simpler:

```py
contract.events = ContractEvents(contract.abi, contract.web3)
```

web3/contract.py
```diff
@@ -965,15 +916,53 @@ def factory(cls, **kwargs):
...
+            decoded_log = get_event_data(self.abi, log)
+            yield decoded_log
+
+    def _decode_log(self, log):
+        return get_event_data(self.abi, log)
+
+    @classmethod
+    def factory(cls, **kwargs):
+        if "event_name" not in kwargs:
+            kwargs["event_name"] = cls.__name__
```
pipermerriam: `kwargs.setdefault('event_name', cls.__name__)` is a slightly more concise way to do this.

pipermerriam: Alternatively, this could be made generic in `PropertyCheckingFactory` if we changed everything to use `class_name` for this argument.

web3/utils/datatypes.py
```diff
@@ -0,0 +1,45 @@
...
+    normalize_address,
+    normalize_bytecode,
+)
+
+
+class PropertyCheckingFactory(type):
+    ens = None
+
+    @to_dict
+    def normalize_properties(attributes, values, ens):
```
pipermerriam: I don't love that this is included in the `PropertyCheckingFactory` since it's specific to the `Contract` class.  Could we change this so that `__new__` accepted an *optional* `normalizers` argument?

web3/utils/datatypes.py
```diff
@@ -0,0 +1,45 @@
...
+        for attribute, value in zip(attributes, values):
+            if attribute in normalizers:
+                normalizer = normalizers[attribute]
+                yield attribute, normalizer(value)
+            else:
+                yield attribute, value
+
+    def __new__(mcs, name, bases, namespace):
+        for key in namespace:
+            if not hasattr(bases[0], key):
```
pipermerriam: *nitpick*

This likely won't cause any issues with the way we're using this, but it should probably check all of the bases or even the set of all `set(concat(base.__mro__ for base in bases))` to cover the full inheritance tree.

web3/utils/normalizers.py
```diff
@@ -124,3 +129,28 @@ def abi_ens_resolver(w3, abi_type, val):
     addresses_checksummed,
     decode_abi_strings,
 ]
+
+
+# ----- Property Normalizers -----
```
pipermerriam: *nitpick*

I don't know it it's actually present in this codebase, but style-wise, for comment headings I prefer

```
#
# The Heading
#
```

I find it provides an easier to spot visual indicator that a new logical section is beginning.

dylanjw:
While working on the doc, I realized this doesnt work how I would like.  When I load transaction receipts with other event logs containing overlapping argument names, I get logs back, instead of an empty log object.

```
>>> tx_hash = contract.functions.logSingle(2, 12345).transact({'to':contract_address})
>>> tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
>>> rich_logs = contract.events.LogSingleArg().processReceipt(tx_receipt)
>>> rich_logs
({'args': {'arg0': 12345}, 'event': 'LogSingleArg', 'logIndex': 0, 'transactionIndex': 0, 'transactionHash': HexBytes('0x112fc27dec9de36ea4cd1266313b12835109fea9d8081314b7b18402aaf18a2c'), 'address': '0xc305c901078781C232A2a521C2aF7980f8385ee9', 'blockHash': HexBytes('0x972751d3ea29a7e6bd86fa73f1765eda03410e9191c0f7fa35f8cf45f5d3d68c'), 'blockNumber': 2},)
>>> tx_hash2 = contract.functions.logDouble(3, 55555, 66666).transact({'to':contract_address})
>>> tx_receipt2 = w3.eth.getTransactionReceipt(tx_hash2)
>>> contract.events.LogSingleArg().processReceipt(tx_receipt2)
({'args': {'arg0': 55555}, 'event': 'LogSingleArg', 'logIndex': 0, 'transactionIndex': 0, 'transactionHash': HexBytes('0xa2de39297d47a1a350cccc00ef007c608b0bdd50e1f2f3d2c5e114a9278ae8e7'), 'address': '0xc305c901078781C232A2a521C2aF7980f8385ee9', 'blockHash': HexBytes('0xdae2dca798d07dcd32d7f39cf0ce81e521b47c59222cbf1defde2af871f6557c'), 'blockNumber': 3},)
>>> contract.events.LogDoubleArg().processReceipt(tx_receipt2)
({'args': {'arg0': 55555, 'arg1': 66666}, 'event': 'LogDoubleArg', 'logIndex': 0, 'transactionIndex': 0, 'transactionHash': HexBytes('0xa2de39297d47a1a350cccc00ef007c608b0bdd50e1f2f3d2c5e114a9278ae8e7'), 'address': '0xc305c901078781C232A2a521C2aF7980f8385ee9', 'blockHash': HexBytes('0xdae2dca798d07dcd32d7f39cf0ce81e521b47c59222cbf1defde2af871f6557c'), 'blockNumber': 3},)
```

Im going to write more tests, and work on improving this.

carver:
> While working on the doc, I realized this doesnt work how I would like.

Hah, this happens to me, too. Occasionally, I think I might like to do doc-first development to deal with this issue, but too many things shake out in the implementation, that I haven't made any real effort at it yet.

> When I load transaction receipts from other function calls containing other events that have overlapping argument names, I get logs back, instead of an empty log object or an error.

Good catch. If something significant changes in the implementation, ping us for another review. If it's a small, simple change, then go ahead and merge.

dylanjw:
I changed `get_event_data` to raise an exception if the provided abi does not match the event signature in the log entry.  I noticed this breaks this profiling script: https://gist.github.com/konradkonrad/87d76dd2526f39e6f5bad19b85374948 from https://github.com/ethereum/web3.py/issues/469, which made me concerned people might be using `get_event_data`, and have issue with making this more strict.

pipermerriam:
@dylanjw our policy is that anything within the `utils` module is a private API and may change in breaking ways at any point.

web3/utils/datatypes.py
```diff
@@ -0,0 +1,50 @@
...
+                yield attribute, normalizer(value)
+            else:
+                yield attribute, value
+
+    def __init__(cls, name, bases, namespace, **kargs):
+        # see PEP487.  To accept kwargs in __new__, they need to be
+        # filtered out here.
+        super().__init__(name, bases, namespace)
+
+    def __new__(mcs, name, bases, namespace, normalizers={}):
```
pipermerriam: My general rule is that mutable data structures are never ok as default arguments for functions.

```
In [1]: def f(a, x=[]):
   ...:     x.append(a)
   ...:     return x
   ...:

In [2]: f(1)
Out[2]: [1]

In [3]: f(2)
Out[3]: [1, 2]

In [4]: f(3)
Out[4]: [1, 2, 3]
```

This specific case doesn't appear to have these problems, but I still think it's a good practice for us to avoid this pattern in favor of the following.

```python
def __new__(mcs, name, bases, namespace, normalizers=None):
    if normalizers is None:
        normalizers = {}
    ...
```

web3/utils/datatypes.py
```diff
@@ -0,0 +1,50 @@
...
+        for key in namespace:
+            verify_key_attr = partial(verify_attr, name, key)
```
pipermerriam: Instead of using `partial` I'll recommend looking into the [`cytoolz.curry`](http://toolz.readthedocs.io/en/latest/api.html#toolz.functoolz.curry) which makes for a nicer partialing API

web3/utils/datatypes.py
```diff
@@ -0,0 +1,50 @@
+from functools import (
+    partial,
```
pipermerriam: Take a look at https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#avoid-from-x-import-y-for-most-standard-library-imports

The general idea is that seeing `functools.partial` somewhere deep in a module makes it pretty clear that I'm looking at the `partial` function from the standard library `functools`.  However, if all I see is `partial` then I have to go back to the top of the module to find out if I'm actually looking at `functools.partial`.  Makes code slightly easier to parse at the cost of a bit of verbosity.

web3/utils/datatypes.py
```diff
@@ -0,0 +1,50 @@
...
+        # see PEP487.  To accept kwargs in __new__, they need to be
+        # filtered out here.
+        super().__init__(name, bases, namespace)
+
+    def __new__(mcs, name, bases, namespace, normalizers={}):
+        for key in namespace:
+            verify_key_attr = partial(verify_attr, name, key)
+            map(verify_key_attr, set(concat(base.__mro__ for base in bases)))
+
+        attributes, values = zip(*namespace.items())
```
pipermerriam: I find it somewhat odd that `normalize_properties` takes the attrs and values as separate arguments.

First, *if* attributes and values were of different lengths then the extra ones from whichever was longer would be silently discarded.  This could be fixed by passing them in as a single iterable of 2-tuples.  `(('key-1', 'value-a'), ('key-2', 'value-2'), ...)`

Second, It seems even cleaner to just pass the `namespace` directly in and change it to be `mcs.normalize_properties(namespace, normalizers)` and then do the `items()` call within that function.

And last, `mcs.normalize_properties` could be replaced with `web3.utils.formatters.apply_formatters_to_dict`.
- it seems like we should be able to instea

web3/utils/events.py
```diff
@@ -147,11 +151,15 @@ def get_event_abi_types_for_decoding(event_inputs):
...
     Given an event ABI and a log entry for that event, return the decoded
+    event data
     """
     if event_abi['anonymous']:
         log_topics = log_entry['topics']
     else:
         log_topics = log_entry['topics'][1:]
+        raw_signature = next(iter(log_entry['topics']), None)
+        if event_abi_to_log_topic(event_abi) != raw_signature:
+            raise MismatchedABI()
```
pipermerriam: Can you add a message to this which indicates what the mismatch was.

web3/utils/events.py
```diff
@@ -147,11 +151,15 @@ def get_event_abi_types_for_decoding(event_inputs):
 def get_event_data(event_abi, log_entry):
     """
     Given an event ABI and a log entry for that event, return the decoded
+    event data
     """
     if event_abi['anonymous']:
         log_topics = log_entry['topics']
     else:
         log_topics = log_entry['topics'][1:]
+        raw_signature = next(iter(log_entry['topics']), None)
```
pipermerriam: Given that I found this line cryptic, I wanted to look at other ways to do this.

```python
# using cytoolz
from cytoolz import first, excepts

raw_signature = excepts(StopIteration, first, lambda v: None)(log_entry['topics'])

# more verbose
try:
    raw_signature = log_entry['topics'][0]
except IndexError:
    raw_signature = None

# or
if log_entry['topics']:
    raw_signature = log_entry['topics'][0]
else:
    raw_signature = None

# and probably my favorite because it produces better error messages
if not log_entry['topics']:
    raise MismatchedABI("Expected event to have a topic but it had not topics")  # this msg could be better
elif event_abi_to_log_topic(event_abi) != log_entry['topics'][0]:
    raise MismatchedABI("<put a good message here>")
```

web3/utils/datatypes.py
dylanjw: Oh no! its the example in the manual <:‑|

pipermerriam: Learning a team's code style takes some time and immersion.

web3/utils/datatypes.py
carver: Castle of Aaaaargh?

Seems like the last comment here got cut off.

dylanjw: i hope he's ok

pipermerriam: No clue

web3/utils/datatypes.py
```diff
@@ -0,0 +1,38 @@
...
+class PropertyCheckingFactory(type):
+    def __init__(cls, name, bases, namespace, **kargs):
+        # see PEP487.  To accept kwargs in __new__, they need to be
+        # filtered out here.
+        super().__init__(name, bases, namespace)
+
+    def __new__(mcs, name, bases, namespace, normalizers=None):
+        for key in namespace:
+            verify_key_attr = verify_attr(name, key)
+            map(verify_key_attr, set(concat(base.__mro__ for base in bases)))
```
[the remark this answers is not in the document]
dylanjw: Thanks for catching that!

PR #579, opened 2018-01-23
Automatic import sorting and compliance testing

carver:
### What was wrong?

Our import style was not being enforced, and was fairly irregular.

### How was it fixed?

See 54573c7 for the key changes, the rest are whitespace/sorting changes.

- Added an `isort` dependency to a new `linter` extra in setup.py
- Configure `isort` very close to our current style approach (they didn't really have the idea of separating sub-packages by a newline)
- Ran `isort` on all the files in the repo
- Add `isort` compatibility test to the new tox `lint` test (which now runs both flake8 and isort)

Automatically run the sort on your files with: 

```sh
isort --recursive --settings-path tox.ini web3 ens tests
```

Probably in a git precommit hook

#### Cute Animal Picture

Cute animal picture


carver:
> I didn't actually check all the changes to the imports.

Understandable. They were all auto-generated, but I did check them in piece by piece with `git add -p`. Then if the patch didn't seem to fit with our current thoughts on best practices, I went to tweak the tox.ini `[isort]` config section and re-generate (except for two unsalvageable files: web3/main.py and web3/utils/windows.py, which I just set to ignore)

PR #584, opened 2018-01-26
remove legacy support for to_hex(dict)

carver:
### What was wrong?

Cruft support is unnecessary.

### How was it fixed?

Removed it, updated error about required type. It was undocumented, so no doc change necessary.

#### Cute Animal Picture

Cute animal picture


carver:
I really just wanted a full suite test against this. Closing and will (later) make a new one with these conversion utilities moved to eth-utils.

PR #585, opened 2018-01-27
Eth tester middleware: conform nonce parameter

monokh:
### What was wrong?

Eth_tester validates nonces as integers however web3.py passes hex when using the eth tester middleware. This causes transactions with a nonce to fail when using this middleware. example:

```
from eth_tester import EthereumTester

from web3.providers.eth_tester import EthereumTesterProvider
web3 = Web3(EthereumTesterProvider(tester))
tx = web3.eth.sendTransaction({
    'from': '0x82A978B3f5962A5b0957d9ee9eEf472EE55B42F1',
    'to': '0x7d577a597B2742b498Cb5Cf0C26cDCD726d39E6e',
    'value': 10000000,
    'gas': 23500,
    'nonce': 1
})
```

results in

`eth_tester.exceptions.ValidationError: Value must be positive integers. Got: 0x1`

### How was it fixed?

Transform hex to integer for the nonce parameter by the eth_tester middleware.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     setup_requires=['setuptools-markdown'],
     extras_require={
         'tester': [
-            "eth-tester[py-evm]==0.1.0b13",
+            "eth-tester[py-evm]==0.1.0b15",
         ],
         'testrpc': ["eth-testrpc>=1.3.3,<2.0.0"],
         'platform_system=="Windows"': [
--- a/web3/providers/eth_tester/middleware.py
+++ b/web3/providers/eth_tester/middleware.py
@@ -109,6 +109,7 @@ TRANSACTION_PARAMS_FORMATTERS = {
     'gas': to_integer_if_hex,
     'gasPrice': to_integer_if_hex,
     'value': to_integer_if_hex,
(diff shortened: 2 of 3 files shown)


carver:
Can you add a test that sends a transaction with a nonce in `utils/module_testing/eth_module.py`, and make sure it fails without this change?

carver:
Looks like geth is complaining about a duplicate transaction. If this is producing the exact same transaction as another test, you could try increasing the gas price on only this test, so the hash will be different.

monokh:
@carver had the same sort of idea but wanted to get tox running to make sure it was fixed before committing. Internet is not great where I am for now so i tried going in blind with your suggestion.

I tried:

different value: `transaction underpriced` 
setting `gas_price` to `web3.eth.gasPrice * 2`: `known transaction`.
both changes above: `transaction underpriced`

So now i'm trying again to get geth tests running locally 😄

monokh:
@carver turns out, we were passing an incorrect `gas_price` (instead of `gasPrice`) to web3 in the tests 😄

PR #586, opened 2018-01-27
Resend transaction API

### LINKED ISSUE #492: API for re-sending transactions
### What was wrong?

There are a few legitimate use cases for wanting to re-submit a transaction that has already been sent with modified parameters.

* Resending the same transaction with a higher gas price during times of network congestion.
* Quickly overriding a transaction sent in error (also by using a higher gas price)
* In a case where you *know* there is a signed transaction that was sent but never included in a block, ensuring that it will not be applied at some point in the future.

### How can it be fixed?

Implement a new API `web3.eth.overrideTransaction(txn_hash, txn_overrides)`

- `txn_hash` is the hash of the transaction you wish to override.
- `txn_overrides` is a dictionary of transaction parameters which should be used in place of the existing transaction parameters.

Behavior is as follows.

- *if* no transaction can be found with the given hash an exception is thrown.
- *if* the transaction has already been mined an exception is thrown.
- The `txn_overrides` may contain any/all fields that are normally accepted as transaction params except `nonce`.
- If `gasPrice` is included in the `txn_overrides`, it **must** be greater than the existing transaction gas price (TODO: should we allow a way to bypass this validation)
- if `gasPrice` is **not** included in the `txn_overrides` we should ???? (TODO: how do we choose an appropriate gas price?).

monokh:
### What was wrong?

There is no convenient way to resend transactions.

### How was it fixed?

Implemented the following APIs:

`w3.eth.replaceTransaction(txn_hash, new_transaction_dict)` : Submits a new transaction that replaces the transaction with `txn_hash`

`w3.eth.modifyTransaction(txn_hash, **individual_fields_to_replace)` : Use new values as specified in `individual_fields_to_replace` but otherwise keep the same parameters when resending the transaction.

In both cases the following rules apply:

- if no transaction can be found with the given hash an exception is thrown.
- if the transaction has already been mined an exception is thrown.
- The new transaction may contain any/all fields that are normally accepted as transaction params except nonce. If nonce is included, it must match the existing transaction.
- If gasPrice is included in the new transaction, it must be greater than the existing transaction gas price (TODO: should we allow a way to bypass this validation)
- if gasPrice is not included in the new transaction we should get two values, and use the maximum price of the two:
  - The gas price of the pending transaction times 1.1 (the minimum required to propagate)
  - The gas price as calculated by the current gas price engine

TODO: 

- [x] Implement transaction existing validation
- [x] Implement nonce validation
- [x] Implement gas price validation and generation
- [x] Integration tests
- [x] Tests for new utils in `utils/transactions.py`
- [x] Investigate geth integration tests
- [x] Documentation
- [x] Fix tests - issues in eth-tester

#### Cute Animal Picture

Cute animal picture

Depends on #585 
Fixes #492


web3/eth.py
```diff
@@ -198,6 +200,25 @@ def getTransactionCount(self, account, block_identifier=None):
             ],
         )
 
+    def replaceTransaction(self, transaction_hash, new_transaction):
+        current_transaction = self.getTransaction(transaction_hash)
+        new_transaction = prepare_replacement_transaction(
+            self.web3, current_transaction, new_transaction
+        )
+
+        # TODO: Does the gas and from defaulting in `sendTransaction` make sense for this?
```
carver: Yes, in the replacement scenario, defaulting the `gas` and `from` fields normally seems reasonable.

web3/eth.py
```diff
@@ -198,6 +200,25 @@ def getTransactionCount(self, account, block_identifier=None):
...
+        return self.sendTransaction(new_transaction)
+
+    def modifyTransaction(self, transaction_hash, **params):
+        current_transaction = self.getTransaction(transaction_hash)
+        new_transaction = merge(current_transaction, params)
+        new_transaction = prepare_replacement_transaction(
+            self.web3, current_transaction, new_transaction
+        )
+
+        # TODO: Does the gas and from defaulting in `sendTransaction` make sense for this?
```
carver: No, but it doesn't matter because there will be no defaults to fill. `current_transaction` will supply `gas` and `from` fields.

web3/utils/transactions.py
```diff
@@ -168,3 +170,28 @@ def get_buffered_gas_estimate(web3, transaction, gas_buffer=100000):
...
+    if current_transaction['blockHash'] is not None:
+        raise ValueError('Supplied transaction with hash {} has already been mined'
+                         .format(current_transaction['hash']))
+    if 'nonce' in new_transaction and new_transaction['nonce'] != current_transaction['nonce']:
+        raise ValueError('Supplied nonce in new_transaction must match the pending transaction')
+
+    if 'nonce' not in new_transaction:
+        new_transaction = assoc(new_transaction, 'nonce', current_transaction['nonce'])
+
+    # TODO: Possibly in a middlware somehow?
```
carver: Not sure this can go in the middleware, because all the middleware can see is the method being called: `eth_sendTransaction`. That's not enough to know that we should use this special replacement gas price approach.

tests/integration/test_ethereum_tester.py
```diff
@@ -190,6 +190,8 @@ def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_acco
             eth_tester.enable_auto_mine_transactions()
             eth_tester.mine_block()
 
+    # TODO: replace auto mining switch with decorator
+    # TODO: replicate for goethereum tests
```
pipermerriam: *if* it turns out that testing this in the integration tests is messy lets look at other options and whether it's even necessary before you burn too much time trying to make it work.  Since this functionality will likely behave differently on different clients, and it's possible that some clients won't reliably accept the replacement transaction I could see these tests getting really messy.

tests/integration/test_ethereum_tester.py
```diff
@@ -190,6 +190,8 @@ def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_acco
             eth_tester.enable_auto_mine_transactions()
             eth_tester.mine_block()
 
+    # TODO: replace auto mining switch with decorator
```
pipermerriam: 👍

tests/integration/test_ethereum_tester.py
monokh: It seems like the `--fakepow` option to geth is what is mining the blocks for the integration tests? There isn't a way to disable that through a RPC or console call so if we wanted to do it, i believe we would need to recreate the geth process for those test specifically. The other option would be to use the miner and call `miner.start()` and `miner.stop()` similar to the `eth-tester` tests. Not sure how this would affect other tests though. Just an initial look, i will need to have a closer look at this.

I guess if the API promises to support a client then we would need to know if the integration breaks so I can't really say that it's unnecessary.

pipermerriam: We had major issues with tests timing out when we tried to do `miner.start/stop()` in our CI environment.  I'd rather skip testing this stuff in integration tests since the tests become unreliable if we're required to mine.

web3/utils/module_testing/eth_module.py
```diff
@@ -213,6 +213,27 @@ def test_eth_sendTransaction(self, web3, unlocked_account):
...
+            'value': 1,
+            'gas': 21000,
+            'gasPrice': web3.eth.gasPrice,
+        }
+        txn_hash = web3.eth.sendTransaction(txn_params)
+        web3.eth.getTransaction(txn_hash)
+
+        txn_params['gasPrice'] = web3.eth.gasPrice * 2
+        replace_txn_hash = web3.eth.replaceTransaction(txn_hash, txn_params)
+        replace_txn = web3.eth.getTransaction(replace_txn_hash)
```
carver: I am pretty sure that this should work, too, after replacement:

```py
assert web3.eth.getTransaction(txn_hash) is None
```

tests/integration/test_ethereum_tester.py
```diff
@@ -190,6 +190,16 @@ def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_acco
             eth_tester.enable_auto_mine_transactions()
             eth_tester.mine_block()
 
+    def test_eth_replaceTransaction(self, eth_tester, web3, unlocked_account):
+        eth_tester.disable_auto_mine_transactions()
+        try:
+            super(TestEthereumTesterEthModule, self).test_eth_replaceTransaction(
```
carver: Now that we're in py3-only, we can do `super()`. Yay :)

tests/integration/test_ethereum_tester.py
```diff
@@ -190,6 +190,16 @@ def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_acco
...
+                web3, unlocked_account,
+            )
+        finally:
+            eth_tester.enable_auto_mine_transactions()
+            eth_tester.mine_block()
```
carver: Would be nice to confirm here that the replacement transaction is the one that gets mined.

web3/utils/transactions.py
```diff
@@ -185,13 +185,15 @@ def prepare_replacement_transaction(web3, current_transaction, new_transaction):
...
-    # TODO: Possibly in a middlware somehow?
-    if 'gasPrice' not in new_transaction:
-            generated_gas_price = web3.eth.generateGasPrice(new_transaction)
-            minimum_gas_price = int(math.ceil(current_transaction['gasPrice'] * 1.1))
-            if generated_gas_price and generated_gas_price > minimum_gas_price:
-                new_transaction = assoc(new_transaction, 'gasPrice', generated_gas_price)
-            else:
-                new_transaction = assoc(new_transaction, 'gasPrice', minimum_gas_price)
+    if 'gasPrice' in new_transaction:
+        if new_transaction['gasPrice'] < current_transaction['gasPrice']:
```
carver: `<=`

carver:
Hm, github notifications showed me an outdated diff. Trying that again...

tests/integration/test_ethereum_tester.py
```diff
@@ -190,6 +190,57 @@ def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_acco
             eth_tester.enable_auto_mine_transactions()
             eth_tester.mine_block()
 
+    # TODO: replace auto mining switch with decorator
+    # TODO: replicate for goethereum tests
+    def test_eth_replaceTransaction(self, eth_tester, web3, unlocked_account):
+        eth_tester.disable_auto_mine_transactions()
+        try:
+            super(TestEthereumTesterEthModule, self).test_eth_replaceTransaction(
```
carver: Now that we're in py3-only, we can do `super()`. Yay :)

Makes the decorator cleaner, too.

tests/integration/test_ethereum_tester.py
carver: ie~ would like to this as part of this PR, eventually :)

web3/utils/module_testing/eth_module.py
```diff
@@ -213,6 +213,145 @@ def test_eth_sendTransaction(self, web3, unlocked_account):
...
+            'to': unlocked_account,
+            'value': 1,
+            'gas': 21000,
+            'gasPrice': web3.eth.gasPrice,
+        }
+        txn_hash = web3.eth.sendTransaction(txn_params)
+
+        txn_params['gasPrice'] = web3.eth.gasPrice * 2
+        replace_txn_hash = web3.eth.replaceTransaction(txn_hash, txn_params)
+        replace_txn = web3.eth.getTransaction(replace_txn_hash)
```
carver: I am pretty sure that this should work, too, after replacement:

```py
assert web3.eth.getTransaction(txn_hash) is None
```

If it works in other nodes, but fails in eth-tester, then please make a tracking issue in eth-tester.

web3/utils/transactions.py
```diff
@@ -168,3 +180,37 @@ def get_buffered_gas_estimate(web3, transaction, gas_buffer=100000):
...
+        raise ValueError('Supplied transaction with hash {} has already been mined'
+                         .format(current_transaction['hash']))
+    if 'nonce' in new_transaction and new_transaction['nonce'] != current_transaction['nonce']:
+        raise ValueError('Supplied nonce in new_transaction must match the pending transaction')
+
+    if 'nonce' not in new_transaction:
+        new_transaction = assoc(new_transaction, 'nonce', current_transaction['nonce'])
+
+    if 'gasPrice' in new_transaction:
+        if new_transaction['gasPrice'] < current_transaction['gasPrice']:
```
carver: The replacement must be strictly greater than current, so this check should be `<=`

web3/utils/module_testing/eth_module.py
monokh: On initial thought, I think the transaction would exist until the block started mining regardless of node in which case it could make the tests really strange to add this.  I'll check the behaviour.

tests/integration/test_ethereum_tester.py
monokh: I've made these tests pass for the geth integration tests :)

tests/core/utilities/test_prepare_transaction_replacement.py
```diff
@@ -0,0 +1,122 @@
...
+        'value': 2,
+    }
+
+    replacement_transaction = prepare_replacement_transaction(
+        web3, current_transaction, new_transaction)
+
+    assert replacement_transaction['gasPrice'] == 20
+
+
+def test_prepare_transaction_replacement_gas_price_defaulting_when_strategy_lower(web3):
```
monokh: Now that we have these tests, could remove some of the integration tests that test the same thing?

monokh:
Just docs and some double checking to do now, so feel free to have a deeper look.

tests/integration/test_ethereum_tester.py
```diff
@@ -175,18 +175,56 @@ def inner(*args, **kwargs):
     return inner
 
 
-class TestEthereumTesterEthModule(EthModuleTest):
-    test_eth_sign = not_implemented(EthModuleTest.test_eth_sign, ValueError)
-
-    def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_account):
+def disable_auto_mine(func):
+    @functools.wraps(func)
+    def func_wrapper(self, eth_tester, web3, unlocked_account):
```
carver: Since you're not doing anything with the other args, this could be `(self, eth_tester, *args)`, for a little more generality if the wrapper gets used on a method with a different signature.

web3/utils/module_testing/eth_module.py
monokh: Tried with geth, indeed the old transaction is removed 🙂 
Raised https://github.com/ethereum/eth-tester/issues/57 
Can have a look at it in the weekend

carver:
Great! It looks like passing the tests is all that's left.

monokh:
Pending on https://github.com/ethereum/eth-tester/pull/60 being fixed

carver:
@monokh This should be unblocked now. eth-tester beta 18 includes https://github.com/ethereum/eth-tester/pull/60

I recommend a rebase.

tests/integration/test_parity.py
```diff
@@ -297,6 +297,46 @@ def test_eth_newBlockFilter(self, web3):
         pytest.xfail('Parity returns latest block on first polling for new blocks')
         super().test_eth_newBlockFilter(web3)
 
+    def test_eth_replaceTransaction(self, web3, unlocked_account):
+        pytest.xfail('Needs ability to efficiently control mining')
+        super().test_eth_replaceTransaction(web3, unlocked_account)
+
+    def test_eth_replaceTransaction_already_mined(self, web3, unlocked_account):
+        pytest.xfail('Parity is not setup to auto mine')
```
monokh: right?

carver: Yes, I believe that's correct. cc @dylanjw

carver:
Nice work taking this to the finish line, @monokh

PR #596, opened 2018-01-28
Check for attribute in ANY base

dylanjw:
### What was wrong?

I introduced a bug in web3.utils.datatypes.PropertyCheckingFactory.  The parameter checking should fail only if the parameter is missing from ALL bases.

### How was it fixed?

It now combines all bases into a single namespace before checking.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/datatypes.py
+++ b/web3/utils/datatypes.py
@@ -28,7 +28,7 @@ class PropertyCheckingFactory(type):
         all_bases = set(concat(base.__mro__ for base in bases))
         for key in namespace:
             verify_key_attr = verify_attr(name, key)
-            verify_key_attr(concat(base.__dict__ for base in all_bases))
+            verify_key_attr(concat(base.__dict__.keys() for base in all_bases))
 
         if normalizers:
             processed_namespace = web3.utils.formatters.apply_formatters_to_dict(
--- a/web3/utils/datatypes.py
+++ b/web3/utils/datatypes.py
@@ -9,8 +9,8 @@ import web3.utils.formatters
 
 
 @curry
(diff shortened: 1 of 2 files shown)


web3/utils/datatypes.py
```diff
@@ -25,9 +25,10 @@ def __init__(cls, name, bases, namespace, **kargs):
         super().__init__(name, bases, namespace)
 
     def __new__(mcs, name, bases, namespace, normalizers=None):
+        all_bases = set(concat(base.__mro__ for base in bases))
         for key in namespace:
             verify_key_attr = verify_attr(name, key)
-            map(verify_key_attr, set(concat(base.__mro__ for base in bases)))
+            verify_key_attr(concat(base.__dict__ for base in all_bases))
```
pipermerriam: personal nitpick here.  I hate the implicit key iteration of dictionaries.  I personally have *never* internalized that dictionary iteration doesn't do `key/value` iteration and thus, have resorted to never using direct iteration on dictionaries, instead preferring `keys()/values()/items()` explicitely.

Can you change to `base.__dict__.keys()`.  Somewhat based on personal preference, but also it helps reduce cognitive load to understand this is pulling all the keys from all the dictionaries.

pipermerriam: If there is any agreement with this approach, I can add it to the tactical manual.

dylanjw: I agree.  It makes it more clear.

dylanjw:
@pipermerriam Ok to merge?

dylanjw:
Just just @carver's gitter comment in gitter requesting a test.  I will write one now.

Issue #577, opened 2018-01-23
Support for fallback and constructor functions


pipermerriam:
Continuation of https://github.com/ethereum/web3.py/issues/353

### What was wrong?

We should support the following API on contract instances for interacting with the fallback function.

- `contract.fallback.estimateGas()`
- `contract.fallback.call()`
- `contract.fallback.transact()`
- `contract.fallback.buildTransaction()`

As well as the following for interacting with the constructor.

- `contract.constructor(*args, **kwargs).estimateGas()`
- `contract.constructor(*args, **kwargs).transact()`
- `contract.constructor(*args, **kwargs).buildTransaction()`

### How can it be fixed?

Hopefully we can leverage the existing `web3.contract.ContractMethod` abstraction.

gitcoinbot:
__This issue now has a funding of 0.4 ETH (402.8 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/577).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/577)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $16127.14 more Funded OSS Work Available at: https://gitcoin.co/explorer

pipermerriam:
@carver @dylanjw thoughts on this?

* Should `contract.deploy()` be deprecated?
* Should `contract.constructor(...).transact()` be named something like `contract.constructor(...).deploy()`?

carver:
> Should contract.deploy() be deprecated?

Sounds good to me. A new API for the local signing of the deploy transaction is needed anyway, and I like that `constructor` consistent with the ABI type name.

> Should contract.constructor(...).transact() be named something like contract.constructor(...).deploy()

I think we should be consistent with the other options here (call, buildTransaction, etc)

gitcoinbot:
__The funding of 0.4 ETH (420.48 USD) attached has been claimed by @zheli.__ 

 @zheli, please leave a comment to let the funder  and the other parties involved your implementation plan.  If you don't leave a comment, the funder may expire your claim at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/577)
[...]

PR #597, opened 2018-01-28
Add fallback support

zheli:
### What was wrong?

Original description from https://github.com/ethereum/web3.py/issues/577

> We should support the following API on contract instances for interacting with the fallback function.
> 
> contract.fallback.estimateGas()
> contract.fallback.call()
> contract.fallback.transact()
> contract.fallback.buildTransaction()
> 
> As well as the following for interacting with the constructor.
> 
> contract.constructor(*args, **kwargs).estimateGas()
> contract.constructor(*args, **kwargs).transact()
> contract.constructor(*args, **kwargs).buildTransaction()
> 

### How was it fixed?

Find fallback method abi by abi type 'fallback', then add some logic in ContractMethod class to handle it.


web3/contract.py
```diff
@@ -268,6 +270,12 @@ def deploy(cls, transaction=None, args=None, kwargs=None):
         txn_hash = cls.web3.eth.sendTransaction(deploy_transaction)
         return txn_hash
 
+    def fallback(self):
+        if self._fallback:
+            return self._fallback
+        else:
+            raise FallbackNotFound("Fallback function does not exist.")
```
pipermerriam: I think a slightly more informative error message would be

> No fallback function was found in the contract ABI

web3/utils/abi.py
```diff
@@ -66,6 +66,11 @@ def get_abi_input_names(abi):
         return [arg['name'] for arg in abi['inputs']]
 
 
+def get_fallback_func_abi(contract_abi):
+    fallback_abis = filter_by_type('fallback', contract_abi)
+    return fallback_abis[0] if fallback_abis else None
```
pipermerriam: Instead of returning `None` here, I think it would be better to raise the `FallbackNotFound` exception.  Otherwise, any use of this function requires that we remember to *check* the return value which is easy to forget.  Forgetting to handle the exception will produce a much clearer error message than an un-expected `None` type floating around where we expected an ABI to be.

web3/utils/contracts.py
```diff
@@ -164,24 +172,32 @@ def prepare_transaction(abi,
...
     if kwargs is None:
         kwargs = {}
 
-    fn_abi = find_matching_fn_abi(abi, fn_name, args, kwargs)
+    if is_fallback_fn:
+        fn_abi = get_fallback_func_abi(abi)
+    else:
+        fn_abi = find_matching_fn_abi(abi, fn_name, args, kwargs)
+
     fn_selector = encode_hex(function_abi_to_4byte_selector(fn_abi))
```
pipermerriam: The fallback function has no function selector and I don't think `function_abi_to_4byte_selector` knows how to handle this correctly.  The function selector should be set to `b''` if we're interacting with the fallback function.

carver: I don't know how sending data to fallback works. If there is data there, isn't there some risk that the first four bytes of the data accidentally matches a function selector that does exist, and then you're not interacting with the fallback when you think you should be?

carver: I won't be able to investigate that right now, so I guess if we don't have a clear answer at hand: the short-term fix is to disallow data of any kind to the fallback. We can revisit later.

web3/contract.py
zheli: Sure, I will change it.

web3/utils/abi.py
zheli: Good point!

web3/utils/contracts.py
pipermerriam: You can send data to the fallback.  If the first 4 bytes of the data matches a function selector the contract will instead call that function.  I think we should do one of the following.

- do a check on whether the first four data bytes matches any known function selectors from the ABI and throw an error.
- do the same check but allow for an extra parameter to the function to explicitly allow the call to go through anyways (this one is my preference)

pipermerriam: By the way, my reasoning for wanting to support sending data that does in-fact match an existing function selector is the following.  It's a minority use case but it's real and shouldn't be too hard to support.

- I want to call an existing function (like the ERC20 `transfer`)
- I want to call it with a custom data packet (there was a recent bug some exchanges exposed with them not padding data correctly, a security researcher would want to be able to use this feature to pen-test).

Admittedly, they can do the same thing with simply `web3.eth.sendTransaction` but I think supporting it through the contract interface would be nice.

We **could** also just expose a new api `contract.sendTransaction` which just auto-fills the `to` address and allows you to specify everything else.

zheli: Sorry for the late reply, I will add the implementation based on your second suggestion.

carver: Yes this seems like the best solution:

> do the same check but allow for an extra parameter to the function to explicitly allow the call to go through anyways (this one is my preference)

@zheli that is the one you chose, right?

zheli: @carver yep.

web3/utils/module_testing/fallback_contract.py
```diff
@@ -0,0 +1,25 @@
...
+FALLBACK_BYTECODE = (
+    "6060604052341561000f57600080fd5b60ae8061001d6000396000f300606060405260043610603f"
+    "576000357c0100000000000000000000000000000000000000000000000000000000900463ffffff"
+    "ff1680633bc5de30146053575b3415604957600080fd5b6001600081905550005b3415605d576000"
+    "80fd5b60636079565b6040518082815260200191505060405180910390f35b600080549050905600"
+    "a165627a7a723058204233f675dc222d76091504fcf32c45f024c9154ad12ae3bc0b411ab33ddc59"
+    "840029"
+)
+
+FALLBACK_ABI = [
```
pipermerriam: I don't think this module is needed.  There isn't anything special about testing fallback functions that would require us to test them across the various integration backends.  Can you remove this file?

web3/utils/contracts.py
```diff
@@ -164,24 +172,32 @@ def prepare_transaction(abi,
...
+def encode_transaction_data(abi, web3, fn_name, args=None, kwargs=None, is_fallback_fn=False):
     fn_abi, fn_selector, fn_arguments = get_function_info(
-        abi, fn_name, args, kwargs,
+        abi, fn_name, args, kwargs, is_fallback_fn
     )
     return add_0x_prefix(encode_abi(web3, fn_abi, fn_arguments, fn_selector))
 
 
-def get_function_info(abi, fn_name, args=None, kwargs=None):
+def get_function_info(abi, fn_name, args=None, kwargs=None, is_fallback_fn=False):
```
pipermerriam: I think that rather than making `get_function_info` support retrieving the fallback function you should just do it one level higher in `encode_transaction_data`.

web3/utils/contracts.py
```diff
@@ -164,24 +172,32 @@ def prepare_transaction(abi,
...
+        fn_is_fallback
     )
     return prepared_transaction
 
 
-def encode_transaction_data(abi, web3, fn_name, args=None, kwargs=None):
+def encode_transaction_data(abi, web3, fn_name, args=None, kwargs=None, is_fallback_fn=False):
     fn_abi, fn_selector, fn_arguments = get_function_info(
-        abi, fn_name, args, kwargs,
+        abi, fn_name, args, kwargs, is_fallback_fn
```
pipermerriam: Rather than using `is_fallback_fn` I'm think we could clean this up using the following pattern.

```
class fallback_fn:
    pass


def encode_transaction_data(abi, web3, fn_identifier, args=None, kwargs=None):
    if fn_identifier is fallback_fn:
        fn_abi, fn_selector, fn_arguments = get_fallback_function_info(...)
    elif is_text(fn_identifier):
        fn_abi, fn_selector, fn_arguments = get_function_info(...)
    else:
        raise TypeError("Unsupported function identifier")

    ...
```

The *new* `get_fallback_function_info` handles all the special cases that were previously handled by `is_fallback` and the `get_function_info` remains simple with less logic forks.

The `fallback_fn` class simply serves as an unambiguous identifier for when we want to specify the fallback function, removing the need for the *extra* flag as well as the need to validate that we aren't passing in both `fn_name` and `is_fallback_fn == True`.

Does this make sense?

web3/contract.py
```diff
@@ -986,11 +1016,14 @@ def call_contract_function(abi,
     Helper function for interacting with a contract function using the
     `eth_call` API.
     """
+    is_fallback_func = function_name is "fallback"
```
pipermerriam: FYI, `is` in python means  `is A the exact same object as B` which should not be confused with equality.

```python
In [45]: x = 'b' * 257

In [46]: y = 'b' * 257

In [47]: x is y
Out[47]: False

In [48]: x == y
Out[48]: True
```

Python makes certain optimizations which cause `is` comparisons to work in some cases since python memoizes some values (like integers 1-256)

```python
In [49]: x = 1

In [50]: y = 1

In [51]: x is y
Out[51]: True

In [52]: x == y
Out[52]: True

In [53]: x = 257

In [54]: y = 257

In [55]: x is y
Out[55]: False

In [56]: x == y
Out[56]: True
```

web3/utils/module_testing/fallback_contract.py
zheli: Sure, actually later I found out this is not needed but forgot to remove it...

web3/contract.py
zheli: Good find, thanks for pointing it out!

web3/utils/contracts.py
zheli: Yeah, I agree this is much cleaner.

zheli: Will make the change according to your suggestion.

web3/utils/contracts.py
zheli: @pipermerriam have some questions regarding to the implementation. Where should I put the check for the first four data bytes? 
I thought I have to add it after we get the result from `encode_abi(web3, fn_abi, fn_arguments, fn_selector`, but then I realize the result will always be `0x` since fallback function doesn't have arguments, and fn_selector is always empty (`b''`). Could you explain a bit more? Thanks!

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -346,3 +352,7 @@ def test_call_undeployed_contract(undeployed_math_contract, call):
     with pytest.raises(BadFunctionCallOutput) as exception_info:
         call(contract=undeployed_math_contract, contract_function='return13')
     assert expected_undeployed_call_error_message in str(exception_info.value)
+
+
+def test_call_fallback_function(fallback_function_contract):
+    fallback_function_contract.fallback.call()
```
pipermerriam: Is there an assertion we can use here on the return value or that `data` was set to `1`?

web3/contract.py
```diff
@@ -209,13 +215,16 @@ def factory(cls, web3, class_name=None, **kwargs):
...
+        setattr(contract, 'fallback', Contract.get_fallback_function(contract.abi, contract.web3))
 
         return contract
 
     #
     # Contract Methods
     #
     @classmethod
+    # TODO add deprecated back
+    # @deprecated_for("contract.constructor.transact")
```
pipermerriam: Just noting the left-in `TODO` comment.

web3/contract.py
```diff
@@ -721,14 +741,23 @@ def __call__(self, *args, **kwargs):
             return super().__call__(*args, **kwargs)
 
 
+class NonExistentFallbackFunction(object):
```
pipermerriam: I'm struggling to find an alternate way to accomplish what this class does.

At first I was thinking of a simple `@property` method which raised the needed exception buts since we need to be able to access the `fallback` property on both the class and instance, that doesn't work.

That lead me to thinking about implementing this as a descriptor instead.

```python
class FallbackFunctionDescriptor:
    fallback_fn = None

    def __init__(self, fallback_fn=None):
        self.fallback_fn = None
    def __get__(self, instance, owner):
        ...
```

While it's still kind of messy, I think it's a little cleaner than raising the exception in `__getattr__` and it causes the exception to be raised at the moment that `c.fallback` is accessed, rather than my understanding of the current behavior which waits until an attribute on `c.fallback` is accessed.

Thoughts?

web3/utils/abi.py
```diff
@@ -140,7 +158,7 @@ def is_encodable(_type, value):
         elif isinstance(value, bytes):
             return len(value) <= max_length
         else:
-            False
```
pipermerriam: 😨 (thanks for cleaning this up.

zheli: np 😛

web3/contract.py
zheli: Haven't thought much when I implemented `NonExistentFallbackFunction` :) so I don't have any preferences. It's up to you. 

I will implement it using `__get__` instead.

tests/core/contracts/test_contract_call_interface.py
zheli: Sorry for the delay. I can assert the return is an empty list()... Not sure what more I can do since fallback function cannot return any value and call() cannot change the data state...

pipermerriam:
@carver going to leave this to you to do final merge.  There are some things that *might should* be cleaned up like some leftover TODO comments, but all in all, this looks good-to-go.

pipermerriam:
@zheli well done and thanks for sticking with this one.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -16,6 +16,7 @@
 pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
 
 
+# TODO use transact to deploy
```
carver: Yeah, this TODO can be removed. The deprecation warnings that show up in the tests should be sufficient.

web3/contract.py
```diff
@@ -745,14 +773,24 @@ def __init__(self, *args, **kwargs):
             self.kwargs = {}
         else:
             self.kwargs = kwargs
-
         self.fn_name = type(self).__name__
         self._set_function_info()
+        # TODO remove this? Seems like it's not used.
```
carver: 👍 to remove `_transaction_data`

web3/utils/contracts.py
```diff
@@ -41,6 +46,13 @@
 )
 
 
+class FunctionNotSpecified(Exception):
```
carver: Can this be dropped entirely? I don't see the usage.

zheli: Good find!

tests/core/contracts/test_contract_call_interface.py
zheli: fixed

carver:
Strange that bf104b9 didn't get run in the CI. 😕

carver:
@zheli let me know if there's anything else you want to do before merge

zheli:
@carver I am thinking about adding few lines in the document about the fallback function, then you can click the merge button :)

zheli:
@carver added some simple description for the fallback api, could you have a look?

carver:
Cool, looks great!

PR #598, opened 2018-01-29
turn off sudo for all but geth

carver:
### What was wrong?

Sudo slows down build startup (adds ~1 minute per job, ~10 min total).

### How was it fixed?

Turn it off for all builds but geth. (Guessed at the right config syntax, because I can't figure out their docs)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 language: python
 dist: trusty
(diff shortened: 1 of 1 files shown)


carver:
Hm, docs suggested this should vastly improve startup time, but no apparent effect. Canceling...

PR #599, opened 2018-01-30
Fixed TypeError in estimate_gas_for_function invocation

tgudlek:
### What was wrong?

Method invocation of `estimate_gas_for_function` consisted of 3 positional arguments, followed by 2 keyword argument, followed by a sequence of unpacked positional arguments etc.

This caused a `TypeError: estimate_gas_for_function() got multiple values for argument 'function_name'` when `estimateGas` was called for a contract function that receives non-zero number of arguments since those arguments would get unpacked after some keyword arguments have already been set.

### How was it fixed?

Removed keyword from method invocation to conform how such methods are called in the rest of the file and fixed the accompanying `TypeError`.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -877,8 +877,8 @@ class ContractFunction(object):
         return estimate_gas_for_function(self.contract_abi,
                                          self.address,
                                          self.web3,
-                                         function_name=self.function_name,
-                                         transaction=estimate_transaction,
+                                         self.function_name,
+                                         estimate_transaction,
                                          *self.args,
                                          **self.kwargs)
 


carver:
Great, thanks for posting this! Seems like the kwargs here should also be dropped then: https://github.com/tgudlek/web3.py/blob/cfd4201924fb7f3b87a4581a6acf648a65f89d51/web3/contract.py#L1070-L1071

Would you also add a test that runs this against a contract function with arguments?

This is the general area, and that `increment` function in the contract also takes a `uint256`, so the new would be almost the same, plus an arg (and you can verify the red/green transition on the test with this fix).
https://github.com/ethereum/web3.py/blob/master/tests/core/contracts/test_contract_estimateGas.py#L32-L33

tgudlek:
Sure, I'll address both of your comments. Thanks!

tgudlek:
Hey, I see there are two different expected gas values, one of which seems like a fallback. I added tests that use `.add` instead of `.increment` and test invocation with `*args` and `**kwargs` respectively. How should I determine those values? My setup returns 30000.

pipermerriam:
> How should I determine those values? 

You can just measure locally and use those values.

tgudlek:
That's what I was thinking, but I'm not sure why's there a try-catch:

```
    try:
        assert abs(gas_estimate - 21472) < 200  # Geth
    except AssertionError:
        assert abs(gas_estimate - 32772) < 200  # eth-tester with py-evm
        pass
```

Why would there ever be multiple correct answers?

carver:
> Why would there ever be multiple correct answers?

Each node can implement estimates however they want. There's no guarantee about how close the estimate will be to the "optimal" amount of gas. (scare quotes, because I'm not certain there's a singular definition for optimal)

I just updated the json-rpc wiki to make that a bit clearer: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_estimategas

tgudlek:
How can I run tests in both settings to get the correct values? I'm currently cloning the git repo, fixing the code, and running with `tox`.

carver:
> How can I run tests in both settings to get the correct values?

Tox will run all of them, but to test them each individually, you can do this for py-evm via eth-tester:

```sh
TOX_POSARGS="-e py35-integration-ethtester" ETHEREUM_TESTER_CHAIN_BACKEND=eth_tester.backends.PyEVMBackend tox
```

and this for geth:

```sh
TOX_POSARGS="-e py36-integration-goethereum" GETH_VERSION=v1.7.2 tox
```

(these are slightly unusual ways of invoking, because I pulled straight from the .travis.yml file. It's just a reordering of a more "typical" way, like `tox -e py35-core`)

carver:
Fixed in #636

PR #604, opened 2018-01-30
Allow attribute access to rich log object

dylanjw:
### What was wrong?

It would be nice to allow attribute style accesses to rich log object keys, like shown in the comment here: https://github.com/ethereum/web3.py/issues/429#issuecomment-344679488

### How was it fixed?

Wrapping the rich log and nested args dict in AttributeDict()

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/events.py
+++ b/web3/utils/events.py
@@ -19,6 +19,9 @@ from eth_utils import (
 from web3.exceptions import (
     MismatchedABI,
 )
+from web3.utils.datastructures import (
+    AttributeDict,
+)
 from web3.utils.encoding import (
     hexstr_if_str,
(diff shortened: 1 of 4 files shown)


web3/utils/events.py
```diff
@@ -210,7 +213,7 @@ def get_event_data(event_abi, log_entry):
     ))
 
     event_data = {
-        'args': event_args,
+        'args': AttributeDict(event_args),
```
carver: Could do an `AttributeDict.recursive(event_args)` here (or even to all of the `event_data`) and skip the one in web3/contract.

Up to you

Issue #381, opened 2017-10-24
Parity Integration Testing


carver:
* Version: 4.*
* Python: 3.5

### What was wrong?

We don't have any integration tests for parity.

It's very difficult to test code like #366 

### How can it be fixed?

Generate some data in a parity fixture database (similar to how the geth one works) and build the integration conftest file.

pipermerriam:
We also *need* some way to install a specific version of parity which should likely be done with a small shell script to start off with until we can iterate towards a generic solution.

carver:
Great point. For determining which versions to test against, I could see which versions are used in the nodes on https://ethstats.net/

Any other ideas? How did you decide for geth?

pipermerriam:
> For determining which versions to test against

I would suggest only the latest point releases excluding any with known bugs, starting with whatever the current latest is.

> How did you decide for geth?

Whatever was *newest* at the time.

carver:
@dylanjw here's the parity integration test issue.

Here's an example integration "setup" file: https://github.com/ethereum/web3.py/blob/master/tests/integration/test_goethereum.py

It ultimately calls all these tests against the fixtures for geth: https://github.com/ethereum/web3.py/tree/master/web3/utils/module_testing

For example: https://github.com/ethereum/web3.py/blob/master/web3/utils/module_testing/eth_module.py

---

So one way to set up the parity integration tests would be to:

1. Generate a blockchain with some specific values, like a new funded account. The geth one is here: https://github.com/ethereum/web3.py/blob/master/tests/generate_go_ethereum_fixture.py
2. Generate fixtures that refer to those values. like a `funded_account_for_raw_txn` fixture that returns the hex address of the account that's funded.
3. Add a method to download/install a specific parity version (maybe a py-parity, equivalent to [py-geth](https://github.com/ethereum/py-geth))
4. Add a tox environment in `tox.ini` with the parity setup
5. Add a travis test run into `.travis.yml` for each parity version to test

I didn't set up the geth integration tests, so it's totally possible that I'm missing some things in this high-level approach, but the above steps should be a good place to start.

pipermerriam:
See here for documentation on the intergration tests: https://github.com/ethereum/web3.py/blob/master/tests/integration/README.md

dylanjw:
I have created fixtures for parity.  Because parity does not have mining functionality, I used the following method to create the generation script:

- Start a geth node on a private network, and run the fixture generation script (adapted from the existing script)
[...]

PR #606, opened 2018-02-01
Parity integration testing

dylanjw:
### What was wrong?

There are no parity integration tests: see https://github.com/ethereum/web3.py/issues/381

### How was it fixed?

Added:
a parity install script
parity fixtures generation script
parity pytests

TODO:
- [x] configure travis CI to run tests against variety of parity versions

#### Cute Animal Picture

Cute animal picture


tests/integration/generate_fixtures/common.py
```diff
@@ -0,0 +1,240 @@
...
+    "coinbase": COINBASE
+}
+
+
+def ensure_path_exists(dir_path):
+    """
+    Make sure that a path exists
+    """
+    if not os.path.exists(dir_path):
+        os.makedirs(dir_path)
```
pipermerriam: FYI, this entire utility is replacable by just using `os.makedirs(the_path, exist_ok=True)`

tests/integration/generate_fixtures/common.py
```diff
@@ -0,0 +1,240 @@
...
+
+def get_open_port():
+    sock = socket.socket()
+    sock.bind(('127.0.0.1', 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    return str(port)
+
+
+def get_geth_binary():
```
pipermerriam: does this belong in `go_ethereum.py`?

tests/integration/generate_fixtures/common.py
```diff
@@ -0,0 +1,240 @@
...
+            sock.connect(ipc_path)
+            sock.settimeout(timeout)
+        except (FileNotFoundError, socket.error):
+            time.sleep(0.01)
+        else:
+            break
+
+
+@contextlib.contextmanager
+def get_geth_process(geth_binary,
```
pipermerriam: Same her, `go_ethereum.py` instead of `common.py`?

dylanjw: I put those in common.py because I use them in the parity script to create the cluster.

pipermerriam: Took me a minute to figure out what was going on.  That makes sense now.  +1 for clever way to de-duplicate the logic for initializing the chain.

dylanjw:
The tests are running in travis now, but failing due to the missing ipc file, which probably means parity closed unexpectedly.  Unfortunately im not seeing any stdout from parity in the travis log.  Next, I will work on getting the parity console output to print to the travis log.

pipermerriam:
> The tests are running in travis now, but failing due to the missing ipc file

I admittedly didn't look at the test, but... you may have to put in some `time.sleep()` statements in to wait for the IPC file to exist as parity initializes.  If you're already doing this then disregard.

dylanjw:
I got this working on a local docker instance of the travis python container.  It seems that the contents of the parity base directory is not sufficient to transfer the private chain data between environments.  I added import/export commands to the fixture creation and test scripts, which solved the first issue.

Now, the build times out while downloading parity due the absence of output for 10 minutes.  I have added a progress bar to install_parity.py to prevent the travis timeout.

.travis.yml
```diff
@@ -58,6 +64,7 @@ before_install:
 install:
   - travis_retry pip install pip setuptools --upgrade
   - travis_retry pip install tox
+  - travis_retry pip install eth_utils tqdm requests
```
carver: Since these are only necessary during parity install, we could skip it in all the tests by inserting the `pip install` as a line with its own: `if [ -n "$PARITY_VERSION" ];` just before the `install_parity` command.

carver:
It will make me so happy to have these in web3!

pipermerriam:
Next up: Trinity ✨

.travis.yml
```diff
@@ -53,11 +60,13 @@ cache:
   - directories:
     - $HOME/.ethash/
     - $GETH_BASE_INSTALL_PATH/
+    - $PARITY_BASE_INSTALL_PATH/
 before_install:
   - if [ -n "$GETH_VERSION" ]; then mkdir -p $HOME/.ethash; fi
 install:
   - travis_retry pip install pip setuptools --upgrade
   - travis_retry pip install tox
+  - if [ -n "$PARITY_VERSION" ]; travis_retry pip install eth_utils tqdm requests
```
carver: looks to be missing ending: `; fi`

.circleci/config.yml
```diff
@@ -15,25 +15,33 @@ common: &common
...
-            curl -O https://storage.googleapis.com/golang/go1.7.4.linux-amd64.tar.gz
-            tar xvf go1.7.4.linux-amd64.tar.gz
-            sudo chown -R root:root ./go
-            sudo mv go /usr/local
-            sudo ln -s /usr/local/go/bin/go /usr/local/bin/go
-            sudo apt-get update;
-            sudo apt-get install -y build-essential;
-            python -m geth.install $GETH_VERSION;
-            sudo ln -s /home/circleci/.py-geth/geth-v1.7.2/bin/geth /usr/local/bin/geth
+          if [ -n "$GETH_VERSION" ]; then
```
carver: I seem to remember there being a problem with doing this, like it messed up the cache. Maybe it would work if you add the GETH_VERSION environment variable to the cache key. Note that because the cache was already built successfully, changes here are dangerous: we might break things and not realize it.

.circleci/config.yml
```diff
@@ -15,25 +15,33 @@ common: &common
...
+              python -m geth.install $GETH_VERSION;
+              sudo ln -s /home/circleci/.py-geth/geth-v1.7.2/bin/geth /usr/local/bin/geth
+            fi
+            $GETH_BINARY version
+            $GETH_BINARY makedag 0 $HOME/.ethash
+          fi
+    - run:
+        name: install parity if needed
+        command: |
+          if [ -n "$PARITY_VERSION" ]; then
```
carver: Same thing here, might want to add PARITY_VERSION to the cache key.

dylanjw:
Im going to move any circle CI config changes from this PR into a new one where they can be tested.  It seems like circle ci is only picking up new pull requests.

PR #615, opened 2018-02-05
fix bug in time-based gas price strategy

pipermerriam:
### What was wrong?

There was a bug in the time-based gas price strategy which was causing it to compute incorrect probabilities.

```
# incorrct from code:
probability = 1 - ((sample_size - num_blocks_accepting_price) / sample_size) ** wait_blocks

# should have been this:
probability = (1 - (sample_size - num_blocks_accepting_price) / sample_size) ** wait_blocks
```

The probability that miner which accepts a given gas price will mine the next block is

```
num_blocks_accepting_price / sample_size
```

So the inverse of this probability is

```
1 - (sample_size - num_blocks_accepting_price) / sample_size
```

And then to compute this across multiple blocks:

```
(1 - (sample_size - num_blocks_accepting_price) / sample_size) ** wait_blocks
```

### How was it fixed?

Updated the formulat to have the correct parenthesis positions.

> The values in the tests are largely *smoke* tests as they are simply computed from the implementation.  Eyeballing them, they look correct (though the previous ones did too 😢 ).

#### Cute Animal Picture

cute-kitten-babies-pets-and-animals-wallpaper

### DIFF
--- a/web3/gas_strategies/time_based.py
+++ b/web3/gas_strategies/time_based.py
@@ -73,7 +73,7 @@ def _compute_probabilities(miner_data, wait_blocks, sample_size):
         min_gas_price = miner_data_by_price[idx].min_gas_price
         num_blocks_accepting_price = sum(m.num_blocks for m in miner_data_by_price[idx:])
         inv_prob_per_block = (sample_size - num_blocks_accepting_price) / sample_size
-        probability_accepted = 1 - inv_prob_per_block ** wait_blocks
+        probability_accepted = (1 - inv_prob_per_block) ** wait_blocks
         yield Probability(min_gas_price, probability_accepted)
 
 
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -103,17 +103,17 @@ def _get_block_by_something(method, params):
     'strategy_params,expected',
     (
         # 120 second wait times
-        (dict(max_wait_seconds=80, sample_size=5, probability=98), 35),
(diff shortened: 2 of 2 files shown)


web3/gas_strategies/time_based.py
```diff
@@ -73,7 +73,7 @@ def _compute_probabilities(miner_data, wait_blocks, sample_size):
         min_gas_price = miner_data_by_price[idx].min_gas_price
         num_blocks_accepting_price = sum(m.num_blocks for m in miner_data_by_price[idx:])
         inv_prob_per_block = (sample_size - num_blocks_accepting_price) / sample_size
-        probability_accepted = 1 - inv_prob_per_block ** wait_blocks
+        probability_accepted = (1 - inv_prob_per_block) ** wait_blocks
         yield Probability(min_gas_price, probability_accepted)
```
carver: Ok, now I'm more confused.

If 1 block accepted the given price in a sample of 4, then the new code says this, which looks wrong:
```py
inv_prob_per_block = 3/4
probability_accepted = (1/4) ^ wait_blocks
```
The old code made more sense to me:
```py
inv_prob_per_block = 3/4
probability_accepted = 1 - (3/4 ^ wait_blocks)
```
Where the parenthetical probability is the chance that the gas price would never be accepted in any of the wait-blocks.

pipermerriam: I'm sleepy and wrong.  You're correct now that I'm looking at it.

> goes back to reviewing code.

raniama:
i cannot understand how could i choose the number of blocs as sample_size because the price could changes if i choose different values

Issue #614, opened 2018-02-05
merge_args_and_kwargs invalid handling of function name in the error message


miohtama:
* Version: 1.9.0
* Python: 3.5
* OS: osx/linux/win

### What was wrong?

`merge_and_kwargs` throws an exception when it tries to construct an error message if you give invalid kwargs for a function constructor.

```
  File "/Users/mikkoohtamaa/code/tm/venv/lib/python3.5/site-packages/web3/contract.py", line 307, in deploy
    deploy_transaction['data'] = cls._encode_constructor_data(args, kwargs)
  File "/Users/mikkoohtamaa/code/tm/venv/lib/python3.5/site-packages/eth_utils/string.py", line 85, in inner
    return force_obj_to_text(fn(*args, **kwargs))
  File "/Users/mikkoohtamaa/code/tm/venv/lib/python3.5/site-packages/web3/contract.py", line 745, in _encode_constructor_data
    arguments = merge_args_and_kwargs(constructor_abi, args, kwargs)
  File "/Users/mikkoohtamaa/code/tm/venv/lib/python3.5/site-packages/eth_utils/string.py", line 71, in inner
    return fn(*text_args, **text_kwargs)
  File "/Users/mikkoohtamaa/code/tm/venv/lib/python3.5/site-packages/web3/utils/abi.py", line 192, in merge_args_and_kwargs
    fn_name=function_abi['name'],
KeyError: 'name'
```

### How can it be fixed?

Assume there is no `function["name"]`, because constructor does not have it, when creating error messages https://github.com/ethereum/web3.py/blob/c32816d74a74be76cbb9987b594ee59826937673/web3/utils/abi.py#L212

gitcoinbot:
__This issue now has a funding of 0.05 ETH (42.52 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/614).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/614)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $7841.51 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 0.05 ETH (41.65 USD) funding by__: 
[...]

PR #618, opened 2018-02-06
fixed name key missing exception raised for error message constructor…

scottydelta:
… incase of invalid kwargs

### What was wrong?

#614 Error constructor was querying function name to display as part of error message but the 'name' key is missing

### How was it fixed?
Now it returns None incase the key is missing

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/abi.py
+++ b/web3/utils/abi.py
@@ -207,9 +207,17 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
 
     unknown_kwargs = {key for key in kwargs.keys() if key not in sorted_arg_names}
     if unknown_kwargs:
+        if function_abi.get('name'):
+            raise TypeError(
+                "{fn_name}() got unexpected keyword argument(s) '{dups}'".format(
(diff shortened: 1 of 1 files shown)


carver:
Thanks for highlighting this issue!

Almost always I'd like to see a test reproducing the problem. In this case of just fixing the message in an error condition, I can see the argument that it's a bit unnecessary... but the error message `None() got unexpected keyword argument(s) ...` would be pretty confusing IMO.

I would love to see the message improved a bit. Maybe it could fall back to showing the 'type' key instead?

If you would like to improve the message a bit, I'm happy to merge it in. If you don't feel like it, I'm happy to take care of it in a different PR.

carver:
Thanks! Only linting left:

https://travis-ci.org/ethereum/web3.py/jobs/338162501
> /home/travis/build/ethereum/web3.py/web3/utils/abi.py:217:9: E265 block comment should start with '# '

carver:
Closes #614

carver:
Thanks!

PR #623, opened 2018-02-07
Add parity jobs to circleci config

dylanjw:
Adds parity jobs to circleci config

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,57 @@ common: &common
         name: install dependencies
         command: pip install --user tox
     - run:
+        name: run tox
+        command: ~/.local/bin/tox
+    - save_cache:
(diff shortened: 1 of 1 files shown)


dylanjw:
@carver Have a look at my solution to the `common` anchor in the circle ci config.  I just duplicated the common anchor and added the extra steps for parity and geth integration tests.  It would be nice to not have to duplicate anything but what is allowed with yaml anchors is limited.  You can only merge key maps, and you can only have one extending alias/anchor per map, since `<<:` counts as a key.

carver:
Ok cool, yeah I don't see a better way than this.

We can re-run the travis pr jobs for parity after merging #624, then this should be good to go.

PR #624, opened 2018-02-08
Increase post chain import wait time before launching parity

dylanjw:
### What was wrong?

Accidental check in of large binary files with parity test fixtures, files have since been removed.  The parity process that imports the rlp file now takes a little bit longer to release the chain db.

### How was it fixed?

Added 1 second wait before launching the parity client process.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/integration/test_parity.py
+++ b/tests/integration/test_parity.py
@@ -149,20 +149,26 @@ def passwordfile():
 def parity_process(
         parity_import_blocks_process,
         parity_binary,
-        ipc_path, datadir,
+        ipc_path,
+        datadir,
         passwordfile,
         author):
-
-    run_parity_command = (
+    command_arguments = (
         parity_binary,
         '--chain', os.path.join(datadir, 'chain_config.json'),
         '--ipc-path', ipc_path,
-        '--base-path', str(datadir),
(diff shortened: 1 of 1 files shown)


dylanjw:
We can close this PR, but I wanted to verify that the integration tests succeed after removing the files.

tests/integration/test_parity.py
```diff
@@ -149,20 +149,26 @@ def passwordfile():
...
         '--chain', os.path.join(datadir, 'chain_config.json'),
         '--ipc-path', ipc_path,
-        '--base-path', str(datadir),
-        '--unlock', str(author),
-        '--password', str(passwordfile),
+        '--base-path', datadir,
+        '--unlock', author,
+        '--password', passwordfile,
     )
+    time.sleep(1)
```
carver: Isn't there a deterministic way to do this check? I'd hate to have tests randomly fail if it happens to take too long.

Maybe we can test for the presence of a file every n ms or something?

dylanjw:
@carver I updated it to poll the `parity import blocks` process for the termination signal.  CircleCI is running, but failing becasue I have it enabled on my fork; but I need the changes in #623 to pass.

PR #628, opened 2018-02-08
fixing no matching functions error

jstoxrocky:
### What was wrong?
https://github.com/ethereum/web3.py/blob/master/web3/utils/contracts.py#L63 will raise a value error indicating "No matching functions" when calling a function that does exist. This happens when the function exists but was passed an incorrect number of argument counts or incorrectly encoded arguments.

I'm not sure my PR is the best fix. I think the problem might lie elsewhere. For example, I feel like https://github.com/ethereum/web3.py/blob/master/web3/utils/contracts.py#L102 should be catching this kind of problem somehow...

Also, calling `myContract.functions.thisFunctionDoesNotExist` will throw an attribute error before ever reaching the block of code in question, so the "No matching functions" error seems to never be correct.  

### How was it fixed?
Not sure if this is the best fix. I just changed the error message.


web3/utils/contracts.py
```diff
@@ -90,7 +90,10 @@ def find_matching_fn_abi(abi, fn_name=None, args=None, kwargs=None):
     if len(function_candidates) == 1:
         return function_candidates[0]
     if not function_candidates:
-        raise ValueError("No matching functions found")
+        raise ValueError(
+            "Incorrect number of arguments supplied or "
+            "one or more arguments could not be encoded to the necessary ABI type"
```
pipermerriam: Still leaves out the possibility that there are no functions by the given name.

- Simple solution: Add more *or* statements to this to add the extra wording.
- Change this logic to narrow down a matrix of what did/did-not match and tailor the error message to exactly what parts had no match.  Logic should not be to difficult to write, but it may take some churn to make it clean and easy to follow.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -346,3 +346,28 @@ def test_call_undeployed_contract(undeployed_math_contract, call):
...
+    with pytest.raises(AttributeError):
+        arrays_contract.functions.iDoNotExist().call()
+
+
+def test_argument_count(arrays_contract, call):
+    with pytest.raises(ValueError) as e:
+        arrays_contract.functions.setBytes32Value().call()
+    message, = e.value.args
+    assert message == (
+        "Incorrect number of arguments supplied or "
```
pipermerriam: I think I would prefer a custom exception type over asserting a specific message.  Since we don't have a `ValidationError` defined in web3.py I think this would be an appropriate time to add one.

jstoxrocky:
@pipermerriam 
Updates: 
I reworked the logic so that error messages displays exactly where the passed arguments failed/succeeded to match functions in the ABI. I also switched the returned error from a ValueError to a ValidationError. Let me know if you think this logic is the right way forward.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -423,3 +424,23 @@ def test_returns_data_from_specified_block(web3, math_contract):
 
     assert output1 == 1
     assert output2 == 2
+
+
+def test_function_does_not_exist(arrays_contract):
+    with pytest.raises(AttributeError):
```
pipermerriam: [`pytest.raises`](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) takes a `match` keyword which allows you to match against the exception message itself.  I suspect that we can do some very basic matching on the function names to be sure we're getting the *right* exception.

pipermerriam:
@carver 👍 from me.

web3/utils/contracts.py
```diff
@@ -70,36 +77,54 @@ def find_matching_event_abi(abi, event_name=None, argument_names=None):
...
+        name_filter,
+        arg_count_filter,
+        encoding_filter,
+    ])
+    function_candidates = pipe(abi, *filters)
     if len(function_candidates) == 1:
         return function_candidates[0]
-    if not function_candidates:
-        raise ValueError("No matching functions found")
+    # No exact match
```
carver: This comment is in an awkward place. I don't really think it's necessary at all, it should be apparent from the if test. But we'll just go ahead and merge it and clean it up in some future PR.

web3/utils/contracts.py
```diff
@@ -70,36 +77,54 @@ def find_matching_event_abi(abi, event_name=None, argument_names=None):
...
     else:
-        raise ValueError("Multiple functions found")
+        matching_identifiers = name_filter(abi)
+        matching_function_signatures = [abi_to_signature(func) for func in matching_identifiers]
+        arg_count_matches = len(arg_count_filter(matching_identifiers))
+        encoding_matches = len(encoding_filter(matching_identifiers))
+        if arg_count_matches == 0:
+            diagnosis = "\nFunction invocation failed due to improper number of arguments."
+        elif encoding_matches == 0 or encoding_matches > 1:
+            diagnosis = "\nFunction invocation failed due to improper argument encoding."
```
carver: So lets say there are two functions: `func1(string)` and `func1(bytes)`. All calls to the function with a string will fail with the message "Function invocation failed due to improper argument encoding", right? We probably want different messages for `encoding_matches == 0` and `encoding_matches > 1`.

PR #630, opened 2018-02-12
Replaced All super(ClassName, self) with super() and  Stopped inheriting from object to define new style classes

voith:
### What was wrong?
Some code changes in python3 that could have been part of #339:
1. `super(ClassName, self)` can be replaced by `super()`.
2.  stopped inheriting from `object`. `ClassName(object):` can be replaced by `ClassName:`.

### How was it fixed?
This PR has been split into two commits:

commit 1:  replaced `super(ClassName, self)` with `super()`. I've skipped super calls that call grandparent instead of the parent. Found only one such occurrence [here](https://github.com/ethereum/web3.py/blob/master/web3/providers/tester.py#L122).

commit 2: stopped inheriting from object to define new style classes.

#### Cute Animal Picture

224142282-image-cute

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,7 +10,7 @@ from web3.utils.threads import (
 from web3.main import Web3
 
 
-class PollDelayCounter(object):
+class PollDelayCounter:
     def __init__(self, initial_delay=0, max_delay=1, initial_step=0.01):
         self.initial_delay = initial_delay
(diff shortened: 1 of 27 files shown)


voith:
cc @carver

carver:
Thanks @voith -- welcome to the repo!

Issue #522, opened 2017-12-22
AttributeError calling watch() on a BlockFilter instance


miller46:
* Version: 4.0.0b4
* Python: 3.6.3
* OS: osx

I am attempting to listen for new blocks as they are added to the network.  I attempt to set up a blockfilter like so:

```
block_filter = self.web3.eth.filter('latest')
block_filter.watch(new_block_callback)
```

and I receive the following error:

`
AttributeError: 'BlockFilter' object has no attribute 'watch'
`

Am I doing something wrong?  This worked for me previously with an earlier version of web3.py, and this is nearly identical to the example in the docs.

carver:
Ah thanks for pointing that out. The examples that use `watch()` should be changed in the latest documentation.

Note the deprecation warning above `watch()` in the stable docs, and that `watch()` is no longer listed in the latest docs:
-http://web3py.readthedocs.io/en/stable/filters.html#web3.utils.filters.Filter.watch
-http://web3py.readthedocs.io/en/latest/filters.html

Instead, you should periodically call `get_new_entries()` to see the latest filter results. It works just like the log filter: http://web3py.readthedocs.io/en/latest/filters.html#web3.utils.filters.LogFilter.get_new_entries

---

I'll leave the issue open until the docs are updated.

Issue #188, opened 2017-06-05
Specifying `topics` to `web3.eth.filter` gives empty list for past events


veox:
* `web3.py` version: 3.8.1 (web3[gevent])
* node: `geth 1.6.5-stable`, local (via IPC)
* Python: 3.5
* OS: Ubuntu 16.04.2 LTS

### What seems wrong?

Retrieving past log events while specifying `'topics'` gives an empty list, if value of `'topics'` is a list with more than one element.

I was under the impression that the filter will match any event that's in `'topics'`. This seems not to be the case.

### Details

Using these variables:

``` python
> # common vars
> import cancelot
> cancelot.enslaunchblock
3648565
> cancelot.registrar
'0x6090a6e47849629b7245dfa1ca21d94cd15878ef'
> list(cancelot.handlers.keys())
['0x7b6c4b278d165a6b33958f8ea5dfb00c8c9d4d0acf1985bef5d10786898bc3e7',
 '0xb556ff269c1b6714f432c36431e2041d28436a73b6c3f19c021827bbdc6bfc29']
```

Not specifying `'topics'` does give a list of events (`dict`s), as expected:

``` python
> # filter for past events, without topics
> filtall =  web3.eth.filter({
    'fromBlock': cancelot.enslaunchblock,
    'toBlock': cancelot.enslaunchblock + 1000,
    'address': cancelot.registrar
    })
> evs = filtall.get(only_changes = False)
> len(evs)
80
> evs[0]
{'address': '0x6090a6e47849629b7245dfa1ca21d94cd15878ef',
 'blockHash': '0x3a2c9e10de45a86d39f52eeaa2beef514d8b0a19ddf6bd231a6d97bca79988b8',
 'blockNumber': 3648565,
 'data': '0x000000000000000000000000000000000000000000000000002386f26fc10000',
 'logIndex': 4,
 'removed': False,
 'topics': ['0xb556ff269c1b6714f432c36431e2041d28436a73b6c3f19c021827bbdc6bfc29',
  '0x1f0e90602b7ed7fe12cd293824cccad3966e676d067df11da128793f5ad97ac3',
  '0x0000000000000000000000000904dac3347ea47d208f3fd67402d039a3b99859'],
 'transactionHash': '0x6e032eed213f1a641af8cdb7fd3770984010a38c04f6c06e393b03dfca30a80d',
 'transactionIndex': 40}
```

[...]

PR #631, opened 2018-02-12
Documentation updates

dylanjw:
This is the collection of updates that Ive made as I read through the web3 docs.  It is a work in progress.

In all the examples,  Im trying out different solutions to questions I have, but may not be good examples to keep, but Im putting them up for comment.  Other examples that seem a bit convoluted might be good candidates for improvements we could make.  I haven't run anything through spell check which I will do before merging.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -106,10 +106,10 @@ Each Contract Factory exposes the following methods.
 
 .. py:method:: Contract.functions.myMethod(*args, **kwargs).transact(transaction)
 
-    Execute the specified function by sending a new public transaction.  
+    Execute the specified function by sending a new public transaction.
 
     This is executed in two steps.
-    
+
     The first portion of this function call ``transact(transaction)`` takes a
     single parameter which should be a python dictionary conforming to
     the same format as the ``web3.eth.sendTransaction(transaction)`` method.
@@ -175,14 +175,14 @@ Each Contract Factory exposes the following methods.
 
 .. py:method:: Contract.functions.myMethod(*args, **kwargs).buildTransaction(transaction)
 
-    Builds a transaction dictionary based on the contract function call specified. 
+    Builds a transaction dictionary based on the contract function call specified.
 
     This method behaves the same as the :py:meth:`Contract.transact` method,
     with transaction details being passed into the first portion of the
     function call, and function arguments being passed into the second portion.
 
     .. note::
-        `nonce` is not returned as part of the transaction dictionary unless it is 
+        `nonce` is not returned as part of the transaction dictionary unless it is
         specified in the first portion of the function call:
 
         .. code-block:: python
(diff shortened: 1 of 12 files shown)


docs/filters.rst
```diff
@@ -121,3 +130,104 @@ logs.  It exposes the following additional methods.
...
+                task = loop.create_task(log_loop(block_filter, 5))
+                try:
+                    loop.run_until_complete(task)
+                finally:
+                    loop.close()
+
+
+            def main():
+                loop = asyncio.new_event_loop()
+                worker = Thread(target=start_log_loop, args=(loop,))
```
carver: We don't have to use asyncio in this other approach, I think it would be clearer if we didn't. For example:

`Thread(target=log_loop, args=(block_filter, 5), daemon=True)`

(where `log_loop` is no longer async, and just uses `time.sleep`)

dylanjw:
I made updates here that fix https://github.com/ethereum/web3.py/issues/522

dylanjw:
TODO: https://github.com/ethereum/web3.py/issues/188

docs/contracts.rst
```diff
@@ -104,22 +114,115 @@ Each Contract Factory exposes the following methods.
...
+
+    The first portion of the function call ``myMethod(*args, **kwargs)``
     selects the appropriate contract function based on the name and provided
     argument.  Arguments can be provided as positional arguments, keyword
     arguments, or a mix of the two.
 
+    The end portion of this function call ``transact(transaction)`` takes a
+    single parameter which should be a python dictionary conforming to
+    the same format as the ``web3.eth.sendTransaction(transaction)`` method.
+    This dictionary may not contain the keys ``data`` or ``to``.
```
pipermerriam: I *think* `to` is allowed *if* the method is being called from the contract class (as opposed to a contract instance).

docs/contracts.rst
```diff
@@ -209,89 +330,33 @@ Each Contract Factory exposes the following methods.
...
 
     .. code-block:: python
 
-        >>> transfer_filter = my_token_contract.eventFilter('Transfer', {'filter': {'_from': '0xdc3a9db694bcdd55ebae4a89b22ac6d12b3f0c24'}})
-        >>> transfer_filter.get_new_entries()
-        [...]  # array of Event Log Objects that match the filter.
-
-        # wait a while...
+        myContract = web3.eth.contract(address=contract_address, abi=contract_abi)
+        myContract.events.myEvent().processReceipt()
```
pipermerriam: This code example seems a bit strange since I thought `processReceipt` expects an instance of a receipt.

docs/filters.rst
```diff
@@ -5,12 +5,33 @@ Filtering
...
+    .. code-block:: python
+
+        event_filter = web3.eth.filter({"address": contract_address})
+
+* Attaching to an existing filter
+
+    .. code-block:: python
+
+        from web3.auto import w3
+        existing_filter = web3.utils.filters.Filter(w3, "0x0")
```
pipermerriam: If we're going to document manual instantiation of `Filter` objects, then the class needs to be moved from within the `web3.utils` directory since `web3.utils` is explicitly non-public API.

dylanjw: What about adding an option to select a filter by `filter_id` to `Eth.filter`?

pipermerriam: I like it.

dylanjw: https://github.com/ethereum/web3.py/pull/641

docs/filters.rst
```diff
@@ -5,12 +5,33 @@ Filtering
...
 
-* Pending Transactions
-* New Blocks
 * Event Logs
 
+    Through the contract instance api:
+
+    .. code-block:: python
+
+        event_filter = myContract.eventFilter('eventName', {'filter': {'arg1':10}})
```
pipermerriam: https://github.com/ethereum/web3.py/issues/642 indicates we could use an example which shows how to do filters that have string values as arguments, including a note about what types they can use for those arguments (do we require bytes?  can they use hex encoded values?)

carver:
Fixes #643

pipermerriam:
Just saw the CI failures.  Can we merge this when it turns green?!

dylanjw:
All green

PR #636, opened 2018-02-13
fix function.myFunction().estimateGas()

dylanjw:
### What was wrong?

```
>>> contracts.StoreVar.functions.setVar(255).estimateGas()
  File "/home/dwilson/Develop/Eth/contract_testing/venv36/lib/python3.6/site-packages/web3/contract.py", line 883, in estimateGas
    **self.kwargs)
TypeError: estimate_gas_for_function() got multiple values for argument 'function_name'
```

### How was it fixed?

Call `estimate_gas_for_function` with positional arguments instead of kwargs.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_estimateGas.py
+++ b/tests/core/contracts/test_contract_estimateGas.py
@@ -36,3 +36,14 @@ def test_contract_estimateGas(web3, math_contract, estimateGas):
     except AssertionError:
         assert abs(gas_estimate - 32772) < 200  # eth-tester with py-evm
         pass
+
+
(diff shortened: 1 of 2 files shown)


pipermerriam:
Any chance we can get a test for this?

dylanjw:
It looks like it is only cases where arguments are passed to the contract function, and we are missing a test for that that case:

```
(Pdb) math_contract.functions.increment().estimateGas()
32772
(Pdb) math_contract.functions.add(5,6).estimateGas()
*** TypeError: estimate_gas_for_function() got multiple values for argument 'function_name'
(Pdb) 
```

PR #641, opened 2018-02-15
Allow instatiating existing filters by id

dylanjw:
### What was wrong?

There was not public method to attach a filter instance to an existing client-side filter.

### How was it fixed?

Now `web3.eth.filter()` accepts either a `filter_params` or `filter_id` argument.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -270,7 +270,11 @@ class Eth(Module):
             [transaction],
         )
 
-    def filter(self, filter_params):
+    def filter(self, filter_params=None, filter_id=None):
+        if filter_id and filter_params:
+            raise ValueError("Ambiguous invocation: provide either a \
+                    `filter_params` or a `filter_id` argument. \
+                    Both were supplied.")
         if is_string(filter_params):
             if filter_params == "latest":
                 filter_id = self.web3.manager.request_blocking(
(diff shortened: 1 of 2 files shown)


web3/eth.py
```diff
@@ -254,7 +254,11 @@ def estimateGas(self, transaction):
             [transaction],
         )
 
-    def filter(self, filter_params):
+    def filter(self, filter_params=None, filter_id=None):
+        if filter_id and filter_params:
+            raise ValueError("Ambiguous invocation: provide either a \
```
carver: IIRC, we're calling this a `TypeError` elsewhere: when the wrong combination of arguments is supplied.

web3/eth.py
```diff
@@ -272,13 +277,17 @@ def filter(self, filter_params):
...
+            _filter_id = self.web3.manager.request_blocking(
                 "eth_newFilter",
                 [filter_params],
             )
+            return LogFilter(self.web3, _filter_id)
+        elif filter_id and not filter_params:
             return LogFilter(self.web3, filter_id)
         else:
-            raise ValueError("Must provide either a string or a valid filter object")
+            raise ValueError("Must provide either filter_params as a \
```
carver: This also looks like it should be a `TypeError`.

carver:
@dylanjw I think the only changes needed to get this merged are:

- That last `ValueError` -> `TypeError`
- Update the `pytest.raises` tests.

PR #649, opened 2018-02-16
[WIP] Middleware to sign transactions locally (continuation of #517)

dylanjw:
### What was wrong?

See https://github.com/ethereum/web3.py/pull/517

Finishing up work done in #517, by extending the tests and adding changes to conform to https://github.com/pipermerriam/ethereum-dev-tactical-manual.

This is work in progress.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_constructor.py
+++ b/tests/core/contracts/test_contract_constructor.py
@@ -20,7 +20,7 @@ def test_contract_constructor_gas_estimate_no_constructor(MathContract):
     try:
         assert abs(gas_estimate - 167412) < 200  # Geth
     except AssertionError:
-        assert abs(gas_estimate - 200321) < 200  # eth-tester with py-evm
+        assert abs(gas_estimate - 177396) < 200  # eth-tester with py-evm
 
 
 def test_contract_constructor_gas_estimate_with_constructor_without_arguments(
@@ -29,7 +29,7 @@ def test_contract_constructor_gas_estimate_with_constructor_without_arguments(
     try:
         assert abs(gas_estimate - 43082) < 200  # Geth
     except AssertionError:
-        assert abs(gas_estimate - 85694) < 200  # eth-tester with py-evm
+        assert abs(gas_estimate - 61594) < 200  # eth-tester with py-evm
 
 
 @pytest.mark.parametrize(
@@ -50,7 +50,7 @@ def test_contract_constructor_gas_estimate_with_constructor_with_arguments(
     try:
         assert abs(gas_estimate - 78572) < 200  # Geth
     except AssertionError:
-        assert abs(gas_estimate - 112574) < 200  # eth-tester with py-evm
+        assert abs(gas_estimate - 89122) < 200  # eth-tester with py-evm
 
 
 def test_contract_constructor_gas_estimate_with_constructor_with_address_argument(
@@ -60,7 +60,7 @@ def test_contract_constructor_gas_estimate_with_constructor_with_address_argumen
     try:
         assert abs(gas_estimate - 50181) < 200  # Geth
     except AssertionError:
-        assert abs(gas_estimate - 95399) < 200  # eth-tester with py-evm
+        assert abs(gas_estimate - 66546) < 200  # eth-tester with py-evm
 
 
 def test_contract_constructor_transact_no_constructor(web3, MathContract, MATH_RUNTIME):
--- a/tests/core/contracts/test_contract_constructor.py
+++ b/tests/core/contracts/test_contract_constructor.py
@@ -15,21 +15,26 @@ def test_contract_constructor_abi_encoding_with_no_constructor_fn(MathContract, 
     assert deploy_data == MATH_CODE
 
 
-def test_contract_constructor_gas_estimate_no_constructor(MathContract):
+def test_contract_constructor_gas_estimate_no_constructor(web3, MathContract):
     gas_estimate = MathContract.constructor().estimateGas()
-    try:
-        assert abs(gas_estimate - 167412) < 200  # Geth
-    except AssertionError:
-        assert abs(gas_estimate - 177396) < 200  # eth-tester with py-evm
+
+    deploy_txn = MathContract.constructor().transact()
+    txn_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
+    gas_used = txn_receipt.get('gasUsed')
+
+    assert abs(gas_estimate - gas_used) < 21000
 
 
 def test_contract_constructor_gas_estimate_with_constructor_without_arguments(
+        web3,
         SimpleConstructorContract):
     gas_estimate = SimpleConstructorContract.constructor().estimateGas()
-    try:
-        assert abs(gas_estimate - 43082) < 200  # Geth
-    except AssertionError:
-        assert abs(gas_estimate - 61594) < 200  # eth-tester with py-evm
+
+    deploy_txn = SimpleConstructorContract.constructor().transact()
+    txn_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
+    gas_used = txn_receipt.get('gasUsed')
+
+    assert abs(gas_estimate - gas_used) < 21000
 
 
 @pytest.mark.parametrize(
@@ -42,25 +47,33 @@ def test_contract_constructor_gas_estimate_with_constructor_without_arguments(
     ),
 )
 def test_contract_constructor_gas_estimate_with_constructor_with_arguments(
+        web3,
         WithConstructorArgumentsContract,
         constructor_args,
         constructor_kwargs):
     gas_estimate = WithConstructorArgumentsContract.constructor(
         *constructor_args, **constructor_kwargs).estimateGas()
-    try:
-        assert abs(gas_estimate - 78572) < 200  # Geth
-    except AssertionError:
-        assert abs(gas_estimate - 89122) < 200  # eth-tester with py-evm
+
+    deploy_txn = WithConstructorArgumentsContract.constructor(
+        *constructor_args, **constructor_kwargs).transact()
+    txn_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
+    gas_used = txn_receipt.get('gasUsed')
+
+    assert abs(gas_estimate - gas_used) < 21000
 
 
 def test_contract_constructor_gas_estimate_with_constructor_with_address_argument(
+        web3,
         WithConstructorAddressArgumentsContract):
     gas_estimate = WithConstructorAddressArgumentsContract.constructor(
         "0x16D9983245De15E7A9A73bC586E01FF6E08dE737").estimateGas()
-    try:
-        assert abs(gas_estimate - 50181) < 200  # Geth
-    except AssertionError:
-        assert abs(gas_estimate - 66546) < 200  # eth-tester with py-evm
+
+    deploy_txn = WithConstructorAddressArgumentsContract.constructor(
+        "0x16D9983245De15E7A9A73bC586E01FF6E08dE737").transact()
+    txn_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
+    gas_used = txn_receipt.get('gasUsed')
+
+    assert abs(gas_estimate - gas_used) < 21000
 
 
 def test_contract_constructor_transact_no_constructor(web3, MathContract, MATH_RUNTIME):
--- a/tests/core/middleware/test_transaction_signing.py
+++ b/tests/core/middleware/test_transaction_signing.py
@@ -96,53 +96,38 @@ def w3_dummy(w3_base, result_generator_middleware):
 
 
 @pytest.mark.parametrize(
-    'method,strict_account_matching,key_object,from_,expected',
+    'method,key_object,from_,expected',
     (
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE, ADDRESS_2, ValueError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE, ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, MIXED_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, MIXED_KEY_MIXED_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, MIXED_KEY_MIXED_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, MIXED_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, SAME_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, SAME_KEY_SAME_TYPE, ADDRESS_2, ValueError),
-        ('eth_sendTransaction', False, SAME_KEY_SAME_TYPE, ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, MIXED_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', True, MIXED_KEY_SAME_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, MIXED_KEY_SAME_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, MIXED_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[0], ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[1], ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[2], ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[3], ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[4], ADDRESS_1, 'eth_sendRawTransaction'),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[0], ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[1], ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[2], ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[3], ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE[4], ADDRESS_2, NotImplementedError),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE[0], ADDRESS_2, ValueError),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE[1], ADDRESS_2, ValueError),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE[2], ADDRESS_2, ValueError),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE[3], ADDRESS_2, ValueError),
-        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE[4], ADDRESS_2, ValueError),
-        ('eth_call', True, MIXED_KEY_MIXED_TYPE, ADDRESS_1, NotImplementedError),
-        ('eth_call', False, MIXED_KEY_MIXED_TYPE, ADDRESS_1, NotImplementedError),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE, ADDRESS_2, NotImplementedError),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', MIXED_KEY_MIXED_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', MIXED_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_SAME_TYPE, ADDRESS_2, NotImplementedError),
+        ('eth_sendTransaction', SAME_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', MIXED_KEY_SAME_TYPE, ADDRESS_2, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', MIXED_KEY_SAME_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[0], ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[1], ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[2], ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[3], ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[4], ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[0], ADDRESS_2, NotImplementedError),
+        ('eth_sendTransaction', SAME_KEY_MIXED_TYPE[1], ADDRESS_2, NotImplementedError),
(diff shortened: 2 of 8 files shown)


dylanjw:
Im hoping to implement a test comparing the signed transaction created through the middleware to a transaction signed with eth.account.signTransaction().  Im assuming that the two should be the same, but I could be misunderstanding something.

carver:
@dylanjw with the way the old PR turned out, it might be easiest to start from scratch

web3/middleware/signing.py
```diff
@@ -0,0 +1,44 @@
...
+
+from cytoolz import (
+    merge,
+)
+
+
+def key_to_account(web3, key):
+    if isinstance(key, eth_account.local.LocalAccount):
+        return key
+    if isinstance(key, (HexBytes, str, bytes)):
```
carver: FWIW, `HexBytes` is already a subclass of `bytes`, so you could drop it.

web3/middleware/signing.py
```diff
@@ -0,0 +1,44 @@
...
+            if method == "eth_sendTransaction":
+
+                transaction_filler = {
+                    'from': account.address,
+                    'nonce': web3.eth.getTransactionCount(account.address)
+                }
+
+                transaction = merge(transaction_filler, params[0])
+                filled_transaction = fill_transaction_defaults(web3, transaction)
+                signed_transaction = account.signTransaction(filled_transaction)
```
carver: If `from` is set in the sending transaction to anything other than the public address of `account`, this is going to crash hard. The safest thing to do is to only sign if the transaction has a `from` value set to the `account` address. Then we just skip empty and mismatched `from` values.

dylanjw:
@pipermerriam or @carver I re-started this from scratch.  Im getting an error that I could use help with.  I have a failing test sending a transaction through the middleware using the provider from ethereum_tester.  py-evm is throwing an error with a recovery id of `37`.

```
venv36/lib/python3.6/site-packages/eth_tester/backends/pyevm/main.py:210: in _insert_transaction_to_pending_block
    _, block = chain.get_vm().apply_transaction(transaction)
venv36/lib/python3.6/site-packages/evm/vm/base.py:80: in apply_transaction
    self.block,
venv36/lib/python3.6/site-packages/evm/vm_state.py:215: in apply_transaction
    computation = self.execute_transaction(transaction)
venv36/lib/python3.6/site-packages/evm/vm/forks/spurious_dragon/vm_state.py:22: in execute_transaction
    computation = _execute_frontier_transaction(self, transaction)
venv36/lib/python3.6/site-packages/evm/vm/forks/frontier/vm_state.py:54: in _execute_frontier_transaction
    transaction.validate()
venv36/lib/python3.6/site-packages/evm/vm/forks/homestead/transactions.py:26: in validate
    super(HomesteadTransaction, self).validate()
venv36/lib/python3.6/site-packages/evm/vm/forks/frontier/transactions.py:52: in validate
    validate_lte(self.v, maximum=28, title="Transaction.v")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

value = 37, maximum = 28, title = 'Transaction.v'

    def validate_lte(value, maximum, title="Value"):
        if value > maximum:
            raise ValidationError(
                "{title} {0} is not less than or equal to {1}".format(
                    value,
                    maximum,
>                   title=title,
                )
            )
E           evm.exceptions.ValidationError: Transaction.v 37 is not less than or equal to 28

maximum    = 28
title      = 'Transaction.v'
value      = 37
```

web3/middleware/signing.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        # TODO Find a better way to check for and handle eth_keys objects
+        return web3.eth.account.privateKeyToAccount(key.to_hex())
+    except AttributeError:
+        raise ValueError("key must be one of the types (...). Was of type {0}".format(type(key)))
+
+
+def construct_in_flight_transaction_signing_middleware(private_key):
+
+    def in_flight_transaction_signing_middleware(make_request, web3):
```
carver: `local_singing_middleware`? (just looking for something a bit shorter)

carver:
> validate_lte(self.v, maximum=28, title="Transaction.v")

I think this is maybe a py-evm bug, where it is being too strict. [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#specification) is where `v` greater than 28 are defined.

> Should I make this an integration test, and not use eth_tester?

This bug would still show up in the py-evm integrations, and I don't think there's much here that (should be) node-dependent, so I think an `eth-tester` test should be fine.

... but it looks like we need more `w3.eth.account` tests to make sure that things like EIP-155 transactions are accepted by eth-tester. (Probably doesn't belong in this PR, though)

web3/middleware/signing.py
pipermerriam: Also, make sure you use `eth_utils.is_same_address` if there is any way that the two addresses are in different formats (checksummed/not-checksummed)

web3/middleware/signing.py
pipermerriam: - `intercept_and_sign_middleware`
- `sign_and_send_raw_middleware`
- `fancy_pants_local_signing_middleware`

web3/middleware/signing.py
carver: Right, because we reject non-checksummed addresses, but that validation may be somewhere deeper in the middleware "onion" :) -- depending on where this gets injected.

dylanjw:
Another issue I found with eth_tester, is the way gas is estimated by sending and revoking the transaction.  This fails when sending a transaction from an external account.

Should I take out the eth_tester tests in this PR, or mark them as XFAIL?

pipermerriam:
I'd suggest `xfail` over removal.

carver:
@dylanjw From conversation this morning: py-evm now has an estimate gas feature at `chain.estimate_gas(transaction)`. The API hasn't had a ton of usage yet, so please ping me if anything seems wonky.

carver:
For tracking, the eth-tester gas estimation fix depends on: https://github.com/ethereum/py-evm/pull/432 (no eth-tester PR yet)

dylanjw:
Here are a few more PR dependencies:

To fix EIP155 tx validation issue: https://github.com/ethereum/py-evm/pull/443
To fix EIP155 tx validation issue: https://github.com/ethereum/eth-tester/pull/72
To fix part of gas estimation issue: https://github.com/ethereum/eth-tester/pull/71

dylanjw:
To complete this PR another issue involving the eth-tester gas estimation must be resolved.  This is the issue im having.

My understanding of py-evm's chain.Chain.estimate_gas() is that it requires a signed py-evm transaction object as the transaction parameter, as `estimate_gas` depends on attributes only in the signed transaction type, and not unsigned:

```
>>> et.backend.chain.estimate_gas(unsigned)
  File "/home/dwilson/Develop/Projects/ethereum/py-evm/evm/chains/base.py", line 478, in estimate_gas
    return self.gas_estimator(state, transaction)
  File "cytoolz/functoolz.pyx", line 232, in cytoolz.functoolz.curry.__call__
  File "/home/dwilson/Develop/Projects/ethereum/py-evm/evm/estimators/gas.py", line 54, in binary_gas_search
    minimum_transaction = SpoofTransaction(transaction, gas=transaction.intrinsic_gas)
AttributeError: 'SpuriousDragonUnsignedTransaction' object has no attribute 'intrinsic_gas'
>>> 
```

But to create a py-evm transaction object, I must provide a signing key.  eth-tester tries to look up the key in the eth-tester managed accounts (obviously can't do that with a local account):

eth_tester/backends/pyevm/main.py:

```python
395     def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='latest'):
\396         signing_key = self._key_lookup[transaction['from']]
397         normalized_transaction = self._normalize_transaction(transaction, block_number)
398         evm_transaction = self.chain.create_unsigned_transaction(**normalized_transaction)
399         signed_evm_transaction = evm_transaction.as_signed_transaction(signing_key)
400         return signed_evm_transaction
```

Im not sure on the way forward with this.

dylanjw:
I just tried it out in the eth-tester PR: https://github.com/ethereum/eth-tester/pull/71/commits/0112470b58488a2439815631fac5dc931fbccbac, and it does seem workable.

dylanjw:
Actually, maybe SpoofTransaction isnt a viable option.  This option still requires a valid address with enough ether to cover gas costs in order for the transaction validation to pass.  See: https://github.com/ethereum/py-evm/pull/454

pipermerriam:
Mind going and doing some digging to see how `go-ethereum` does gas estimations?  I feel like we're re-inventing something that's already solved elsewhere.

dylanjw:
Sure.  Sounds like a good exercise for me.

carver:
@dylanjw this looks ready for a rebase, now that #737 is merged (as you already mentioned)

dylanjw:
This is now pending: https://github.com/ethereum/eth-tester/pull/71 and another eth-tester version bump.

pipermerriam:
@dylanjw where is this at?  Anything blocking it?

dylanjw:
Im testing with eth-tester:master.  The newest version of py-evm requires the gas to be set, whereas older versions did not (maybe?).  Should I adjust tests, or is the gas supposed to get filled in if left blank?  I thought that was the purpose of estimate_gas.

pipermerriam:
> he newest version of py-evm requires the gas to be set when creating an unsigned transaction. Should I fill in a value in eth_tester, or should py-evm allow unset gas? I think the other forks allowed unset gas.

I think it's ok to auto-fill in the `gas` value on the eth-tester side.  Seems like using the max gas available in the block would be an appropriate default.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+
+from web3.utils.formatters import (
+    apply_formatter_if,
+)
+from web3.utils.transactions import (
+    fill_nonce,
+    fill_transaction_defaults,
+)
+
+to_hexstr_from_eth_key = lambda x: x.to_hex()  # noqa: E731
```
pipermerriam: You can replace this with `operator.methodcaller('to_hex')`

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    return formatter(private_key)
+
+
+def to_account(web3, private_key):
+
+    if isinstance(private_key, LocalAccount):
+        return private_key
+
+    elif isinstance(private_key, (PrivateKey, str, bytes,)):
```
pipermerriam: The acceptance of both `str` and `bytes` seems off here, or at least like they should require different logic (unless maybe `web3.eth.account.privateKeyToAccount` gracefully handles hex encoded private keys in which case ignore I said anything).

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def raw_key_formatter(private_key):
+    formatter = compose(
+        apply_formatter_if(is_eth_key, to_hexstr_from_eth_key),
+    )
+    return formatter(private_key)
+
+
+def to_account(web3, private_key):
```
pipermerriam: We've been using the convention of `w3` as the instance name for web3 instances.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+    raise TypeError(
+        "key must be one of the types: "
+        "eth_keys.datatype.PrivateKey, eth_account.local.LocalAccount, "
+        "or raw private key as a string or bytestring. "
+        "Was of type {0}".format(type(private_key)))
+
+
+def construct_sign_and_send_raw_middleware(private_key):
+
+    def sign_and_send_raw_middleware(make_request, web3):
```
pipermerriam: Same here for using `w3` as the instance variable name.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+        return web3.eth.account.privateKeyToAccount(normalized_key)
+
+    raise TypeError(
+        "key must be one of the types: "
+        "eth_keys.datatype.PrivateKey, eth_account.local.LocalAccount, "
+        "or raw private key as a string or bytestring. "
+        "Was of type {0}".format(type(private_key)))
+
+
+def construct_sign_and_send_raw_middleware(private_key):
```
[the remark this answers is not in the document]
pipermerriam: This actually looks like a great use case for [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch).

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def sign_and_send_raw_middleware(make_request, web3):
+        account = to_account(web3, private_key)
+
+        def middleware(method, params):
+            if method == "eth_sendTransaction":
+
+                fill_tx = compose(
```
pipermerriam: This looks like it could be done outside of the `middleware` function closure as to not re-execute it every time.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+                    fill_transaction_defaults(web3),
+                    fill_nonce(web3))
+
+                transaction = fill_tx(params[0])
+
+                if 'from' in transaction:
+                    try:
+                        assert is_same_address(account.address, transaction['from'])
```
pipermerriam: This`try/except` that catches the assertion seems like it would be clearer as `if not is_same_address(...): raise ValueError(...)`.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+                    except AssertionError:
+                        raise ValueError("Sending account address mismatch. \
+                            Transaction 'from' parameter does not match the \
+                            address to the private key used to construct signing \
+                            middleware.")
+                else:
+                    raise ValueError("Transaction signing middleware parameter \
```
pipermerriam: We've been using the following convention for multiline strings (and in general have avoided `\` based multiline statements).

```python
raise ValueError(
    "This string spans "
    "multiple lines "
    "but python magically concatenates the lines for "
    "us which is kind of nice."
)
```

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
...
+                        requirement not met: 'from' address was not included in\
+                        transaction parameters.")
+
+                raw_tx = account.signTransaction(transaction).rawTransaction
```
pipermerriam: In our internal code, I've preferred we *not* rely on the `AttrDict` API and to instead use explicit dictionary lookups.  Curious what others feel about this convention.  cc @carver ?

web3/utils/transactions.py
```diff
@@ -30,6 +30,18 @@
 }
 
 
+@curry
+def fill_nonce(web3, transaction):
+    if 'from' in transaction and 'nonce' not in transaction:
+        return dict(
```
pipermerriam: can you change this to use `cytoolz.assoc`

web3/middleware/signing.py
carver: Yup, `privateKeyToAccount` assumes hex-encoding if a `str`: http://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.privateKeyToAccount

web3/middleware/signing.py
carver: I don't feel strongly enough to push for a different convention, so I'm good with internal code preferring to use item lookup. But...

This *kind of* isn't internal code, it's the API exposed by `eth-account`. The critical difference is that even if the `attrdict` middleware is removed, this attribute access would still work. So I'm :+1: with leaving as is.

web3/middleware/signing.py
```diff
@@ -0,0 +1,92 @@
+from eth_account.local import (
+    LocalAccount,
+)
+from eth_keys.datatypes import (
+    PrivateKey,
+)
+from eth_utils import (
+    is_same_address,
+)
+from toolz import (
```
voith: have a look at #809. This should be change to `from web3.utils.toolz import compose`

web3/utils/transactions.py
voith: *`web3.utils.toolz.assoc`

tests/core/contracts/test_contract_constructor.py
```diff
@@ -20,7 +20,7 @@ def test_contract_constructor_gas_estimate_no_constructor(MathContract):
     try:
         assert abs(gas_estimate - 167412) < 200  # Geth
     except AssertionError:
-        assert abs(gas_estimate - 200321) < 200  # eth-tester with py-evm
+        assert abs(gas_estimate - 177396) < 200  # eth-tester with py-evm
```
carver: An alternative to consider: set the actual gas cost on the left, and the 21000 tolerance on the right (instead of 200).

tests/core/middleware/test_transaction_signing.py
```diff
@@ -0,0 +1,188 @@
...
+    w3.eth.sendTransaction({
+        'to': w3.eth.accounts[0],
+        'from': account.address,
+        'gas': 21000,
+        'gasPrice': 0,
+        'value': 1})
+    assert w3.eth.getBalance(account.address) <= start_balance - 1
+
+
+def test_signed_transaction_unset_gas(w3, key_object, fund_account):
```
carver: Looks like this and the one above it could be compressed into one with parametrize.

tests/core/middleware/test_transaction_signing.py
```diff
@@ -0,0 +1,188 @@
...
+    w3.middleware_stack.add(construct_sign_and_send_raw_middleware(key_object))
+    with pytest.raises(ValueError):
+        w3.eth.sendTransaction({
+            'to': w3.eth.accounts[0],
+            'gas': 21000,
+            'from': '!@#',
+            'value': 10})
+
+
+def test_missing_address_signed_transaction(w3, key_object):
```
carver: Looks like these last three could be compressed into one with parametrize.

web3/middleware/signing.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    return formatter(private_key)
+
+
+@singledispatch
+def to_account(val):
+    raise TypeError(
+        "key must be one of the types: "
+        "eth_keys.datatype.PrivateKey, eth_account.local.LocalAccount, "
+        "or raw private key as a string or bytestring. "
```
carver: Suggestion: "~as a string~ as a hex string"

web3/middleware/signing.py
```diff
@@ -0,0 +1,107 @@
...
+
+to_account.register(PrivateKey, private_key_to_account)
+to_account.register(str, private_key_to_account)
+to_account.register(bytes, private_key_to_account)
+
+
+def construct_sign_and_send_raw_middleware(private_key_or_account):
+
+    def sign_and_send_raw_middleware(make_request, w3):
+        account = to_account(private_key_or_account)
```
carver: This could move into the outer method, so it's not run every time the middleware gets rebuilt. (Although that's still pretty rare).

web3/middleware/signing.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                transaction = fill_tx(params[0])
+
+                if 'from' in transaction:
+                    if not is_same_address(account.address, transaction['from']):
+                        raise ValueError(
+                            "Sending account address mismatch."
+                            "Transaction 'from' parameter does not match the"
+                            "address to the private key used to construct signing"
+                            "middleware.")
```
carver: Hm, if I'm reading this right: if I specify a `from` address that my node manages, and I have the signing middleware, it will crash with this error.

Maybe check the `from` address against `w3.eth.accounts` first?

web3/middleware/signing.py
```diff
@@ -0,0 +1,107 @@
...
+                else:
+                    raise ValueError(
+                        "Transaction signing middleware parameter"
+                        "requirement not met: 'from' address was not included in"
+                        "transaction parameters.")
```
carver: Since the node can have its own default sender, I think the best thing here is to let this pass through...

Or maybe we could have a strict keyword on the middleware constructor. If strict is on, then it crashes on an empty from field, else it lets the original transaction pass through. I think I could get behind it being default strict (since it's not a default middleware).

carver:
Just realized there's no way to have multiple keys. I was expecting that you could stack a few of these middlewares to support multiple keys. So maybe I'm slightly :-1: on strict mode being on (or supporting it at all). I think we can just let all send transactions that don't match the account through, and the node will error out if it doesn't recognize the account.

web3/middleware/signing.py
dylanjw: I didnt consider the multiple keys use case.  It makes more sense to pass it through when from doesnt match.  That would allow one to stack multiple signing middlewares, unless Im missing something.

pipermerriam:
@carver @dylanjw thoughts on having multiple keys that people want in-flight signing.

- support multiple keys in the same middleware.
- incur overhead of multiple middlewares and keep each middleware isolated to a single key.

I'm inclined to go with multiple keys in the same middleware option since it **should** be very clean to implement and the overhead could get large-ish if someone say wants 100+ keys for in flight signing.

At first I was inclined to say that's a minority use case but here are two use cases which might necessitate that.  Both are different versions of the same use case.

1. The execution client for ethereum alarm clock sets up multiple accounts.  This allows the client to issue many execution transactions which aren't dependent on one another due to nonces to execute scheduled transactions that are all scheduled at about the same time.  **if** the alarm service ends up with a lot of usage, it'd be reasonable to have 50-100 accounts being used for execution.
2. Programatic trading on a decentralized exchange.  Same issue, wanting to be able to issue orders concurrently that aren't nonce dependent.

Thoughts?

carver:
:+1: for multiple keys in one middleware

tests/core/contracts/test_contract_constructor.py
dylanjw: @carver Replacing the hardcoded value with the actual gas usage is showing the gas estimation to be >21000 off in most cases.

dylanjw:
Ive added support for multiple keys in a single middleware object.  Fixing up the tests to pass ended up being a bit ugly, so I would like to clean that up and add a few more tests.

web3/middleware/signing.py
```diff
@@ -0,0 +1,127 @@
...
+
+def raw_key_formatter(private_key):
+    formatter = compose(
+        apply_formatter_if(is_eth_key, to_hexstr_from_eth_key),
+    )
+    return formatter(private_key)
+
+
+@to_dict
+def to_accounts(val):
```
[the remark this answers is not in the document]
dylanjw: No, just bad naming.

web3/middleware/signing.py
```diff
@@ -0,0 +1,127 @@
...
+    normalized_key = raw_key_formatter(val)
+    return Account.privateKeyToAccount(normalized_key)
+
+
+to_account.register(PrivateKey, private_key_to_account)
+to_account.register(str, private_key_to_account)
+to_account.register(bytes, private_key_to_account)
+
+
+def construct_sign_and_send_raw_middleware(private_keys_or_accounts, strict=False):
```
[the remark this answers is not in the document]
dylanjw: I have it so a single key or an iterable of keys are both an accepted first parameter.  I figured the use case with a large array of keys would be easier this way.

dylanjw:
This is ready to go aside from a py-evm --> eth-tester version bump to fix the gas estimation errors.

web3/middleware/signing.py
```diff
@@ -0,0 +1,138 @@
...
+    def sign_and_send_raw_middleware(make_request, w3):
+
+        fill_tx = compose(
+            fill_transaction_defaults(w3),
+            fill_nonce(w3))
+
+        def middleware(method, params):
+            if method != "eth_sendTransaction":
+                return make_request(method, params)
+
```
carver: *super nitpick* unnecessary whitespace

web3/middleware/signing.py
```diff
@@ -0,0 +1,138 @@
...
+            else:
+                transaction = fill_tx(params[0])
+
+            if 'from' not in transaction:
+                if strict:
+                    raise ValueError(
+                        "Transaction signing middleware parameter"
+                        "requirement not met: 'from' address was not included in"
```
carver: Looks like there should be spaces at the end of the above two strings, or we get:
- parameterrequirement
- intransaction

web3/middleware/signing.py
```diff
@@ -0,0 +1,138 @@
...
+                        "transaction parameters.")
+                else:
+                    return make_request(method, params)
+
+            elif transaction.get('from') not in accounts:
+                if strict:
+                    raise ValueError(
+                        "Sending account address mismatch."
+                        "Transaction 'from' parameter does not match the"
+                        "address to the private key used to construct signing"
```
carver: More strings missing spaces

pipermerriam:
@dylan is it correct that you need:

1. a py-evm release
2. an eth-tester release with latest py-evm

andrevmatos:
Hope it's not too late to ask, but looks like this middleware only intercept `eth_sendTransaction`. Any chance of this also providing `eth_sign`? Would this be enough for a complete signer middleware (i.e. any other missing methods besides this)?

pipermerriam:
@andrevmatos +1 to a middleware for signing but:

1. I prefer it be implemented indepenent of ths middleware.
2. I prefer it be implemented once a signing standard has been established.

Right now `eth_sign` is largely broken and there is yet to be a standard.  Until one of the signing standards EIPs moves forward we're hesitant to build any more signing tooling.

carver:
@andrevmatos check out http://web3py.readthedocs.io/en/stable/web3.eth.account.html#sign-a-message for local signing examples

dylanjw:
@pipermerriam Yes to requiring a version bump of py-evm, and a version bump of eth-tester with the latest py-evm.

pipermerriam:
The CircleCI outage is blocking this. otherwise will have it done as soon as CI is back up.

tests/core/middleware/test_transaction_signing.py
```diff
@@ -0,0 +1,304 @@
...
+        ),
+        (
+            #  Un-strict transaction with mismatched sender
+            # expect a key error with sendTransaction + unmanaged account
+            {
+                'gas': 21000,
+                'value': 10
+            },
+            False,
+            KeyError,
```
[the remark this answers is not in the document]
dylanjw: `eth_tester` throws a key error in that case.  In this set of tests (`test_signed_transaction`), Im using a web3 instance with `EthereumTesterProvider`, no longer using the dummy provider.

tests/core/middleware/test_transaction_signing.py
```diff
@@ -0,0 +1,304 @@
...
+    w3_base.middleware_stack.add(result_generator_middleware)
+    return w3_base
+
+
+@pytest.mark.parametrize(
+    'method,strict_account_matching,key_object,from_,expected',
+    (
+        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE, ADDRESS_1, 'eth_sendRawTransaction'),
+        ('eth_sendTransaction', True, SAME_KEY_MIXED_TYPE, ADDRESS_2, ValueError),
+        ('eth_sendTransaction', False, SAME_KEY_MIXED_TYPE, ADDRESS_2, NotImplementedError),
```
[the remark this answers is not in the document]
dylanjw: There are other places in pytests where we do this.  I agree it would be clearer to define our own exception.  I will make a note of this, and add it to another PR that includes the other tests as well.

pipermerriam:
New `eth-tester` and `py-evm` released for you.  Latest alphas should be good to go.

pipermerriam:
I'm regretting even posting this because of the massive amount of thrash this issue has had but.

I'm inclined to drop the `strict` logic.

- The behavior is potentially surprising and non-intuitive
- It can be implemented as a stand-alone middleware

stand-alone middleware decouples these two bits of functionality which I think is good, and we can still provide them bundled together through something like a `merge_middleware` function which wraps up multiple middlewares into a single middleware.

pipermerriam:
:shipit: when green!

Issue #576, opened 2018-01-23
hex-encode the block number in eth_call


carver:
* Version: *?

### What was wrong?

As reported by @chfast :

```
>>> web3.eth.call(tx, 5)
ValueError: {'message': 'invalid argument 1: hex string without 0x prefix', 'code': -32602}
```

To fill in:

- web3 version: _
- web3 provider: _

### How can it be fixed?

Hex encode the block number before making rpc call, probably through the block number formatter in the pythonic middleware: https://github.com/ethereum/web3.py/blob/master/web3/middleware/pythonic.py

Workaround: supply block number already hex-encoded, like: `web3.eth.call(tx, "0x5")`

PR #656, opened 2018-02-20
Pop and validate chainId, if provided with txn

carver:
### What was wrong?

`chainId` is a part of a transaction only used during local signing. Remote nodes don't know what to do with `chainId` as a field in `eth_sendTransaction`, according to spec.

#595 should be after merging this change.

### How was it fixed?

If sending a transaction to a remote node, strip out the `chainId` field, after validating the value. Added a new middleware layer that make make sense as a place to move more validation, in general.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/middleware/validation.py
+++ b/web3/middleware/validation.py
@@ -1,0 +1,47 @@
+from cytoolz import (
+    compose,
+    curry,
+    dissoc,
+)
+from eth_utils.curried import (
+    apply_formatter_at_index,
+    apply_formatters_to_dict,
+)
+
+from web3.exceptions import (
+    ValidationError,
+)
+
+
+@curry
+def validate_chain_id(web3, chain_id):
+    if chain_id == web3.version.network:
+        return None
+    else:
+        raise ValidationError(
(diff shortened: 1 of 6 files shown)


web3/middleware/validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+def validation_middleware(make_request, web3):
+    transaction_validator = apply_formatters_to_dict({
+        'chainId': validate_chain_id(web3),
+    })
+
+    transaction_sanitizer = compose(transaction_normalizer, transaction_validator)
+
+    def middleware(method, params):
+        if method in ('eth_sendTransaction', 'eth_estimateGas', 'eth_call'):
```
pipermerriam: *semantics nitpick*

Should be a `set` instead of a `tuple`

carver:
Waiting to merge until the eth-utils beta.2 API changes are all cleaned up.

carver:
The next PR (for #576 ) depends on this, so I'm just going to merge this, and fixup the `set` in the new one.

PR #663, opened 2018-02-21
 accept integers for block ids in w3.eth.call

### LINKED ISSUE #576: hex-encode the block number in eth_call

carver:
### What was wrong?

Fixes #576 

### How was it fixed?

- convert int block numbers to hex in `pythonic` middleware
- convert hex back to int for eth-tester provider
- a little leftover fix from #656 
- bugfix in eth-tester middleware: only convert to integer if value is valid hex in `to_integer_if_hex` (was converting any `str` previously)
- pin to eth-utils beta1 until all the cleanup is complete

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -7,6 +7,7 @@ from hexbytes import (
 from web3.exceptions import (
     BadFunctionCallOutput,
     InvalidAddress,
+    BlockNumberOutofRange
 )
 from web3.utils.ens import (
     contract_ens_addresses,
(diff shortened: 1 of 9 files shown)


carver:
Probably will close after #552 merges (except for the little tuple -> set tweak)

carver:
Bummer, this thing has a Heisenbug. Repeated runs alternate between failing and not, with logs showing up where they're not supposed to (or not showing up when they're supposed to, or all tests passing).

carver:
Ah, looks like all the transactions were leaking from test to test. Reverting all changes in eth-tester (when mining is disabled), seems to resolve it in local testing.

PR #664, opened 2018-02-21
upgrade eth-abi for eth-utils b2 compatibility

carver:
### What was wrong?

eth-abi v1b0 was not compatible with eth-utils v1b2

### How was it fixed?

upgraded to beta 1

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "cytoolz>=0.9.0,<1.0.0",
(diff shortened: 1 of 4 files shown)


carver:
Ah, still stuck on eth-tester compatibility with eth-utils

carver:
Watch for one of these to close and then eth-tester to be released: https://github.com/ethereum/eth-tester/pull/64 or https://github.com/ethereum/eth-tester/pull/63

PR #669, opened 2018-02-23
added waitForTransactionReceipt to web3.eth which returns the transac…

### LINKED ISSUE #574: Add function w3.eth.waitForTransactionReceipt(tx_hash)
* Version: 4

### What was wrong?

It's a bit cumbersome to wait for a transaction to be mined. For example, after deploying a transaction, example code typically involves a small while loop, like in #573 

This is a very common need, and should have an easy one-liner.

### How can it be fixed?

Something like:

```py
receipt = w3.eth.waitForTransactionReceipt(tx_hash)
contract = MyContract(receipt['contractAddress'])

# or

with w3.waitFor(lambda: w3.eth.getTransactionReceipt(tx_hash)) as receipt:
    contract = MyContract(receipt['contractAddress'])

```

Very open to alternative API suggestions.

This could potentially let us do cool things eventually, like use pubsub under the hood. Note that pubsub integration would be significantly easier in the first API than the second. But the first API implies adding other new functions that start with `waitFor` at some point in the future.

scottydelta:
…tion after its mined

### What was wrong?
Fixes #574 

### How was it fixed?
Added `waitForTransactionReceipt` to `web3.eth` which waits for transaction to mine and returns the transaction once it is mined.

#### Cute Animal Picture

Cute animal picture


scottydelta:
Test and docs yet to be added. please add `Work in Progress` tag to it. thanks

scottydelta:
how can I go about adding a test since it needs to have a pending transaction before it returns receipt? i did test it manually on Ropsten Network by creating a transaction manually and then running the function to test if it actually waits and returns the receipt after transaction is mined. thanks

carver:
There is some pretty new functionality in eth-tester to enable this. One approach would be something like:

```py
def disable_in_1s():
    time.sleep(1)
    eth_tester.enable_auto_mine_transactions()

eth_tester.disable_auto_mine_transactions()
txn_hash = w3.eth.sendTransaction(...)

with pytest.raises(TimeoutException):
    w3.eth.waitForTransactionReceipt(txn_hash, timeout=0.5)

Thread(disable_in_1s, daemon=True).start()

txn_receipt = eth.waitForTransactionReceipt(txn_hash, timeout=2)
assert txn_receipt is not None and txn_receipt['block_hash']
```

But yeah, it's a little hairy with the thread blocking. There's probably a better approach, but that would get it done.

pipermerriam:
@carver what do you think about testing via something like this using middleware to simulate a transaction not yet being mined.

```
counter = itertools.counter()
def unmined_receipt_simulator_middleware(make_request, web3):
    receipt_counters = collections.defaultdict(itertools.counter)

        if method == 'eth_getTransactionReceipt':
            txn_hash = params[0]
            if next(receipt_counters[txn_hash]) < 2:
                return {'result': None}
    return middleware
```

carver:
Yes, I like that much better :+1:  It's simpler, and takes less time to run.

For @scottydelta: `itertools.count` instead of `counter`

It would be nice to confirm that the middleware is working with a manual check of the empty result first, then the wait (but with a short timeout so that the test completes quickly even if it fails):

```py
assert w3.eth.getTransactionReceipt(txn_hash) == None

receipt = w3.eth.waitForTransactionReceipt(txn_hash, timeout=1)
assert receipt['transaction_hash'] == txn_hash
assert receipt['block_hash']
```

carver:
You can run `make lint-roll` locally to cleanup and check for linting issues.  Loo

Based on the [circle lint run](https://circleci.com/gh/ethereum/web3.py/1619?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), the only issue left is the ordering of the imports:

```
ERROR: /home/circleci/repo/web3/eth.py Imports are incorrectly sorted.
--- /home/circleci/repo/web3/eth.py:before	2018-02-26 07:03:47.343238
+++ /home/circleci/repo/web3/eth.py:after	2018-02-26 07:04:02.690568
@@ -41,9 +41,9 @@
     assert_valid_transaction_params,
     extract_valid_transaction_params,
     get_buffered_gas_estimate,
-    wait_for_transaction_receipt,
     get_required_transaction,
     replace_transaction,
+    wait_for_transaction_receipt,
 ```

File: web3/eth.py [hunk 1]
```diff
@@ -187,6 +188,10 @@ def getTransactionFromBlock(self, block_identifier, transaction_index):
             [block_identifier, transaction_index],
         )
 
+    def waitForTransactionReceipt(self, transaction_hash):
```
carver: I would expect this to have a timeout, like the underlying method.

scottydelta: you got it 👍

scottydelta:
@carver while I understand your initial [suggested method](https://github.com/ethereum/web3.py/pull/669#issuecomment-368099932) of testing, I am finding @pipermerriam middleware method to stimulate un-mined transaction hard to understand. I haven't worked with tests alot.

is the actual test executed like this?

    make_request = Mock()
    txn_receipt = unmined_receipt_simulator_middleware(make_request, web3)
    txn_hash = web3.eth.sendTransaction({
        'from': web3.eth.coinbase,
        'to': '0xd3CdA913deB6f67967B99D67aCDFa1712C293601',
    })
    assert txn_receipt('eth_getTransactionReceipt', [txn_hash]) is None
    ..

pipermerriam:
@scottydelta 

Not exactly.  You would add the `unmined_receipt_simulator_middleware` to the web3 middlewares:

```python
w3.middleware_stack.add(unmined_receipt_simulator_middleware)
```

Then you can call `web3.eth.getTransactionReceipt` and the first two calls will always return `None`, after which it will return the receipt.

scottydelta:
@pipermerriam thanks for the guidance. Wouldn't this test `getTransactionReceipt` instead of newly added method `waitForTransactionReceipt`? I know in the end `getTransactionReceipt` get called when `waitForTransactionReceipt` is invoked.

pipermerriam:
@scottydelta the middlewares don't know the difference between `web3.eth.getTransactionReceipt` and `web3.eth.waitForTransactionReceipt`.  The only thing that middlewares are aware of are the underlying JSON-RPC methods.

So, by constructing this middleware which responds to `eth_getTransactionReceipt` JSON-RPC requests, you're emulating the behavior of an unmined transaction by intercepting the request and returning `None`.

So, the idea is that you are testing `web3.eth.waitForTransactionReceipt` by artificially creating the JSON-RPC behavior for an unmined transaction.

Make sense?

scottydelta:
It makes sense but we are not testing the code we just added in any way. The method `waitForTransaction` is never tested and instead JSON-RPC method `eth_getTransactionReceipt` gets tested which must already be getting tested in some other tests. Is that a normal behavior? 

This is my first experience with tests so finding it a bit confiusing

scottydelta:
@pipermerriam It makes sense but the method `waitForTransaction` which I just added never gets tested, instead the JSON-RPC method `eth_getTransactionReceipt` gets tested which must already be getting tested in some other test. is that a normal behavior?

carver:
@scottydelta it looks like this is almost at the finish line. Are you able to work on it more?

PR #673, opened 2018-02-25
Add EIP155 Transaction test to eth_account tests

dylanjw:
### What was wrong?

It was noted in #649 that EIP155 transactions fail against eth_tester/py-evm, and that were no existing tests cases covering this.

### How was it fixed?

A test was added to tests/core/eth-module/test_accounts.py sending the raw transaction to eth_tester.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/eth-module/test_accounts.py
+++ b/tests/core/eth-module/test_accounts.py
@@ -383,7 +383,7 @@ def test_eth_account_prepared_encrypt(acct, web3js_key, web3js_password):
 
 @pytest.mark.xfail
 @pytest.mark.parametrize(
-    'txn, private_key, expected_raw_tx, tx_hash, r, s, v',
+    'expected_txn, raw_tx, expected_tx_hash, r, s, v',
     (
         (
             {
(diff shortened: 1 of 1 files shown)


dylanjw:
I marked the failing test xfail, since this is not a web3.py bug, and probably requires changes to py-evm to be less strict with the recovery id and to conform to eip155.

tests/core/eth-module/test_accounts.py
```diff
@@ -371,3 +379,52 @@ def test_eth_account_prepared_encrypt(acct, web3js_key, web3js_password):
...
+    ids=['web3js_example', '31byte_r_and_s'],
+)
+def test_eth_account_sign_and_send_EIP155_transaction_to_eth_tester(
+        w3,
+        txn,
+        private_key,
+        expected_raw_tx,
+        tx_hash,
+        r, s, v):
+    w3.eth.sendRawTransaction(expected_raw_tx)
```
carver: We can catch a lot more problems by checking the value after sending, like:

```py
actual_txn_hash = w3.eth.sendRawTransaction(expected_raw_tx)
assert actual_txn_hash == expected_txn_hash
actual_txn = w3.eth.getTransaction(txn_hash)
for key in {'to', 'value', ..., 'chainId'}:
    assert actual_txn[key] == txn[key]
assert actual_txn.r == r
assert actual_txn.s == s
assert actual_txn.v == v
```

tests/core/eth-module/test_accounts.py
```diff
@@ -371,3 +379,52 @@ def test_eth_account_prepared_encrypt(acct, web3js_key, web3js_password):
     decrypted_key = acct.decrypt(encrypted, web3js_password)
 
     assert decrypted_key == to_bytes(hexstr=web3js_key)
+
+
+@pytest.mark.xfail
+@pytest.mark.parametrize(
+    'txn, private_key, expected_raw_tx, tx_hash, r, s, v',
```
carver: A couple of these variable names could be updated in this context:

- `txn` -> `expected_txn`
- `expected_raw_tx` -> `raw_txn`
- `private_key` can be deleted
- `tx_hash` -> `expected_txn_hash`

carver:
The Travis run is only breaking because it's using the `.travis.yml` from the branch instead of master. Since everything else looks good, including the same tests on circleci, I'll go ahead and merge.

PR #676, opened 2018-02-28
Eth utils v1.0.1 compatible

carver:
### What was wrong?

Oh boy, another eth-utils deletion. Last one until v2 :+1: 

### How was it fixed?

Delete all usages of `force_*` and `coerce_*`, replacing with the appropriate `to_text` or `to_bytes`.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/utils/blocks.py
+++ b/web3/utils/blocks.py
@@ -1,16 +1,17 @@
 from eth_utils import (
-    force_text,
     is_hex,
     is_integer,
     is_string,
     remove_0x_prefix,
+    text_if_str,
+    to_text,
 )
 
 
 def is_predefined_block_number(value):
     if not is_string(value):
         return False
-    return force_text(value) in {"latest", "pending", "earliest"}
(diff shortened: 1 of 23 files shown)


carver:
Sorry for all the churn on this PR, it's actually faster to run the tests in CircleCI than locally, and this PR is time-sensitive.

web3/utils/blocks.py
```diff
@@ -1,16 +1,24 @@
...
     remove_0x_prefix,
 )
 
 
 def is_predefined_block_number(value):
-    if not is_string(value):
+    if is_text(value):
+        value_text = value
+    elif is_bytes(value):
+        value_text = value.decode('latin-1')
```
pipermerriam: I'm assuming the use of `latin-1` is to ensure that we can in fact decode the bytes to a text value.  In that case, I think this deserves a comment.

carver: Right, this is speculatively decoding arbitrary byte values. I'll comment :+1:

PR #681, opened 2018-03-03
Eth account unstable

### LINKED ISSUE #678: Warn that eth-account has not been audited
* Version: 4.0.0

## What was wrong?

`eth-account` is pending audit, but there is an expectation that `w3.eth.account` will be included with web3.py v4. We don't want people who see that v4 is going stable to mistake that as a signal that `eth-account` is ready for prod use.

### Do all of these:

- Include a big warning box at the top of http://web3py.readthedocs.io/en/latest/web3.eth.account.html saying something like: "This has not gone through security review"
- Some kind of explicit opt-in process
- Add a section to the docs explaining how to enable `w3.eth.account`

#### Approaches

1. Install "extra" to opt-in
2. Flip runtime flag to opt in
3. ??

#### Approach Detail

1. Don't install `eth-account` unless user installs web3 with `pip install web3[unaudited-private-keys]`. At startup, if `eth_account` is not available, create a `w3.eth.account` property that immediately throws on access, explaining:
    a. what the property is
    b. ask if they meant `w3.eth.accounts`
    c. why it's not enabled by default
    d. how to enable it
2. Always install `eth-account`, but immediately install the throwing `w3.eth.account` property  (see: a, b, c, d from 1). Add a `w3.eth.i_know_private_key_code_is_unaudited = True` property setter that installs `eth-account` to `w3.eth.account`.

#### Tradeoffs

1. Install "extra" to opt-in
   - Benefit: easy to get started; one-time opt-in
   - Downside: if eth-account was already installed (eg~ because they were on the beta), then they can "accidentally" opt in without noticing
2. Flip runtime flag to opt in
   - Benefit: No way to "accidentally" opt in, error message offers a straightforward solution
   - Downside: Have to flip the flag on every run, annoying at terminal

#### Current favorite

Between these, I'm leaning to 2. We can get rid off all this nonsense after the audit. In the meantime, I'd rather have it be a bit more annoying than to have people accidentally opt-in.

In the short term, for people at the terminal, we could guide them to use `from eth_account import Account` directly.

#### Something else?

@pipermerriam any other ideas or thoughts on these?

carver:
### What was wrong?

Fixes #678 

### How was it fixed?

- raise `AttributeError` on `w3.eth.account` by default, saying "This feature is disabled, pending security audit. ..."
- Re-enable the feature with `w3.eth.enable_unaudited_features()`
- Add big docs warning: "**Do not use** this module in production. It is still in beta. A security audit is pending."
- The docs build had started to generate some warning cruft again. Cleaned it up.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -99,4 +99,15 @@ def wait_for_transaction():
 @pytest.fixture()
 def web3():
     provider = EthereumTesterProvider()
-    return Web3(provider)
+    w3 = Web3(provider)
+
+    # Delete this whole block after eth-account has passed security audit
+    try:
+        w3.eth.account
+    except AttributeError:
+        pass
+    else:
+        raise AssertionError("Unaudited features must be disabled by default")
+    w3.eth.enable_unaudited_features()
+
+    return w3
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -83,6 +83,7 @@ exclude_patterns = [
     'web3.rst',
     'modules.rst',
     'web3.auto.rst',
+    'web3.gas_strategies.rst',
     'web3.middleware.rst',
     'web3.providers.rst',
     'web3.providers.eth_tester.rst',
--- a/docs/gas_price.rst
+++ b/docs/gas_price.rst
@@ -104,14 +104,14 @@ Available gas price strategies
       recommended that a caching solution be used to reduce the amount of chain
       data that needs to be re-fetched for each request.
 
-.. code-block:: python
+    .. code-block:: python
 
-    from web3 import Web3, middleware
-    from web3.gas_strategies.time_based import medium_gas_price_strategy
+        from web3 import Web3, middleware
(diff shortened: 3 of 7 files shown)


dangell7:
Can I get some info on this @carver? Whats the reason for this notification/error

dangell7:
@carver what type of security audit are you looking for? What kinds of security vulnerabilities does using eth.account open up?

I'm using local private keys with a hosted node. Really hoping to get this feature pushed to production as I was already using it in my own fork of web3 but decided to upgrade. Now you've got me worried... lol

I would be willing to put in a week or so of work on this to get it to the stable version.

carver:
> I would be willing to put in a week or so of work on this to get it to the stable version.

Thanks, but most of what's left is organization/consensus. Not much coding to do. Won't be long now.

Especially if you're not signing messages, there won't be much in the way of changes -- just some extra validation that might catch if some inputs were messed up.

Issue #666, opened 2018-02-22
Constructor Functions


carver:
Continuation of #577 

Add constructor method to replace and expand on the current `contract.deploy()` capability. Proposed API:

- `contract.constructor(*args, **kwargs).estimateGas()`
- `contract.constructor(*args, **kwargs).transact()`
- `contract.constructor(*args, **kwargs).buildTransaction()`

carver:
@dangell7 started in on this with #595 -- are you interested in continuing the work? You should be unblocked now that the `chainID` issue is fixed. I think we can get another bounty on just this piece, since the fallback work in #577 was more than expected.

pipermerriam:
confirmed that I'll bounty this issue with an equivalent ETH amount as the previous one.

carver:
How about you @zheli - are you interested and available to work on this? (given that an equivalent bounty is coming soon)

zheli:
I would be happy to work on this issue :)

pipermerriam:
@zheli still waiting for txn to confirm, but i just attached a `0.4` eth bounty to this.  We'd like to get this closed out quick so if you're available to work on it and want to do it.... go for it!

dangell7:
Yeah sorry about going mia. I’ve been busy, the team switched to .js 😔but it really should just need to be tested. I’ve been using the forked repo for awhile now. You’re welcome to copy anything I wrote.

pipermerriam:
I used too low a gas price yesterday evening, trying with a higher one.  Either way, this issue will have a bounty.

gitcoinbot:
__This issue now has a funding of 0.4 ETH (346.49 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/666).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/web3.py/issues/666)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $7705.12 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 0.4 ETH (346.49 USD) funding by__: 
 1. [@zheli](https://gitcoin.co/profile/zheli) 

[...]

PR #693, opened 2018-03-12
Constructor functions

zheli:
### What was wrong?

Pull request for #666.

Original text:

> Continuation of #577
>
>Add constructor method to replace and expand on the current contract.deploy() capability. Proposed API:
>
>contract.constructor(*args, **kwargs).estimateGas()
>contract.constructor(*args, **kwargs).transact()
>contract.constructor(*args, **kwargs).buildTransaction()

### How was it fixed?

Based on the code from @dangell7 (https://github.com/ethereum/web3.py/pull/595), introduced a new `ContractConstructor` class.

#### Cute Animal Picture

Cute animal picture


zheli:
This is still work in progress, just want to get some feedback before I finish.

web3/contract.py
```diff
@@ -275,6 +275,28 @@ def deploy(cls, transaction=None, args=None, kwargs=None):
...
+        :param args: The contract constructor arguments as positional arguments
+        :param kwargs: The contract constructor arguments as keyword arguments
+        :return: a contract constructor object
+        """
+        if cls.bytecode is None:
+            raise ValueError(
+                "Cannot call constructor on a contract that does not have 'bytecode' associated "
+                "with it"
+            )
+        # is bytecode not None required by estimateGas and other methods?
```
dylanjw: `bytecode` is not required for `estimateGas`, `call`, or `transact` on deployed contracts, so checking for a bytecode value here is good, since it wont happen further down.

web3/contract.py
```diff
@@ -884,21 +1003,21 @@ def transact(self, transaction=None):
...
         else:
-            estimate_transaction = dict(**transaction)
+            estimate_gas_transaction = dict(**transaction)
 
-        if 'data' in estimate_transaction:
-            raise ValueError("Cannot set data in call transaction")
-        if 'to' in estimate_transaction:
-            raise ValueError("Cannot set to in call transaction")
+        if 'data' in estimate_gas_transaction:
+            raise ValueError("Cannot set data in estimateGas transaction")
```
dylanjw: Thanks for cleaning this up :+1:

web3/contract.py
```diff
@@ -638,6 +660,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
         return deploy_data
 
 
+class ContractConstructor:
```
dylanjw: Is there enough code overlap between this class and the ContractFunction class, to attempt to unify them in some way? If the main difference is the transaction parameter checking, you could use your strategy of putting the parameter check into a function (`check_forbidden_keys_in_transaction`) in a shared base class that can be overrided or have class or instance level `forbidden_keys` set to fit the different situation.

web3/contract.py
```diff
@@ -638,6 +660,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+
+        if 'to' in transact_transaction:
+            raise ValueError("Cannot specify `to` for transact transaction")
+
+        if self.web3.eth.defaultAccount is not empty:
+            transact_transaction.setdefault('from', self.web3.eth.defaultAccount)
+
+        transact_transaction['data'] = self.data_in_transaction
+
+        return self.web3.eth.sendTransaction(transact_transaction)
```
dylanjw: Is the other main difference between the ContractContructor and ContractFunction methods due to the `prepare_transaction` function? What about making a `prepare_constructor_transaction` variant, and setting a class variable with the desired function to use to prepare the transaction?

zheli:
@dylanjw how would you determine the expected gas value in the tests? Right now I just assume that the number from `estimateGas()` function is correct and change the test data based on that.

pipermerriam:
Feel free to just measure how much gas is used and then assert with that value.

web3/contract.py
```diff
@@ -645,6 +665,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+    def estimateGas(self, transaction=None):
+        if transaction is None:
+            estimate_gas_transaction = {}
+        else:
+            estimate_gas_transaction = dict(**transaction)
+            self.check_forbidden_keys_in_transaction(estimate_gas_transaction,
+                                                     ["data", "to"])
+
+        if self.address:
+            estimate_gas_transaction.setdefault('to', self.address)
```
carver: These two lines don't make sense to me in the context of a constructor, which should always have a blank `to` field.

web3/contract.py
```diff
@@ -645,6 +665,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+        else:
+            transact_transaction = dict(**transaction)
+            self.check_forbidden_keys_in_transaction(transact_transaction,
+                                                     ["data", "to"])
+
+        if 'data' in transact_transaction:
+            raise ValueError("Cannot set data in transact transaction")
+
+        if 'to' in transact_transaction:
+            raise ValueError("Cannot specify `to` for transact transaction")
```
carver: These two field checks look redundant after the `check_forbidden_keys_in_transaction()` call

web3/contract.py
carver: I think there might be some cuft left over from copying things from ContractFunction, and once it's cleared out it will look less similar, to the point where I'm not sure merging them adds clarity. I'm happy to be shown wrong on that, though, because it does *feel* like there's duplication here.

web3/contract.py
```diff
@@ -645,6 +665,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+        """
+        Build the transaction dictionary without sending
+        """
+
+        if transaction is None:
+            build_transaction = {}
+        else:
+            build_transaction = dict(**transaction)
+            self.check_forbidden_keys_in_transaction(build_transaction,
+                                                     ["data", "to", "nonce"])
```
carver: Why are we not allowing the user to specify a `nonce` here?

web3/contract.py
```diff
@@ -937,25 +1054,25 @@ def buildTransaction(self, transaction=None):
         Build the transaction dictionary without sending
         """
         if transaction is None:
-            built_transaction = {}
+            build_transaction = {}
```
carver: I disagree that this wording is clearer, but if you like `build` a lot better, I am ok with keeping it. Here's why I like `built` better: `build` is a verb, but the variable is holding a thing (noun) -- the transaction that is being built.

Counter-point: the analogous calls, like `estimate_gas_transaction` sound more like verbs, too, so `build` is a bit more consistent with those. On that point, I'd argue that the best solution would be to change them all to nouns, but maybe it's less obvious what the other nouns should be, like: `gas_estimate_transaction`?

Takeaway: this is a semantic nitpick, and a bit murky, so you make the call.

web3/contract.py
```diff
@@ -645,6 +665,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
...
+    """
+    def __init__(self, web3, address, abi, bytecode, args, kwargs):
+        self.web3 = web3
+        self.address = address
+        self.abi = abi
+        self.bytecode = bytecode
+        self.data_in_transaction = self._encode_data_in_transaction(args, kwargs)
+
+    @combomethod
+    def _encode_data_in_transaction(self, args=None, kwargs=None):
```
carver: _Just an idea_

Maybe this whole method could be moved to the contract utils, as:

```py
def encode_constructor_data(web3, abi, bytecode, args, kwargs):
```

and remove it from this class and the contract class, using the utility method instead.

web3/contract.py
```diff
@@ -645,6 +665,103 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
         return deploy_data
 
 
+class ContractConstructor:
+    """
+    Class for contract constructor API.
+    """
+    def __init__(self, web3, address, abi, bytecode, args, kwargs):
+        self.web3 = web3
+        self.address = address
```
carver: When is the address going to be helpful in the constructor? Maybe we want to raise an exception if the user tries to issue a deployment transaction for a contract that already has an address?

I don't see any other good way to use the address value with the constructor.

web3/contract.py
zheli: Thanks for the feedback! 

@carver That is what I am thinking about as well. I would like the code to be as DRY as possible, but at the same time too much abstraction could make it more difficult to maintain. I will see what I can do.

web3/contract.py
zheli: That's a good point. I will try to do that.

web3/contract.py
zheli: True. I will remove it.

web3/contract.py
zheli: Actually I am not so sure about it myself as well, I will revert this change and rename estimate_gas_transactions.

zheli: I think I will leave `estimate_gas_transaction` and `transact_transaction` be, focus on the main task for now. Doesn't seem to provide more values any way :)

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,134 @@
...
+    gas_estimate = SimpleConstructorContract.constructor().estimateGas()
+    try:
+        assert abs(gas_estimate - 21472) < 200  # Geth
+    except AssertionError:
+        assert abs(gas_estimate - 43082) < 200  # eth-tester with py-evm
+
+
+def test_contract_constructor_gas_estimate_with_constructor_with_arguments(
+        WithConstructorArgumentsContract):
+    gas_estimate = WithConstructorArgumentsContract.constructor(args=[1234, 'abcd']).estimateGas()
```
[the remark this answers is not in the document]
zheli: It doesn't work current implementation since `constructor()`'s signature is
```
    def constructor(cls, args=None, kwargs=None)
```
And we will then get `args=1234, kwargs=b'abcd'`...

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,134 @@
...
+    assert txn_receipt is not None
+
+    assert txn_receipt['contractAddress']
+    contract_address = txn_receipt['contractAddress']
+
+    blockchain_code = web3.eth.getCode(contract_address)
+    assert blockchain_code == decode_hex(SIMPLE_CONSTRUCTOR_RUNTIME)
+
+
+def test_contract_constructor_transact_with_constructor_with_arguments(
```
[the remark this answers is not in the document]
zheli: I modified some tests with `pytest.mark.parametriz` decorator so that we can test different args and kwargs combinations. Can you have a look?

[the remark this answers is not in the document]
dylanjw: Im not seeing this change, have you pushed it yet?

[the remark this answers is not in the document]
zheli: Weird, my changes in the branch is not showing up in this pull request:

https://github.com/zheli/web3.py/commits/constructor-functions

Something wrong with github ATM?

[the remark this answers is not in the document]
zheli: Now it's there :) Changed the author and hash for the last commit.

web3/contract.py
```diff
@@ -278,10 +278,29 @@ def deploy(cls, transaction=None, args=None, kwargs=None):
 
         deploy_transaction['data'] = cls._encode_constructor_data(args, kwargs)
 
-        # TODO: handle asynchronous contract creation
         txn_hash = cls.web3.eth.sendTransaction(deploy_transaction)
         return txn_hash
 
+    @classmethod
+    def constructor(cls, args=None, kwargs=None):
```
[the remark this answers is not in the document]
zheli: Yes, of course! Now I follow 👍

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,134 @@
...
+    deploy_txn = WithConstructorArgumentsContract.constructor(args=[1234, 'abcd']).transact()
+
+    txn_receipt = web3.eth.getTransactionReceipt(deploy_txn)
+    assert txn_receipt is not None
+
+    assert txn_receipt['contractAddress']
+    contract_address = txn_receipt['contractAddress']
+
+    blockchain_code = web3.eth.getCode(contract_address)
+    assert blockchain_code == decode_hex(WITH_CONSTRUCTOR_ARGUMENTS_RUNTIME)
```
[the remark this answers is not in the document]
zheli: Fixed.

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,134 @@
...
+        web3, WithConstructorAddressArgumentsContract, WITH_CONSTRUCTOR_ADDRESS_RUNTIME):
+    deploy_txn = WithConstructorAddressArgumentsContract.constructor(
+        args=['0x16D9983245De15E7A9A73bC586E01FF6E08dE737']
+    ).transact()
+    txn_receipt = web3.eth.getTransactionReceipt(deploy_txn)
+    assert txn_receipt is not None
+    assert txn_receipt['contractAddress']
+    contract_address = txn_receipt['contractAddress']
+    blockchain_code = web3.eth.getCode(contract_address)
+    assert blockchain_code == decode_hex(WITH_CONSTRUCTOR_ADDRESS_RUNTIME)
```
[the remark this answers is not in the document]
zheli: Fixed

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,134 @@
...
+        assert abs(gas_estimate - 43082) < 200  # eth-tester with py-evm
+
+
+def test_contract_constructor_gas_estimate_with_constructor_with_arguments(
+        WithConstructorArgumentsContract):
+    gas_estimate = WithConstructorArgumentsContract.constructor(args=[1234, 'abcd']).estimateGas()
+    try:
+        assert abs(gas_estimate - 21472) < 200  # Geth
+    except AssertionError:
+        assert abs(gas_estimate - 78572) < 200  # eth-tester with py-evm
```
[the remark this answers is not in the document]
zheli: alright :)

web3/contract.py
zheli: fixed.

web3/contract.py
zheli: Thanks! It is removed.

docs/contracts.rst
```diff
@@ -109,18 +109,18 @@ Methods
 
 Each Contract Factory exposes the following methods.
 
-.. py:classmethod:: Contract.deploy(transaction=None, args=None)
```
pipermerriam: This documentation shouldn't be removed yet, but rather marked as deprecated in the documentation.

```
.. warning:: `Contract.deploy` has been deprecated in favor of .....
```

tests/core/contracts/test_contract_constructor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    assert txn_receipt['contractAddress']
+    contract_address = txn_receipt['contractAddress']
+
+    blockchain_code = web3.eth.getCode(contract_address)
+    assert blockchain_code == decode_hex(WITH_CONSTRUCTOR_ARGUMENTS_RUNTIME)
+    assert EXPECTED_DATA_A == WithConstructorArgumentsContract(
+        address=contract_address).functions.data_a().call()
+    assert EXPECTED_DATA_B == WithConstructorArgumentsContract(
+        address=contract_address).functions.data_b().call()
```
carver: The global expected value is problematic, because it means the parametrization cannot test a variety of A and B values. It's better to put the expected values as their own parametrized values.

docs/contracts.rst
carver: :+1: to leaving and deprecating the old deploy method

zheli:
Awesome!

pipermerriam:
@carver just checking in to be sure we're good to pay out the bounty on this.

PR #682, opened 2018-03-03
Issue #675

boneyard93501:
added the ConciseContract namespace collision check, test function, and mangled a simple cut & paste job to update the reminder that ConciseContract leads to loss of Contract methods; move the _none_addr from ConciseContract to standalone.
there is some room for improvement in the collision test case for staging the AssertionError and i revisit that sunday or monday.

PR #697, opened 2018-03-14
Add auto-complete support for contract methods. 

### LINKED ISSUE #654: Autocomplete doesn't work, exploring in shell is hard
* Version: 4.0.0b9
* Python: 3.6
* OS: osx

### What was wrong?

Autocomplete is useless in python shell (ipython, ipdb, dir(), etc.).
I expected that I'll get autocompletion when I, for example, type `contract_instance.[TAB]`.

```python
contract_instance = w3.eth.contract(
    contract_address,
    abi=contract_interface['abi'],
    ContractFactoryClass=ConciseContract)
```

I expected that triggering `contract_instance.` would suggest `contract_instance.myMethod()`.
Contract is just an example, I've noticed that most of the code w/ metaclasses don't have this.

### How can it be fixed?
Looking at the source code it could be possible to add `__all__` to `PropertyCheckingFactory` or something/somewhere similar.

Exploring the API from shell is something that works really well for other python library, would be great if it would work for web3.

rakanalh:
### What was wrong?
According to #654, auto-complete didn't work in environments like ipython as loading contracts does not attach `ConciseMethod` callables to the current contract instance.

### How was it fixed?

Introduced iterable functionality to `ContractFunctions` so that `ConciseFunction` could go through the list of function names, fetch the contract function and attach it to the current instance.

#### Cute Animal Picture
This was the hardest part about the PR.

Cute animal picture

Fixes #654

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -106,6 +106,13 @@ class ContractFunctions:
                         address=address,
                         function_identifier=func['name']))
 
+    def __iter__(self):
+        if not hasattr(self, '_functions') or not self._functions:
+            return
+
+        for func in self._functions:
+            yield func['name']
+
 
 class ContractEvents:
     """Class containing contract event objects
@@ -656,13 +663,21 @@ class ConciseContract:
         self._classic_contract = classic_contract
         self.address = self._classic_contract.address
 
+        for func_name in self._classic_contract.functions:
+            func = getattr(self._classic_contract.functions, func_name)
+
+            setattr(
+                self,
+                func_name,
+                ConciseMethod(
+                    func,
+                    self._classic_contract._return_data_normalizers
+                )
+            )
+
     @classmethod
     def factory(cls, *args, **kwargs):
         return compose(cls, Contract.factory(*args, **kwargs))
-
-    def __getattr__(self, attr):
-        contract_function = getattr(self._classic_contract.functions, attr)
-        return ConciseMethod(contract_function, self._classic_contract._return_data_normalizers)
 
     @staticmethod
     def _none_addr(datatype, data):
--- a/tests/core/contracts/test_implicit_contract.py
+++ b/tests/core/contracts/test_implicit_contract.py
@@ -58,13 +58,13 @@ def test_implicitcontract_call_default(math_contract, get_transaction_count):
     assert get_transaction_count("pending") == (blocknum, 0)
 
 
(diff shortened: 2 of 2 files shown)


rakanalh:
I'll look into failing tests tonight... need to figure out an issue running tests as eth-utils throws an error. Related issue https://github.com/ethereum/eth-utils/issues/73

web3/contract.py
```diff
@@ -106,6 +107,19 @@ def __init__(self, abi, web3, address=None):
                         address=address,
                         function_identifier=func['name']))
 
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        self._iter_current_func += 1
```
pipermerriam: This stateful implementation isn't going to work for us.

I believe the following should work.

```python
def __iter__(self):
    for func in self._functions:
        yield func['name']
```

web3/contract.py
```diff
@@ -656,13 +670,24 @@ def __init__(self, classic_contract):
...
+
     @classmethod
     def factory(cls, *args, **kwargs):
         return compose(cls, Contract.factory(*args, **kwargs))
 
-    def __getattr__(self, attr):
-        contract_function = getattr(self._classic_contract.functions, attr)
-        return ConciseMethod(contract_function, self._classic_contract._return_data_normalizers)
+    def _populate_functions(self):
+        for func_name in self._classic_contract.functions:
```
pipermerriam: This logic should just be inlined in the `__init__` method of this class.

pipermerriam:
Upgrading your `eth-tester` version should fix the `eth-utils` import issue.

rakanalh:
@pipermerriam thanks... you were right, my eth-tester seems to have required an upgrade. I've fixed the issue and managed to also fix the test issue. One approach would have been to initialize `self._functions = []` but i thought there was a reason there's no state kept in `ContractFunctions` so i managed with checking for the presence of the attribute.

Running the tests, i have faced 3 failures related to gasEstimate, i ran the tests on the latest web3 master and the same tests fail there as well, am not sure they're related to this PR, but i'll try to help out later double checking those failures.

rakanalh:
I noticed this PR: https://github.com/ethereum/web3.py/pull/682 which is approved and makes changes to the same file as my PR. I can rebase my PR on top of that one or wait until it's merged first and then rebase. Thoughts?

PR #702, opened 2018-03-17
move contract.eventFilter to contract.events.*.createFilter

dylanjw:
### What was wrong?

The contract.events.* api doesnt do very much yet.

### How was it fixed?

Add deprecation warning to `contract.eventFilter`, and create a new method `contract.events.<event_name>.createFilter()`

#### Cute Animal Picture

image


docs/contracts.rst
```diff
@@ -129,7 +129,7 @@ Each Contract Factory exposes the following methods.
 
     Returns the transaction hash for the deploy transaction.
 
-.. py:classmethod:: Contract.eventFilter(event_name, filter_params=None)
+.. py:classmethod:: Contract.events.<event name>.createFilter(filter_params=None)
```
pipermerriam: Just realizing we're using a legacy API that was never really re-evaluated for whether we could improve it.


What if instead of `filter_params` as a single argument you could pass in all of the individual values allowed in filter params, probably as keyword arguments.

```python
c.events.Transfer.createFilter(topics=..., args=..., _from=..., fromBlock=..., toBlock=...)
```

Note that `_from` kind of awkward due to the collision with the `from` keyword in python so maybe we could even change that one to `address=...`?

dylanjw: What if we kept the filter param dict as an option to give raw arguments, and had the new kwargs do some argument processing.  For example, if providing `topics` through the keyword arg, you wouldnt need to include the event signature as the first arg and could enter the other topics as strings or integers values, rather than hex string representations.

Also, the AND or OR syntax for topics is a little odd:

from the json-rpc doc:

>A note on specifying topic filters:

>Topics are order-dependent. A transaction with a log with topics [A, B] will be matched by the following topic filters:

>   [] "anything"
>   [A] "A in first position (and anything after)"
>   [null, B] "anything in first position AND B in second position (and anything after)"
>   [A, B] "A in first position AND B in second position (and anything after)"
>   [[A, B], [A, B]] "(A OR B) in first position AND (A OR B) in second position (and anything after)"

Can you think of a clearer syntax? Maybe a single string of topics with "," for AND (e.g, followed by...) "|" for OR and "*" for ANY?  The above examples would look like:

"*"
"A"
"*, B"
"A, B"
"A|B, A|B"

Im not sure that this is an improvement. It would be cool if the "and anything after" matching could be controlled with the syntax, but I dont think there is a way to have "and nothing after".

pipermerriam: Maybe rather than a DLS (which is what the string based version you proposed) we could do this with a custom helper type.

```
>>> t(A) & t(B)) , t(A) | t(B)
[[A, B], [A, B]]
```

This fancy new `t` class would have custom written `__and__` and `__or__` methods which evaluated to the list form for the topic.

pipermerriam: I just looked up the spec and I don't think my fancy new type idea will work.

carver: So far, when people get stuck on events, it's not because they are confused by the topic filters. We should fix the things we know are broken, and keep transparency to the json-rpc layer where we're not sure.

Some examples of things that I have seen trip people up: 1) default fromBlock value is `latest` and the event they're looking for already happened, 2) and disappearing filters on the client.

Some crazy options for the first:
- deviate from the `fromBlock` default in json-rpc, so that it is 0
- *require* that people specify a `fromBlock` when creating a filter?
- always return a result with log queries, with a value like `{fromBlock: 1234, toBlock:1237, events: [], allEvents: [{event-that-did-not-match}]}`, so that it's clear what blocks are being inspected. There are negative performance implications, but maybe you could opt out of that behavior for production.
-

carver: I take it back: https://github.com/ethereum/web3.py/issues/188 is an example of confusion on filter topics. Although I'd still argue it's less common that the others.

dylanjw: I like the idea of requiring a fromBlock, and including some metadata about the search in the results.  Im less keen on adding the non-matching events to the filter results.  

I think forcing users to think about a choice for the `fromBlock` parameter should be enough. I would prefer keeping the default as `latest` than changing to 0. Typically you need to set up an event listener before catching events. It seems like a more unique situation to listen for events from the past.

dylanjw: I added an explanation of how the topic matching works to the doc, so maybe thats enough for #188

docs/contracts.rst
```diff
@@ -129,7 +129,7 @@ Each Contract Factory exposes the following methods.
 
     Returns the transaction hash for the deploy transaction.
 
-.. py:classmethod:: Contract.eventFilter(event_name, filter_params=None)
```
pipermerriam: Given that this method is still present, we should leave it in the documentation and just mark it as having been deprecated.

```python
.. warning:: The `Contract.eventFilter` method has been deprecated in favor of .......
```

web3/providers/eth_tester/middleware.py
```diff
@@ -35,7 +36,11 @@ def is_named_block(value):
     return value in {"latest", "earliest", "pending"}
 
 
-to_integer_if_hex = apply_formatter_if(is_string, hex_to_integer)
+def is_str_is_hex(value):
```
pipermerriam: Would naming this `is_hexstr` be more consistent with our existing naming convention?

carver:
I've seen at least one person who wants to catch *all* events coming out of a contract, so I'm not so sure we want to drop `contract.eventFilter()`. But I don't feel strongly about it, especially since we're not removing the feature for a while. We can always un-deprecate it later.

web3/contract.py
```diff
@@ -1132,6 +1133,46 @@ def _parse_logs(self, txn_receipt):
                 continue
             yield decoded_log
 
+    @combomethod
+    def createFilter(
+            self, *,  # PEP 3102
+            argument_filters=dict(),
```
pipermerriam: This should get fixed.  Mutable values as defaults for function arguments is an antipattern:

http://docs.python-guide.org/en/latest/writing/gotchas/

This specific case, I don't think it causes any problems, but still something we should avoid.

web3/contract.py
```diff
@@ -1132,6 +1133,46 @@ def _parse_logs(self, txn_receipt):
...
+            fromBlock=None,
+            toBlock="latest",
+            address=None,
+            topics=list()):
```
pipermerriam: Same here, should be changed to `None` and then defaulted to an empty list conditionally if it is `None`

web3/contract.py
ti6oti6o: createEvent does not yield the same result as eventFilter...For example, the below gives no results with createFilter and some results with eventFilter....
```
myfilter_new=contract_instance.events.Deposit.createFilter(fromBlock=2058615)
myfilter_old=contract_instance.eventFilter('Deposit',{'fromBlock':2058615,'toBlock':'latest'})
print("     filterDeposit (OLD METHOD)--",myfilter_old.get_all_entries())
print("     filterDeposit (NEW METHOD)--",myfilter_new.get_all_entries())
```

ti6oti6o:
I think I posted the below in the wrong spot (apologies for that):
createFilter does not yield the same result as eventFilter...For example, the below gives no results with createFilter and some results with eventFilter.... Why is that? Below is rinkeby and web3(4.0.0)

```
myfilter_new=contract_instance.events.Deposit.createFilter(fromBlock=2058615)
myfilter_old=contract_instance.eventFilter('Deposit',{'fromBlock':2058615,'toBlock':'latest'})
print("     filterDeposit (OLD METHOD)--",myfilter_old.get_all_entries())
print("     filterDeposit (NEW METHOD)--",myfilter_new.get_all_entries())
```

dylanjw:
@ti6oti6o Can you have a look at [this comment](https://github.com/ethereum/web3.py/issues/730#issuecomment-379136038), in a thread discussing issues with the createFilter API.  Can you post more enough information to reproduce the issue you are experiencing with createFilter?

pipermerriam:
> I've seen at least one person who wants to catch all events coming out of a contract

This strikes me as something that deserves it's own dedicated API:

- `contract.all_events.createFilter(...)`

Issue #691, opened 2018-03-09
Multiple Providers Call to highest block number


MichaelRazum:
* Version: latest
* Python:3.5
* OS: win

### What was wrong?
I think it is wrong (not 100% sure )! Right now I'm using infura as HTTP Provider and want to add a localhost provider running geth. So the natural way would be: 
`web3 = Web3([geth_local_host_provider, infura_provider])`

Now as I understand web3 makes always the first request to geth_local_host_provider and if it fails it tries infura_provider. In my case sometimes geth get out of sync. So the actual block number is different. I think the right strategy would be to make the "request" to the HTTP Provider with the highest block.

### How can it be fixed?
Check first the block's of every provider and filter if any provider in the list has a lower block number.

pipermerriam:
I don't think your proposed solution will work for us, but I do think this is an issue that we should look into providing at minimum the tools for solving.  This strikes me as something that *should* be solveable with our middleware layer, but I think it may be tough as things currently stand because middlewares don't have visibility across multiple providers.

cc @carver

boneyard93501:
i was looking into something similar to not select infura for stateful interactions. not quite done but checking for blocks as a (user) add-on is a little more manageable.
assuming something like this:
```python
def get_w3() -> Web3:
    uris = [('local','http://127.0.0.1:8545'),
            ('infura','https://ropsten.infura.io/1vy5tbsVeNil32nbTKl5')
           ]
    providers = []
    for uri in uris:
        f = HTTPProvider(uri[1])
        f.__name__ = uri[0]
        providers.append(f)
    return Web3(providers)
```

something like that will get you the winning provider:

```python
def select_provider(providers: list) -> HTTPProvider :
    selection = {}
    for p in providers:
        result = p.make_request('eth_blockNumber',[])
        block_int = int(result['result'],16)
        if selection.get('winner') is None:
            selection['winner'] = p
            selection['block'] = block_int
        if block_int > selection['block']:
            selection['winner'] = p
            selection['block'] = block_int

    return selection['winner']
``` 
if select_providers() actually returned a (ordered by block size provider) list preceding the request loop in  _make_request in manager.py,  you'd end up with the desired result ... at least that's the entry point i was eying for avoiding the infura provider if need be.

pipermerriam:
The issue with pushing this logic into the manager is that it's forces it onto everyone.  While it's going to be *common* for people to want to use the provider with the highest block number, there are other selection criteria that people may want to use.

[...]

PR #703, opened 2018-03-17
add provider ranking and method routing to manager #691

boneyard93501:
### What was wrong?
no options in manager to customize provider selection and provider-based method routing/broadcasting (see #691)

### How was it fixed?
WIP ... add features for custom(izable) manager

### DIFF
--- a/web3/custom_manager_outline.py
+++ b/web3/custom_manager_outline.py
@@ -1,0 +1,439 @@
+#!/usr/bin/env python3
+# -*- coding: utf8 -*-
+import sys
+import threading
+# from copy import deepcopy
+
+from web3 import Web3, HTTPProvider
+"""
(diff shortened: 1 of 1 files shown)


pipermerriam:
I want to verify that this current PR is *currently* intended to be demonstrative of what can be done with some theoretical new manager APIs.

I like some of the *built-in* strategies that you've written and think they'll be something we can include *eventually*.

I however want to let you know how this process will likely need to go.

1. Define the use cases we intend to support.
2. Define the minimal new APIs to support these (probably on the `Manager` class)
3. PR implementing **only** the new APIs
4. Individual PRs to implement each strategy that we want to include in the codebase.

Thanks for working on this.  Excited to have more flexibility in this realm.

marcgarreau:
Fast forward a couple years: Web3.py only supports one provider per instance. Thanks for the original exploration.

Issue #566, opened 2018-01-19
Websocket provider


banteg:
I think it's time we have a talk about adding a websocket provider.
- web3.js v.1.0 deprecated `HttpProvider` in favor of `WebsocketProvider`
- infura [added](https://github.com/INFURA/infura/issues/29) public websocket endpoints
- there is an excellent [websockets](https://websockets.readthedocs.io/en/stable/) library
- websockets give 4-8x better performance compared to HTTP/IPC from my tests

downsides:
- lots of code to rewrite
- supporting both sync and async code is hard

pipermerriam:
> I think it's time we have a talk about adding a websocket provider

I found this absolutely hilarious for some reason....

pipermerriam:
I suspect we can get by with a minimal initial implementation which operates synchronously much the same way the IPC provider does?  I am not well versed in websockets so would need to validate this assertion.

banteg:
I experimented with this a while ago, but no, it can't be easily added. Websockets require a persistent connection and a running event loop. Here it gets tricky as async calls spread to the very top. Other than that, the API is the same.

carver:
> websockets give 4-8x better performance compared to HTTP/IPC from my tests

The performance spread between HTTP and IPC is higher than 2x in my experience. Did you separately test against IPC, or are you just bundling them together after running HTTP tests?

BTW, I'm not arguing against a WebSockets provider; it's valuable enough to implement if it's only faster than HTTP, for services that are only available remotely. But I am genuinely curious how it compares specifically to IPC.

carver:
> Websockets require a persistent connection and a running event loop.

This isn't super different from IPC. A websockets implementation could probably use something similar to IPC's `PersistantSocket` implementation: https://github.com/ethereum/web3.py/blob/master/web3/providers/ipc.py#L31

owocki:
@pipermerriam @carver id love to see websockets available in web3py.  in fact, it looks like [infura wants to see it too](http://bits.owocki.com/2S0n1C3k2p2J/Screen%20Shot%202018-03-12%20at%2011.48.36%20AM.png)

is this on the roadmap soon?  is it a good bounty-able issue?  perhaps given the route that @pipermerriam suggests here? 

> I suspect we can get by with a minimal initial implementation

pipermerriam:
I'm not sure if they **need** to be done in conjunction with each other, but this issue compliments the websockets one.

https://github.com/ethereum/web3.py/issues/657

[...]

PR #708, opened 2018-03-19
Added WebsocketProvider

voith:
### What was wrong?
web3.py does not support websockets as discussed in  #566

### How was it fixed?
A `WebsocketProvider` has been implemented using `websockets` library.

TODO list:

- [x] Address all todo comments.

- [x] Fix integrations tests

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     setup_requires=['setuptools-markdown'],
     extras_require={
         'tester': [
-            "eth-tester[py-evm]==0.1.0-beta.21",
+            "eth-tester[py-evm]==0.1.0-beta.22",
             "py-geth>=2.0.1,<3.0.0",
         ],
         'testrpc': ["eth-testrpc>=1.3.3,<2.0.0"],
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
         "lru-dict>=1.1.6,<2.0.0",
         "eth-hash[pycryptodome]",
         "requests>=2.16.0,<3.0.0",
+        "websockets>=4.0.1",
     ],
     setup_requires=['setuptools-markdown'],
     extras_require={
@@ -40,9 +41,6 @@ setup(
         'platform_system=="Windows"': [
             'pypiwin32'  # TODO: specify a version number, move under install_requires
         ],
-        'websocket': [
-            "websockets==4.0.1"
-        ]
     },
     py_modules=['web3', 'ens'],
     license="MIT",
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,9 @@ setup(
         'platform_system=="Windows"': [
             'pypiwin32'  # TODO: specify a version number, move under install_requires
         ],
+        'websocket': [
+            "websockets==4.0.1"
+        ]
     },
     py_modules=['web3', 'ens'],
     license="MIT",
--- a/tests/integration/go_ethereum/test_goethereum_http.py
+++ b/tests/integration/go_ethereum/test_goethereum_http.py
@@ -3,12 +3,12 @@ import pytest
 from web3 import Web3
 
 from .common import (
-    get_open_port,
     GoEthereumEthModuleTest,
     GoEthereumNetModuleTest,
     GoEthereumPersonalModuleTest,
     GoEthereumTest,
     GoEthereumVersionModuleTest,
+    get_open_port,
 )
 from .utils import (
     wait_for_http,
--- a/tests/integration/go_ethereum/test_goethereum_http.py
+++ b/tests/integration/go_ethereum/test_goethereum_http.py
@@ -1,9 +1,9 @@
 import pytest
-import random
 
 from web3 import Web3
 
 from .common import (
+    get_open_port,
     GoEthereumEthModuleTest,
     GoEthereumNetModuleTest,
     GoEthereumPersonalModuleTest,
@@ -17,12 +17,12 @@ from .utils import (
 
 @pytest.fixture(scope="module")
 def rpc_port():
-    return random.choice(range(8000, 9000))
+    return get_open_port()
 
 
 @pytest.fixture(scope="module")
 def endpoint_uri(rpc_port):
-    return 'http://localhost:{}'.format(rpc_port)
+    return 'http://localhost:{0}'.format(rpc_port)
 
 
 @pytest.fixture(scope='module')
@@ -33,7 +33,7 @@ def geth_command_arguments(geth_binary, datadir, rpc_port):
         '--nodiscover',
         '--fakepow',
         '--rpc',
-        '--rpcport', str(rpc_port),
+        '--rpcport', rpc_port,
         '--rpcapi', 'db,eth,net,web3,personal,web3',
         '--ipcdisable',
     )
--- a/tests/integration/go_ethereum/test_goethereum_http.py
+++ b/tests/integration/go_ethereum/test_goethereum_http.py
@@ -1,7 +1,5 @@
-import os
 import pytest
 import random
-import tempfile
 
 from web3 import Web3
 
--- a/web3/providers/websocket.py
+++ b/web3/providers/websocket.py
@@ -12,14 +12,17 @@ from web3.providers.base import (
     JSONBaseProvider,
 )
 
-THREAD_POLLING_INTERVAL = 2  # secs
+THREAD_POLLING_INTERVAL = 0.5  # secs
 
 
 async def _stop_loop(loop, parent_thread):
     while parent_thread.is_alive():
         await asyncio.sleep(THREAD_POLLING_INTERVAL)
     else:
-        print('stopping loop')
+        all_tasks = asyncio.Task.all_tasks(loop)
+        current_task = asyncio.Task.current_task(loop)
+        pending_tasks = all_tasks - {current_task}
+        loop.run_until_complete(asyncio.gather(*pending_tasks))
         loop.stop()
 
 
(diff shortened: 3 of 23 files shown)


voith:
Submitted early to get early feedback.
I have added the `WebsocketProvider` to the same tests that run integrity tests for parity which earlier used `IPCProvider` only. The transactions tests will fail because the transactions will be repeated by both providers. 
IMO, integrity tests should test all providers. @pipermerriam If supporting all providers in the integrity tests is the way to go then I'd be happy to add them in a separate PR.

setup.py
```diff
@@ -40,6 +40,9 @@
         'platform_system=="Windows"': [
             'pypiwin32'  # TODO: specify a version number, move under install_requires
         ],
+        'websocket': [
+            "websockets==4.0.1"
```
pipermerriam: cc @carver 

I think this should be added as a core requirement that is installed by default.

web3/providers/websocket.py
```diff
@@ -0,0 +1,67 @@
...
+    asyncio.set_event_loop(loop)
+    loop.run_forever()
+
+
+# TODO: Think of a better name
+def _get_threaded_loop():
+    new_loop = asyncio.new_event_loop()
+    t = Thread(target=_start_event_loop, args=(new_loop,))
+    # TODO: figure out if daemon is the way to go
+    t.daemon = True
```
pipermerriam: Ideally we want to have the life cycle of the thread and the provider be explicitly bound together such that if an instance of the provider goes away, so does the thread.  This implies to me that we need *both* the thread **and** the loop from this function.

web3/providers/websocket.py
```diff
@@ -0,0 +1,67 @@
...
+
+def _start_event_loop(loop):
+    asyncio.set_event_loop(loop)
+    loop.run_forever()
+
+
+# TODO: Think of a better name
+def _get_threaded_loop():
+    new_loop = asyncio.new_event_loop()
+    t = Thread(target=_start_event_loop, args=(new_loop,))
```
pipermerriam: *nitpick*

We've generally avoided single letter variables.  Can you rename this to something like `thread` or `loop_thread`

web3/providers/websocket.py
```diff
@@ -0,0 +1,67 @@
...
+            self.endpoint_uri = get_default_endpoint()
+        if self._loop is None:
+            self._loop = _get_threaded_loop()
+        super().__init__()
+
+    def __str__(self):
+        return "WS connection {0}".format(self.endpoint_uri)
+
+    async def coro_make_request(self, request_data):
+        # TODO: Find out if there's a way of not establishing a connection every time
```
pipermerriam: I would look into how the `IPCProvider` does it with the [`PersistantSocket`](https://github.com/voith/web3.py/blob/0970ee3ee5fe1e1fb4fc75ab3521c3c0febefdd8/web3/providers/ipc.py#L33) object.

setup.py
voith: will do

web3/providers/websocket.py
voith: I had taken the same approach before submitting, However I couldn't manage to close the event_loop ad the thread successfully. I tried doing this in `__del__`  method.
I admit that I have limited knowledge about the life cycle of python objects and threads. I will give this another try. 
However, If you could give me some of guidance then It'd be very helfpful.

web3/providers/websocket.py
voith: will do

web3/providers/websocket.py
voith: will do

voith:
@pipermerriam thanks for the review!
I will try to submit a PR with tests separated for ipc and http either tonight or tomorrow. This will help me add tests for the `WebsocketProvider` easily.

voith:
depends on #710

setup.py
carver: :+1: core requirement seems fine here

web3/providers/websocket.py
```diff
@@ -0,0 +1,67 @@
...
+        async with websockets.client.connect(uri=self.endpoint_uri, loop=self._loop) as conn:
```
gsalgado: Since the coroutines scheduled here (or indirectly by `websockets`) will be in a different thread than the one running the event loop, so I believe that has to be done using `asyncio.run_coroutine_threadsafe()`, no? At least that's what https://docs.python.org/3/library/asyncio-dev.html#concurrency-and-multithreading states

boneyard93501:
@voith i think with a very minor tweak, you'll eliminate a few issues. specifically, you may want to consider running one event loop, one thread only. you could easily get there switching from instance to class attributes and add little housekeeping. e.g., for a [quick poc] (https://gist.github.com/boneyard93501/b4aaf612513fb07ed5c4664eab6b2dd2)

now, i have encountered problems shutting the loop down gracefully, as you'll see in the comments, which means you get a bunch of exceptions when the daemonized thread force closes the loop. it's a common, well-known issue; still, it's ugly. maybe @gsalgado can have a look at the async_ws_thread_closer function.

voith:
> consider running one event loop, one thread only. you could easily get there switching from instance to class attributes and add little housekeeping. e.g

@boneyard93501 Although having one event loop was what I intended to implement, It was a bug in my code. Switching to `ClassName` to self did the trick. I had seen this problem before and I thought implementing a Singleton Eventloop class would do the trick. However, you saved me a lot of trouble :)

>  now, i have encountered problems shutting the loop down gracefully, as you'll see in the comments, which means you get a bunch of exceptions when the daemonized thread force closes the loop. it's a common, well-known issue; still, it's ugly

I have already tried steps taken by you to shut it down gracefully. you can have a look at [code snippet](https://github.com/ethereum/web3.py/issues/566#issuecomment-375948105)

 > which means you get a bunch of exceptions when the daemonized thread force closes the loop

I agree and I know that its ugly but I couldn't find a better solution and atleast the user doesn't get to see the errors.

voith:
@pipermerriam @carver @gsalgado I have addressed all your comments and I've added a bare minimum version of `WebsocketProvider`. Let me know if more needs to be done. 

However, I won't be available until Monday in case more work needs to be done

tests/integration/go_ethereum/test_goethereum_http.py
```diff
@@ -41,16 +39,6 @@ def geth_command_arguments(geth_binary, datadir, rpc_port):
     )
 
 
-@pytest.fixture(scope='module')
```
voith: Unrelated to this PR. But this fixture wasn't needed here

voith:
Also tested:
```
In [10]: w = Web3(Web3.WebsocketProvider('wss://mainnet.infura.io/_ws'))

In [11]: w.eth.getBlock(0)
Out[11]: 
AttributeDict({'difficulty': 17179869184,
 'extraData': HexBytes('0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa'),
 'gasLimit': 5000,
 'gasUsed': 0,
 'hash': HexBytes('0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3'),
 'logsBloom': HexBytes('0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'),
 'miner': '0x0000000000000000000000000000000000000000',
 'mixHash': HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000'),
 'nonce': HexBytes('0x0000000000000042'),
 'number': 0,
 'parentHash': HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000'),
 'receiptsRoot': HexBytes('0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'),
 'sha3Uncles': HexBytes('0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347'),
 'size': 540,
 'stateRoot': HexBytes('0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544'),
 'timestamp': 0,
 'totalDifficulty': 17179869184,
 'transactions': [],
 'transactionsRoot': HexBytes('0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'),
 'uncles': []})
```

pipermerriam:
@carver leaving this for you to make final review/merge/whatever

boneyard93501:
@void4 
re graceful shutdown: i was rummaging around aiohttp and uvloop to see if i could find anything better. no dice. not saying it's not there just that i couldn't make it work.
re documentation: as i was looking through uvloop, i stumbled upon some multiprocessing issues which reminded me of a somewhat heated discussion a couple years ago. upshot is, with the threading design at hand, the GIL would spit the bit, i.e. deadlock, if one tried to use wsprovider in processes. hence, you probably should add a big "this won't work with multiprocessing" warning to the very top of the documentation. if you're bored, curious, or both and enjoy a good GIL debate, have a look at [this](https://bugs.python.org/issue27422)

voith:
@pipermerriam Thanks for clean up. I have pushed a commit to fix the linting issues. 
However, I have no idea why the tests are failing now. The failing tests seem to be completely unrelated. I tried rebasing with master, but tests are still failing.
cc @carver.

@boneyard93501 I will write some code to fully understand what you're trying to say and then I'll get back to you.

voith:
https://github.com/ethereum/py-evm/pull/483 might be cause for the failing tests, not sure though!

carver:
@voith can you upgrade this PR to eth-tester 0.1.0-beta.22? I believe that should fix at least one of the problems: The older eth-tester version had a bad py-evm dependency.

carver:
Ok, looks like everything is going to pass. The commit history is pretty messy right now. Are interested in squashing the fixups into a few cohesive commits? If not, I'm happy to just do a squash into a single commit.

voith:
@carver feel free to squash it into a single commit!

voith:
thanks everyone for helping me complete this.
I will submit work for the bounty as soon as gitcoin approves my request for faucet.

@boneyard93501 I'm sorry that I didn't address your comments but If anyone complains about the problems that you've forecasted, I'll resolve them myself!

PR #710, opened 2018-03-20
Refactored integration tests to support all providers

voith:
### What was wrong?

Integration tests use `IPCProvder` only  at the moment. 

### How was it fixed?

As discussed in https://github.com/ethereum/web3.py/pull/708#pullrequestreview-105086683, separate tests files have been created for `HTTPProvider` and `IPCProvider`.

TODO:

- [x] refactor `geth` integration tests to have similar support

- [x] Add a `wait_for_http` utility similar to `wait_for_socket`

- [x] Instead of using default `endpoint_uri` for `HTTPProvider`, start the http server at a different port.(better to be explicit rather than implicit). 

#### Cute Animal Picture

Cute animal picture


voith:
@dylanjw It'll be great if you could review this as I've touched some of your code.

I had two options to run these tests:
1. create separate tox environment for each test. something like `TOX_POSARGS="-e py36-integration-parity-ipc"` and `TOX_POSARGS="-e py36-integration-parity-http"`

OR

2. Add both the files to the same tox environment like I've done in this PR.
`integration-parity: py.test -s {posargs:tests/integration/parity/test_parity_ipc.py tests/integration/parity/test_parity_http.py}`. However, this needs changing scope of all tests to module.

I chose the 2nd approach because I didn't want extra circle/tox instances to be spawned for the same kind of tests, which would otherwise require additional time to run the test suite.

All the integration tests have a separate module and also run in separate processes(on different instances in this case). So is there any good reason why the tests were scoped for `session` and not `module`?

dylanjw:
> So is there any good reason why the tests were scoped for session and not module?

Scoping by module is good.

voith:
@carver @dylanjw I have addressed both your comments. This PR is now ready for review!

PR #724, opened 2018-03-27
Add eth-tester dependency for vanilla web3

carver:
### What was wrong?

web3py's `__init__.py` imports `EthereumTesterProvider`, which imports `eth_tester`, which isn't there in a vanilla install.

### How was it fixed?

Add `eth_tester` dependency to install (but not py-evm).

#### Cute Animal Picture

Cute animal picture


carver:
Abandoning for a new approach in #725

Issue #551, opened 2018-01-12
Filters that don't rely on server state


pipermerriam:
### What was wrong?

Filters suck.

- Sometimes the server loses the `filter_id`
- You don't really have any clarity to what is actually happening under the hood on the server.
- You can't use filters with infura
- They don't work if you're load balanced across multiple nodes.

### How can it be fixed?

I propose we implement a configurable *filter backend* engine.  For now, the default engine will retain the current functionality of using the node to handle filters.

However, we also implement a new local filter backend that looks something like the following.

```python
from cytoolz import (
    merge,
    sliding_window,
)


def log_filter(web3, filter_params, page_size=10):
    # TODO: handle pending/latest/earliest and cases where these are not included in the filter params.
    from_block = filter_params['fromBlock']
    to_block = filter_params['toBlock']

    # TODO: handle dynamic block ranges to enforce reasonable request sizes.
    for block_number in range(from_block, to_block + 1):
        params = merge(filter_params, {'fromBlock': block_number, 'toBlock': block_number})
        logs = web3.eth.getLogs(params)

        for page in sliding_window(page_size, logs):
            yield page
```

This `log_filter` generator can be used under the hood in place of using the node based filters to handle all filter logic locally, only relying on the node for `eth_getLogs` which is stateless.  Each time a new filter is created, the filter backend will create one of these generators for it and keep track of it internally.  Then, any calls to `eth_getFilterChanges` can just return the `next(...)` for the appropriate filter generator.  We *should* be able to do something similar with `eth_getFilterLogs` by re-generating the original filter generator and then returning all of the values up-to-the-current state of the generator.

pipermerriam:
I just realized this fits perfectly into the middleware framework.  No need for a new API, just a new middleware.

carver:
It's worth taking a look at the issues that this project has with filters in v3, and making sure that this is compatible. https://github.com/makerdao/pymaker/blob/7f2047b87d6982ce3cfadae07b8fad0026896202/pymaker/lifecycle.py

I *think* this proposal does a great job of addressing them all, but a thorough look before implementing this approach would be wise.

lpassos:
Any idea when this could get in?

pipermerriam:
I suspect it will be part of the next major release cycle but there's no firm timeline for when that will be.

dylanjw:
Im starting on this.

jasonrhaas:
[...]

PR #732, opened 2018-04-01
node independent filter middleware

dylanjw:
### What was wrong?
See #551 (tldr; node filters suck)

### How was it fixed?
Added a middleware that stores a "filtering generator" keyed by filter id.  Calls to `getFilterChanges` will retrieve the stored generator and retrieve logs via `eth_getLogs`.

Features implemented:
- [x] Limit the eth_getLogs request size by set number of blocks.
- [x] Works with dynamic block ranges
    - [x] toBlock = 'latest'
    - [ ] ~~toBlock = 'pending' (need to test)~~
- [x] `eth_newFilter`
- [x] `eth_getFilterChanges`
- [x] `eth_getFilterLogs`
- [x] block filters
- ~~[ ] pending transaction filters~~

TODO:
- [x] Clean up tests: split tests up; unittests to tests/core/middleware, functional/integration to tests/core/filtering
- [x] Write more tests.  In particular:
    - [x] Check for missing or overlapping blocks in the results.  (assumptions were made)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/middleware/test_filter_middleware.py
+++ b/tests/core/middleware/test_filter_middleware.py
@@ -1,0 +1,147 @@
+import pytest
+
+from web3 import Web3
+from web3.middleware import (
+    construct_result_generator_middleware,
+    local_filter_middleware,
+)
+from web3.middleware.filter import (
+    block_ranges,
+    iter_latest_block_ranges,
+)
+from web3.providers.base import (
+    BaseProvider,
+)
+
+
+class DummyProvider(BaseProvider):
+    def make_request(self, method, params):
+        raise NotImplementedError("Cannot make request for {0}:{1}".format(
+            method,
+            params,
+        ))
+
+
+@pytest.fixture(scope='function')
+def iter_block_number(start=0):
+    def iterator():
+        block_number = start
+        while True:
+            sent_value = (yield block_number)
+            if sent_value is not None:
+                block_number = sent_value
+    block_number = iterator()
+    next(block_number)
+    return block_number
+
+
+@pytest.fixture(scope='function')
+def result_generator_middleware(iter_block_number):
+    return construct_result_generator_middleware({
+        'eth_getLogs': lambda *_: ["middleware"],
+        'eth_getBlockByNumber': lambda *_: type('block', (object,), {'hash': 'middleware'}),
+        'net_version': lambda *_: 1,
+        'eth_blockNumber': lambda *_: next(iter_block_number),
+    })
+
+
+@pytest.fixture(scope='function')
+def w3_base():
+    return Web3(providers=[DummyProvider()], middlewares=[])
+
+
+@pytest.fixture(scope='function')
+def w3(w3_base, result_generator_middleware):
+    w3_base.middleware_stack.add(result_generator_middleware)
+    w3_base.middleware_stack.add(local_filter_middleware)
+    return w3_base
+
+
+@pytest.mark.parametrize("start, stop, expected", [
+    (0, 12, [
+        (0, 4),
+        (5, 9),
+        (10, 12)
+    ]),
+    (0, 15, [
+        (0, 4),
+        (5, 9),
+        (10, 14),
+        (15, 15)
+    ]),
+    (0, 0, [
+        (0, 0),
+    ]),
+    (1, 1, [
+        (1, 1),
+    ]),
+    (5, 0, TypeError),
+])
+def test_block_ranges(start, stop, expected):
+    if isinstance(expected, type) and issubclass(expected, Exception):
+        with pytest.raises(expected):
+            block_ranges(start, stop)
+    else:
+        for actual, expected in zip(block_ranges(start, stop), expected):
+            assert actual == expected
+
+
+@pytest.mark.parametrize("from_block,to_block,current_block,expected", [
+    (0, 10, [10], [
+        (0, 10),
+    ]),
+    (0, 55, [0, 19, 55], [
+        (0, 0),
+        (1, 19),
+        (20, 55),
+    ]),
+])
+def test_iter_latest_block_ranges(
+        w3,
+        iter_block_number,
+        from_block,
+        to_block,
+        current_block,
+        expected):
+    latest_block_ranges = iter_latest_block_ranges(w3, from_block, to_block)
+    for index, block in enumerate(current_block):
+        iter_block_number.send(block)
+        expected_tuple = expected[index]
+        try:
+            actual_tuple = next(latest_block_ranges)
+        except StopIteration as e:
+            if e.value:
+                actual_tuple = e.value
+
+        assert actual_tuple == expected_tuple
+
+
+def test_pending_block_filter_middleware(w3):
+    with pytest.raises(NotImplementedError):
+        w3.eth.filter('pending')
+
+
+def test_local_filter_middleware(w3, iter_block_number):
+    block_filter = w3.eth.filter('latest')
+    iter_block_number.send(1)
+
+    log_filter = w3.eth.filter(filter_params={'fromBlock': 'latest'})
+
+    assert w3.eth.getFilterChanges(block_filter.filter_id) == ["middleware"]
+
+    iter_block_number.send(2)
+    results = w3.eth.getFilterChanges(log_filter.filter_id)
+    assert results == ["middleware"]
+
+    assert w3.eth.getFilterLogs(log_filter.filter_id) == ["middleware"]
+
+    filter_ids = (
+        block_filter.filter_id,
+        log_filter.filter_id
+    )
+
+    # Test that all ids are str types
+    assert all(isinstance(_filter_id, (str,)) for _filter_id in filter_ids)
+
+    # Test that all ids are unique
+    assert len(filter_ids) == len(set(filter_ids))
--- a/tests/core/middleware/test_filter_middleware.py
+++ b/tests/core/middleware/test_filter_middleware.py
@@ -58,6 +58,10 @@ def w3(w3_base, result_generator_middleware):
 
 
 @pytest.mark.parametrize("start, stop, expected", [
+    (2, 7, [
+        (2, 6),
+        (7, 7)
+    ]),
     (0, 12, [
         (0, 4),
         (5, 9),
--- a/tests/core/middleware/test_filter_middleware.py
+++ b/tests/core/middleware/test_filter_middleware.py
@@ -86,7 +86,9 @@ def test_block_ranges(start, stop, expected):
         with pytest.raises(expected):
             block_ranges(start, stop)
     else:
-        for actual, expected in zip(block_ranges(start, stop), expected):
+        actual = tuple(block_ranges(start, stop))
+        assert len(actual) == len(expected)
+        for actual, expected in zip(actual, expected):
             assert actual == expected
 
 
@@ -111,12 +113,7 @@ def test_iter_latest_block_ranges(
     for index, block in enumerate(current_block):
         iter_block_number.send(block)
         expected_tuple = expected[index]
-        try:
-            actual_tuple = next(latest_block_ranges)
-        except StopIteration as e:
-            if e.value:
-                actual_tuple = e.value
-
+        actual_tuple = next(latest_block_ranges)
         assert actual_tuple == expected_tuple
 
 
--- a/web3/middleware/filter.py
+++ b/web3/middleware/filter.py
@@ -88,7 +88,7 @@ def block_ranges(start_block, last_block=None, step=5):
     )
 
 
-def iter_latest_block(web3, to_block=None):
+def iter_latest_block(w3, to_block=None):
     """Returns a generator that dispenses the latest block, if
     any new blocks have been mined since last iteration.
 
@@ -97,7 +97,7 @@ def iter_latest_block(web3, to_block=None):
     If ``to_block`` is defined, ``StopIteration`` is raised
     after to_block is reached.
 
-    >>> mined_blocks = dispense_mined_blocks(web3, 0, 10)
+    >>> mined_blocks = dispense_mined_blocks(w3, 0, 10)
     >>> next(new_blocks)  # Latest block = 0
     0
     >>> next(new_blocks)  # No new blocks
@@ -119,7 +119,7 @@ def iter_latest_block(web3, to_block=None):
     )
 
     while True:
-        latest_block = web3.eth.blockNumber
+        latest_block = w3.eth.blockNumber
         if is_bounded_range and latest_block > to_block:
             return
         #  No new blocks since last iteration.
@@ -130,7 +130,7 @@ def iter_latest_block(web3, to_block=None):
         _last = latest_block
 
 
-def iter_latest_block_ranges(web3, from_block, to_block=None):
+def iter_latest_block_ranges(w3, from_block, to_block=None):
     """Returns an iterator unloading ranges of available blocks
 
     starting from `fromBlock` to the latest mined block,
@@ -146,7 +146,7 @@ def iter_latest_block_ranges(web3, from_block, to_block=None):
     (46, 50)
     """
     _from_block = from_block
-    for _latest_block in iter_latest_block(web3, to_block):
+    for _latest_block in iter_latest_block(w3, to_block):
         if _latest_block is None:
             yield (None, None)
         else:
@@ -159,7 +159,7 @@ def drop_items_with_none_value(params):
 
 
 def get_logs_multipart(
-        web3,
+        w3,
         startBlock,
         stopBlock,
         address,
@@ -178,14 +178,14 @@ def get_logs_multipart(
             'address': address,
             'topics': topics
         }
-        yield web3.eth.getLogs(
+        yield w3.eth.getLogs(
             drop_items_with_none_value(params))
 
 
 class RequestLogs:
     def __init__(
             self,
-            web3,
+            w3,
             from_block=None,
             to_block=None,
             address=None,
@@ -193,9 +193,9 @@ class RequestLogs:
 
         self.address = address
         self.topics = topics
-        self.web3 = web3
+        self.w3 = w3
         if from_block is None or from_block == 'latest':
-            self._from_block = web3.eth.blockNumber + 1
+            self._from_block = w3.eth.blockNumber + 1
         else:
             self._from_block = from_block
         self._to_block = to_block
@@ -203,8 +203,8 @@ class RequestLogs:
 
     @property
     def from_block(self):
-        if self._from_block > self.web3.eth.blockNumber:
-            from_block = self.web3.eth.blockNumber
+        if self._from_block > self.w3.eth.blockNumber:
+            from_block = self.w3.eth.blockNumber
 
         else:
             from_block = self._from_block
@@ -214,13 +214,13 @@ class RequestLogs:
     @property
     def to_block(self):
         if self._to_block is None:
-            to_block = self.web3.eth.blockNumber
+            to_block = self.w3.eth.blockNumber
 
         elif self._to_block == 'latest':
-            to_block = self.web3.eth.blockNumber
+            to_block = self.w3.eth.blockNumber
 
-        elif self._to_block > self.web3.eth.blockNumber:
-            to_block = self.web3.eth.blockNumber
+        elif self._to_block > self.w3.eth.blockNumber:
+            to_block = self.w3.eth.blockNumber
 
         else:
             to_block = self._to_block
@@ -228,14 +228,14 @@ class RequestLogs:
         return to_block
 
     def _get_filter_changes(self):
-        for start, stop in iter_latest_block_ranges(self.web3, self._from_block, self._to_block):
+        for start, stop in iter_latest_block_ranges(self.w3, self._from_block, self._to_block):
             if None in (start, stop):
                 yield []
 
             yield list(
                 concat(
                     get_logs_multipart(
-                        self.web3,
+                        self.w3,
                         start,
                         stop,
                         self.address,
@@ -246,7 +246,7 @@ class RequestLogs:
         return list(
             concat(
                 get_logs_multipart(
-                    self.web3,
+                    self.w3,
                     self.from_block,
                     self.to_block,
                     self.address,
@@ -269,9 +269,9 @@ FILTER_CHANGES_METHODS = set([
 
 
 class RequestBlocks:
-    def __init__(self, web3):
-        self.web3 = web3
-        self.start_block = web3.eth.blockNumber + 1
+    def __init__(self, w3):
+        self.w3 = w3
+        self.start_block = w3.eth.blockNumber + 1
 
     @property
     def filter_changes(self):
@@ -280,22 +280,22 @@ class RequestBlocks:
     def get_filter_changes(self):
 
         block_range_iter = iter_latest_block_ranges(
-            self.web3,
+            self.w3,
             self.start_block,
             None)
 
         for block_range in block_range_iter:
-            yield(block_hashes_in_range(self.web3, block_range))
+            yield(block_hashes_in_range(self.w3, block_range))
 
 
 @to_list
-def block_hashes_in_range(web3, block_range):
+def block_hashes_in_range(w3, block_range):
     from_block, to_block = block_range
     for block_number in range(from_block, to_block + 1):
-        yield getattr(web3.eth.getBlock(block_number), 'hash', None)
+        yield getattr(w3.eth.getBlock(block_number), 'hash', None)
 
 
-def local_filter_middleware(make_request, web3):
+def local_filter_middleware(make_request, w3):
     filters = {}
     filter_id_counter = map(to_hex, itertools.count())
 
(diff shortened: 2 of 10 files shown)


web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+
+    while toBlock is None or toBlock > current_block:
+        # return blocks up to current block
+        yield _fromBlock, current_block
+        _fromBlock = current_block + 1
+        current_block = int(web3.eth.blockNumber)
+        # no new blocks
+        if current_block < _fromBlock:
+            yield (None, None)
+        # reach stop_block after being passed by current_block
```
pipermerriam: There is no `stop_block` in the context here.  Is this comment outdated?

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+        for _start in self._count_from:
+            if self.stop and _start > self.stop:
+                raise StopIteration
+            _end = _start + (self.step - 1)
+            if self.stop and self.stop < _end:
+                _end = self.stop
+            return _start, _end
+
+
+def unload_latest_blocks(web3, fromBlock, toBlock):
```
pipermerriam: We've maintained using `fromBlock` instead of `from_block` in most public APIs, but for internal APIs I'd prefer we stick with `from_block` style casing.

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def log_filter(
+        web3,
+        fromBlock=None,
+        toBlock=None,
+        address=None,
+        topics=None,):
+
+    if not fromBlock or fromBlock == 'latest':
```
pipermerriam: `not fromBlock` will result in `fromBlock == 0` not iterating from the genesis block as expected, but rather just starting at the current block which looks like a bug.

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+        _fromBlock = int(web3.eth.blockNumber)
+    else:
+        _fromBlock = fromBlock
+
+    if _fromBlock > int(web3.eth.blockNumber):
+        raise ValueError("fromBlock ({0}) has not yet been mined".format(fromBlock))
```
pipermerriam: I'm not sure this needs to be an error case.  Why should we forbid someone setting up a filter in advance?  If you want to punt on implementation can you just change this to a `NotImplementedError` and change the error messaging to indicate that this just hasn't been implemented yet.

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+            start,
+            stop,
+            address,
+            topics,
+            chunk_size=5))
+
+
+def filter_middleware(make_request, web3):
+    filters = {}
+    filter_id_counter = map(to_hex, itertools.count())
```
pipermerriam: oooooh, fancy! infinite map!

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def middleware(method, params):
+        if method in (
```
pipermerriam: It'd be good to both use a `set` in places where you are only doing membership checking **and** to elevate this to a module level constant so we don't have to allocate it every time the filter is called.

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+                'eth_newFilter',):
+
+            filter_id = next(filter_id_counter)
+
+            if method == 'eth_newFilter':
+                _filter = log_filter(web3, **params[0])
+
+            filters[filter_id] = _filter
+            return {'result': filter_id}
+        elif method == 'eth_getFilterChanges' or method == 'eth_getFilterLogs':
```
pipermerriam: It looks like `eth_getFilterLogs` is supposed to have it's own separate implementation? (based on your PR checklist and the differences in their respective functionality)

web3/middleware/filter.py
dylanjw: I can implement future block ranges.

web3/middleware/filter.py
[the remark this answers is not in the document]
dylanjw: I'll add a docstring.  I was having trouble naming this function.

It returns block ranges starting from `fromBlock` to the latest mined block, until reaching `toBlock`. e.g.:

>>> blocks_to_filter = unload_latest_blocks(w3, 0, 50)
>>> next(blocks_to_filter)  # latest block number = 11
(0, 11)
>>> next(blocks_to_filter)  # latest block number = 45
(12, 45)
>>> next(blocks_to_filter)  # latest block number = 50
(46, 50)

web3/middleware/filter.py
dylanjw: yep

web3/middleware/filter.py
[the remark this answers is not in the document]
pipermerriam: - `get_block_ranges`
- `get_latest_block_ranges` 
- `iter_block_rangest`
- `iter_latest_block_ranges`

Don't particularly love any of those.  The `iter_` prefixed ones seems a bit more descriptive/semantically correct.

web3/middleware/filter.py
```diff
@@ -0,0 +1,131 @@
...
+            _end = _start + (self.step - 1)
+            if self.stop and self.stop < _end:
+                _end = self.stop
+            return _start, _end
+
+
+def unload_latest_blocks(web3, fromBlock, toBlock):
+    _fromBlock = fromBlock
+
+    current_block = int(web3.eth.blockNumber)
```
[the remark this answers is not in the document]
dylanjw: No.  I misread an error and added this and didnt remember to take them out.

dylanjw:
Im starting to suspect that a pending transaction filter is not possible.  Some of the reasons:

- I think it would involve polling the transaction pool, but that carries the risk of missing transactions.
- Infura does not yet have any tx pool apis.
- API is node dependent (not an impossible problem).

Maybe we could add a recipe to the docs for an alternate way from node filters to get information on pending transactions, in lieu of having the filtering middleware handle it.

pipermerriam:
@dylanjw re pending transactions.

1. This middleware doesn't need to be functional for anything but log filtering so feel free to drop block and pending transaction filters from scope.
2. Pending transaction filters provide no guarantees that you'll see all of them anyways, a node can only provide info on pending transactions they know about and there will be plenty that they never learn about prior to being mined.
3. since there's no guarantees around what pending transactions will be visible to the node and thus exposed in filters, I think it's *ok* for us to do the following.

- A middleware specific to the transaction pool APIs which will only work if they are available (and fails hard if the endpoints are not available).
- A middleware that only monitors for `eth_sendTransaction` and `eth_sendRawTransaction` and only returns transaction hashes that traverse those APIS

I still think I'd recommend separating this functionality from this PR and doing it as an independent PR to try and limit the scope of this feature.

carver:
If you rebase this on master, some of these errors will be fixed, because master is upgraded to eth-tester devnum 22 @dylanjw

dylanjw:
@pipermerriam I implemented the block filter in the PR because it was easy to do; most of the work is done with the `iter_latest_block_ranges()`.  Is it ok to leave it here?

web3/middleware/filter.py
```diff
@@ -0,0 +1,240 @@
...
+)
+
+
+def range_counter(start, stop=None, step=5):
+    if stop is not None and start > stop:
+        raise TypeError(
+            "Incompatible start and stop arguments.",
+            "Start must be less than or equal to stop.")
+
+    def range_counter(start, stop, step):
```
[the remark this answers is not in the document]
dylanjw: The wrapper is so the exception is raised when the generator is created.  e.g.  

without the wrapper:

```
>>> s = segment_count(5, 0)
>>> next(s)
  File "web3/middleware/filter.py", line 45, in segment_count
    "Start must be less than or equal to stop.")
TypeError: ('Incompatible start and stop arguments.', 'Start must be less than or equal to stop.')
>>> 
```

With the wrapper:

```
>>> s = segment_count(5, 0)
  File "web3/middleware/filter.py", line 45, in segment_count
    "Start must be less than or equal to stop.")
TypeError: ('Incompatible start and stop arguments.', 'Start must be less than or equal to stop.')
>>> 
```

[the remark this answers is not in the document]
pipermerriam: Interesting, I wasn't aware that generators were lazy in that way.  Neat and also kind of awkward.  This deserves a comment imho as it's not clear just by reading the code.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+    (3, 6)
+    >>> next(segment_counter)
+    (6, 9)
+    >>> next(segment_counter) #  Remainder is also returned
+    (9, 10)
+    """
+    def finite(start, stop, step):
+        #  If the initial range is less than the step
+        #  just return (start, stop)
+        if stop < step:
```
pipermerriam: Should this be `if start + step >= stop`?

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+            yield (start, stop)
+            return
+        for i in zip(
```
pipermerriam: *naming nitpick*

```
for segment in zip(...):
    ...
```

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+                range(start + step, stop, step)):
+            yield i
+        else:
+            remainder = (stop - start) % step
+            #  Handle the remainder
+            if remainder:
+                yield (stop - remainder, stop)
+
+    def infinite(start, step):
+        for i in zip(
```
pipermerriam: same *naming nitpick* `s/i/segment`

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+                range(start, stop - step, step),
+                range(start + step, stop, step)):
+            yield i
+        else:
+            remainder = (stop - start) % step
+            #  Handle the remainder
+            if remainder:
+                yield (stop - remainder, stop)
+
+    def infinite(start, step):
```
pipermerriam: This could be elevated to a module level function.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+    >>> next(segment_counter)
+    (0, 3)
+    >>> next(segment_counter)
+    (3, 6)
+    >>> next(segment_counter)
+    (6, 9)
+    >>> next(segment_counter) #  Remainder is also returned
+    (9, 10)
+    """
+    def finite(start, stop, step):
```
pipermerriam: This could be elevated to a module level function.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+                count(start + step, step)):
+            yield i
+
+    if stop is None:
+        return infinite(start, step)
+    else:
+        return finite(start, stop, step)
+
+
+def block_range_chunker(start_block, last_block=None, step=5):
```
pipermerriam: Maybe use the *segment* name here too:  `get_segmented_block_range`.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+       Segments do not overlap to facilitate use as ``toBlock``, ``fromBlock``
+       json-rpc arguments, which are both inclusive.
+    """
+
+    if last_block is not None and start_block > last_block:
+        raise TypeError(
+            "Incompatible start and stop arguments.",
+            "Start must be less than or equal to stop.")
+
+    def block_range_chunker(start_block, last_block, step):
```
pipermerriam: Generally good to not have this have the same name as the outer function (just for naming clarity).  Also,  I think you can accomplish this without the inner function using:

```
def outer_iter():
    if thing:
        raise Exception(...)
    return (value for value in iter_something())
```

This will hit the validation step when called (as opposed to when `next` is first called.  Do you think readability seems better with that pattern?

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+    return block_range_chunker(start_block, last_block, step)
+
+
+def iter_latest_block(web3, to_block=None):
+    """Returns a generator that dispenses the latest block, if
+    any new blocks have been mined since last iteration.
+
+    If there are no new blocks None is returned.
+
+    If ``to_block`` is defined, ``StopIteration`` is raised
```
pipermerriam: This no longer appears to be accurate (and `StopIteration` seems to have been deprecated in favor of `return` with iterators.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+    >>> next(new_blocks)
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+      StopIteration
+    >>>
+    """
+    _last = None
+    while True:
+        latest_block = web3.eth.blockNumber
+        if (latest_block is not None and
```
pipermerriam: I'm a big fan of avoiding multiline `if` statements in favor of a local variable.

```
is_condition_true = (
    something is not None and
    other_thing is None and
    this_is_a_long_statement()
)
if is_conditional_true:
    ...
```

I feel like the readability is better.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+        #  No new blocks since last iteration.
+        if _last is not None and _last == latest_block:
+            yield None
+        else:
+            yield latest_block
+        _last = latest_block
+
+
+def count(start=0, step=1):
+    """A counter you can update with send()
```
pipermerriam: This seems like a code example in the docstring would be very informative because even reading this it's not entirely clear how I would go about using it.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def get_logs_multipart(
+        web3,
+        startBlock,
+        stopBlock,
+        address,
+        topics,
+        max_blocks):
+    """Use to make large requests to ``eth_getLogs``
```
pipermerriam: *language nitpick*

> Used to split up requests to `eth_getLogs` into smaller segments to ensure returned results are not unreasonably large.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+            self,
+            web3,
+            from_block=None,
+            to_block=None,
+            address=None,
+            topics=None):
+
+        self.address = address
+        self.topics = topics
+        self.web3 = web3
```
pipermerriam: We've been using `w3` as the variable name for a `Web3` instance.  Maybe switch to use this convention.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+        else:
+            self._from_block = from_block
+        self._to_block = to_block
+        self.filter_changes = self._get_filter_changes()
+
+    @property
+    def from_block(self):
+        if self._from_block > self.web3.eth.blockNumber:
+            from_block = self.web3.eth.blockNumber
+
```
pipermerriam: Any reason for the extra whitespace?  Visually confusing to me but mostly because I'm not used to it.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+            yield list(
+                concat(
+                    get_logs_multipart(
+                        self.web3,
+                        start,
+                        stop,
+                        self.address,
+                        self.topics,
+                        max_blocks=5)))
```
pipermerriam: This value should probably be configurable.  Think about requesting logs over the span of 10,000 blocks, that's 2,000 API requests.  I'd at least up the default to around 100.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+            None)
+
+        for block_range in block_range_iter:
+            yield(block_hashes_in_range(self.web3, block_range))
+
+
+@to_list
+def block_hashes_in_range(web3, block_range):
+    for block_number in range(block_range[0], block_range[1] + 1):
+        yield web3.eth.getBlock(block_number).hash
```
pipermerriam: I *think* that `web3.eth.getBlock` returns `None` if `block_number` is greater than the current HEAD block number, which means this will raise an `AttributeError` if that case is met.

web3/middleware/filter.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def middleware(method, params):
+        if method in NEW_FILTER_METHODS:
+
+            filter_id = next(filter_id_counter)
+
+            if method == 'eth_newFilter':
+                _filter = RequestLogs(web3, **apply_key_map(FILTER_PARAMS_KEY_MAP, params[0]))
+
+            if method == 'eth_newBlockFilter':
```
pipermerriam: `s/if/elif` and probably good to add an `else` that raises `NotImplimented`.

tests/core/middleware/test_filter_middleware.py
```diff
@@ -0,0 +1,136 @@
...
+
+class DummyProvider(BaseProvider):
+    def make_request(self, method, params):
+        raise NotImplementedError("Cannot make request for {0}:{1}".format(
+            method,
+            params,
+        ))
+
+
+@pytest.fixture(scope='function')
```
pipermerriam: Not highly opposed to this but `"function"` is the default `scope` for `pytest.fixture`

tests/core/middleware/test_filter_middleware.py
```diff
@@ -0,0 +1,136 @@
...
+        assert actual_tuple == expected_tuple
+
+
+def test_local_filter_middleware(w3, iter_block_number):
+    block_filter = w3.eth.filter('latest')
+    assert block_filter.filter_id == '0x0'
+    iter_block_number.send(1)
+
+    with pytest.raises(NotImplementedError):
+        w3.eth.filter('pending')
```
pipermerriam: Maybe move this to it's own test?

tests/core/middleware/test_filter_middleware.py
```diff
@@ -0,0 +1,136 @@
...
+        except StopIteration as e:
+            if e.value:
+                actual_tuple = e.value
+
+        assert actual_tuple == expected_tuple
+
+
+def test_local_filter_middleware(w3, iter_block_number):
+    block_filter = w3.eth.filter('latest')
+    assert block_filter.filter_id == '0x0'
```
pipermerriam: Maybe instead of asserting the exact `id` since this is somewhat testing the implementation, we could have a test that spams a number of different filter creations and asserts that the set of `filter_id` values that are returned are all strings and are unique.

web3/middleware/filter.py
dylanjw: That is true, I'll handle that case.

web3/middleware/filter.py
dylanjw: Im using "segment" to denote this kind of output (with 1-length segments): (0, 1), (1, 2), (2, 3), (3, 4), etc.

Then there is the fromBlock -> toBlock type ranges, which the equivalent of above would be: (0, 0), (1, 1), (2, 2), etc.

Because the fromBlock/toBlock type ranges seem a little less conventional, I decided to make the basic function output the more conventional "segment" ranges (where given a range (a, b): b - a = step), and making them conform to the fromBlock/toBlock expected values as a separate step when iterating over the segments.

Chunker is confusing, because it is the same concept as segmenting.  When Im dealing with the block type ranges, I'll call them, "block ranges".

dylanjw:
> We could potentially fuzz test this against an eth-tester backed filter via two separate web3 instances (one eth-tester and one middleware backed) to verify that they return the correct log entries.

@pipermerriam I added parametrized web3 fixture to tests/core/filtering/conftest.py that runs all the normal filtering tests against the middleware filter and the node filter.  Is that what you mean?  Or do you want tests with automated inputs?  Im not too experienced with fuzz testing.

pipermerriam:
> I added a parametrized web3 fixture to tests/core/filtering/conftest.py

Looking at the tests that live under `tests/core/filtering` I'm not sure we're testing the things I'd like to test.  The tests in there look to be more *smoke* tests that things that really provide coverage for the filter APIs.

Ideally, I think we need a test or set of tests that cover the following.

- Filtering across a range of ~20 or more blocks.
- Multi-address, Multi-topic (with combinations of *and* and *or* topic combinations).
- Variants that are *open* ended on the `toBlock` side to see that it correctly handles new block mining.

dylanjw:
> Check for missing or overlapping blocks in the results. (assumptions were made)

Im getting a heisenbug, where in some cases all tests pass, and in others there are missing log results.  Im guessing this could be due to skipping blocks in the ranged block getLogs calls.

dylanjw:
I added a test that emits an event to 100 blocks in a row, and show that the a "local" filter finds everyone.  Doesnt add up with my theory that there are breaks in the eth_getLogs block ranges.

dylanjw:
Not ready to merge.  Test results are still flaky.

dylanjw:
Function scoped pytest fixtures are not respected by Hypothesis, so the failures have to do with logs in the overlapping blocks between runs, where the chain state is not reset.  Im surprised this hasnt been a bigger issue with flaky tests.

carver:
Wow, I can confirm that only `hypothesis` fails here, reusing integer 2:

```py
import pytest
import itertools

from hypothesis import given, strategies as st

counter = itertools.count()
hypothesis_already_seen = set()
parametrized_already_seen = set()


@pytest.fixture
def next_int():
    return next(counter)


@pytest.mark.parametrize('ignored_text', ('', 'a'), )
def test_parametrize_and_fixtures(next_int, ignored_text):
    assert next_int not in parametrized_already_seen
    parametrized_already_seen.add(next_int)


@given(st.text())
def test_hypothesis_and_fixtures(next_int, ignored_text):
    assert next_int not in hypothesis_already_seen
    hypothesis_already_seen.add(next_int)
```

I would consider this a hypothesis bug.

carver:
Nevermind, it's already there, and old... :cry: https://github.com/HypothesisWorks/hypothesis/issues/377

carver:
There is a workaround, so at least you can check locally if the scoping issue is the *only* problem you're seeing: https://github.com/untitaker/pytest-subtesthack/

dylanjw:
I gotten around the issue by taking a snapshot and restoring at the beginning of the test function.  The issue with hypothesis actually uncovered a bug, where the local filters were picking up events from the last block of the previous hypothesis run.

dylanjw:
Im not getting flaky test runs anymore.

pipermerriam:
@carver assigned to you, since this was a concept I came up with I think it's valuable for you to do primary review (cause bias...), I can do another pass when you are done if you think it will help.

web3/middleware/filter.py
```diff
@@ -0,0 +1,326 @@
...
+
+    if last_block is not None and start_block > last_block:
+        raise TypeError(
+            "Incompatible start and stop arguments.",
+            "Start must be less than or equal to stop.")
+
+    return (
+        (from_block, to_block - 1)
+        for from_block, to_block
+        in segment_count(start_block, last_block + 1, step)
```
carver: The default value `last_block` could be `None` which blows up here.

Looking at the invocation, it looks like it's always defined, and never `None`. So maybe just drop the `None` default value in the signature, and you can drop the `last_block is not None` check above.

web3/middleware/filter.py
```diff
@@ -0,0 +1,326 @@
...
+
+    @property
+    def to_block(self):
+        if self._to_block is None:
+            to_block = self.w3.eth.blockNumber
+
+        elif self._to_block == 'latest':
+            to_block = self.w3.eth.blockNumber
+
+        else:
```
carver: *nit* I'm not generally a fan of this gap before the else or elif statement. Up to you, though.

web3/middleware/filter.py
```diff
@@ -0,0 +1,326 @@
...
+    def get_logs(self):
+        return list(
+            concat(
+                get_logs_multipart(
+                    self.w3,
+                    self.from_block,
+                    self.to_block,
+                    self.address,
+                    self.topics,
+                    max_blocks=50)))
```
carver: Probably belongs in a constant with the one from line 235 also.

web3/middleware/filter.py
```diff
@@ -59,8 +59,8 @@ def gen_bounded_segments(start, stop, step):
         yield (start, stop)
         return
     for segment in zip(
-            range(start, stop - step, step),
-            range(start + step, stop, step)):
+            range(start, stop - step + 1, step),
+            range(start + step, stop + 1, step)):
         yield segment
     else:
```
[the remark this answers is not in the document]
dylanjw: Its a for/else loop, so it runs at the end of the for loop.  It would work the same to remove the else clause, but probably safer to keep it in case anything is added to this method.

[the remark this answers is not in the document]
pipermerriam: `for/else` is typically reserved for loops that have a `break` somewhere in them, in order to *only* run the `else` in whatever case the `break` does not get triggered (or inversely, having the ability to `break`, thus skipping the `else` code.

tests/core/middleware/test_filter_middleware.py
```diff
@@ -0,0 +1,151 @@
...
+        (1, 1),
+    ]),
+    (5, 0, TypeError),
+])
+def test_block_ranges(start, stop, expected):
+    if isinstance(expected, type) and issubclass(expected, Exception):
+        with pytest.raises(expected):
+            block_ranges(start, stop)
+    else:
+        for actual, expected in zip(block_ranges(start, stop), expected):
```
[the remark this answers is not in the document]
pipermerriam: pre-check on length gives an early exit and is probably preferrable to complicating the comparison loop with an abnormal exit condition which is that implicitely, nothing will evaluate equal to `missing_flag`.

PR #734, opened 2018-04-02
First draft of v4 migration docs

carver:
### What was wrong?

There was no guidance on how to migrate from v3 to v4 stable.

### How was it fixed?

Add a migration page to the docs.

#### Cute Animal Picture

Cute animal picture


carver:
These are pretty weak, but better than nothing. @dylanjw would you mind taking a look and seeing if you can do a better job? :)

(Don't worry about the test failures, that seems to be an eth-tester issue that I'm looking into now)

carver:
> I'm not sure all of the following are accurate but I think they are...

Mostly, with some nitpicks:

> string abi types now require text strings

Not quite, `bytes` is accepted, but you must be able to decode them using utf-8.

> bytes abi types now require byte strings

`str` is accepted, but it must be a hex-encoded string.

> string abi return values are now returned as utf8 encoded values

*nittiest of picks* -- now returned as a `str`, "utf8 **decoded**"

> API for toHex, toText, toBytes all changed to use the new primative/text/hexstr argument pattern.

The [new keyword arguments _were_ added into v3](http://web3py.readthedocs.io/en/v3.16.5/overview.html#Web3.toHex), but still accepted `str` in the `primitive` position, for backward compatibility. The main change here is that the first argument no longer accepts `str` in v4.

dylanjw:
> backward compatibility. The main change here is that the first argument no longer accepts str in v4.

seems like this is still allowed:

```
>>> w3 = web3.Web3(web3.EthereumTesterProvider())
>>> w3.version.api
'4.0.0'
>>> w3.toHex("cowmö")
'0x636f776dc3b6'
>>> 
```

dylanjw:
I rebased this with master, fyi.

carver:
>> The main change here is that the first argument no longer accepts str in v4.

> seems like this is still allowed:

Thanks for pointing that out! `to_bytes` and `sha3` don't accept text in the first position. But this seems presumptuous: https://github.com/ethereum/eth-utils/blob/master/eth_utils/conversions.py#L41-L42

I opened https://github.com/ethereum/eth-utils/issues/85

PR #738, opened 2018-04-04
Update ImplicitContract

fubuloubu:
### What was wrong?
ConciseContract changed how it was operating significantly, ImplicitContract was inheriting from it.
I use ImplicitContract for a testing tool so that users don't execute a state-changing function as a call by accident.

### How was it fixed?
Updated ConciseContract to use a class variable when it creates a Method handler object
Updated ImplicitContract so it just changes this class variable, removing all custom code.
NOTE: Had to switch the order of the *Method and *Contract classes for that to work.

#### Cute Animal Picture

Who, me?

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -766,6 +766,30 @@ class ContractConstructor:
             raise ValueError("Cannot set {} in transaction".format(', '.join(keys_found)))
 
 
+class ConciseMethod:
+    ALLOWED_MODIFIERS = set(['call', 'estimateGas', 'transact', 'buildTransaction'])
+
+    def __init__(self, function, normalizers=None):
+        self._function = function
+        self._function._return_data_normalizers = normalizers
+
+    def __call__(self, *args, **kwargs):
+        return self.__prepared_function(*args, **kwargs)
+
+    def __prepared_function(self, *args, **kwargs):
+        if not kwargs:
+            modifier, modifier_dict = 'call', {}
+        elif len(kwargs) == 1:
+            modifier, modifier_dict = kwargs.popitem()
+            if modifier not in self.ALLOWED_MODIFIERS:
+                raise TypeError(
+                    "The only allowed keyword arguments are: %s" % self.ALLOWED_MODIFIERS)
+        else:
+            raise TypeError("Use up to one keyword argument, one of: %s" % self.ALLOWED_MODIFIERS)
+
+        return getattr(self._function(*args), modifier)(modifier_dict)
+
+
 class ConciseContract:
     '''
     An alternative Contract Factory which invokes all methods as `call()`,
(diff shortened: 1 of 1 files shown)


web3/contract.py
```diff
@@ -766,6 +766,30 @@ def check_forbidden_keys_in_transaction(transaction, forbidden_keys=None):
             raise ValueError("Cannot set {} in transaction".format(', '.join(keys_found)))
 
 
+class ConciseMethod:
+    ALLOWED_MODIFIERS = set(['call', 'estimateGas', 'transact', 'buildTransaction'])
```
carver: Since we're touching this code anyway, how about:

```py
ALLOWED_MODIFIERS = {'call', 'estimateGas', 'transact', 'buildTransaction'}
```

web3/contract.py
```diff
@@ -779,6 +803,8 @@ class ConciseContract:
 
     > contract.functions.withdraw(amount).transact({'from': eth.accounts[1], 'gas': 100000, ...})
     '''
+    method_class = ConciseMethod
```
carver: A key point of the refactor was to reduce/remove name conflicts. This adds another potential conflict. But I think we can get to your goal in almost the same way. (see following comments for pseudocode)

web3/contract.py
```diff
@@ -779,6 +803,8 @@ class ConciseContract:
...
+
     def __init__(self, classic_contract):
```
carver: ```def __init__(self, classic_contract, method_class=ConciseMethod):```

web3/contract.py
```diff
@@ -798,7 +824,7 @@ def __init__(self, classic_contract):
                     self._classic_contract.functions,
                     fn_name)
 
-                _concise_method = ConciseMethod(
+                _concise_method = self.method_class(
```
carver:
```
                _concise_method = method_class(
```

web3/contract.py
```diff
@@ -862,25 +880,7 @@ class ImplicitContract(ConciseContract):
...
-
-        return function_abi['constant'] if 'constant' in function_abi.keys() else False
-
-    def __call__(self, *args, **kwargs):
-        # Modifier is not provided and method is not constant/pure do a transaction instead
-        if not kwargs and not self.__call_by_default(args):
-            return super().__call__(*args, transact={})
-        else:
-            return super().__call__(*args, **kwargs)
+    method_class = ImplicitMethod
```
carver:
```py
def __init__(self, classic_contract, method_class=ImplicitMethod):
    super().__init__(self, classic_contract, method_class=method_class)
```

web3/contract.py
fubuloubu: :+1: did not know about this syntax

web3/contract.py
fubuloubu: I was thinking about that way too, this way was just easier so I figured I'd give it a shot

fubuloubu:
The TravisCI failure was that something didn't execute when it was expected to. Nothing to do with the changes.

pipermerriam:
I restarted the failed travis build.

PR #743, opened 2018-04-04
Set transaction['to'] to empty when building a transaction.

### LINKED ISSUE #739: How to connect to Ropston via infura and deploy contract use web3.py？
* Version: 4.0.0
* Python: 3.4
* OS: win

### What was wrong?
How to connect to Ropston and deploy contract use web3.py？
I have deploy a contract to Ropston use remix and MetaMask.
I want know how to  deploy a contract to Ropston via infura using web3.py **without geth or parity.**
I read the Documention and find there is no how-to chapter.
A small demo code will be very helpful.

this is my code
and it can't work at all.
````python
from solc import compile_source
from web3.contract import ConciseContract

privateKey = "xxxxx"
# Solidity source code
contract_source_code = '''
pragma solidity ^0.4.0;

contract Greeter {
    string public greeting;

    function Greeter() {
        greeting = 'Hello';
    }

    function setGreeting(string _greeting) public {
        greeting = _greeting;
    }

    function greet() constant returns (string) {
        return greeting;
    }
}
'''

compiled_sol = compile_source(contract_source_code) 
contract_interface = compiled_sol['<stdin>:Greeter']
w3 = Web3(HTTPProvider("https://ropsten.infura.io/"))
w3.eth.enable_unaudited_features()
account = Account()
acct = account.privateKeyToAccount(privateKey)  # !!!!重要
contract_ = w3.eth.contract(abi=contract_interface['abi'], bytecode=contract_interface['bin'])
nonce = w3.eth.getTransactionCount(acct.address)
contract_data = contract_.buildTransaction({'from': acct.address,
                                                                     'gas': 1728712,
                                                                     'gasPrice': w3.toWei('1', 'gwei'),
                                                                     'nonce': nonce,})
signed = w3.eth.account.signTransaction(contract_data, privateKey)
w3.eth.sendRawTransaction(signed.rawTransaction)
````

### How can it be fixed?

I don't know

### PR DESCRIPTION
**Pull request by zheli**
### What was wrong?
Cannot use result from `buildTransaction()` for `signTransaction()` because the key `to` is missing.

### How was it fixed?
Set built_transaction['to'] to b''. This fixes #739 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/test_contract_constructor.py
+++ b/tests/core/contracts/test_contract_constructor.py
@@ -143,9 +143,9 @@ def test_contract_constructor_build_transaction_to_field_error(MathContract):
 
 
 def test_contract_constructor_build_transaction_no_constructor(web3, MathContract):
-    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     txn_hash = MathContract.constructor().transact({'from': web3.eth.accounts[0]})
     txn = web3.eth.getTransaction(txn_hash)
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     unsent_txn = MathContract.constructor().buildTransaction({'nonce': nonce})
     assert txn['data'] == unsent_txn['data']
 
@@ -157,9 +157,9 @@ def test_contract_constructor_build_transaction_no_constructor(web3, MathContrac
 
 def test_contract_constructor_build_transaction_with_constructor_without_argument(web3,
                                                                                   MathContract):
-    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     txn_hash = MathContract.constructor().transact({'from': web3.eth.accounts[0]})
     txn = web3.eth.getTransaction(txn_hash)
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     unsent_txn = MathContract.constructor().buildTransaction({'nonce': nonce})
     assert txn['data'] == unsent_txn['data']
 
@@ -183,10 +183,10 @@ def test_contract_constructor_build_transaction_with_constructor_with_argument(
         WithConstructorArgumentsContract,
         constructor_args,
         constructor_kwargs):
-    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     txn_hash = WithConstructorArgumentsContract.constructor(
         *constructor_args, **constructor_kwargs).transact({'from': web3.eth.accounts[0]})
     txn = web3.eth.getTransaction(txn_hash)
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase)
     unsent_txn = WithConstructorArgumentsContract.constructor(
         *constructor_args, **constructor_kwargs).buildTransaction({'nonce': nonce})
     assert txn['data'] == unsent_txn['data']
--- a/web3/middleware/pythonic.py
+++ b/web3/middleware/pythonic.py
@@ -3,7 +3,6 @@ import operator
 
 from cytoolz import (
     curry,
-    dissoc,
 )
 from cytoolz.curried import (
     keymap,
@@ -43,6 +42,7 @@ from web3.utils.formatters import (
     integer_to_hex,
     is_array_of_dicts,
     is_array_of_strings,
+    remove_key_if,
 )
 
 from .formatting import (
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by pipermerriam** [comment]
Can you also verify that passing in `nonce` to the `buildTransaction(..)` call works.

**Comment by carver** [comment]
@zheli I was looking at this as well. Here's an approach for fixing the test failures: remove the `to` key from transactions, if it is a "valid" way to specify an empty to field: `to in {'', b'', None}`.

Specifically, in `web3/middleware/pythonic.py`:
```
-transaction_param_formatter = apply_formatters_to_dict(TRANSACTION_PARAM_FORMATTERS)
+
+@curry
+def remove_key_if(key, remove_if, input_dict):
+    if key in input_dict and remove_if(input_dict):
+        return dissoc(input_dict, key)
+    else:
+        return input_dict
+
+
+transaction_param_formatter = compose(
+    remove_key_if('to', lambda txn: txn['to'] in {'', b'', None}),
+    apply_formatters_to_dict(TRANSACTION_PARAM_FORMATTERS),
+)
```

**Comment by zheli** [comment]
@pipermerriam @carver I got an error from `validate_transaction()` when calling buildTransaction() with nonce, do I need to add more things in pythonic.py middleware? :)

**Comment by carver** [comment]
> do I need to add more things in pythonic.py middleware? :)

This is mostly just a problem that eth_tester doesn't support nonce's. Maybe you could:

1. move the `remove_key_if` method to a utils module
2. delete the nonce key in an eth_tester-specific middleware (web3/providers/eth_tester/middleware.py)
3. open an issue on eth-tester to support nonces in transactions (with a reminder to remove the patch you're adding in step 2 ^)

File: tests/core/contracts/test_contract_constructor.py [hunk 1]
```diff
@@ -184,13 +183,15 @@ def test_contract_constructor_build_transaction_with_constructor_with_argument(
         WithConstructorArgumentsContract,
         constructor_args,
         constructor_kwargs):
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase) + 1
```
pipermerriam: Is this supposed to  be an invalid nonce?  Otherwise, I *think* that the `+ 1` should be removed.

zheli: If the nonce is 0, it will be changed to 1 in the return transaction. I am not sure why...

pipermerriam: Hrm, that sounds wrong, but likely a bug elsewhere (eth-tester)?.  Either way, it doesn't need to block this PR. 👍

carver: Another transaction is being sent in 187-188. Probably the most natural way to write this is to remove the `+ 1` and move this line down to just before 190.

tests/core/contracts/test_contract_constructor.py
```diff
@@ -142,32 +142,31 @@ def test_contract_constructor_build_transaction_to_field_error(MathContract):
         MathContract.constructor().buildTransaction({'to': '123'})
 
 
-def test_contract_constructor_build_transaction_nonce_field_error(MathContract):
-    with pytest.raises(ValueError):
-        MathContract.constructor().buildTransaction({'nonce': '123'})
-
-
 def test_contract_constructor_build_transaction_no_constructor(web3, MathContract):
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase) + 1
```
carver: Let's move this down to after sending the transaction so we don't have the confusing `+ 1`

tests/core/contracts/test_contract_constructor.py
```diff
@@ -142,32 +142,31 @@ def test_contract_constructor_build_transaction_to_field_error(MathContract):
...
 
     new_txn_hash = web3.eth.sendTransaction(unsent_txn)
     new_txn = web3.eth.getTransaction(new_txn_hash)
     assert new_txn['data'] == unsent_txn['data']
+    assert new_txn['nonce'] == nonce
 
 
 def test_contract_constructor_build_transaction_with_constructor_without_argument(web3,
                                                                                   MathContract):
+    nonce = web3.eth.getTransactionCount(web3.eth.coinbase) + 1
```
carver: Move nonce

tests/core/contracts/test_contract_constructor.py
zheli: Of course! Stupid me :D

pipermerriam: The transaction nonce should always be the number of transactions the account has sent (prior to sending the current transaction).  So the first txn an account sends should have a nonce of `0`, second transaction `1`, and so on.

If you check the `nonce` against `web3.eth.getTransactionCount` *after* the transaction has been mined then it should satisfy `txn["nonce"] == web3.eth.getTransactionCount(addr) - 1`.

If the `nonce` value for a transaction is wrong (too large, too small), then the transaction should be rejected by the node.

zheli: @pipermerriam thx for the explanation! Just one more question:D why do we want to specify nonce in transaction manually if it will be automatically assigned later anyway?

PR #744, opened 2018-04-04
Update filters.rst

Lebski:
### What was wrong?
The contracts.rst states: 
``fromBlock`` is a mandatory field.

### How was it fixed?
Added fromBlock example to filters.rst


carver:
Thanks!

Issue #730, opened 2018-03-31
Empty event log 


jfdelgad:
* Version: 4.b13
* Python: 3.6
* OS: win
* Geth 
* Rinkeby

I have a simple contract that produces events. The events are registered on mist correctly.
I can interact with the contract without problems but can't get the events. I am simply doing this:

"RegisterCall" is the name of the event in the contract.

```
event_filter = mycontract.events.RegisterCall.createFilter(fromBlock=1)
event_filter.get_all_entries()
>>>[]
```

Any suggestion will be appreciated.

EralpB:
Can you please put all your code? after instantiating the contract object you need to force update it's address to lowercase address (no EIP-55).

jfdelgad:
Hi, thanks for your answer

I am not forcing the address to lower case, I will try and will post the results. In the mean time this is the code. I remove the contract_soruce_code because is long.
 
compiled_sol = compile_source(contract_source_code) 
mycontract.Interface = compiled_sol['<stdin>:MyContract']

```
mycontractAddress = '0x36e87aBE249D4146fA2FED32Ce13f592122c611E'
mycontract = w3.eth.contract(address=mycontractAddress, abi=mycontractInterface['abi'],ContractFactoryClass=ConciseContract)
w3.personal.unlockAccount(w3.eth.accounts[0], '*********', duration=None)

event_filter = mycontract.events.RegisterCall.createFilter(fromBlock=1)
while True:
    print(event_filter.get_all_entries())
    time.sleep(2)

```

dylanjw:
Can you try the old api, so I can see if this is a problem specific to the new api.

```python
event_filter = mycontract.eventFilter('RegisterCall', filter_params={'fromBlock': 1})
```

Are you able to see results with a simple block filter?

```python
block_filter = w3.eth.filter("latest")
```

If you could post more of your code, it would be helpful.

EralpB:
@jfdelgad please try this one:

```
mycontractAddress = '0x36e87aBE249D4146fA2FED32Ce13f592122c611E'
mycontract = w3.eth.contract(address=mycontractAddress, abi=mycontractInterface['abi'],ContractFactoryClass=ConciseContract)
mycontract.address = '0x36e87abe249d4146fa2fed32ce13f592122c611e'  # THIS LINE!
w3.personal.unlockAccount(w3.eth.accounts[0], '*********', duration=None)

# I'm not sure if the following change is required, but just to be on the safe side:
[...]

PR #745, opened 2018-04-06
Replace mutable argument defaults in createFilter()

dylanjw:
### What was wrong?
createFilter() had arguments with mutable default values.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1167,9 +1167,6 @@ class ContractEvent:
         if argument_filters is None:
             argument_filters = dict()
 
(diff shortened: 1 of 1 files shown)


dylanjw:
This fixes #730. This was missed in tests because the arguments are explicitly set to None.

The utility function `construct_event_filter_params` expects None when no topics are set, otherwise the event signature topic does not get created.

carver:
I would feel a whole lot better about this if we add a test that fails before this patch and passes after.

PR #749, opened 2018-04-07
Update filters.rst

ambrozic:
Fix for broken url


carver:
Thanks ambrozic! Looks like #754 did the same thing (although that one came in after yours). Please send in more if you see them, and I'll try to get them in order next time. :)

PR #750, opened 2018-04-07
Example in filters.rst broken

DanielVF:
### What was wrong?

The example of event filtering in filters.rst is broken because `web3.sha3` has changed it's behavior.

`web3.sha3` now requires that you specify a `text` parameter to pass a string, and it now returns a HexBytes object instead of a string.

### How was it fixed?

I've updated the example to pass the correct parameters to the sha method and to convert the HexBytes output to a string like the filter topic is expecting.

Before

    event_signature_hash = web3.sha3("eventName(uint32)")

After

    event_signature_hash = web3.sha3(text="eventName(uint32)").hex()

Relevant docs:

- http://web3py.readthedocs.io/en/stable/overview.html#cryptographic-hashing
- https://github.com/carver/hexbytes/blob/master/hexbytes/main.py

#### Cute Animal Picture

Swissy Puppy

### DIFF
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -124,7 +124,7 @@ equivalent filter creation would look like:
 
     .. code-block:: python
 
-        event_signature_hash = web3.sha3("eventName(uint32)")
+        event_signature_hash = web3.sha3(text="eventName(uint32)").hex()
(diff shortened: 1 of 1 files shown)


carver:
Thanks!

PR #760, opened 2018-04-11
Update contract deployment logic in tests to use contract.constructor

jstoxrocky:
### What was wrong?
Tests were throwing warnings since `contract.deploy` has been deprecated.  Also, it looked like in many places `web3.eth.getTransactionReceipt` was being used where `web3.eth.waitForTransactionReceipt` is more appropriate. I made replacements here as well. I also think that it might be time to remove the `@pytest.fixture(params=['func_first', 'func_last'])` decorators, since "func comes last" has been deprecated. Should we be maintaining tests for deprecated code(?). Removal of 'func_last' will get rid of most (all?) warnings being thrown at the moment.

### How was it fixed?
`contract.deploy` has been replaced in tests with `contract.constructor`. `web3.eth.getTransactionReceipt` has been replaced with `web3.eth.waitForTransactionReceipt`.

### DIFF
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -12,8 +12,8 @@ pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
 
 @pytest.fixture()
 def math_contract(web3, MathContract):
-    deploy_txn = MathContract.deploy()
-    deploy_receipt = web3.eth.getTransactionReceipt(deploy_txn)
+    deploy_txn = MathContract.constructor().transact()
+    deploy_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
     assert deploy_receipt is not None
     _math_contract = MathContract(address=deploy_receipt['contractAddress'])
     return _math_contract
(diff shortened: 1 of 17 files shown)


carver:
> contract.deploy has been replaced in tests with contract.constructor. web3.eth.getTransactionReceipt has been replaced with web3.eth.waitForTransactionReceipt.

Thanks!

> I also think that it might be time to remove the @pytest.fixture(params=['func_first', 'func_last']) decorators, since "func comes last" has been deprecated. Should we be maintaining tests for deprecated code(?).

Until they are officially removed in v5, the methods are still supported. So we should leave the tests in, to make sure they continue to work.

> Removal of 'func_last' will get rid of most (all?) warnings being thrown at the moment.

I believe @voith is starting to look into filtering out warnings (when we are specifically testing deprecated methods). Maybe hop in gitter and work it out with him to reduce duplicate work?

tests/core/contracts/test_contract_buildTransaction.py
```diff
@@ -12,24 +12,25 @@
...
+    deploy_txn = FallballFunctionContract.constructor().transact()
+    deploy_receipt = web3.eth.waitForTransactionReceipt(deploy_txn)
     assert deploy_receipt is not None
     _fallback_contract = FallballFunctionContract(address=deploy_receipt['contractAddress'])
     return _fallback_contract
 
 
 def test_build_transaction_with_contract_no_arguments(web3, math_contract, buildTransaction):
     txn = buildTransaction(contract=math_contract, contract_function='increment')
+    # txn = math_contract.functions.increment().buildTransaction()
```
carver: Please remove

tests/core/contracts/test_contract_transact_interface.py
```diff
@@ -9,36 +9,34 @@
 from web3.utils.empty import (
     empty,
 )
-from web3.utils.transactions import (
-    wait_for_transaction_receipt,
-)
+
```
carver: Removing this line will fix the lint error

jstoxrocky: @carver thanks, I was having some troubling running tests on my computer so I just pushed and hoped for the best with Travis. Something with parity integration tests. Is there some documentation on how to make that work? Some kind of FileNotFound error.

jstoxrocky:
@carver thanks for the heads up, I'll hop on gitter!

Issue #705, opened 2018-03-19
Get contract functions with item lookup


carver:
* Version: 4

### What was wrong?

If a contract's function name is stored in a variable, and we want to work with that function, we have to do an awkward:

`contract_func = getattr(contract.functions, function_name)`

### How can it be fixed?

Add a function lookup using:

`contract_func = contract.functions[function_name]`

PR #761, opened 2018-04-11
Implement item lookup for ContractFunctions and ContractEvents

gcarq:
### What was wrong?

See issue #705

### How was it fixed?

Implemented `__getitem__(...)` for `ContractFunctions` and `ContractEvents` to allow a simple function/event lookup, i.e.:

```
contract_func = contract.functions[function_name]
contract_event = contract.events[event_name]
```

I would be happy to mention it in the docs, but I couldn't find a relevant section.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/contracts/conftest.py
+++ b/tests/core/contracts/conftest.py
@@ -498,7 +498,7 @@ def invoke_contract(api_style=None,
         raise ValueError("allowable_invoke_method must be one of: %s" % allowable_call_desig)
 
     if api_style == 'func_first':
-        function = getattr(contract.functions, contract_function)
+        function = contract.functions[contract_function]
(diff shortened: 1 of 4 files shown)


gcarq:
@carver thanks, I will take a look.

At some places the `functions` attribute is called directly, i.e.:
```
        myContract = web3.eth.contract(address=contract_address, abi=contract_abi)
        myContract.functions.multiply7(3).call()
```

Is this some special case or are the docs just outdated?

carver:
```
myContract.functions.multiply7(3).call()
```

This is a preferred way of calling functions now. Maybe the alternative you're thinking of is the [`ConciseContract` variant](http://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.ConciseContract)?

PR #762, opened 2018-04-12
Remove deprecation warnings in unit tests

jstoxrocky:
### What was wrong?
There were a lot of deprecation warnings thrown after running unit tests. @voith Sorry for jumping on this, had to work off some steam and took it out on the unit tests.

### How was it fixed?
Function calls were either brought up to date to remove deprecation warnings, or pytest assertions were used to explicitly state that a deprecation warning should be raised. Used in conjunction with #760, there should be no more warnings throw when running tests.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 import time
+import warnings
 
 from web3.providers.eth_tester import (
     EthereumTesterProvider,
@@ -111,3 +112,8 @@ def web3():
     w3.eth.enable_unaudited_features()
 
     return w3
+
+
+@pytest.fixture(autouse=True)
+def print_warnings():
+    warnings.simplefilter('always')
--- a/tests/core/contracts/conftest.py
+++ b/tests/core/contracts/conftest.py
@@ -502,7 +502,8 @@ def invoke_contract(api_style=None,
         result = getattr(function(*func_args, **func_kwargs), api_call_desig)(tx_params)
     elif api_style == 'func_last':
         api_call_cls = getattr(contract, api_call_desig)
-        result = getattr(api_call_cls(tx_params), contract_function)(*func_args, **func_kwargs)
+        with pytest.deprecated_call():
+            result = getattr(api_call_cls(tx_params), contract_function)(*func_args, **func_kwargs)
     else:
         raise ValueError("api_style must be 'func_first or func_last'")
 
--- a/tests/core/contracts/test_extracting_event_data_old.py
+++ b/tests/core/contracts/test_extracting_event_data_old.py
@@ -75,9 +75,9 @@ def test_event_data_extraction(web3,
                                event_name,
                                call_args,
                                expected_args):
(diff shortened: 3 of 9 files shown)


voith:
Hi @jstoxrocky,
I was planning to suppress the logs using [pytest.mark.filterwarnings](https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings]), however I could not get it working. I was curious why It was not working. So I went ahead and read the source of `pytest` and `warninngs`. It turns out that `filters`  that are [added last](https://github.com/python/cpython/blob/04e82934659487ecae76bf4a2db7f92c8dbe0d25/Lib/warnings.py#L178) get higher priority. 
So the filter `'@pytest.mark.filterwarnings(ignore::DeprecationWarning')` was always getting lower priority than the following code:
https://github.com/ethereum/web3.py/blob/6b783c5ed18ad47bb89c0ce2119037c981d2d27c/web3/utils/decorators.py#L53
Here's a hack I came up with to get it working:
```python
import functools
import warnings

import pytest

class Filter(list):
    def insert(self, index, p_object):
        self.append(p_object)


def filterwarnings(filters):
    def decorator(to_wrap):
        @functools.wraps(to_wrap)
        def wrapper(*args, **kwargs):
            with warnings.catch_warnings(record=True) as logs:
                warning_filters = warnings.filters[:]
                warnings.filters = Filter()
                warnings._filters_mutated()
                for _filter in filters:
                    warnings._setoption(_filter)
                return_val = to_wrap(*args, **kwargs)
                # log all uncaught logs
                # for warning in logs:
                #     msg = warnings.formatwarning(
                #         warning.message, warning.category,
                #         warning.filename, warning.lineno, warning.line)
                #     item.warn(msg)  # item obj in a pytest hook
                warnings.filters = warning_filters
                warnings._filters_mutated()
                return return_val
        return wrapper
    return decorator
```
and then use it as `@filterwarnings(['ignore::DeprecationWarning'])`. However, this is a hack and it can change intended behaviour.
 I was planning to implement this as a [pytest hook](https://docs.pytest.org/en/latest/writing_plugins.html#writing-hook-functions), but I don't think that it's worth.

`with pytest.deprecated_call()` seems like a better option to me because once the deprecated code is removed it will  error  if tests are left unmodified. However, `pytest.mark.filterwarnings` can remain as a stale check.

carver:
Cool, thanks for digging into this.

Let's just remove the warning tweaks in the decorator, and establish it once during the pytest run, like in: https://github.com/ethereum/eth-utils/pull/66/files

Then `with pytest.deprecated_call()` seems like the right way to assert the deprecated functionality. :+1:

jstoxrocky:
@carver Not sure what the issue is with geth ipc 1.7.2 integration tests. They seem to run fine locally on my machine. I don't understand the circleci output. Also looks like some duplicate work in #763 that I did not see.

pipermerriam:
@jstoxrocky I restarted that job.  Lets see what happens.

carver:
@akolotov :+1:

PR #763, opened 2018-04-12
removed warning filter adjustement

akolotov:
removed filter adjustement since it disallows to filter out deprication warnings

### What was wrong?

The warning filter configuration applied in decorator switched off any filtering and then turned it on again. It caused impossibility to suppress deprication warning by contstruction like this: 

```
warnings.simplefilter("ignore", category=DeprecationWarning)
```

### How was it fixed?

The warning filter adjustment was removed.

#### Cute Animal Picture

Cute animal picture


carver:
Resolved in #762

PR #756, opened 2018-04-09
Added a check for POA extraData

voith:
### What was wrong?

see #712

### How was it fixed?
 
A check for POA extra has been added with an error message directing to the docs.

#### Cute Animal Picture
vulpix full 1774430

PR #777, opened 2018-04-20
Undoing the mixup between Chain ID and Network ID

carver:
### What was wrong?

web3py was returning `networkId` when asked for `chainId`.

### How was it fixed?

There is no quick/foolproof way to determine the value, and `None` is valid in all cases, so we just return that.

- There were incorrect `chainId` validations in place, remove them
- Always return `None` from `w3.net.chainId`
- Document

#### Cute Animal Picture

### DIFF
--- a/tests/core/eth-module/test_transactions.py
+++ b/tests/core/eth-module/test_transactions.py
@@ -12,20 +12,13 @@ from web3.middleware.simulate_unmined_transaction import (
     'make_chain_id, expect_success',
     (
         (
-            lambda web3: web3.version.network,
+            lambda web3: web3.net.chainId,
             True,
         ),
-        (
-            lambda web3: int(web3.version.network),
-            True,
-        ),
-        (
-            lambda web3: int(web3.version.network) + 1,
+        pytest.param(
+            lambda web3: 999999999999,
             False,
-        ),
-        (
-            lambda web3: str(int(web3.version.network) + 1),
-            False,
+            marks=pytest.mark.xfail,
         ),
     ),
 )
--- a/web3/middleware/validation.py
+++ b/web3/middleware/validation.py
@@ -15,8 +15,8 @@ from web3.exceptions import (
 
 @curry
 def validate_chain_id(web3, chain_id):
-    if chain_id is None:
-        return None
+    if chain_id == web3.net.chainId:
+        return chain_id
     else:
         raise ValidationError(
             "The transaction declared chain ID %r, "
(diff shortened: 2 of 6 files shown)


web3/middleware/validation.py
```diff
@@ -15,14 +15,14 @@
 
 @curry
 def validate_chain_id(web3, chain_id):
-    if chain_id == web3.version.network:
+    if chain_id is None:
```
voith: what?? This is going away?? Just because this middleware I spent the last two hours trying to figure out https://github.com/voith/web3.py/blob/b3a30a7374cb5f205d7e96a326533dc8394c7a73/web3/utils/formatters.py#L168-L180.
I would have got #756 right the first time I did it if I knew this change was coming.

Also you might want to remove `web3` from `validate_chain_id(web3, chain_id)` as its no longer needed

carver: I guess the clearer thing to do is his to set it to:

```py
if chain_id == web3.net.chainId:
```

Right now, that's the same thing. But once we get nodes to make `chainId` available in a standard way, it should start working again.

I'll review #756 today.

tests/core/eth-module/test_transactions.py
```diff
@@ -12,20 +12,13 @@
...
-        (
-            lambda web3: int(web3.version.network) + 1,
-            False,
-        ),
-        (
-            lambda web3: str(int(web3.version.network) + 1),
+        pytest.param(
+            lambda web3: 999999999999,
             False,
+            marks=pytest.mark.xfail,
```
voith: @carver why did you mark this as xfail?
```python
from web3.auto import w3
transaction = {'to': w3.eth.accounts[0], 'chainId': 9999999, 'from': w3.eth.accounts[1]}
w3.personal.unlockAccount(w3.eth.accounts[1], 'passphrase')
w3.eth.sendTransaction(transaction)
```
shouldn't the above code give `ValidationError` for `chainId`?  In fact validation middleware doesn't seem to do what it is expected to do.
I'm not very sure if this is a regression but I'm fixing this in #756.

web3/middleware/validation.py
```diff
@@ -33,7 +33,6 @@ def transaction_normalizer(transaction):
 
 def validation_middleware(make_request, web3):
     transaction_validator = apply_formatters_to_dict({
-        'chainId': validate_chain_id(web3),
```
voith: I see that this is the cause for using xfail! 
Read the description in the PR.  However If `validate_chain_id` is not needed anymore, why is the code left behind?
Is this something you plan to return back to

carver: > Is this something you plan to return back to?

Yes. Once we can reliably discover the chainID from the connected node, then we should validate chainID again.

PR #784, opened 2018-04-23
Update SHH module for Whisper v6

ConnorChristie:
### What did I do?

Updated the shh module to conform to the Whisper v6 protocol. Much of the work was just changing the signatures and methods to use the new revised ones from v6.

I also added async watch functionality to the filter protocol.

_If there is anything missing or additional details that need to be updated, I would be glad to make those changes!_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,5 @@ crashlytics.properties
 crashlytics-build.properties
 fabric.properties
 
+# VS Code
+.vscode/
--- a/conftest.py
+++ b/conftest.py
@@ -8,7 +8,7 @@ from web3.providers.eth_tester import (
 from web3.utils.threads import (
     Timeout,
 )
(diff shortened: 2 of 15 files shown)


ConnorChristie:
I have mainly been going off of the official [v6 implementation](https://github.com/ethereum/go-ethereum/blob/master/whisper/whisperv6/api.go) as well as the [documentation from web3.js](http://web3js.readthedocs.io/en/1.0/web3-shh.html).

carver:
Ok, thanks for the links. It partially confirmed my expectation that this is based on a (maybe) shifting API.

I hesitate a little bit to commit to our own stable version API on top of of a beta one. But... the current `shh` module is ridiculously outdated, it seems clear that almost no one is using it. Plus it's not a default module so :+1:

PR #785, opened 2018-04-23
Better error messages when the RPC request fails encoding

### LINKED ISSUE #771: Better error message on failed encoding
* Version: 4.1.0

### What was wrong?

A field of type `HexBytes` slipped through the pythonic middleware and the web3 provider tried to encode it with json. The only error message was:

> TypeError: Object of type 'HexBytes' is not JSON serializable

### How can it be fixed?

Catch type errors in:

```
~/web3/providers/base.py in encode_rpc_request(self, method, params)
     65             "method": method,
---> 66             "params": params or [],
     67             "id": next(self.request_counter),
     68         }))
     69
```
In case of a `TypeError`, inspect `params` and raise another `TypeError` saying which parameter index was failing to json encode, and (if applicable) which key in the `dict` at that index.

boneyard93501:
### What was wrong?

Fixes #771 

### How was it fixed?
update encode_rpc_request with try block for json encoding
add test in test_provider.py using JsonBaseProvider

### DIFF
--- a/web3/providers/base.py
+++ b/web3/providers/base.py
@@ -59,15 +59,16 @@ class JSONBaseProvider(BaseProvider):
         self.request_counter = itertools.count()
 
     def decode_rpc_response(self, response):
-        decoded = FriendlyJsonSerde().json_decode(to_text(response))
-        return decoded
+        text_response = to_text(response)
+        return FriendlyJsonSerde().json_decode(text_response)
 
     def encode_rpc_request(self, method, params):
-        rpc_dict = {"jsonrpc": "2.0",
-                    "method": method,
-                    "params": params or [],
-                    "id": next(self.request_counter),
-                    }
+        rpc_dict = {
+            "jsonrpc": "2.0",
+            "method": method,
+            "params": params or [],
+            "id": next(self.request_counter),
+        }
         encoded = FriendlyJsonSerde().json_encode(rpc_dict)
         return to_bytes(text=encoded)
 
--- a/web3/providers/base.py
+++ b/web3/providers/base.py
@@ -1,5 +1,4 @@
 import itertools
-import json
 
 from eth_utils import (
     to_bytes,
@@ -8,6 +7,9 @@ from eth_utils import (
 
 from web3.middleware import (
     combine_middlewares,
+)
+from web3.utils.encoding import (
+    FriendlyJsonSerde,
 )
 
 
@@ -57,15 +59,17 @@ class JSONBaseProvider(BaseProvider):
         self.request_counter = itertools.count()
 
     def decode_rpc_response(self, response):
-        return json.loads(to_text(response))
+        decoded = FriendlyJsonSerde().json_decode(to_text(response))
+        return decoded
 
     def encode_rpc_request(self, method, params):
-        return to_bytes(text=json.dumps({
-            "jsonrpc": "2.0",
-            "method": method,
-            "params": params or [],
-            "id": next(self.request_counter),
-        }))
+        rpc_dict = {"jsonrpc": "2.0",
+                    "method": method,
+                    "params": params or [],
+                    "id": next(self.request_counter),
+                    }
+        encoded = FriendlyJsonSerde().json_encode(rpc_dict)
+        return to_bytes(text=encoded)
 
     def isConnected(self):
         try:
--- a/web3/utils/encoding.py
+++ b/web3/utils/encoding.py
@@ -214,37 +214,38 @@ def hexstr_if_str(to_type, hexstr_or_primitive):
 
 
 class FriendlyJsonSerde:
-    ''' friendly json serde '''
-    def __init__(self,):
-        self.type_error = False
-        self.type_error_msg = 'Could not encode: '
+    '''
+    Friendly JSON serializer & deserializer
 
+    When encoding or decoding fails, this class collects
+    information on which fields failed, to show more
+    helpful information in the raised error messages.
+    '''
     def _json_mapping_errors(self, mapping):
         for key, val in mapping.items():
             try:
                 self._friendly_json_encode(val)
             except TypeError as exc:
-                yield "value at %r because (%s)" % (key, exc)
+                yield "%r: because (%s)" % (key, exc)
 
     def _json_list_errors(self, iterable):
         for index, element in enumerate(iterable):
             try:
                 self._friendly_json_encode(element)
             except TypeError as exc:
-                yield "item at index %d because (%s)" % (index, exc)
+                yield "%d: because (%s)" % (index, exc)
 
     def _friendly_json_encode(self, obj):
         try:
             encoded = json.dumps(obj)
             return encoded
         except TypeError as full_exception:
-            self.type_error = True
             if hasattr(obj, 'items'):
                 item_errors = '; '.join(self._json_mapping_errors(obj))
-                self.type_error_msg += item_errors
+                raise TypeError("dict had unencodable value at keys: {{{}}}".format(item_errors))
             elif is_list_like(obj):
                 element_errors = '; '.join(self._json_list_errors(obj))
-                self.type_error_msg += element_errors
+                raise TypeError("list had unencodable value at index: [{}]".format(element_errors))
             else:
                 raise full_exception
 
@@ -257,8 +258,7 @@ class FriendlyJsonSerde:
             raise ValueError(err_msg)
 
     def json_encode(self, obj):
-            encoded = self._friendly_json_encode(obj)
-            if self.type_error:
-                raise TypeError(self.type_error_msg)
-            else:
-                return encoded
+        try:
+            return self._friendly_json_encode(obj)
+        except TypeError as exc:
+            raise TypeError("Could not encode to JSON: {}".format(exc))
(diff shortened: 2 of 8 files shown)


tests/core/providers/test_provider.py
```diff
@@ -51,3 +52,18 @@ def must_not_call():
...
+    import datetime
+
+    provider = JSONBaseProvider()
+
+    method = 'test_method'
+    params = []
+    expected_bstr = b'{"jsonrpc": "2.0", "method": "test_method", "params": [], "id": 0}'
+    assert provider.encode_rpc_request(method, params) == expected_bstr
+    params = [datetime.datetime.utcnow()]
+    with pytest.raises(TypeError, match=r'.* failure .*'):
```
carver: Maybe match against `datetime` to show that the repr of the field we care about was included?

web3/providers/base.py
```diff
@@ -60,12 +60,25 @@ def decode_rpc_response(self, response):
...
+                                "params": rpc_params,
+                                "id": rpc_id})
+        except TypeError:
+            err_msg = ''
+            for i, p in enumerate(params):
+                try:
+                    json.dumps(p)
+                except TypeError:
+                    err_msg += ' ({},{});'.format(i, repr(p))
+            msg = 'encode_rpc_request failure for method {} for invalid (index, parameter): '
```
pipermerriam: Instead of embedding this functionality within the provider, this strikes me as something that could be extracted to a stand-alone function and tested much easier in isolation.

```python
def json_encode_with_helpful_error_message(payload):
    try:
        return json.dumps(payload)
    except:
        ...  # awesome error messaging code.
```

Thoughts?  Should help cleanup this method body a lot and lower the overhead to testing.

boneyard93501:
@carver absolutely. 
@pipermerriam sure. where do you want it ?  still in base.py or move it into utils, say, serializers.py? ditto for  decode_rpc_response ?

pipermerriam:
I'd be good with `web3.utils.json`.  Open to other options for a better namespace assuming we want to avoid collision with the stdlib `json` but I don't think it's much of a problem.

carver:
`make lint-roll` locally will check all your lint issues @boneyard93501 

Also, please do a rebase and force push. That will make the diff readable again.

boneyard93501:
it's been working locally.
``` python
(web3) 93501@iMac27 ~/localdev/source-code/forks/web3.py (fix-771) $ make lint-roll
<snip>
________________________________________________________________________________________ summary ________________________________________________________________________________________
  lint: commands succeeded
  congratulations :)
```

web3/utils/encoding.py
```diff
@@ -209,3 +210,34 @@ def hexstr_if_str(to_type, hexstr_or_primitive):
...
+    try:
+        j_dict = json.loads(to_text(response))
+    except json.decoder.JSONDecodeError:
+        msg = 'decode_rpc_response failure for "{}".'.format(repr(response))
+        raise ValueError(msg)
+    return j_dict
+
+
+@curry
+def encode_rpc_request(rpc_id, method, params):
```
carver: Right now, this is tightly coupled to the json-rpc request. What if we wanted to do a similar error message when encoding a genesis block into json format for a file? It would be nice to have something like a:

```py
def friendly_json_encode(obj_to_encode):
    ...
```

This could add helpful messages across this project (and others), pretty much anywhere that json encoding is happening.

Instead of using our special knowledge of `params` here, it could just iterate through all dicts and lists, emitting warnings for each offending piece. Something like:

```py
def friendly_json_encode(obj):
    try:
        encoded = json.dumps(obj)
    except TypeError full_exception:
        if hasattr(obj, 'items'):
            item_errors = '; '.join(json_mapping_errors(obj))
            raise TypeError("Could not encode dictionary because of: %s" % item_errors)
        elif eth_utils.is_list_like(obj):
            element_errors = '; '.join(json_list_errors(obj) 
            raise TypeError("Could not encode list because of: %s" % element_errors)
        else: 
            raise full_exception

                 
def json_mapping_errors(mapping):
    for key, val in mapping.items():
        try:
            friendly_json_encode(val)
        except TypeError as exc:
            yield "value at %r because (%s)" % (key, exc)
 

def json_list_errors(iterable):
    for index, element in enumerate(iterable):
        try:
            friendly_json_encode(element)
        except TypeError as exc:
            yield "item at index %d because (%s)" % (index, exc) 
```

(Totally untested)

web3/providers/base.py
```diff
@@ -55,17 +57,8 @@ def isConnected(self):
...
-
-    def encode_rpc_request(self, method, params):
-        return to_bytes(text=json.dumps({
-            "jsonrpc": "2.0",
-            "method": method,
-            "params": params or [],
-            "id": next(self.request_counter),
-        }))
+        self.decode_rpc_response = decode_rpc_response
+        self.encode_rpc_request = partial(encode_rpc_request(next(self.request_counter)))
```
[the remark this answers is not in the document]
boneyard93501: f'ing hell. i made one "small" change from the off-project test ... should have known better. dang. sorry.

web3/utils/encoding.py
boneyard93501: makes sense. let me have a look. hopefully, i'll have some time tomorrow.

boneyard93501: @carver do want to accommodate  (optional) encoder/decoder hooks ?
```python
def json_encoder(obj, CustomEncoder=None):
    try:
        encode = json.dumps(obj, cls=CustomEncoder)
         ...
# and
def json_decoder(json_string, obj_hook=None):
    try:
        decode = json.loads(json_string, object_hook=obj_hook)
        ...
```

carver: Yeah, sounds great :+1:

boneyard93501: @carver very sorry but i'm out of pocket this week for both 770 and 771. back next week.

boneyard93501: @carver  just about to re-start this . just looking at the yield approach outlined, it seems to me we'll end up pushing two errors in just about every invalid case. e.g., *{'date': datetime.datetime.utcnow()}* -- one for mapping/sequence types and one for other types leading to the full exception raised. that in turn prevents the full traversal of mapping or sequence objects. just want to make sure that is the intended behavior. is it?

carver: The idea is that all of the fields should be traversed to make a message about all the fields that have errors. If that's not what is happening, then it's a bug. But I think it works.

carver:
`make lint-roll` should fix up any import spacing and/or ordering issues. (although the important part is pretty simple to call directly: `isort --recursive web3 ens tests`)

boneyard93501:
@carver i don't think those failures are coming from my end. i just cloned a clean master instance to double check and:
```bash
93501@iMac27 ~/localdev/web3/web3.py (master) $ workon web3-4.2.1
(web3-4.2.1) 93501@iMac27 ~/localdev/web3/web3.py (master) $ make lint-roll
isort --recursive web3 ens tests
Fixing /Users/93501/localDev/web3/web3.py/tests/integration/go_ethereum/test_goethereum_ws.py
Fixing /Users/93501/localDev/web3/web3.py/tests/integration/parity/test_parity_http.py
Fixing /Users/93501/localDev/web3/web3.py/tests/integration/parity/test_parity_ipc.py
Fixing /Users/93501/localDev/web3/web3.py/tests/integration/parity/test_parity_ws.py
Skipped 17 files
/Library/Developer/CommandLineTools/usr/bin/make lint
tox -elint

<snip>

from web3 import Web3
from web3.utils.module_testing import (
    NetModuleTest,
ERROR: InvocationError for command '/Users/93501/localDev/web3/web3.py/.tox/lint/bin/isort --recursive --check-only --diff /Users/93501/localDev/web3/web3.py/web3/ /Users/93501/localDev/web3/web3.py/ens/ /Users/93501/localDev/web3/web3.py/tests/' (exited with code 1)
____________________________________________________________________________________________________________ summary ____________________________________________________________________________________________________________
ERROR:   lint: commands failed
make[1]: *** [lint] Error 1
make: *** [lint-roll] Error 2
(web3-4.2.1) 93501@iMac27 ~/localdev/web3/web3.py (master) $
```
any suggestions?  thanks.

carver:
Hm, usually when that happens to me, it means that one of my environments is not completely up to date: either my tox or my venv.

boneyard93501:
@carver thanks. i'll just have to reset my pyenv. so i can do the os x and ubuntu portions for issue #840, dev env, if you want. make up for being a pita :)

carver:
Heh, sounds good. For more background: `isort` checks your environment to see which libraries are installed. It groups your imports roughly by builtins, 3rd-party, then local. But if it can't find a module installed in the environment, it assumes local. So if it is trying to group an import down with the locals that looks like a 3rd party library, then that one is probably not installed in the environment you ran `isort` inside.

boneyard93501:
@carver  thanks. this' been driving my crazy. i can get everything to work up to 4.2.0; when i create a virtualenv from 4.2.1, i get  the lint roll failure. darnest thing. tried it on clean ubuntu, os x. anyway, i worked-off the 4.2 venv, we'll see if that sticks.

boneyard93501:
@carver no worries and thanks for all your your help. appreciate it.

... just found that unsent in a very stale chrome tab. pressing the green button ... pavlov would have to say something about this :) thanks again.

PR #789, opened 2018-04-25
specified a version for pypiwin32 and moved it to install_requires

voith:
### What was wrong?

`pypiwin32` did not have a version number.

### How was it fixed?
specified a version for pypiwin32 and moved it to install_requires.

I have tested this change on a windows machine. Also tested on Mac to make sure that the `platform_system` check works as expected.

#### Cute Animal Picture
980x

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup(
         "eth-hash[pycryptodome]",
         "requests>=2.16.0,<3.0.0",
         "websockets>=4.0.1",
+        "pypiwin32>=223;platform_system=='Windows'",
     ],
     setup_requires=['setuptools-markdown'],
     python_requires='>=3.5, <4',
@@ -38,9 +39,6 @@ setup(
         'linter': [
             "flake8==3.4.1",
             "isort>=4.2.15,<5",
-        ],
-        'platform_system=="Windows"': [
-            'pypiwin32'  # TODO: specify a version number, move under install_requires
         ],
     },
     py_modules=['web3', 'ens'],


voith:
Test failure seems unrelated.

carver:
Restarted the test

setup.py
```diff
@@ -26,6 +26,7 @@
         "eth-hash[pycryptodome]",
         "requests>=2.16.0,<3.0.0",
         "websockets>=4.0.1",
+        "pypiwin32>=223; platform_system == 'Windows'",
```
pipermerriam: A brief search shows that we may want to remove this whitespace to conform to existing conventions

- pytest: https://github.com/pytest-dev/pytest/blob/master/setup.py#L71-L72
- pipenv: https://github.com/pypa/pipenv/blob/master/setup.py#L30-L32

carver:
@voith or anyone know how big an impact this had for Windows? I'm working on the release notes. Some options:

- Now works on Windows
- Better Windows support
- Less broken on Windows

I don't want to oversell the change :laughing:

voith:
> @voith or anyone know how big an impact this had for Windows? 

I hope it makes the lives of windows users easier 😃  Already checked the release notes, so my opinion doesn't matter now.

BTW, I just tested v4.2.0 on a windows machine and it works fine. Made sure that the `IPCProvider` works as expected.

The only things that I had to struggle though was with installing `visual studio build tools`. cytoolz needs it. But I guess every developer on windows is expected to have these tools.

carver:
> BTW, I just tested v4.2.0 on a windows machine and it works fine. Made sure that the `IPCProvider` works as expected.

Great!

> The only things that I had to struggle though was with installing visual studio build tools. cytoolz needs it. But I guess every developer on windows is expected to have these tools.

:man_shrugging: I was a Windows guy many moons ago, but I eagerly forgot everything as soon as possible. Maybe this could be improved with a fancy windows install that detects if the build tools are available and if not, switches to installing toolz instead. (plus changing all the cytoolz imports to use an internal shim that picks between toolz and cytoolz).

pipermerriam:
`from web3.utils.toolz import ...`

```python
# web3/utils/__init__.py
try:
   import cytoolz as toolz
except ImportError:
   import toolz
```

I'd be game to do this across the board so that we can do the same thing with our `toolz/cytoolz` dependency and swap what we install based on whether it's a Cypthon or pypy install.

voith:
@carver @pipermerriam I tried to install toolz when build tools aren't available. Have a look at my [commit](https://github.com/ethereum/web3.py/commit/bffae8bfac363c02e5d0e8ddd502d458a4c423b2). 

However, this is not as simple as I thought it would be. Firstly, `cytoolz` is not the only package that `web3.py` uses that uses c extensions. `lru-dict` uses c extensions too. Moreover, things were fairly simple for switching  `cytoolz` --> `toolz` as both have similar module and API structure. The python equivalent for `lru-dict` is `pylru`, however their API's differ from each other. Besides packages like 
`eth-utils` also have `cytoolz` as a dependency.  If at all we had to make this switch, we'd have to make all other dependencies compatible too.

I don't think this is worth doing. Every time `web3.py` needs a dependency that uses c extension we'd have to explore a python equivalent package and write extra code for it just so that a windows user doesn't have to bother installing build tools?!?! And why compromise on speed offered by c extensions? Also `distutils` give a good enough error message upon failure:
```
error: Setup script exited with error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": http://landinghub.visualstudio.com/visual-cpp-build-tools
```
Also, ubuntu users have to install `python-dev` tools whenever they use c extensions:
```
sudo apt-get install python3.6-dev
```

pipermerriam:
@voith somewhat related: https://github.com/ethereum/eth-utils/pull/86

Doesn't fix the need for something to handle c-extensions.  I think our best route is to make sure the documentation has this stuff clearly laid out in the installation section.

As for the LRU dependencies: we use `lru-dict` instead of `pylru` because of licensing.

voith:
> I think our best route is to make sure the documentation has this stuff clearly laid out in the installation section.

Yeah documenting this seems good enough to me at the moment.  I'll try to submit a PR, but only once I finish work on invoking ambiguous contact functions.

PS: I haven't started writing code for it yet.  Just started to dig code in order to have a better understanding.

Issue #770, opened 2018-04-17
Dropped data field in modifyTransaction


carver:
* Version: 4.1.0
* Python: 3.6
* OS: linux

### What was wrong?

Used `web3.eth.modifyTransaction(txn_hash, gasPrice=new_gas_price)`, and it dropped the data field from the original transaction.

### How can it be fixed?

Rename `input` field to `data` when copying data from old transaction.

boneyard93501:
@carver 

want to do it in the param extraction ?
```python
# transactions.py
def extract_valid_transaction_params(transaction_params):
    t_params = {key: transaction_params[key] 
                for key in VALID_TRANSACTION_PARAMS if key in transaction_params}
    if t_params.get('data') is None and t_params.get('input') is not None:
        t_params['data'] = deepcopy(t_params['input'])  
    return t_params
```

pipermerriam:
Um, maybe just use the [`eth_utils.apply_key_map`](https://github.com/ethereum/eth-utils#apply_key_mapformatter_dict-dict_like---dict)?  At least with the latest `eth-utils` it should protect against situations where both `data` and `input` are included and prevent overwritting.  (not super familiar with this issue so disregard if this isn't the right solution).

boneyard93501:
i looked at that but wasn't entirely sure whether data and input could be instantiated; i kinda doubt it. but if its a highly improbably/impossible edge-case, re-keying is definitely the way to go.

PR #793, opened 2018-04-26
Accomodate input param in transaction processing (#770)

boneyard93501:
### What was wrong?
transactions with only input but not data param fail to re-key thereby dropping data. see # 770

Related to Issue # 770

### How was it fixed?
update extract_valid_transaction_params to test for input param

### x-module issue:
the (same) extract_valid_transaction_params() is also in eth-tester.utils.transactions.py. assuming the PR passes, want this updated commensurately ?

### DIFF
--- a/tests/core/utilities/test_valid_transaction_params.py
+++ b/tests/core/utilities/test_valid_transaction_params.py
@@ -38,31 +38,47 @@ def test_assert_valid_transaction_params_invalid_param():
 
 
 def test_extract_valid_transaction_params():
-    from copy import deepcopy
-    base_input = {
+    ''' baseline test for clean param dict '''
+    input = {
         'from': '0x0',
         'to': '0x0',
         'gas': 21000,
         'gasPrice': 5000000,
+        'data': '0x0',
         'value': 1,
         'nonce': 2,
         'chainId': 1,
     }
-
-    input = deepcopy(base_input)
-    input['data'] = '0x0'
     valid_transaction_params = extract_valid_transaction_params(input)
     assert valid_transaction_params == input
 
-    in_input = deepcopy(base_input)
-    in_input['input'] = '0x0'
-    in_input['data'] = '0x0'
-    valid_transaction_params = extract_valid_transaction_params(in_input)
-    assert valid_transaction_params == input
 
-    input['input'] = '0x1'
-    with pytest.raises(AttributeError, match=r'.* You need to resolve this conflict.'):
-        extract_valid_transaction_params(input)
+test_data = [({'data': '0x0', 'input': '0x0'}, {'data': '0x0'}), ]
+@pytest.mark.parametrize("transaction_params, expected", test_data)  # noqa E302
+def test_extract_valid_transaction_params_2(transaction_params, expected):
+    valid_transaction_params = extract_valid_transaction_params(transaction_params)
+    assert valid_transaction_params == expected
+
+
+test_data = [({'input': '0x0'}, {'data': '0x0'}), ]
+@pytest.mark.parametrize("transaction_params, expected", test_data)  # noqa E302
+def test_extract_valid_transaction_params_3(transaction_params, expected):
+    valid_transaction_params = extract_valid_transaction_params(transaction_params)
+    assert valid_transaction_params == expected
+
+
+test_data = [({}, {}), ]
+@pytest.mark.parametrize("transaction_params, expected", test_data)  # noqa E302
+def test_extract_valid_transaction_params_4(transaction_params, expected):
+    valid_transaction_params = extract_valid_transaction_params(transaction_params)
+    assert valid_transaction_params == expected
+
+
+
+@pytest.mark.xfail(raises=AttributeError, strict=True)  # noqa E302
+def test_extract_valid_transaction_params_5():
+    transaction_params = {'data': '0x0', 'input': '0x1'}
+    extract_valid_transaction_params(transaction_params)
 
 
 def test_extract_valid_transaction_params_includes_invalid():
--- a/tests/core/utilities/test_valid_transaction_params.py
+++ b/tests/core/utilities/test_valid_transaction_params.py
@@ -38,18 +38,31 @@ def test_assert_valid_transaction_params_invalid_param():
 
 
 def test_extract_valid_transaction_params():
-    input = {
+    from copy import deepcopy
+    base_input = {
         'from': '0x0',
         'to': '0x0',
         'gas': 21000,
         'gasPrice': 5000000,
         'value': 1,
-        'data': '0x0',
         'nonce': 2,
         'chainId': 1,
     }
+
+    input = deepcopy(base_input)
+    input['data'] = '0x0'
     valid_transaction_params = extract_valid_transaction_params(input)
     assert valid_transaction_params == input
+
+    in_input = deepcopy(base_input)
+    in_input['input'] = '0x0'
+    in_input['data'] = '0x0'
+    valid_transaction_params = extract_valid_transaction_params(in_input)
+    assert valid_transaction_params == input
+
+    input['input'] = '0x1'
+    with pytest.raises(AttributeError, match=r'.* You need to resolve this conflict.'):
+        extract_valid_transaction_params(input)
 
 
 def test_extract_valid_transaction_params_includes_invalid():
--- a/tests/core/utilities/test_valid_transaction_params.py
+++ b/tests/core/utilities/test_valid_transaction_params.py
@@ -37,48 +37,42 @@ def test_assert_valid_transaction_params_invalid_param():
         })
 
 
-def test_extract_valid_transaction_params():
-    ''' baseline test for clean param dict '''
-    input = {
-        'from': '0x0',
-        'to': '0x0',
-        'gas': 21000,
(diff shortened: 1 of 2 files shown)


tests/core/utilities/test_valid_transaction_params.py
```diff
@@ -38,19 +38,32 @@ def test_assert_valid_transaction_params_invalid_param():
...
+    in_input = deepcopy(base_input)
+    in_input['input'] = '0x0'
+    in_input['data'] = '0x0'
+    valid_transaction_params = extract_valid_transaction_params(in_input)
+    assert valid_transaction_params == input
+
+    input['input'] = '0x1'
+    with pytest.raises(AttributeError, match=r'.* You need to resolve this conflict.'):
+        extract_valid_transaction_params(input)
+
```
carver: This looks like it should really be just like the old test, but with parametrized inputs and expected values. I think there are a couple more cases to cover, too. Will be easier to see in paramaterized inputs.

web3/utils/transactions.py
```diff
@@ -92,8 +92,19 @@ def get_required_transaction(web3, transaction_hash):
 
 
 def extract_valid_transaction_params(transaction_params):
-    return {key: transaction_params[key]
-            for key in VALID_TRANSACTION_PARAMS if key in transaction_params}
+    t_params = {key: transaction_params[key]
```
carver: *style nitpick*

We almost never abbreviate to a single letter. Some alternatives:
- `extracted_params`
- `cleaned_params`

web3/utils/transactions.py
```diff
@@ -92,8 +92,19 @@ def get_required_transaction(web3, transaction_hash):
...
+                for key in VALID_TRANSACTION_PARAMS if key in transaction_params}
+
+    if t_params.get('data') is not None and transaction_params.get('input') is not None:
+        if t_params['data'] != transaction_params['input']:
+            msg = 'failure to handle this transaction due to both "input: {}" and'
+            msg += ' "data: {}" are populated. You need to resolve this conflict.'
+            raise AttributeError(msg.format(transaction_params['input'], t_params['data']))
```
carver: I find it easier to read when all of the alternative cases are explicitly handled in else blocks, even if it means a bit of duplication. eg~

```py
    if t_params.get('data') is not None and transaction_params.get('input') is not None:
        if t_params['data'] != transaction_params['input']:
            msg = 'failure to handle this transaction due to both "input: {}" and'
            msg += ' "data: {}" are populated. You need to resolve this conflict.'
            raise AttributeError(msg.format(transaction_params['input'], t_params['data']))
        else:
            return dissoc(t_params, 'input')
    elif t_params.get('data') is None and transaction_params.get('input') is not None:
        remapped_data =  t_params.pop('input')
        return assoc(
            dissoc(t_params, 'input'),
            'data',
            remapped_data,
        )
    else:
        return dissoc(t_params, 'input')
```

carver: Maybe even splitting up the `and` tests into nested ifs:
```py
if transaction_params.get('input') is None:
    return dissoc(t_params, 'input')
elif t_params.get('data') is None:
        ...
else:

File: tests/core/utilities/test_valid_transaction_params.py (hunk 1 above)
Comment (boneyard93501) [reply tests/core/utilities/test_valid_transaction_params.py]: the test envelope right now is:
1: data, no input :: data
2: data, input, data == input :: data
3: data, input,  data != input :: Error  (for that i need to change the value, e.g., '0x1')

to add:
4: no data, input :: data  (from input)
adding 5: no data, no input seems overkill but i'm happy to oblige.

File: web3/utils/transactions.py (hunk 2 above)
Comment (boneyard93501) [reply web3/utils/transactions.py]: k.

File: web3/utils/transactions.py (hunk 3 above)
Comment (boneyard93501) [reply web3/utils/transactions.py]: i didn't add *"input"* to *VALID_TRANSACTION_PARAMS*, so it's never in *extracted_params* (formerly *t_params*). instead, i'm checking  *transaction_params*  for *"input"* (throughout the function) so there's no managing of keys required; just value assignment or raise. do you want me to add it to the *VALID_TRANSACTION_PARAMS* ?

File: tests/core/utilities/test_valid_transaction_params.py [hunk 4]
```diff
@@ -38,20 +38,49 @@ def test_assert_valid_transaction_params_invalid_param():
...
     }
     valid_transaction_params = extract_valid_transaction_params(input)
     assert valid_transaction_params == input
 
 
+test_data = [({'data': '0x0', 'input': '0x0'}, {'data': '0x0'}), ]
+@pytest.mark.parametrize("transaction_params, expected", test_data)  # noqa E302
+def test_extract_valid_transaction_params_2(transaction_params, expected):
+    valid_transaction_params = extract_valid_transaction_params(transaction_params)
+    assert valid_transaction_params == expected
```
carver: parameterize allows us to collapse tests 1-4 into one, like:

```py
# all of these values should probably be different, so we can't accidentally cross-verify values.
FULL_TXN_DICT =  {
        'from': '0x0',
        'to': '0x1',
        'gas': 21000,
        'gasPrice': 5000000,
        'data': '0x2',
        'value': 3,
        'nonce': 2,
        'chainId': 1,
    }


@pytest.mark.parametrize(
    "transaction_params, expected",
    (
        (FULL_TXN_DICT, FULL_TXN_DICT),
        ({'data': '0x0', 'input': '0x0'}, {'data': '0x0'}),
        ({'input': '0x0'}, {'data': '0x0'}),
        ({}, {}),
    ),
)
def test_extract_valid_transaction_params(transaction_params, expected):
    valid_transaction_params = extract_valid_transaction_params(transaction_params)
    assert valid_transaction_params == expected
```

This also nicely sums up the the functionality side-by-side, IMO.

tests/core/utilities/test_valid_transaction_params.py
```diff
@@ -38,20 +38,49 @@ def test_assert_valid_transaction_params_invalid_param():
...
+def test_extract_valid_transaction_params_4(transaction_params, expected):
+    valid_transaction_params = extract_valid_transaction_params(transaction_params)
+    assert valid_transaction_params == expected
+
+
+
+@pytest.mark.xfail(raises=AttributeError, strict=True)  # noqa E302
+def test_extract_valid_transaction_params_5():
+    transaction_params = {'data': '0x0', 'input': '0x1'}
+    extract_valid_transaction_params(transaction_params)
```
carver: Sometimes we combine the exceptional cases into the standard cases, but I'm personally fine with having this be a separate test, which keeps the test body of both tests simple.

However, we have been using `xfails` exclusively for known bugs. Testing that we intentionally raise exceptions should be done with `with pytest.raises(AttributeError):` instead. Bonus points if you use the `match` keyword arg to confirm that the words `data` and `input` are included. (loose enough to let us modify the text, but strict enough to make sure we are clear enough about what happened)

web3/utils/transactions.py
```diff
@@ -92,8 +92,28 @@ def get_required_transaction(web3, transaction_hash):
...
+        else:
+            return extracted_params
+    elif extracted_params.get('data') is None:
+        if transaction_params.get('input') is not None:
+            extracted_params['data'] = transaction_params['input']
+            return extracted_params
+        else:
+            return extracted_params
+    else:
+        return extracted_params
```
carver: Cool, this whole structure looks very complete, and easy to verify all the paths :+1: 

Since this last else block is an unreachable code path, let's just raise an exception.

web3/utils/transactions.py
```diff
@@ -92,8 +92,28 @@ def get_required_transaction(web3, transaction_hash):
...
+                err_vals = (transaction_params['input'], extracted_params['data'])
+                raise AttributeError(msg.format(*err_vals))
+            else:
+                return extracted_params
+        else:
+            return extracted_params
+    elif extracted_params.get('data') is None:
+        if transaction_params.get('input') is not None:
+            extracted_params['data'] = transaction_params['input']
+            return extracted_params
```
carver: Just one last piece of mutability that we can eliminate and collapse into one line:

```py
return assoc(extracted_params, 'data', transaction_params['input'])
```

tests/core/utilities/test_valid_transaction_params.py
```diff
@@ -38,20 +38,49 @@ def test_assert_valid_transaction_params_invalid_param():
...
 def test_extract_valid_transaction_params():
+    ''' baseline test for clean param dict '''
     input = {
         'from': '0x0',
         'to': '0x0',
         'gas': 21000,
         'gasPrice': 5000000,
-        'value': 1,
         'data': '0x0',
+        'value': 1,
```
carver: Moving `value` here wasn't meaningful, right? If it affects the test somehow, we've got a bigger problem.

boneyard93501:
@carver  thanks for bearing with me and the feedback. feel free to squash it .

web3/utils/transactions.py
boneyard93501: @ carver i knew i forgot something ... AttributeError or general Exception ?
```python
exc_msg = 'extract_valid_transaction_params failure for transaction parameters: {}'
Exception({exc_msg".format(transaction_parameters))
```

Issue #719, opened 2018-03-21
EthCC 2018 Paris Workshop Results


carver:
My apologies to the workshop group. You unintentionally became documentation testers. The good news is that your struggles were not in vain! I marked down what I could about where people got stuck, and we'll see what we can do about improving the situation.

### What was wrong?

1. Old version of pip, setuptools
2. Run virtualenv setup in python console
3. Virtualenv not installed
4. Using python 2, ended up with web3 v3, because there is no v4 version available
5. Using web3 v3 (missed "--pre" during install)
6. Connection/`Provider` is very confusing for people: why do they have to connect to anything? how do they specify what to connect to? what are all these different providers for and why are we making them pick one?
7. Tried to use web3.auto for infura
8. Can't deploy via infura
9. Faucet down
10. Ethtoken example only works on mainnet
11. Ethtoken docs did not show how to customize w3
12. People not clear when w3 is calling the node and when working locally
13. Didn't know how to fix address checksum error
14. Pip wasn't installed
15. Could not understand solc missing error (installed separate from pysolc)

### How can it be fixed?

1. ~~Add pip & setuptools update to quickstart instructions~~
2. ~~Clearer distinction between shell and python interpreter~~
3. ~~Add (link to) instructions for installing `virtualenv`~~
4. ~~Use the new pip setup.py attribute `python_requires` to give a cleaner error message~~
5. ~~Wait for v4 release~~
6. ~~Deep dive on basic questions, like "Why do I need to connect to a node?"~~
7. Maybe link to new "choosing a node" docs in exception when `web3.auto.w3` can't find the node?
8. ~~Wait for #666~~
9. ~In docs, link to a list of several faucets for different testnets. (There's probably already a good ESE post with this~
10. Write a working ethtoken example for testnet
1. Add ethtoken docs for customizing w3 in ethtoken for testnet
2. In v5, consider a separate `Web3.utils` attribute that is clearly never hitting the network (can host things like: type conversions, sha3, eth-account) #722 (similar to web3.js v1 approach)
3. ~#716 -> #803~
4. ~Add (link to) pip install instructions for a variety of platforms~
5. https://github.com/ethereum/py-solc/issues/40

PR #795, opened 2018-04-27
Documentation updates inspired by EthCC 2018 Workshop

carver:
### What was wrong?

Related to Issue #719 

People are still pretty confused. Trying to do better, at least for the people who actually read the docs.

### How was it fixed?

A lot of new documentation. Simplify the quickstart a bit, but link out to bigger explanations for common problems. Some explainers/tutorials on what Web3.py (and Ethereum) are all about.

- [x] Add pip & setuptools update to quickstart instructions
- [x] Clarify distinction between shell and python interpreter
- [x] Add (link to) instructions for installing virtualenv
- [x] Use the new pip setup.py attribute python_requires to give a cleaner error message
- [x] Deep dive on basic questions, like "Why do I need to connect to a node?"
- [x] In docs, link to a list of several faucets for different testnets. (There's probably already a good ESE post with this)
- [x] Add (link to) pip install instructions for a variety of platforms

#### Cute Animal Picture

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -10,26 +10,32 @@ check out this example:
 
 .. _contract_example:
 
-Simple Contract Example
------------------------
+Contract Deployment Example
+----------------------------------------------
+
+To run this example, you will need to install a few extra features:
+
+- The sandbox node provided by eth-tester. You can install it with ``pip install -U web3[tester]``.
+- The ``solc`` solidity compiler. See `Installing the Solidity Compiler
+  <http://solidity.readthedocs.io/en/latest/installing-solidity.html#binary-packages>`_
 
 .. code-block:: python
 
     import json
     import web3
 
-    from web3 import Web3, TestRPCProvider
+    from web3 import Web3
     from solc import compile_source
     from web3.contract import ConciseContract
 
     # Solidity source code
     contract_source_code = '''
-    pragma solidity ^0.4.0;
+    pragma solidity ^0.4.21;
 
     contract Greeter {
         string public greeting;
 
-        function Greeter() {
+        function Greeter() public {
             greeting = 'Hello';
         }
 
@@ -37,7 +43,7 @@ Simple Contract Example
             greeting = _greeting;
         }
 
-        function greet() constant returns (string) {
+        function greet() view public returns (string) {
             return greeting;
         }
     }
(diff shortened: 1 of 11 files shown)


docs/quickstart.rst
```diff
@@ -3,33 +3,24 @@ Quickstart
...
-    # with virtualenv active, install...
+.. NOTE:: All code starting with a ``$`` is meant to run on your terminal.
+    All code starting with a ``>>>`` is meant to run in a python interpreter,
+    like `ipython <https://pypi.org/project/ipython/>`_.
 
 Installation
 ------------
 
-Web3.py can be installed using ``pip`` as follows.
+Web3.py can be installed (preferably in a virtualenv) using ``pip`` as follows:
```
pipermerriam: making `virtualenv` a link to the `setup-environment` ref might help

docs/contracts.rst
```diff
@@ -1,8 +1,72 @@
...
+Simple Contract Example
+-----------------------
+
+.. code-block:: python
+
+    import json
+    import web3
+
+    from web3 import Web3, TestRPCProvider
+    from solc import compile_source
```
voith: Most beginners just install `py-solc` without installing the solidity complier. Should there be a note pointing to [Installing the Solidity Compiler](solidity.readthedocs.io/en/latest/installing-solidity.html)?

docs/contracts.rst
```diff
@@ -1,8 +1,72 @@
...
+    contract_interface = compiled_sol['<stdin>:Greeter']
+
+    # web3.py instance
+    w3 = Web3(TestRPCProvider())
+
+    # Instantiate and deploy contract
+    contract = w3.eth.contract(abi=contract_interface['abi'], bytecode=contract_interface['bin'])
+
+    # Get transaction hash from deployed contract
+    tx_hash = contract.deploy(transaction={'from': w3.eth.accounts[0], 'gas': 410000})
```
voith: if `deploy` is being deprecated, why is its use promoted here?

docs/contracts.rst
```diff
@@ -1,8 +1,72 @@
...
+
+    # Get tx receipt to get contract address
+    tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
```
voith: May be use `waitForTransactionReceipt` instead? I think it would be better to give an example that would work on the main network too.

docs/contracts.rst
carver: Thanks for checking this. :+1:  I was moving the example out of the quickstart, and hadn't dug into the code itself. Now is a good time to clean it up.

Issue #698, opened 2018-03-14
Better error message if `abi` is not provided for contract


pipermerriam:
* Version: 3.x and 4.x

### What was wrong?

In `3.x` if no ABI is provided when instantiating a contract instance you would get an unhelpful error message as seen here: https://github.com/ethereum/web3.py/issues/696

In `4.x` this type of error won't occur, however, the situation can still be approved because the `contract.functions` API simply won't have any accessible functions.

### How can it be fixed?

I don't think we need to fix `3.x` but we *could* provide a simple patch to raise a better error message if `self.abi` is `None`.

In `4.x` I think that we can improve the situation.  Maybe accessing `contract.functions` when `abi is None` should raise an exception that no contract ABI was provided (same for accessing `contract.events`).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/698/312).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $4,678.51 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

nemaniarjun:
Hey! I am new to Gitcoin, Web3 and blockchain in general. I do have python coding experience and I think I can get this done.

For `3.x` : Add an Exception with a proper message (here)[https://github.com/ethereum/web3.py/blob/1cb19a0e2116d3d2a7075b28683290d3b60c76c7/web3/contract.py#L186]

For: `4.x`: This is bit more complex, on my first go through the code, I think raising exceptions on creation of `ContractFunction` and `ContractEvent` Class objects should be enough.

I see for testing, PyTest is being used, I will write tests for the changes I make :)

Please let me know if this plan seems suitable, I will start reading the docs and more about web3.py, web3.js and Ethereum in general.

pipermerriam:
Yes, feel free to take this on.

For `4.x` (which is `master`) the best I can think of is to raise a custom `AttributeError` when an attempt is made to access a function via `c.functions.myFunction` in the case where the `abi` contains no function definitions.

```python
[...]

PR #800, opened 2018-04-29
Contract.py: `ABI` not provided error message

nemaniarjun:
Now raises TypeError if ABI is not provided with the contract.
### What was wrong?

Part of #698 

### How was it fixed?
Added check for abi, if `None` raises `TypeError`

#### Cute Animal Picture


pipermerriam:
Closing in favor of #801 

It's valid to instantiate a contract without an `abi`.  The use case for this is a contract which you **only** intend to deploy.

PR #801, opened 2018-04-30
Better error messaging when no ABI is provided.

mikery:
### What was wrong?

Related to Issue #698 

### How was it fixed?

If the user attempts to access a function/event when there is no ABI, or there are no functions/events defined in the ABI, raise `NoABIFunctionsFound`/`NoABIEventsFound` respectively. Attributes are checked/accessed through `self.__dict__` to avoid recursion errors.

If the user attempts to access a function/event that does not exist (but we do have an ABI with functions/events), raise `MismatchedABI`.

It might also make sense to have `MismatchedABI` inherit from `AttributeError`, so that anyone who was relying on `AttributeError` being raised when calling a non-existent function can continue using the same code.

#### Cute Animal Picture

Cat tax


pipermerriam:
> It might also make sense to have MismatchedABI inherit from `AttributeError`, so that anyone who was relying on `AttributeError` being raised when calling a non-existent function can continue using the same code.

I'm not inclined to do this since it would make using this exception for anything other than attribute access *wrong-ish*.  For example, you can lookup functions as `contract.functions['myFunction']`.  Being able to catch that exception with an `AttributeError` feels wrong to me.

web3/contract.py
```diff
@@ -117,6 +119,19 @@ def __iter__(self):
...
+            raise NoABIFunctionsFound(
+                "The abi for this contract contains no function definitions. ",
+                "Are you sure you provided the correct contract abi?"
+            )
+        elif function_name not in self.__dict__['_functions']:
+            raise MismatchedABI(
+                "The function '{}' was not found in this contract's abi. ".format(function_name),
+                "Are you sure you provided the correct contract abi?"
+            )
+        return getattr(self, function_name)
```
pipermerriam: This seems like it should be a `super()` call.

web3/contract.py
```diff
@@ -143,6 +158,19 @@ def __init__(self, abi, web3, address=None):
...
+            raise NoABIEventsFound(
+                "The abi for this contract contains no event definitions. ",
+                "Are you sure you provided the correct contract abi?"
+            )
+        elif event_name not in self.__dict__['_events']:
+            raise MismatchedABI(
+                "The event '{}' was not found in this contract's abi. ".format(event_name),
+                "Are you sure you provided the correct contract abi?"
+            )
+        return getattr(self, event_name)
```
pipermerriam: same here. should this be a `super()` call?

web3/contract.py
mikery: Actually it should raise `AttributeError`. `__getattr__` is only called when the normal attribute lookup fails. If we reach this point the lookup has already failed, we are just deciding which error message to show. I'll push an update with this behaviour.

pipermerriam: My thought was that the `super()` would indeed raise an `AttributeError` but that we shouldn't presume that there isn't additional `__getitem__` logic to be executed.

mikery: Makes sense. I didn't add it initially because both `__init__` and `__getitem__` already presume this is a top-level class and do not call `super()`.

I'll need to implement it as `return super().__getattribute__(function_name)` - because it is a top-level class `__getattr__` is not exposed through super calls. If `super().__getattribute__` can't find the attribute it will fall back to `super().__getattr__` if it is defined, so the effect is identical.

PR #807, opened 2018-05-01
eth-tester upgrade

carver:
### What was wrong?

Stuck on old versions of eth-tester/py-evm/pyrlp.

### How was it fixed?

Use the latest ones.

#### Cute Animal Picture


carver:
The connection to the docs is that I had `grep`d for `rlp` usages. That example was the only one.

PR #809, opened 2018-05-03
conditionally install cytoolz or toolz  based on python implementation

voith:
### What was wrong?
@pipermerriam This is for you 😃 
addresses https://github.com/ethereum/web3.py/pull/789#issuecomment-385447090

### How was it fixed?

Conditionally install cytoolz or toolz  based on python implementation. 

In my first attempt I tried to avoid imports from `dicttoolz`, `curried` by importing directly from `cytoolz` or `toolz`. However `from cytoolz import keymap` is different from `from cytoolz.curried import keymap`. So I kept the import order the same as before: `from web3.utils.toolz.dictoolz ...`

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -72,7 +72,7 @@ from web3.utils.normalizers import (
     normalize_address,
     normalize_bytecode,
 )
-from web3.utils.toolz.functoolz import (
+from web3.utils.toolz import (
     compose,
     partial,
 )
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -17,10 +17,6 @@ from eth_utils import (
     is_text,
     to_tuple,
 )
-from toolz.functoolz import (
-    compose,
-    partial,
-)
 
 from web3.exceptions import (
     BadFunctionCallOutput,
@@ -75,6 +71,10 @@ from web3.utils.normalizers import (
     normalize_abi,
     normalize_address,
     normalize_bytecode,
+)
+from web3.utils.toolz.functoolz import (
+    compose,
+    partial,
 )
 from web3.utils.transactions import (
     fill_transaction_defaults,
--- a/web3/utils/toolz.py
+++ b/web3/utils/toolz.py
@@ -1,0 +1,47 @@
+import sys
+
+try:
+    from cytoolz import (
+        assoc,
+        complement,
+        compose,
+        curried,
+        curry,
+        dicttoolz,
+        dissoc,
+        functoolz,
+        groupby,
+        identity,
+        itertoolz,
+        merge,
+        partial,
+        pipe,
+        sliding_window,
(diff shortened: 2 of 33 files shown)


voith:
I don't know how good the test coverage is. I haven't tested this manually, so if anything breaks blame @pipermerriam 😂

web3/utils/toolz.py
```diff
@@ -0,0 +1,47 @@
...
+        pipe,
+        sliding_window,
+        valmap,
+    )
+
+
+modules = [curried, dicttoolz, functoolz, itertoolz]
+for module in modules:
+    module_name = module.__name__.split('.')[-1]
+    sys.modules['web3.utils.toolz.{0}'.format(module_name)] = module
```
voith: had to do this to make imports like`from web3.utils.toolz.curried import ...` work.

web3/contract.py
```diff
@@ -17,10 +17,6 @@
     is_text,
     to_tuple,
 )
-from toolz.functoolz import (
```
voith: Also replaced imports where toolz was used instead of cytoolz.
Is there any good reason why cytoolz wasn't used?
I am relying on unit tests to tell me if there's something incompatible between cytoolz and toolz.

pipermerriam: There was a historical issue where `cytoolz.compose` threw an error if you wrapped it in a `functools.partial`.  I think this has been resolved now.

web3/utils/toolz.py
pipermerriam: I think I'd prefer to remove this `sys.modules` stuff and just create two modules so that all imports come from either.
```
from web3.utils.toolz import (...)
from web3.utils.toolz.curried import (...)
```

This would be done with the following directory structure

```
./web3/utils/toolz/__init__.py  # place all `web3.utils.toolz` imports here
./web3/utils/toolz/curried.py  # place all `web3.utils.toolz.curried` imports here
```

Any of the places where we were doing `from cytoolz.functoolz import thing` can be converted to just `from web3.utils.toolz import thing`

voith: Yeah, I was just lazy to create separate modules for each and I also wanted to look cool. 
But +1 for simplicity!

web3/contract.py
voith: Ok, I'll test this `toolz` stuff manually just to be sure!

web3/utils/toolz.py
pipermerriam: TIL you can shove modules into `sys.modules` and then import them.  Fancy.

web3/contract.py
voith: Done! Tested all occurrences of toolz imports manually. Did not find any incompatibility issues!

PR #827, opened 2018-05-08
Fix address being added twice with `createFilter`

### LINKED ISSUE #826: createFilter event syntax does not work while eventFilter works
* Version: 4.2.0
* Python: 3.6
* OS: linux

### What was wrong?
Filtering for an event with Syntax 2 does not work, while the deprecated Syntax 1 works:

Syntax 1:
```python
event_filt= instance.eventFilter(
        'EventName',
        {
            'fromBlock': 0,
            'toBlock': 'latest',
        })
```

Syntax 2
```python
event_filt = instance.events.EventName.createFilter(fromBlock=0, toBlock='latest')
```

### How can it be fixed?
By inspecting `event_filt.filter_params`, when using Syntax 2, the `address` field is populated with a list which includes the contract address twice. This happens because in the following snippet, the call is made to `construct_event_filter_params` with both `address` and `contract_address` which have the same value. 

https://github.com/ethereum/web3.py/blob/ee8e217fec8231e6262473d1f72fa9dad68ff3c8/web3/contract.py#L1211-L1219

Commenting out contract.py:1217 solves the issue for me. 
If you are using ganache as a node you need to monkey-patch web3 to bypass trufflesuite/ganache-cli#494, and create the contract object with a lowercase address https://github.com/ethereum/web3.py/issues/674, https://github.com/jordanjambazov/web3.py/commit/a61b382dd5c11de1102779868d377408c7590839

gakonst:
Fixes #826

### DIFF
--- a/tests/core/filtering/test_contract_on_event_filtering.py
+++ b/tests/core/filtering/test_contract_on_event_filtering.py
@@ -1,7 +1,27 @@
 import pytest
 
+from eth_utils import (
+    is_address,
+)
+
 # Ignore warning in pyethereum 1.6 - will go away with the upgrade
 pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
+
+
+@pytest.mark.parametrize('call_as_instance', (True, False))
+def test_create_filter_address_parameter(web3, emitter, Emitter, call_as_instance):
+    if call_as_instance:
+        event_filter = emitter.events.LogNoArguments.createFilter(fromBlock="latest")
+    else:
+        event_filter = Emitter.events.LogNoArguments.createFilter(fromBlock="latest")
+
+    if call_as_instance:
+        # Assert this is a single string value, and not a list of addresses
+        assert isinstance(event_filter.filter_params['address'], str)
+        assert is_address(event_filter.filter_params['address'])
+    else:
+        #  Undeployed contract shouldnt have address...
+        assert 'address' not in event_filter.filter_params
 
 
 @pytest.mark.parametrize('call_as_instance', (True, False))
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1296,9 +1296,6 @@ class ContractEvent:
         if argument_filters is None:
             argument_filters = dict()
 
-        if not address:
-            address = self.address
-
         _filters = dict(**argument_filters)
 
         data_filter_set, event_filter_params = construct_event_filter_params(
@@ -1307,7 +1304,7 @@ class ContractEvent:
             argument_filters=_filters,
             fromBlock=fromBlock,
             toBlock=toBlock,
-            # address=address,
+            address=address,
             topics=topics,
         )
 
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1307,7 +1307,7 @@ class ContractEvent:
             argument_filters=_filters,
             fromBlock=fromBlock,
             toBlock=toBlock,
-            address=address,
+            # address=address,
             topics=topics,
         )
 


carver:
If this is the route to go, then the line should be deleted instead of commented out. I'm curious to hear what Dylan thinks.

carver:
Sweet. One last thing I would love to see: can we reproduce this in a test?

Could be a core test, but maybe even an integration test, since we have some that are already emitting events IIRC.

dylanjw:
Added a test for the double address, shown to fail on `master` below:

```
tests/core/filtering/test_contract_on_event_filtering.py::test_create_filter_address_parameter[True] FAILED
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

web3 = <web3.main.Web3 object at 0x7f8f02c92160>, emitter = <web3.utils.datatypes.Contract object at 0x7f8f053a0160>, Emitter = <class 'web3.utils.datatypes.Contract'>, call_as_instance = True

    @pytest.mark.parametrize('call_as_instance', (True, False))
    def test_create_filter_address_parameter(web3, emitter, Emitter, call_as_instance):
        if call_as_instance:
            event_filter = emitter.events.LogNoArguments.createFilter(fromBlock="latest")
        else:
            event_filter = Emitter.events.LogNoArguments.createFilter(fromBlock="latest")
    
        if call_as_instance:
            # Assert this is a single string value, and not a list of addresses
>           assert isinstance(event_filter.filter_params['address'], str)
E           AssertionError: assert False
E            +  where False = isinstance(['0xF2E246BB76DF876Cef8b38ae84130F4F55De395b', '0xF2E246BB76DF876Cef8b38ae84130F4F55De395b'], str)

Emitter    = <class 'web3.utils.datatypes.Contract'>
call_as_instance = True
emitter    = <web3.utils.datatypes.Contract object at 0x7f8f053a0160>
event_filter = <web3.utils.filters.LogFilter object at 0x7f8f02c1a0f0>
web3       = <web3.main.Web3 object at 0x7f8f02c92160>

tests/core/filtering/test_contract_on_event_filtering.py:16: AssertionError
```

carver:
I'm sure the parity failures aren't related (see #882 ), but our commit history will be much cleaner if we rebase on master. @dylanjw your test is getting buried in that merge commit.

tests/core/filtering/test_contract_on_event_filtering.py
```diff
@@ -1,9 +1,26 @@
...
+def test_create_filter_address_parameter(web3, emitter, Emitter, call_as_instance):
+    if call_as_instance:
+        event_filter = emitter.events.LogNoArguments.createFilter(fromBlock="latest")
+    else:
+        event_filter = Emitter.events.LogNoArguments.createFilter(fromBlock="latest")
+
+    if call_as_instance:
+        # Assert this is a single string value, and not a list of addresses
+        assert isinstance(event_filter.filter_params['address'], str)
+        assert is_address(event_filter.filter_params['address'])
```
carver: Do these asserts need to be in the `if` check? Right now it's not asserting anything if `call_as_instance` is `False`.

dylanjw: If call_as_instance is False, I think the contract is not deployed, so there should be no address field in the parameters.  I suppose I should assert that as well.

dylanjw: ...but wait.  If the class contract is an undeployed contract with no address, how are transactions sent.  :confused:

dylanjw: because you cant.

```
(Pdb) Emitter.address
(Pdb) emitter.address
'0xF2E246BB76DF876Cef8b38ae84130F4F55De395b'
(Pdb)
(Pdb) Emitter.functions.logString("hi").transact()
*** ValueError: Please ensure that this contract instance has an address.
(Pdb) 
```

worldview restored. :relieved:

PR #828, opened 2018-05-08
Support fixedMxN ABI type

carver:
### What was wrong?

No support for Fixed ABI type (needed by Casper)

### How was it fixed?

- upgrade to eth-abi v1.1
- Move most of the is_encodable test to eth-abi
- Special handling of ENS addresses and hex-encoded bytes values specially
- Update minimum allowed int256 value in tests (from `-2 ** 255 + 1` to `-2 ** 255`)

#### Cute Animal Picture

### DIFF
--- a/tests/core/utilities/test_abi_is_encodable.py
+++ b/tests/core/utilities/test_abi_is_encodable.py
@@ -21,8 +21,8 @@ from web3.utils.abi import (
         (True, 'bytes32', False),  # no wrong types
         (0, 'bytes32', False),  # no wrong types
         # int
-        (-1 * 2**255, 'int256', False),
-        (-1 * 2**255 + 1, 'int256', True),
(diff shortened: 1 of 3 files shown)


carver:
Ok, think it's ready for a real review @pipermerriam

tests/core/utilities/test_abi_is_encodable.py
```diff
@@ -21,8 +21,8 @@
         (True, 'bytes32', False),  # no wrong types
         (0, 'bytes32', False),  # no wrong types
         # int
-        (-1 * 2**255, 'int256', False),
-        (-1 * 2**255 + 1, 'int256', True),
+        (-1 * 2**255 - 1, 'int256', False),
```
davesque: Were you meaning to check the positive bound here?

Issue #825, opened 2018-05-08
generateGasPrice division by zero exception


kolosovvs:
At this method generateGasPrice() I have division by zero exception
It's because `_get_avg_block_time` will return zero in some cases
```
def _get_avg_block_time(w3, sample_size):
    latest = w3.eth.getBlock('latest')

    if latest['number'] == 0 or sample_size == 0: # <-- here
        return 0

    constrained_sample_size = min(sample_size, latest['number'])
    oldest = w3.eth.getBlock(latest['number'] - constrained_sample_size)
    return (latest['timestamp'] - oldest['timestamp']) / constrained_sample_size
```

pipermerriam:
This should be a reasonably easy thing to fix.  I think that if the average block time returned is `0` then we need to default to a *safe* default block time.  For now I'm fine with this being hard coded to `15` unless someone has a better idea.

PR #838, opened 2018-05-11
Raising exception when constrained_sample_size becomes 0

glaksmono:
### What was wrong?
`_get_avg_block_time` will return zero in some cases

Fix https://github.com/ethereum/web3.py/issues/825

### How was it fixed?
For now we can hardcode this to 15 per suggestion

#### Cute Animal Picture

### DIFF
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -1,6 +1,9 @@
 import pytest
 
 from web3 import Web3
+from web3.exceptions import (
+    ValidationError,
+)
 from web3.gas_strategies.time_based import (
     construct_time_based_gas_price_strategy,
 )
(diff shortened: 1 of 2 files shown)


web3/gas_strategies/time_based.py
```diff
@@ -20,7 +20,8 @@ def _get_avg_block_time(w3, sample_size):
     latest = w3.eth.getBlock('latest')
 
     if latest['number'] == 0 or sample_size == 0:
-        return 0
+        # a safe default block time
+        return 15
```
pipermerriam: @carver what do you think about allowing this to be configured via environment variable?

pipermerriam: cc @dylanjw too

carver: I think we should raise if you try to construct a 0 sample size middleware: https://github.com/glaksmono/web3.py/blob/fb464356a644445559d798258e585f13aa16964f/web3/gas_strategies/time_based.py#L123

Otherwise, estimating with 0 blocks available seems too trivial to worry about otherwise. I can't imagine anyone bothering to set an env variable for what to do when estimating gas for only the first block.

web3/gas_strategies/time_based.py
```diff
@@ -20,7 +20,8 @@ def _get_avg_block_time(w3, sample_size):
...
 
     constrained_sample_size = min(sample_size, latest['number'])
```
carver: Maybe the check could be moved to a simpler: `if constrained_sample_size == 0`?

web3/gas_strategies/time_based.py
```diff
@@ -19,10 +19,10 @@
 def _get_avg_block_time(w3, sample_size):
     latest = w3.eth.getBlock('latest')
 
-    if latest['number'] == 0 or sample_size == 0:
-        return 0
-
     constrained_sample_size = min(sample_size, latest['number'])
+    if constrained_sample_size == 0:
+        raise ZeroDivisionError('Constrained sample size is 0')
```
pipermerriam: I think I'd use a `web3.exceptions.ValidationError` here.  Since this isn't actual division by zero, but rather pre-validation.

glaksmono:
@carver added the tests on https://github.com/ethereum/web3.py/pull/838/commits/5f93dbed0f81dd181fcca6c0fd5a0a11ce13cb69

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
```diff
@@ -139,3 +142,56 @@ def test_time_based_gas_price_strategy(strategy_params, expected):
...
+        (dict(max_wait_seconds=40, sample_size=0, probability=50),
+            'Constrained sample size is 0'),
+        # 20 second wait times, 0 sample_size
+        (dict(max_wait_seconds=20, sample_size=0, probability=98),
+            'Constrained sample size is 0'),
+        (dict(max_wait_seconds=20, sample_size=0, probability=90),
+            'Constrained sample size is 0'),
+        (dict(max_wait_seconds=20, sample_size=0, probability=50),
+            'Constrained sample size is 0'),
+    ),
```
carver: Hah, maybe a little overkill with all these different other parameter values, but it's not worth holding up the PR to tighten this up.

dangell7:
FWI this is "Constrained sample size is 0" is being thrown on a brand new truffle-cli aka a 0 block test rpc....

dangell7:
Also now receiving a "tuple index out of range". Are you guys using afre

> FWI this is "Constrained sample size is 0" is being thrown on a brand new truffle-cli aka a 0 block test rpc....

Nevermind, I found the solution in the issues and docs. For future, I was forced to set the gasprice if the block was 0. IDK if you want to handle that as this is only a testing env and in prod you would never see that issue. But it does make me wonder how you guys test? Do you use ganache or ropsten or your own chain?

carver:
The best recommendation I remember hearing was to fall back to an RPC `eth_gasPrice` call if there are 0 blocks available.

PR #846, opened 2018-05-15
Use pytest instead of py.test per upstream recommendation, #dropthedot

scop:
### What was wrong?

Used `py.test` to run tests.

### How was it fixed?

By using `pytest` instead.

http://blog.pytest.org/2016/whats-new-in-pytest-30/
https://twitter.com/hashtag/dropthedot

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ lint-roll:
 	$(MAKE) lint
 
 test:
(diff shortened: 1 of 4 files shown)


pipermerriam:
test failure seems unrelated. re-running.

PR #861, opened 2018-05-23
Renamed getUncleFromBlock and added alias to getTransactionFromBlock

ankitchiplunkar:
### What was wrong?

The method names were not consistent with the JSON-RPC method names. 
For more details refer to this conversation: https://github.com/ethereum/web3.py/pull/857#pullrequestreview-121916432

### How was it fixed?
1. Changed the name of `getUncleFromBlock` method to `getUncleByBlock`
2. Added alias to the method `getTransactionFromBlock`
3. Added tests for the alias

#### Cute Animal Picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -197,6 +197,13 @@ class Eth(Module):
 
     def getTransactionFromBlock(self, block_identifier, transaction_index):
         """
+        Alias for the method getTransactionByBlock
+        Depreceated to maintain naming consistency with the json-rpc API
+        """
+        return self.getTransactionByBlock(block_identifier, transaction_index)
+
+    def getTransactionByBlock(self, block_identifier, transaction_index):
+        """
         `eth_getTransactionByBlockHashAndIndex`
         `eth_getTransactionByBlockNumberAndIndex`
         """
@@ -210,13 +217,6 @@ class Eth(Module):
             method,
             [block_identifier, transaction_index],
         )
-
-    def getTransactionByBlock(self, block_identifier, transaction_index):
-        """
-        Alias for the method getTransactionFromBlock
-        More details here: https://github.com/ethereum/web3.py/pull/857
-        """
-        return self.getTransactionFromBlock(block_identifier, transaction_index)
 
     def waitForTransactionReceipt(self, transaction_hash, timeout=120):
         return wait_for_transaction_receipt(self.web3, transaction_hash, timeout)
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -173,7 +173,7 @@ class Eth(Module):
             [block_identifier],
         )
 
-    def getUncleFromBlock(self, block_identifier, uncle_index):
+    def getUncleByBlock(self, block_identifier, uncle_index):
         """
         `eth_getUncleByBlockHashAndIndex`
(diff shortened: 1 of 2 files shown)


web3/utils/module_testing/eth_module.py
```diff
@@ -501,16 +501,21 @@ def test_eth_getTransactionByHash_contract_creation(self,
...
+    def test_eth_getTransactionFromBlockNumberAndIndex(self, web3, block_with_txn, mined_txn_hash):
         transaction = web3.eth.getTransactionFromBlock(block_with_txn['number'], 0)
         assert is_dict(transaction)
         assert transaction['hash'] == HexBytes(mined_txn_hash)
 
+    def test_eth_getTransactionByBlockNumberAndIndex(self, web3, block_with_txn, mined_txn_hash):
+        transaction = web3.eth.getTransactionByBlock(block_with_txn['number'], 0)
+        assert is_dict(transaction)
+        assert transaction['hash'] == HexBytes(mined_txn_hash)
+
```
carver: Could you add one more for `test_eth_getTransactionByBlockHashAndIndex`?

web3/eth.py
```diff
@@ -211,6 +211,13 @@ def getTransactionFromBlock(self, block_identifier, transaction_index):
             [block_identifier, transaction_index],
         )
 
+    def getTransactionByBlock(self, block_identifier, transaction_index):
+        """
+        Alias for the method getTransactionFromBlock
+        More details here: https://github.com/ethereum/web3.py/pull/857
+        """
+        return self.getTransactionFromBlock(block_identifier, transaction_index)
```
carver: I think it's cleaner to put the implementation in this new method, and alias from the deprecated version to this implementation . Then removing the deprecated name in v5 will be as simple as deleting the deprecated function.

web3/eth.py
```diff
@@ -211,6 +211,13 @@ def getTransactionFromBlock(self, block_identifier, transaction_index):
             [block_identifier, transaction_index],
         )
 
+    def getTransactionByBlock(self, block_identifier, transaction_index):
+        """
+        Alias for the method getTransactionFromBlock
+        More details here: https://github.com/ethereum/web3.py/pull/857
```
carver: I would recommend putting the full description from `getTransactionFromBlock` here, and then the docs in `getTransactionFromBlock` can say that they are an alias to `getTransactionByBlock`. If you want to include the reason why it's deprecated, I think it can be summarized neatly as "naming consistency with the json-rpc API" (rather than linking to the PR thread). Linking to a github PR feels wrong, but I can't put my finger on why.

web3/utils/module_testing/eth_module.py
ankitchiplunkar: Done

web3/eth.py
ankitchiplunkar: Agreed, have made the changes

web3/eth.py
ankitchiplunkar: Made the corrections in descriptions.

> Linking to a github PR feels wrong, but I can't put my finger on why.

I have removed the link, but I personally prefer keeping the links of stackoverflow and other discussions in documentation

PR #873, opened 2018-05-30
Update parity binary for integration testing - libssl version differences

dylanjw:
### What was wrong?

Related to https://github.com/ethereum/web3.py/pull/865

### How was it fixed?
debian package version is built with libssl1.1, changing to the "debian" build rather than the "linux" build will hopefully fix the issue.

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #875, opened 2018-05-31
Fix CircleCI Parity integration failures

veox:
### What was wrong?

Parity integration tests are failing due to the used (dynamically-linked!) Parity binary not finding its dependencies.

Conflicts with PR #873.

### How was it fixed?

(There were 42 (or so) commits in the "debug" branch. It was a messy tangle, so I squashed them. Sorry for that, but I really think it's better this way.)

There are four main changes.

#### First

Debian version (Jessie, 8; or Stretch, 9) is specified excplicitly to CircleCI for Parity integration jobs.

Before this commit, Python 3.6 jobs might get run on either, which introduces a heisenbug (see "Second" point below).

(Python 3.5 jobs were always run on Jessie, since CircleCI doesn't have a Debian 9 + Python 3.5 docker image.)

After this commit, Python 3.5 jobs are _explicitly_ run on Jessie, and Python 3.6 jobs are _always_ run on Stretch.

#### Second and third

The Parity node/client version has been updated to v1.10.4.

(Note that this is a pre-compiled, _dynamically linked_ executable, downloaded from vanity-service.parity.io.)

Before this commit, when testing Parity integration on Jessie, the pre-compiled `parity` binary may have failed to start, due to the `openssl` or `libstdc++6` dependencies being too old.

Updating from v1.8.x to v1.10.y helps avoid the `openssl` dependency, as this version is no longer built against `openssl`.

After this commit, tTo work around the remaining `libstdc++6` incompatibility, Jessie builds enable the Debian 'testing' repository, and update the package from there.

#### Fourth

Some tests are marked flaky.

It seems that the Parity v1.10.x release family do things faster or in a more asynchronous way than v1.{8,9}.

Before this commit, two tests in the 'eth' module would sporadically fail:

* `test_eth_call_old_contract_state(...)` - about 1 in 3 times;
* `test_eth_getTransactionReceipt_unmined(...)` - about 1 in 10 times.

After this commit, the tests are allowed to succeed just once in 3 runs.

#### Cute Animal Picture

Squash!

Source: [/r/aww](https://www.reddit.com/r/aww/comments/18lc91/my_guinea_pigs_stunt_double_is_this_butternut/)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ jobs:
       - image: circleci/python:3.5
     environment:
       TOXENV: py35-integration-parity-ipc
-      PARITY_VERSION: v1.8.8
+      PARITY_VERSION: v1.9.7
 
   py35-integration-parity-http:
     <<: *parity_steps
@@ -178,7 +178,7 @@ jobs:
       - image: circleci/python:3.5
     environment:
       TOXENV: py35-integration-parity-http
-      PARITY_VERSION: v1.8.8
+      PARITY_VERSION: v1.9.7
 
   py35-integration-parity-ws:
     <<: *parity_steps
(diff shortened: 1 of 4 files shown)


veox:
Ping @dylanjw as conflicting PR.

PR #886, opened 2018-06-05
Replace travis ci build status with circle ci

dylanjw:
### What was wrong?
README show build failing.

### How was it fixed?
Show circleci status

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Join the chat at https://gitter.im/ethereum/web3.py](https://badges.gitter.im/ethereum/web3.py.svg)](https://gitter.im/ethereum/web3.py?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-[![Build Status](https://travis-ci.org/ethereum/web3.py.png)](https://travis-ci.org/ethereum/web3.py)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Join the chat at https://gitter.im/ethereum/web3.py](https://badges.gitter.im/ethereum/web3.py.svg)](https://gitter.im/ethereum/web3.py?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-[![Build Status](https://travis-ci.org/ethereum/web3.py.png)](https://travis-ci.org/ethereum/web3.py)
+[![Build Status](https://circleci.com/gh/ethereum/web3.py.svg?style=shield&circle-token=11aaa7dbcd9d4b46f8abeed7b0cc6ed70b9c548d)](https://circleci.com/gh/ethereum/web3.py.svg?style=shield&circle-token=11aaa7dbcd9d4b46f8abeed7b0cc6ed70b9c548d)
```
carver: I don't think the token is necessary. In fact, it may have just leaked some kind of authentication info. I'd recommend logging out and back in to circle.

dylanjw: I followed their guide that said to make a read only token for the status bar.  It is interesting that it isnt required.  I'll delete the token from circleci and remove it here.

dylanjw: I see now, the token is required if the project is private.

carver: Ah, since their guide said to put it in, then I'm happy to do that. :+1: I was just surprised, and thought it might have been a copy/paste from an image you loaded in your session.

PR #890, opened 2018-06-06
middleware/pythonic: only apply formatter to 'publicKey' if not null.

veox:
### What was wrong?

While figuring out y-days `geth`/`parity` consensus failure on Ropsten, I couldn't retrieve the "malformed" block from the Parity node via `web3.py`; here's a gist of the traceback (the last line is 128 KiB long):

https://gist.github.com/veox/8a24b90b0abe4efc89e7b37c06b6bcb1#file-trace-py

(The gist also includes system info - scroll up if needed.)

The error happens with `full_transactions=True` (through `TypeError: Could not format value None as field 'publicKey'`).

### How was it fixed?

Via @carver on https://gitter.im/ethereum/web3.py?at=5b17eb9916f649612410092f:

> `publicKey` will not be `None` in correctly formed mainnet blocks (for now), since every transaction requires a signer. But, `publicKey` also isn't a standardized (consensus-critical) piece of info that's reported -- I don't think `geth` reports it at all.

#### Cute Animal Picture

Source: [d2r2 on peepeth](https://peepeth.com/d2r2/peeps/QmX6yu32aFyHGyXCN8YX2P2ckX4UfZuvmZDSp5WZoS1Uyb)

### DIFF
--- a/web3/middleware/pythonic.py
+++ b/web3/middleware/pythonic.py
@@ -92,7 +92,7 @@ TRANSACTION_FORMATTERS = {
     'gasPrice': to_integer_if_hex,
     'value': to_integer_if_hex,
     'from': to_checksum_address,
-    'publicKey': to_hexbytes(64),
+    'publicKey': apply_formatter_if(is_not_null, to_hexbytes(64)),
     'r': to_hexbytes(32, variable_length=True),
     'raw': HexBytes,
     's': to_hexbytes(32, variable_length=True),


pipermerriam:
Sounds like a pretty simple fix, however, do we need to do any negotiation on whether the key is actually present?

carver:
> do we need to do any negotiation on whether the key is actually present?

`apply_formatters_to_dict` only applies formatters if the key exists. The most complete name might be `apply_formatters_to_values_of_keys_present_in_dict`, but that's a little wordy :).

I can imagine us wanting a strict mode of the eth-utils method, but such a thing doesn't exist yet, afaik. Maybe something like: `def apply_formatters_to_dict(formatters, dict, skip_missing_keys=True)` (defaulting to `True` for backwards compatibility).

Anyhow, this should work, so I'll merge.

Issue #884, opened 2018-06-05
Permit bytes addresses for programmatic API usage


carver:
Update Web3.py to accept length=20 `bytes` values as an ABI address. This will be the only format for addresses that Web3.py accepts without a *checksum*.

- The `pythonic_middleware` and `abi_middleware` will need to be updated to handle this.
- The code used for contract parameter validation will need to be updated.
- The code related to filters may need to be changed too.
- Tests need to be written for contract and base API (`web3.eth.getBalance`, etc) use cases.
- Documentation needs to be updated to indicate that raw bytes addresses are accepted.

also see #883 

The implementation should aim to be as simple as possible, potentially normalizing to a common address format using a middleware or something rather than adjusting every API to support this new format.  Use your discretion on whether this is the right approach.

pipermerriam:
Change likely involves updating `pythonic_middleware` and `abi_middleware`

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.3 ETH (175.93 USD @ $586.42/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/884/574).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $21,419.41 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [iamonuwa](https://gitcoin.co/profile/iamonuwa) has started work. 	
 * Q: @carver @pipermerriam, you need the docs to be updated or the code itself?

pipermerriam:
@iamonuwa PR should include implementation, tests, and documentation updates.

LefterisJP:
So with this PR we will be able submit byte encoded addresses to web3.py function?

pipermerriam:
Main description of this issue has been updated with a more complete spec.  Whoever implements this will still need to do some digging to ensure they cover all of the APIs.

iamonuwa:
[...]

PR #893, opened 2018-06-09
Permitted bytes addresses in web3

voith:
### What was wrong?
Web3 did not permit bytes addresses.
see  #884

### How was it fixed?
Permitted bytes addresses in web3.

#### TODO LIST

- [x]  Add tests

- [x]  update docs


voith:
@pipermerriam @carver I have added some tests to check that my code is functionally complete.
The integration tests some parts of contracts and filters. However, I might need more tests there.
Let me know if I'm missing any tests or if you'd like me to add more tests. 
Currently, I have modified the existing tests to accept both binary and hex addresses. I'm not sure if this is the right approach. I wanted to verify that the code works, so to get it done fast I modified the existing tests. If you guys want this tested in a completely separate module then I'm fine with that. 

Also, I had some trouble with `nonce`. `w3.eth.getTransactionCount` doesn't seem to increment the count, once a transaction has been sent. I tried using `waitForTransactionReceipt` but that didn't change anything.
https://github.com/ethereum/web3.py/blob/70b7ded70c7ff324a4ad65524297f3df226ef36e/web3/utils/module_testing/eth_module.py#L223
I was getting this error with geth: 
```
ValueError: {'code': -32000, 'message': 'known transaction: 157d2598aa226207eb4711e04511266f0ad50817d6654c0a0835ae2b3dd3fa7e'}
```


In such cases I left the tests as is.

PS: I'll update docs once I get some feedback on this!

tests/core/contracts/test_contract_init.py
```diff
@@ -11,12 +15,21 @@
...
     assert deploy_receipt is not None
-    return deploy_receipt['contractAddress']
+    return deploy_receipt
+
+
+@pytest.fixture(params=['hex', 'bytes'])
+def math_addr(request, math_deploy_receipt):
+    addr = math_deploy_receipt['contractAddress']
+    if request.param == 'bytes':
+        return to_bytes(hexstr=addr)
```
carver: I considered recommending `eth_utils.to_canonical_address(addr)` here, but I find it hard to remember that canonical means the `bytes` version (to me the semantically "canonical" version of the address is the checksum version).

So I actually like `to_bytes` for clarity here.

tests/integration/test_ethereum_tester.py
```diff
@@ -63,6 +64,14 @@ def math_contract(web3, math_contract_factory, math_contract_deploy_txn_hash):
     return math_contract_factory(contract_address)
 
 
+@pytest.fixture(scope="module", params=['bytes', 'hex'])
+def math_contract_address(math_contract, request):
+    if request.param == 'bytes':
+        return math_contract.address
```
carver: conversion missing?

tests/integration/test_ethereum_tester.py
```diff
@@ -151,6 +168,21 @@ def unlocked_account(web3, unlockable_account, unlockable_account_pw):
...
+@pytest.fixture(params=['bytes', 'hex'])
+def unlockable_account_dual_type(unlockable_account, request):
+    if request.param == 'bytes':
+        return to_bytes(hexstr=unlockable_account)
+    else:
+        return unlockable_account
+
+
+@pytest.fixture
+def unlocked_account_dual_type(web3, unlockable_account_dual_type, unlockable_account_pw):
```
carver: I liked the other approach in this PR: replace the existing fixture with a parametrized one, as opposed to leaving the old one and adding a new one with a different name.

web3/utils/rpc_abi.py
```diff
@@ -24,6 +24,7 @@
 
 FILTER_PARAMS_ABIS = {
     'to': 'address',
+    'address': 'address'
```
carver: *nitpick* add trailing comma

web3/utils/rpc_abi.py
```diff
@@ -47,6 +48,9 @@
     'eth_sign': ['address', 'bytes'],
     # personal
     'personal_sendTransaction': TRANSACTION_PARAMS_ABIS,
+    'personal_lockAccount': ['address'],
+    'personal_unlockAccount': ['address', None, None],
+    'personal_sign': [None, 'address', None]
```
carver: *nitpick* add trailing comma

carver:
> `w3.eth.getTransactionCount` doesn't seem to increment the count, once a transaction has been sent

That doesn't sound right. Something else must be going on. Can you leave in an example of a broken test with `getTransactionCount`?

> I was getting this error with geth

Right, I think the integration tests don't reset state in between tests.

> In such cases I left the tests as is.

That should be fine, although it should be possible to parametrize non-state-changing tests with both kinds of addresses, like a `getBalance(address)` test.

tests/integration/test_ethereum_tester.py
voith: I had to do this because some tests were failing with the parametrized approach. Its the same issue that I've mentioned in https://github.com/ethereum/web3.py/pull/893#issuecomment-396145172.

voith:
Thanks @carver for the review!

> That doesn't sound right. Something else must be going on. Can you leave in an example of a broken test with getTransactionCount?

Quite possible that something else is going wrong. I did not spend any time digging it. But simply using paramaterized `unlocked_account` caused the test failures for me. I'll try to come up with a test case to replicate the issue tomorrow.

Besides do you have any ideas what tests I might be missing? Or any other tests that you would like to see added. Although this code won't break anything, I want to make sure that I've covered most cases where binary addresses will be used.

carver:
> Besides do you have any ideas what tests I might be missing? Or any other tests that you would like to see added.

Maybe search through this API list for `account`? The first one I found that I'm not sure is covered is `getStorageAt`.

Maybe I missed it, but tests for addresses being passed in as arguments to contract function calls.

Checking that `w3.eth.account.signTransaction` accepts transactions with from & to addresses in `bytes`.

Also, the `ens` module.

Filters: contract address, or indexed address arguments.

voith:
@carver I tried to figure out what was going wrong with [test_eth_sendTransaction_with_nonce](https://github.com/ethereum/web3.py/blob/70b7ded70c7ff324a4ad65524297f3df226ef36e/web3/utils/module_testing/eth_module.py#L215). This time I tested with parity. Like I said before `web3.eth.getTransactionCount(unlockable_account_dual_type)` returns the same value every time.
On inspection, `web3.eth.waitForTransactionReceipt(txn_hash)` returned the following
```
AttributeDict({'blockHash': None, 'blockNumber': None, 'contractAddress': None, 'cumulativeGasUsed': 21000, 'gasUsed': 21000, 'logs': [], 'logsBloom': HexBytes('0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'), 'root': '0x979928a2ecf81da90713b24af9b83cbc4ad59e7e7096ce86ff62ad1780c1d677', 'status': None, 'transactionHash': HexBytes('0x66a9cc4cf67fadda3d0691756d8c355b9ad77beb71ca1311467b6d8ad580dd01'), 'transactionIndex': 0})
```
As you can see `blockNumber` is `None` in this case. My guess is that the private chain instantiated doesn't mine(Not very sure though). @dylanjw Can you validate this statement?(since you wrote the parity tests).

Also, I got the test passing by doing the following:
```python
count = 0


def _counter():
    global count
    while True:
        yield count
        count += 1

counter = _counter()

class EthModuleTest:
   def test_eth_sendTransaction_with_nonce(self, web3, unlockable_account_dual_type):
        txn_params = {
          ...
         'nonce': web3.eth.getTransactionCount(unlockable_account_dual_type) + next(counter),
        }

```
which clearly shows that the `state` is not being updated! I'll leave such tests as is and focus on the missing tests to get this PR completed.

voith:
Checklist of tests that are needed to call this feature complete. I will update this list as I find more and point to the line number in the diff where the test was added. Hopefully, this will help the reviewer!

- [ ] `eth.getBalance`
- [ ] `eth.getStorageAt`
- [ ] `eth.getCode`
- [ ] `eth.getTransactionCount`
- [ ] `eth.sign`
- [ ] `eth.estimateGas`
- [ ] `eth.getLogs`
- [ ] `eth.newFilter`
- [ ] `eth.sendTransaction`
- [ ] `personal.lockAccount`
- [ ] `personal.unlockAccount`
- [ ] `personal.sign`
- [ ] `personal_sendTransaction`
- [ ]  `parity.traceCall`
- [ ] `eth.contract`
- [ ] `filters`
- [ ] `ens`

carver:
> My guess is that the private chain instantiated doesn't mine

Yeah, that sounds right to me. I don't think we're using the `geth --dev` equivalent for parity on the integration tests. (Actually, I don't think we're using `geth --dev` on the geth tests either)

voith:
The only tests pending now are the ones for ENS. Also the files might seem to be a lot. But the code changes are very limited. Most these changes are just tests.  

Simply parametrizing tests for ens didn't work. I need some digging to do. I'll try my best to finish this in the next few hours.

tests/core/contracts/conftest.py
```diff
@@ -15,6 +20,11 @@
 CONTRACT_ABI = json.loads('[{"constant":false,"inputs":[],"name":"return13","outputs":[{"name":"result","type":"int256"}],"type":"function"},{"constant":true,"inputs":[],"name":"counter","outputs":[{"name":"","type":"uint256"}],"type":"function"},{"constant":false,"inputs":[{"name":"amt","type":"uint256"}],"name":"increment","outputs":[{"name":"result","type":"uint256"}],"type":"function"},{"constant":false,"inputs":[{"name":"a","type":"int256"},{"name":"b","type":"int256"}],"name":"add","outputs":[{"name":"result","type":"int256"}],"type":"function"},{"constant":false,"inputs":[],"name":"increment","outputs":[{"name":"","type":"uint256"}],"type":"function"},{"constant":false,"inputs":[{"name":"a","type":"int256"}],"name":"multiply7","outputs":[{"name":"result","type":"int256"}],"type":"function"},{"anonymous":false,"inputs":[{"indexed":false,"name":"value","type":"uint256"}],"name":"Increased","type":"event"}]')  # noqa: E501
 
 
+@pytest.fixture(params=[lambda x: to_bytes(hexstr=x), identity])
+def address_conversion_func(request):
```
carver: Cool, this approach seems like a clean way to test the different accepted address formats. :+1: 

Instead of repeating it in several conftests, it seems like a reasonable thing to put it in the root `tests/conftest.py`.

voith:
@carver Ready for review!

Issue #894, opened 2018-06-10
Gas price generation using time based method generating low gas prices


robertmagier:
* Version: 4.3.0
* Python: 3.6.5
* OS: win
* Node: https://mainnet.infura.io/FiZkoo##Isk3ke######

### What was wrong?
I am using sendRawTransaction to send my transaction to main network using infura node. 
Function returns valid transaction hash, but this hash can't be found on etherscan. 
Few times I got an error message that transaction is under priced, but this is not always the case. 
In most cases it just looks like transaction was send without problems. No error is being returned.

To estimate gas price I am using this code:
```python
            web3.eth.setGasPriceStrategy(medium_gas_price_strategy)
            gasprice = web3.eth.generateGasPrice()

```
It always returns 1 gwei. This is below Infura gas price limit. 

I did a test and changed my code to use 4 gwei, 10 gwei and 20 gwei. In this case transaction was successfully send to blockchain, but then after using generateGasPrice my problem returned. 

Below is the code I use to call sendRawTransaction. 

```python
from web3 import Web3, HTTPProvider
from web3.contract import ConciseContract
from web3.auto import w3
from web3.middleware import geth_poa_middleware
from web3.gas_strategies.time_based import fast_gas_price_strategy, slow_gas_price_strategy,medium_gas_price_strategy


    file = open('contractdef.js',"r")
    abiData = json.loads(file.read())
    contract_address = we3.toChecksumAddress(app.config['ETHEREUM_CONTRACT'])
    contract_instance = we3.eth.contract(contract_address,abi=abiData)

    w3.eth.enable_unaudited_features()
    we3 = Web3(HTTPProvider(app.config['PROVIDER']))
    we3.eth.enable_unaudited_features()
    we3.middleware_stack.inject(geth_poa_middleware, layer=0)

    we3.eth.setGasPriceStrategy(medium_gas_price_strategy)
    gasprice = we3.eth.generateGasPrice()
    
    owner_address = app.config['PUBLIC_ADDRESS']
    claimed_address = app.config['CLAIMED_ADDRESS']

    amount = 10
    
    gas = contract_instance.functions.transfer(claimed_address,amount).estimateGas({'from':owner_address})
    nonce = we3.eth.getTransactionCount(we3.toChecksumAddress(owner_address),'pending')

    txn = contract_instance.functions.transfer(claim.claimed_address,amount).buildTransaction({
            'from':owner_address,
            'chainId':1,
            'gas':gas,
[...]

PR #895, opened 2018-06-11
Replace min price with low percentile for gas price estimation

dylanjw:
### What was wrong?

Fixes https://github.com/ethereum/web3.py/issues/894

The current time based gas price estimation strategy uses the minimum gas price from a set of blocks from a particular miner.  Doing this makes the assumption that there is an equal chance for all transactions at or above the minimum price to have been accepted by the miner.  This isn't true, as the miners will accept higher priced transactions before those close to their set minimum.

### How was it fixed?

Using a low percentile instead of the minimum should give a better estimate.  We don't want to use the record minimum, we want the reasonable minimum price to make our estimate.  In lieu of divining a reasonable minimum, using a lower percentile should get us a bit closer.

#### Cute Animal Picture

### DIFF
--- a/web3/gas_strategies/time_based.py
+++ b/web3/gas_strategies/time_based.py
@@ -9,13 +9,19 @@ from eth_utils import (
 from web3.exceptions import (
     ValidationError,
 )
+from web3.utils.math import (
+    percentile,
+)
 from web3.utils.toolz import (
     curry,
     groupby,
     sliding_window,
 )
 
-MinerData = collections.namedtuple('MinerData', ['miner', 'num_blocks', 'min_gas_price'])
+MinerData = collections.namedtuple(
+    'MinerData',
+    ['miner', 'num_blocks', 'min_gas_price', 'low_percentile_gas_price'])
+
 Probability = collections.namedtuple('Probability', ['gas_price', 'prob'])
 
 
@@ -54,7 +60,11 @@ def _aggregate_miner_data(raw_data):
 
     for miner, miner_data in data_by_miner.items():
         _, block_hashes, gas_prices = map(set, zip(*miner_data))
-        yield MinerData(miner, len(set(block_hashes)), min(gas_prices))
+        yield MinerData(
+            miner,
+            len(set(block_hashes)),
+            min(gas_prices),
+            percentile(list(gas_prices), percentile=15))
 
 
 @to_tuple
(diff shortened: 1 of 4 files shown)


dylanjw:
Im using the 15th percentile, which is just something I picked by comparing the result to https://ethgasstation.info/.  If we decide this is a good approach, than I can spend more time choosing an optimal percentile.

web3/utils/math.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def percentile(values, percentile):
+    if not values:
+        return None
+    index = len(values) * percentile / 100
+    if index % 2 == 0:
+        return (sorted(values)[index] + sorted(values)[index + 1]) / 2
+
+    return sorted(values)[math.ceil(index)]
```
carver: Sorting might be slow, as `values` gets bigger. It's probably worth sorting once and reusing the sorted list for performance, instead of sorting three times in the worst case.

web3/gas_strategies/time_based.py
```diff
@@ -54,7 +60,11 @@ def _aggregate_miner_data(raw_data):
 
     for miner, miner_data in data_by_miner.items():
         _, block_hashes, gas_prices = map(set, zip(*miner_data))
-        yield MinerData(miner, len(set(block_hashes)), min(gas_prices))
+        yield MinerData(
+            miner,
+            len(set(block_hashes)),
+            min(gas_prices),
+            percentile(list(gas_prices), percentile=15))
```
pipermerriam: why the `list` cast?

dylanjw: I will remove this.  I originally had pasted a percentile function from stackexchange that only accepted lists.  I replaced the function and didnt remove this.

pipermerriam:
Test failures are due to the algorithm now producing different numbers for the test input data.

- acceptable easy solution: update numbers with new numbers as generated in the tests.
- best solution: take the time to write new tests which are easier to reason about what the *right* numbers should be (or even a range).
- probably worst solution: take the time to compute the right numbers with the current tests.

dylanjw:
thanks @benediamond!

web3/utils/math.py
```diff
@@ -0,0 +1,12 @@
+import math
+
+
+def percentile(values, percentile):
+    if not values:
+        return None
```
pipermerriam: This looks like a place where an exception should be thrown.  Return value checking is inherently error prone.

raniama:
i cannot understand how could i choose the number of blocs as sample_size because the price could changes if i choose different values

carver:
This is not the right place to ask, try https://ethereum.stackexchange.com or our gitter channel at https://gitter.im/ethereum/web3.py

PR #897, opened 2018-06-11
Lay down initial framework for ethpm integration

njgheorghita:
### What was wrong?
No PM in web3.py

### How was it fixed?
Started initial steps to implement PM

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
         "eth-abi>=1.1.1,<2",
         "eth-account>=0.2.1,<0.3.0",
-        "eth-pm>=0.1.0,<1.0.0",
+        "eth-pm>=0.1.0-beta.17,<1.0.0",
         "eth-utils>=1.0.1,<2.0.0",
         "hexbytes>=0.1.0,<1.0.0",
         "lru-dict>=1.1.6,<2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
         "eth-abi>=1.1.1,<2",
         "eth-account>=0.2.1,<0.3.0",
-        "eth-pm==0.1.0-beta.17",
+        "eth-pm==0.1.0-alpha.17",
         "eth-utils>=1.0.1,<2.0.0",
         "hexbytes>=0.1.0,<1.0.0",
         "lru-dict>=1.1.6,<2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ setup(
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
         "eth-abi>=1.1.1,<2",
         "eth-account>=0.2.1,<0.3.0",
+        "eth-pm>=0.1.0,<1.0.0",
         "eth-utils>=1.0.1,<2.0.0",
         "hexbytes>=0.1.0,<1.0.0",
         "lru-dict>=1.1.6,<2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
         "eth-abi>=1.1.1,<2",
         "eth-account>=0.2.1,<0.3.0",
-        "eth-pm>=0.1.0-beta.17,<1.0.0",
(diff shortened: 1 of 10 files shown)


web3/main.py
```diff
@@ -108,12 +109,15 @@ class Web3:
...
 
-    def __init__(self, providers=empty, middlewares=None, modules=None, ens=empty):
+    def __init__(self, providers=empty, middlewares=None, modules=None, pm=False, ens=empty):
         self.manager = RequestManager(self, providers, middlewares)
 
         if modules is None:
             modules = get_default_modules()
 
+        if pm is True:
+            modules['pm'] = PM
```
njgheorghita: Not sure if this is the best way to allow "opt-in" usage of the `pm` module? @pipermerriam suggested using `enable_unaudited_features` but it looks like that does nothing currently.

dylanjw: You can have a look at the history of `enable_unaudited_features`, before it was changed to do nothing: https://github.com/ethereum/web3.py/blob/838b51d3e73200f7db961b5b6c1a9fc486d947ad/web3/eth.py#L59

The module path/structure is different, `web3.eth.account`, but maybe you could adapt something similar for `web3.pm`.

njgheorghita: Sounds good, thanks!

carver: Another option that is reasonable is to build it like the `Ssh` module. So it's just a free floating module, and if you want to opt-in, you just attach it to your `w3` like:

```
tests/core/shh-module/conftest.py:10:    Shh.attach(web3, "shh")
```

njgheorghita: Thanks! I'm leaning towards #2 since it reduces code and follows the `ssh` pattern

njgheorghita:
@carver @dylanjw Any clues why circle is failing? It's unable to import `HeaderNotFound` exception from `py-evm` - but I'm having trouble figuring out the fix

dylanjw:
seems like a dependency issue with eth-typing.  Not exactly sure what is going on yet.  My pip environment is haunted; so Im starting over.  Not sure what this was about:

```sh
pip show eth-typing
pip freeze | grep "typing"
python -c "import eth_typing; print(eth_typing.__file__)"   
/home/dwilson/Develop/Projects/ethereum/py-evm/venv-evm/lib/python3.6/site-packages/eth_typing/__init__.py
```

Somehow eth-typing was installed in my env but not managed by pip.

dylanjw:
I see.  eth-typing used to be a local py-evm module.

dylanjw:
@njgheorghita eth-pm has py-evm version pinned to 0.2.0a16, but eth-tester needs 0.2.0a18.

setup.py
```diff
@@ -21,6 +21,7 @@
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
         "eth-abi>=1.1.1,<2",
         "eth-account>=0.2.1,<0.3.0",
+        "eth-pm>=0.1.0-beta.17,<1.0.0",
```
pipermerriam: I'd pin to a specific beta unless you're confident that breaking changes won't be introduced in subsequent betas.

njgheorghita:
@pipermerriam @carver I notice that all the `web3.*.rst` files are in `.gitignore`. Is there a reason why? I added `web3.pm.rst` to `index.rst` - should I force add `web3.pm.rst` to get around the `.gitignore`? Or is there a better way?

carver:
>  all the web3.*.rst files are in .gitignore

Yeah, there are some files that would duplicate definitions in the docs. Force-adding `web3.pm.rst` works for me. :+1:

smlee729:
`>> pip install web3 --force-reinstall
Collecting web3
  Using cached https://files.pythonhosted.org/packages/a4/8d/430666983507fa0557e596975f0cd9133264f64c70e498f48266fdfe2366/web3-4.4.0-py3-none-any.whl
Collecting eth-account<0.3.0,>=0.2.1 (from web3)
  Using cached https://files.pythonhosted.org/packages/1c/f4/752b666464002dbd30d30443426cd25c045f941ae5f5e7b08eda1aee3c68/eth_account-0.2.3-py3-none-any.whl
Collecting eth-pm==0.1.0-alpha.17 (from web3)
  Could not find a version that satisfies the requirement eth-pm==0.1.0-alpha.17 (from web3) (from versions: )
No matching distribution found for eth-pm==0.1.0-alpha.17 (from web3)`

I cannot install web3 using pip. Is it related to this?

carver:
Yes, sorry @smlee729 -- Upgrading with `pip install web3>=4.4.1` should get you going again. (and now, if you want eth-pm, you'll have to install it separately)

PR #905, opened 2018-06-16
Create new list subclass for event topic management

dylanjw:
This this first piece of an improved event filter api.

Topics will be managed with a list like object that has some particular features:

1. The list prohibits repeat assignment
  e.g.:

```python
>>> from web3.utils.datastructures import SingleIndexAssignmentList
>>> test_list = SingleIndexAssignmentList()
>>> test_list.append("0x10101")
>>> test_list[0] = "0x11111"
  File "/home/dwilson/Develop/Projects/ethereum/web3.py/web3/utils/datastructures.py", line 255, in __setitem__
    raise ItemResetException("Values can't be reset.")
web3.exceptions.ItemResetException: Values can't be reset.
>>> 
```
2. Accepts out of range indexed assignments, filling in `None` for the preceding items.  For example:

```python
>>> from web3.utils.datastructures import SingleIndexAssignmentList
>>> test_list = SingleIndexAssignmentList()
>>> test_list[12] = "0x12345"
>>> test_list
[None, None, None, None, None, None, None, None, None, None, None, None, '0x12345']
>>> 
```

3.  The list can be copied, which does not carry over the state prohibiting repeat value setting.

```python
>>> from web3.utils.datastructures import SingleIndexAssignmentList
>>> test_list = SingleIndexAssignmentList()
>>> test_list.append("0x10101")
>>> test_list[0] = "0x11111"
  File "/home/dwilson/Develop/Projects/ethereum/web3.py/web3/utils/datastructures.py", line 255, in __setitem__
    raise ItemResetException("Values can't be reset.")
web3.exceptions.ItemResetException: Values can't be reset.
>>> new_test_list = test_list.copy()
>>> new_test_list
["0x10101"]
>>> new_test_list[0] = "0x11111
>>> new_test_list
["0x11111"]
>>> new_test_list[0] = "0x22222"
  File "/home/dwilson/Develop/Projects/ethereum/web3.py/web3/utils/datastructures.py", line 255, in __setitem__
    raise ItemResetException("Values can't be reset.")
```

These features will be leveraged to create a filter creation object that will (1) after having the initial parameters fed in, be essentially immutable, (2) allow setting only some topics in the topic argument list, while leaving those unfilled to match on all, and (3) allow existing filters to be used as templates to build new filters.

#### Cute Animal Picture

### DIFF
--- a/tests/core/utilities/test_single_index_assignment_list.py
+++ b/tests/core/utilities/test_single_index_assignment_list.py
@@ -1,0 +1,94 @@
+import pytest
+
+from hypothesis import (
+    given,
+    strategies as s,
+)
+
+from web3.exceptions import (
+    ItemResetException,
+)
+from web3.utils.datastructures import (
+    SingleIndexAssignmentList,
+)
+from web3.utils.hypothesis import (
+    hexstr_strategy,
+)
+
+elements = s.one_of(
+    s.integers(),
+    s.booleans(),
+    s.floats(),
+    s.tuples(),
+    s.lists(),
+    s.sets(),
+    s.dictionaries(s.integers(), s.integers()),
+    s.characters(),
+    s.text(),
+    s.binary(),
+    hexstr_strategy()
+)
+
+
+@s.composite
+def values_indexes_resets(draw, elements=elements):
+    values = draw(s.lists(elements, min_size=1))
+    indexes = draw(
+        s.lists(
+            max_size=len(values),
+            min_size=len(values),
+            elements=s.integers(min_value=0, max_value=len(values) - 1), unique=True))
+
+    bools = draw(
+        s.lists(
+            max_size=len(values),
+            min_size=len(values),
+            elements=s.booleans()))
+    return (values, indexes, bools)
+
+
+@s.composite
+def filled_list_not_prohibited(draw):
+    values = draw(s.lists(elements, min_size=1))
+    not_prohibited_list = SingleIndexAssignmentList(values)
+    return not_prohibited_list
+
+
+@s.composite
+def filled_list_setting_prohibited(draw):
+    values = draw(s.lists(elements, min_size=1))
+    prohibited_list = SingleIndexAssignmentList(values)
+    prohibited_list.set_prohibited = [True for i in range(len(prohibited_list))]
+    return prohibited_list
+
+
+@given(values_indexes_resets())
+def test_SingleIndexAssignmentList(values_indexes_resets):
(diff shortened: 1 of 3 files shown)


carver:
I sort of followed, until this:

```py
>>> new_test_list = test_list.copy()
>>> new_test_list
["0x10101"]
>>> new_test_list[0] = "0x11111
>>> new_test_list
["0x11111"]
```

If I can try to summarize the immutability model: you can mutate a value only once, until after you copy, and then you can reset a value in the list, but only once after the copy.

I think we can find something cleaner. Maybe look for inspiration in the `pyrlp` API, where the object is fully immutable, but with convenience methods for building modified copies.

The new class could be an implementation of [`Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence), so `__setitem__` and `__delitem__` wouldn't ever be implemented.

 A first stab might look like:

```py
>>> test_list = NoneFilledSequence([(0, "0x10101"), (2, "0x22222")])
>>> test_list
["0x10101", None, "0x22222"]
>>> new_test_list = test_list.copy((0, "0x11111"), (5, "0x55555"))
>>> new_test_list
["0x11111", None, "0x22222", None,"0x55555"]
>>> new_test_list[0] = "0x22222"
TypeError
```

pipermerriam:
@carver I agree that the mutate once pattern is not ideal, but I don't feel like the proposed solution is better.  That API feels cumbersome and not very intuitive.

If we drop use case 3, using an existing filter to template multiple other new filters, we can remove the `copy()` API.

Not obviously better ideas right now for creating topic lists so maybe tomorrow after standup we can do a brainstorming session.

dylanjw:
I will drop the copy support.  Maybe a better idea would be a list that you can freeze, then the one can make whatever updates to the filter settings prior to deploying, after the deploying the settings are frozen:

```
>>> f = myContract.events.MyEvent.newFilter()
>>> f.topics
>>> ["0x1s34fd78...."]
>>> f.topics[1] = "0x3l34588"
>>> f.deploy()
>>> f.topics[2] = "0xkja88hf"
  File "/home/dwilson/Develop/Projects/ethereum/web3.py/web3/utils/datastructures.py", line 255, in __setitem__
    raise FrozenListException("This is a frozen list.")
web3.exceptions.FrozenListException: Values can't be reset.
```

carver:
@dylanjw should this be closed in favor of the new API we talked about on the call?

dylanjw:
Yes.

PR #914, opened 2018-06-20
Allowed passing websockets connection kwargs

### LINKED ISSUE #913: Erorr in websockets.py: '<=' not supported between instances of 'int' and 'NoneType'
* web3 (4.3.0)
* websockets (4.0.1)
* Python: 3.6
* OS: osx HighSierra

### What was wrong?

`web3 = Web3(Web3.WebsocketProvider("ws://10.224.12.6:8546"))`
`web3.eth.syncing //returns data`

The websocket is clearly open but when I run a filter which is supposed to have many entries, I get the following error trace:

Upon running: `data = web3.eth.getFilterLogs(new_block_filter.filter_id)`, I get:

```
~/Desktop/contracts-py/contracts/lib/python3.6/site-packages/web3/providers/websocket.py in make_request(self, method, params)
     81             WebsocketProvider._loop
     82         )
---> 83         return future.result()

/anaconda3/lib/python3.6/concurrent/futures/_base.py in result(self, timeout)
    430                 raise CancelledError()
    431             elif self._state == FINISHED:
--> 432                 return self.__get_result()
    433             else:
    434                 raise TimeoutError()

/anaconda3/lib/python3.6/concurrent/futures/_base.py in __get_result(self)
    382     def __get_result(self):
    383         if self._exception:
--> 384             raise self._exception
    385         else:
    386             return self._result

~/Desktop/contracts-py/contracts/lib/python3.6/site-packages/web3/providers/websocket.py in coro_make_request(self, request_data)
     71         async with self.conn as conn:
     72             await conn.send(request_data)
---> 73             return json.loads(await conn.recv())
     74 
     75     def make_request(self, method, params):

~/Desktop/contracts-py/contracts/lib/python3.6/site-packages/websockets/protocol.py in recv(self)
    321             next_message.cancel()
    322             if not self.legacy_recv:
--> 323                 raise ConnectionClosed(self.close_code, self.close_reason)
    324 
    325     @asyncio.coroutine

~/Desktop/contracts-py/contracts/lib/python3.6/site-packages/websockets/exceptions.py in __init__(self, code, reason)
    145         self.reason = reason
    146         message = "WebSocket connection is closed: "
--> 147         if 3000 <= code < 4000:
    148             explanation = "registered"
    149         elif 4000 <= code < 5000:

TypeError: '<=' not supported between instances of 'int' and 'NoneType'
```

The same filter runs fine (albeit a bit slow) using `Web3.HTTPProvider()`

voith:
### What was wrong?
The current implementation of `WebsocketProvider` did not allow configuring extra configurations that a `websockets` connection takes
fixes #913

### How was it fixed?
Allowed passing websockets connection kwargs

### DIFF
--- a/web3/providers/websocket.py
+++ b/web3/providers/websocket.py
@@ -8,9 +8,14 @@ from threading import (
 
 import websockets
 
+from web3.exceptions import (
+    ValidationError,
+)
 from web3.providers.base import (
     JSONBaseProvider,
 )
+
+RESTRICTED_WEBSOCKET_KWARGS = {'uri', 'loop'}
 
 
 def _start_event_loop(loop):
@@ -32,14 +37,17 @@ def get_default_endpoint():
 
 class PersistentWebSocket:
 
-    def __init__(self, endpoint_uri, loop):
+    def __init__(self, endpoint_uri, loop, websocket_kwargs):
         self.ws = None
         self.endpoint_uri = endpoint_uri
         self.loop = loop
+        self.websocket_kwargs = websocket_kwargs
 
     async def __aenter__(self):
         if self.ws is None:
-            self.ws = await websockets.connect(uri=self.endpoint_uri, loop=self.loop)
+            self.ws = await websockets.connect(
+                uri=self.endpoint_uri, loop=self.loop, **self.websocket_kwargs
+            )
         return self.ws
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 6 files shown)


voith:
In case this PR needs a test case then:
```python
import pytest
from websockets.exceptions import ConnectionClosed
from web3 import Web3

w3 = Web3(Web3.WebsocketProvider(max_size=1))
with pytest.raises(ConnectionClosed):
    w3.eth.getBlock(0)
```
However, I'm not sure if web3 should testing functionality that should be tested by `websockets`!

web3/providers/websocket.py
```diff
@@ -55,13 +56,13 @@ class WebsocketProvider(JSONBaseProvider):
     logger = logging.getLogger("web3.providers.WebsocketProvider")
     _loop = None
 
-    def __init__(self, endpoint_uri=None):
+    def __init__(self, endpoint_uri=None, **kwargs):
```
carver: I'm not wild about letting the `websockets` connection parameters dictate the whole keyword argument namespace here. How about adding a single dict variable, like `connection_params={'max_size': 1}`?

That way there is no chance of a future namespace conflict or confusing mismatch. (we already have one mismatch: a value is called `uri` in websockets, but `endpoint_uri` here).

web3/providers/websocket.py
```diff
@@ -32,14 +32,15 @@ def get_default_endpoint():
 
 class PersistentWebSocket:
 
-    def __init__(self, endpoint_uri, loop):
+    def __init__(self, endpoint_uri, loop, **kwargs):
```
carver: Same thing here with the kwargs.

web3/providers/websocket.py
```diff
@@ -32,14 +32,15 @@ def get_default_endpoint():
...
         self.ws = None
         self.endpoint_uri = endpoint_uri
         self.loop = loop
+        self.kwargs = kwargs
 
     async def __aenter__(self):
         if self.ws is None:
-            self.ws = await websockets.connect(uri=self.endpoint_uri, loop=self.loop)
+            self.ws = await websockets.connect(uri=self.endpoint_uri, loop=self.loop, **self.kwargs)
```
carver: Maybe guard against people passing in `loop` or `uri` as one of the `connection_params` here.

web3/providers/websocket.py
voith: I chose the name `websocket_kwargs` instead, to be consistent with `request_kwargs` in `HttpProvider`.

PR #921, opened 2018-06-22
IPCProvider: Accept pathlib.Path in constructor.

cryptomental:
### What was wrong?

IPCProvider accepted only str paths.

### How was it fixed?

If ipc_path is an instance of pathlib.Path, resolve the path
and cast to string. Modify IPCProvider sync test to use pathlib
Path.

#### Cute Animal Picture

Fixes: #867


cryptomental:
Thank you for the approval!

PR #924, opened 2018-06-22
Add documentation for simple PM user stories

davesque:
### What was wrong?

Seemed like a good place to start with implementing simple PM user stories in web3.py was to write some documentation for how I imagined it should work.

### How was it fixed?

Added new article for PM documentation.

#### Cute Animal Picture


davesque:
@pipermerriam @njgheorghita This is *very* much in flux.  However, feel free to comment with ideas/concerns.

docs/package_management.rst
```diff
@@ -0,0 +1,56 @@
...
+filesystem.  In the code snippet below, a number of different approaches are
+used to get a package object for a ``greeter`` package:
+
+.. code-block:: python
+
+    >>> from web3 import Web3, HTTPProvider
+    >>> w3 = Web3(HTTPProvider('https://ropsten.infura.io/<token>'))
+
+    >>> # Get a package from a registry
+    >>> greeter_pkg = w3.pm.get_package_from_registry('greeter', registry='ens://packages.ethereum.eth')
```
njgheorghita: It's not necessary - but should we add a version param and a note that it's optional (or if left out will return the most recent version)?

docs/package_management.rst
```diff
@@ -0,0 +1,56 @@
...
+
+    >>> # Get a package from a URI
```
njgheorghita: Not sure if this is the right place to do it - or in `py-ethpm`'s docs - but a list somewhere of the different support c-a uri's would be useful

docs/pm_examples/greeter/1.0.0-pretty.json
```diff
@@ -0,0 +1,33 @@
+{
```
njgheorghita: Haha, it's awesome finally seeing a new manifest that's not one of the use-case examples 👍

carver:
Would you mind slipping in the `pmp` -> `pm` typo fix in the web3.pm.rst doc, too?

docs/package_management.rst
davesque: We could maybe add a note below the code snippet that more c-a URI formats are supported and link to the `py-ethpm` docs.  It seems like there will probably be a section in the `py-ethpm` docs with that information and it seems wrong to duplicate it across the `py-ethpm` and `web3.py` docs.

davesque:
@carver Thx for pointing that out.  Done.

PR #942, opened 2018-07-01
Skip ethpm test until dependency stabilized

njgheorghita:
### What was wrong?
`PM` module test depended on `ethpm` which has been removed until it's stablized.
Some other fixes to clean up Circle build

### How was it fixed?
Skip automatically running the `PM` test for the time being to clean up Circle build

#### Cute Animal Picture
image


njgheorghita:
Seems as though there's an error with the parity installation checksum validation, not sure if that's an issue on our end or theirs 🤷‍♂️

tests/core/pm-module/test_pm_init.py
```diff
@@ -1,8 +1,7 @@
 import pytest
 
-from web3.pm import (
-    PM,
-)
+# Skip all test that depend on `ethpm` until dependency is stabilized.
+pytest.importorskip("ethpm")
```
pipermerriam: Our general preference is to fully remove and then re-add later when ready in cases like this.  Git makes digging this stuff from history pretty easy.

njgheorghita: Sounds good

voith: Looks like @carver has something else in mind: https://github.com/ethereum/web3.py/pull/893/commits/7a0fb00e5ab47dbfdde146329754b51bfe969667

njgheorghita: @voith Good to know - If it doesn't make a big difference, I'm leaning towards removing the tests from master, and moving `ethpm` integration to it's own branch on `web3` until we reach a stable checkpoint. @davesque ?

Issue #943, opened 2018-07-03
Log filter unable to match entries by unindexed argument values when argument is value is not 32bytes


avatar-lavventura:
- Web3.py version 4.2.1
-  geth version: `1.8.0-unstable`

-------

https://web3py.readthedocs.io/en/latest/filters.html?highlight=eventFilter#module-web3.utils.filters

Says:
`event_filter = mycontract.events.myEvent.createFilter(fromBlock='latest', {'filter': {'arg1':10}})`

But:
https://web3py.readthedocs.io/en/latest/contracts.html?highlight=createfilter

Says:
`Contract.events.<event name>.createFilter(fromBlock=block, [toBlock=block, argument_filters={"arg1": "value"}, topics=[]])`

--------------

More detail please see: https://ethereum.stackexchange.com/q/52521/4575

I have followed [web3.py guide][1] for filtering.

> Through the contract instance api:
> 
>     event_filter = mycontract.events.myEvent.createFilter(fromBlock='latest', {'filter':{'arg1':10}})

------

But I get an error for this line:

    myContract.events.LogJob.createFilter(fromBlock='latest', {'filter': {'arg1':10}})

My main goal is the filter events based on their arguments.

Error I am having:

    myContract.events.LogJob.createFilter(fromBlock='latest', {'filter': {'arg1':10}})
                                                                     ^
    SyntaxError: positional argument follows keyword argument

**[Q]** How could I fix this error?

----------

Please also note that `argument_filters` seems like does not work. After I use `argument_filters` I do not get any error, but filter returns `[]` even I provide correct event's keyword (dictionary of argument name and its value). 

*Please see an example:*

    blockReadFrom = 1899201;
    myFilter = eBlocBroker.events.LogJob.createFilter(fromBlock=blockReadFrom)
    print(myFilter.get_all_entries())

Returns, here you can see that there is emitted event with `'storageID': 0` under `AttributeDict`:

[...]

PR #946, opened 2018-07-03
Add hack to make regex data filtering work

dylanjw:
### What was wrong?

Related to Issue #943 
The regex data filters assume all values will be 32 bytes, which is a false assumption.

### How was it fixed?
I replaced the regex with one that will match any abi length value, which is a pretty bad idea. Maybe there is a better short term solution, or maybe this should be left alone pending the replacement for data filters.

#### Cute Animal Picture

image

### DIFF
--- a/web3/utils/filters.py
+++ b/web3/utils/filters.py
@@ -127,7 +127,7 @@ class TransactionFilter(Filter):
     pass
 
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
Since there is already a bug, and this solution probably also has a bug, I think it'd be best to leave it be and focus on the real solution.

PR #949, opened 2018-07-05
[pm-api] Instantiate Package with web3 in get_package_from_manifest

njgheorghita:
### What was wrong?
`Package` objects should be instantiated with a `web3` by default in `PM.get_package_from_manifest`

Related to Issue # [48 in py-ethpm](https://github.com/ethpm/py-ethpm/issues/48)

### How was it fixed?
Passed `web3` instance as an arg when instantiating a `Package` object in `PM.get_package_from_manifest`

#### Cute Animal Picture
image


tests/core/pm-module/test_pm_init.py
```diff
@@ -14,11 +20,23 @@ def web3():
         from web3.pm import PM
```
voith: if `ethpm` is a strict dependency of web3 then is these try except block needed?

njgheorghita: @voith We've decided to continue with `ethpm` development on a separate branch `pm-api` until we reach a stable v1. So this strict dependency will only exist on the experimental `pm-api` branch - and once v1 is reached we will ensure all dependency issues are stable before merging `pm-api` into `master`

davesque: You're right.  Nick and I decided to just delete the `ethpm` requirement from `setup.py`.  That way, we can still possibly later merge the `pm-api` branch into master and have the web3 PM module be optional alpha functionality.  Thanks for noticing this discrepancy.

pipermerriam:
Maybe good to go ahead and add docs (with a big warning at the top that these APIs are not stable)

njgheorghita:
@pipermerriam 👍 docs have already started [here](https://github.com/ethereum/web3.py/blob/master/docs/web3.pm.rst)

davesque:
I think we had some discussion about this previously.  However, I can't remember what we decided.

I'm feeling at this moment that it's not a good idea to add the `ethpm` package to the `install_requires` key in `setup.py`.  I think we should make it so that any tests that depend on that package will be marked as expected failures.  This allows us to continue to muck around with the package name if need be during the alpha phase without breaking the current release of `web3.py`.

carver:
> I think we should make it so that any tests that depend on that package will be marked as expected failures.

:+1:

njgheorghita:
@davesque  I was thinking that since we're keeping ethpm development on the `pm-api` branch until we reach a stable release, then it wouldn't matter writing the tests without `xfail` and putting the dependency in `install_requires` since that's where it will end up eventually. Is there a reason to maintain the more defensive approach to integration (i.e. using `xfail`, import statement warnings, etc.) even though we're not pushing to master?

carver:
> Is there a reason to maintain the more defensive approach to integration (i.e. using xfail, import statement warnings, etc.) even though we're not pushing through to master?

At some point the branch should be merged with master, and that might happen before we're ready to include ethpm as a default dependency. (For example, that could be a reasonable approach when it's `beta` level)

njgheorghita:
@carver Gotcha, thanks. What's the better strategy here? Add `ethpm` as a dependency in an `extras_require`? Or simply leave it out of `setup.py` and skip all the tests and just run them locally.

davesque:
@njgheorghita I think it's fine to leave it out of `setup.py` for now since we talked about modifying the relevant testing module to skip tests if the package has not been installed.

Issue #872, opened 2018-05-30
Discussion: How to improve filter api


dylanjw:
### What was wrong?

See #836.  Current web3 filter api's are not intuitive and are just unable to do certain things.  An example is creating a filter for array event arguments.  Current argument encoding can not differentiate between argument options and arguments that _are_ arrays:

Current api:

```python
some_contract.events.eventWithArrayArg.createFilter(fromBlock='latest', filter={'array_arg': [b'00', b'11'])
```

The following lines in web3/utils/events.py parse the above array incorrectly into argument options:

```
 67  	    encoded_args = [
 68  	        [
 69  	            None if option is None else encode_hex(encode_single(arg['type'], option))
 70  	            for option in arg_options]
 71  ->	        for arg, arg_options in zipped_abi_and_args
 72  	    ]
```

### How can it be fixed?

Argument values to filter on should have abi type aware handling, so that the filter parameters can be validated, and array type arguments can be distinguished from lists of multiple match targets.  A few ideas for how this could look:

**Option 1**. A filter argument type, that evaluates to the encoded topic or topics.  Can 
be supplied to the topic parameter list when creating an event filter.

```python
my_array_arg_single_value = contract.events.SomeEvent.args['my_array_argument']([b'11', b'22'])
my_array_arg_single_value
b'\xac{\x94\xf1X\x12S\xf5\x02\xa5\xfa \xba\xeb.\xcf<\x8b\xfa\xd2G6\xcbw\xe4,\x84\x99\x0eMew'
    
my_array_arg_multiple_values = contract.events.SomeEvent.args['my_array_argument']([b'00'], [b'11', b'22'])
[b'\xd3R@\xdf\xb7\xd5\xf3\xe7\xa3?i\x87+\x84Y\x04g\xe5\xd4\xce.z3\x12\xff\xa8\xe5\xc5\xab6{\xbe', b'\xac{\x94\xf1X\x12S\xf5\x02\xa5\xfa \xba\xeb.\xcf<\x8b\xfa\xd2G6\xcbw\xe4,\x84\x99\x0eMew']
        
contract.events.SomeEvent.createFilter(fromBlock='latest', topics=[my_array_arg_multiple_values])
```

Although this is a bit awkward, it is nice that the argument object could be immutable.

**Option 2.** The createFilter method to accept kwargs for event arguements

Single value:
```python
contract.events.SomeEvent.createFilter(fromBlock='latest', my_array_argument=[b'11', b'22'])
```    
[...]

PR #953, opened 2018-07-10
Introduce a better filter argument api with EventFilterBuilder

dylanjw:
### What was wrong?

Related to Issue #872 

The existing filter apis are unable to properly differentiate array arguments and multiple match arguments.

### How was it fixed?
This PR introduces a new "filter builder" api, where parameters can be set prior to deploying the filter.  example:

```
filter_builder = myContract.events.myEvent.build_filter()
filter_builder.args['array_of_states'].match_single(("CA", "WA", "TX",))
filter_builder.args['array_of_states'].match_any(("CA", "WA", "TX",), ("MN", "OH",))
filter_builder.fromBlock = 0
filter_bulder.toBlock = "latest"
my_filter = filter_builder.deploy() 
```

Additionally this PR contains a bug fix to the data argument filters, by replacing the regex mechanism with an abi decoding and direct value comparison.  EDIT: Replacing the `data_filter_set_function` created the need to update the contract.py filter api's, as they are required to pass in the data filter set with the abi type.

#### Cute Animal Picture

image


web3/utils/normalizers.py
```diff
@@ -58,12 +56,6 @@ def addresses_checksummed(abi_type, data):
         return abi_type, to_checksum_address(data)
 
 
-@implicitly_identity
-def decode_abi_strings(abi_type, data):
-    if abi_type == 'string':
-        return abi_type, codecs.decode(data, 'utf8', 'backslashreplace')
```
dylanjw: Im getting various DecodingErrors when using the updated eth_abi package that has the change to return strings as string types.

Likely the string data in these tests are not encodable.

carver: Hm, well feel free to copy in a trace if you want another pair of eyes on it.

BTW, there might be other things besides contract arguments/return values that depend on this `decode_abi_strings()` method, so it's not clear it should be deleted.

web3/contract.py
```diff
@@ -1290,34 +1282,41 @@ def createFilter(
...
             argument_filters = dict()
+        else:
+            for arg, arg_value in argument_filters.items():
+                filter_builder.args[arg].match_single(arg_value)
 
-        _filters = dict(**argument_filters)
+        if topics is not None:
+            raise NotImplementedError(
+                "Directly setting topics is not currently supported. "
+                "Use the argument_filters parameter instead.")
```
dylanjw: I added this not implemented error to simplify things, as a temporary measure. The ability to set topics directly should be added back to createFilter

dylanjw: TODO

dylanjw: I was hoping to merge the manually entered and the generated topic arguments together, but I realize that that would yield filter results that are entirely different than adding the results from the two sets of filtered data.  e.g. merging (A, B) with (C, D) to get a topic list like ((A, C), (B, D)).  The merged topic list would match AB, AD, CB, CD instead of just AC, BD.  For this reason, manually entering topics needs to either be disabled, or throw an exception if topics have already been generated.

web3/utils/normalizers.py
dylanjw: @carver I moved this to another PR since it needs it own discussion: https://github.com/ethereum/web3.py/pull/974

web3/utils/encoding.py
```diff
@@ -275,3 +278,34 @@ def to_4byte_hex(hex_or_str_or_bytes):
...
+    is_dynamic = True
+
+    def encode(self, value):
+        encoded_elements = self.encode_elements(value)
+        encoded_value = encoded_elements
+
+        return encoded_value
+
+
+def encode_single_packed(_type, value):
```
dylanjw: This will be replaced with the eth-abi packed encoder once we get https://github.com/ethereum/eth-abi/pull/89 for web3.py

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -51,9 +52,10 @@ def array_values(draw):
     return (matching, non_matching)
 
 
-@pytest.mark.xfail(reason="The regex based data filters do not work with dynamic sized types")
 @pytest.mark.parametrize('call_as_instance', (True, False))
+@pytest.mark.parametrize('api_style', ('v3', 'v4', 'build_filter'))
 @given(vals=dynamic_values())
+@settings(max_examples=5)
```
dylanjw: Im getting hypothesis warnings because these tests run slow; 500 - 2000 ms

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -162,37 +177,45 @@ def test_data_filters_with_list_arguments(
...
+
+        for txn_hash in txn_hashes:
+            wait_for_transaction(web3, txn_hash)
+
+        log_entries = event_filter.get_new_entries()
+        assert len(log_entries) == 2
+        assert log_entries[0]['transactionHash'] == txn_hashes[0]
+        assert log_entries[1]['transactionHash'] == txn_hashes[2]
+    else:
+        with pytest.raises(TypeError):
```
dylanjw: List arguments are forbidden for v3/v4 api's to prevent ambiguity between list arguments and lists match options.

tests/core/filtering/test_contract_on_event_filtering.py
```diff
@@ -146,49 +168,23 @@ def test_filter_with_contract_address(web3, emitter, emitter_event_ids, wait_for
...
-    assert len(seen_logs) == 4
-
-    post_event_filter = contract.events.LogTripleWithIndex.createFilter(
-        argument_filters={'arg1': [1, 2], 'arg2': [1, 2]},
-        fromBlock=0,
-    )
-
-    old_logs = post_event_filter.get_all_entries()
-    assert len(old_logs) == 4
+    with pytest.raises(TypeError):
```
dylanjw: Filter argument/topic options are also not allowed with v3/v4 api's.

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -32,20 +32,28 @@ def single_transaction(w3):
...
     matching_transact = emitter.functions.logNoArgs(
         which=1).transact
     non_matching_transact = emitter.functions.logNoArgs(
         which=0).transact
 
-    event_filter = emitter.events.LogNoArguments().createFilter(fromBlock='latest')
+    if api_style == 'build_filter':
+        builder = emitter.events.LogNoArguments.build_filter()
+        builder.fromBlock = 'latest'
+        builder.address = emitter.address
```
dylanjw: address should be filled in when creating the builder object from a deployed contract.

tests/core/utilities/test_event_filter_builder.py
```diff
@@ -0,0 +1,52 @@
...
+@given(st.text())
+def test_match_single_string_type_properties_data_arg(value):
+    ea = DataArgumentFilter(arg_type="string")
+    ea.match_single(value)
+
+
+@given(st.text())
+def test_match_single_string_type_properties_topic_arg(value):
+    ea = TopicArgumentFilter(arg_type="string")
+    ea.match_single(value)
```
dylanjw: Should assert the encoded topic value is correct.

web3/contract.py
```diff
@@ -1322,21 +1338,46 @@ def createFilter(
...
+    for filter_name, filter_value in _filters.items():
+        _input = name_indexed_inputs[filter_name]
+        if is_array_type(_input['type']):
+            raise TypeError(
+                "createFilter no longer supports array type filter arguments. "
+                "see the build_filter method for filtering array type filters.")
+        if is_list_like(filter_value):
+            raise TypeError(
+                "createFilter no longer supports setting filter argument options implicitly. "
+                "see the build_filter method for filtering setting filters with the match_any "
```
dylanjw: typo

web3/utils/filters.py
```diff
@@ -166,16 +161,51 @@ def format_entry(self, entry):
         return entry
 
     def set_data_filters(self, data_filter_set):
+        """(('uint256', [12345, 54321]), ('string', ('a-single-string',)))
```
dylanjw: improve doc string

web3/utils/filters.py
```diff
@@ -166,16 +161,51 @@ def format_entry(self, entry):
...
 
     def is_valid_entry(self, entry):
-        if not self.data_filter_set_regex:
+        if not self.data_filter_set:
             return True
-        return bool(self.data_filter_set_regex.match(entry['data']))
+        return bool(self.data_filter_set_function(entry['data']))
+
+
+def normalize_data_values(type_string, data_value):
```
dylanjw: converting bytes to strings can go away with eth-abi v2

carver: We'll have to support old versions of eth-abi at least until web3.py v5, though.

tests/core/contracts/test_contract_events_buildFilter.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_build_filter(web3):
+    contract = web3.eth.contract(abi=CONTRACT_ABI)
+    filter_builder = contract.events.Increased.build_filter()
+    filter_builder.args['value'].match_any(100, 200, 300)
+    _filter = filter_builder.deploy(web3)
+    assert _filter.filter_params == {
+        'topics': (
+            '0x3496c3ede4ec3ab3686712aa1c238593ea6a42df83f98a5ec7df9834cfa577c5',)}
```
pipermerriam: I think this value is going to be very opaque later.  Can you either:

- add a comment indicating what this value represents
- replace with the code that derives this value

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -162,37 +177,45 @@ def test_data_filters_with_list_arguments(
...
+        api_style,
         vals):
 
     matching, non_matching = vals
 
     if call_as_instance:
-        event_filter = create_filter(emitter, [
-            'LogListArgs', {
-                'filter': {"arg1": matching}}])
+        contract = emitter
```
pipermerriam: Somewhat orthogonal but this could be to use the [`request`](https://docs.pytest.org/en/latest/reference.html#_pytest.fixtures.FixtureRequest.getfixturevalue) fixture to lazily request these fixtures:

```python
def the_test(request):
    if call_as_instance:
        contract = request.getfixturevalue('emitter')
    else:
        contract = request.getfixturevalue('Emitter')
```

It'd give us *some* increase in test speed since it won't need to deploy the `emitter` contract for every single run.

tests/core/filtering/test_filters_against_many_blocks.py
pipermerriam: I'm not sure I understand?  As a user, I would be surprised if the address was not automatically populated when I created a filter from a contract instance.

tests/core/utilities/test_event_filter_builder.py
pipermerriam: What does `ea` stand for?

web3/contract.py
```diff
@@ -383,20 +386,29 @@ def eventFilter(self, event_name, filter_params={}):
...
+
+        _, event_filter_params = construct_event_filter_params(
             event_abi,
             contract_address=self.address,
             argument_filters=argument_filters,
             **filter_meta_params
         )
 
+        filter_builder = EventFilterBuilder(event_abi)
+        filter_builder.address = event_filter_params.get('address')
```
pipermerriam: This seems like the place where it should default to the contract address if an explicit address was not provided.

Additionally, it seems like when building a filter from a contract instance we should **require** the address be the contract address and only allow the address to be specified for contract factory based filters.

web3/utils/encoding.py
pipermerriam: It's normally good to leave this type of information in an embedded comment linking to the issue that must be closed before the code can be updated.

web3/utils/encoding.py
```diff
@@ -275,3 +278,34 @@ def to_4byte_hex(hex_or_str_or_bytes):
...
+    abi_type = abi_type_parser.parse(_type)
+    if has_arrlist(_type):
+        item_encoder = registry.get_encoder(str(abi_type.item_type))
+        if abi_type.arrlist[-1] != 1:
+            return DynamicArrayPackedEncoder(item_encoder=item_encoder).encode(value)
+        else:
+            raise NotImplementedError(
+                "Fixed arrays are not implemented in this packed encoder prototype")
+    else:
+        if abi_type.base == "string":
```
pipermerriam: You could drop an indentation level and do this as:

```python
if has_arrlist(_type):
    ...
elif abi_type.base == 'string':
    ...
elif abi_type.base == 'bytes'
    ...
```

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
     pop_singlets,)
+
+
+def is_indexed(arg):
+    if isinstance(arg, TopicArgumentFilter) is True:
+        return True
+    return False
+
+
+not_indexed = complement(is_indexed)
```
pipermerriam: maybe `is_not_indexed` to conform to a common suffix.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+
+        log_filter = web3.eth.filter(self.filter_params)
+        log_filter.filter_params = self.filter_params
+        log_filter.set_data_filters(self.data_argument_values)
+        log_filter.builder = self
+        if self.formatter is not None:
+            log_filter.log_entry_formatter = self.formatter
+        return log_filter
+
+    def _initial_event_topic(self):
```
pipermerriam: I'd take this off the class and just make it a stand-alone private module function and pass the `event_abi` in as an input.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+    def toBlock(self, value):
+        self._toBlock = value
+
+    @property
+    def address(self):
+        return self._address
+
+    @address.setter
+    def address(self, value):
+        self._address = value
```
pipermerriam: Since the getter and setter are both just the defaults, why not just use a normal class attribute and drop the `property` stuff.  Same with the rest of these `property` functions.  If they don't do any special functionality and have both a `getter` and `setter` then a normal property is probably easier to reason about.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+        return filter(
+            is_indexed,
+            map(lambda x: self.args[x], self._ordered_arg_names))
+
+    @property
+    @to_tuple
+    def data_args(self):
+        return filter(
+            not_indexed,
+            map(lambda x: self.args[x], self._ordered_arg_names))
```
pipermerriam: Maybe this could be a computed property?

```python
@property
def ordered_args(self):
    return tuple(map(self.args.__getitem__, self.ordered_arg_names))
```

Then these functions become:

```python
@property
def data_args(self):
    return tuple(filter(not_indexed, self.ordered_args))
```

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+            "topics": self.topics,
+            "fromBlock": self.fromBlock,
+            "toBlock": self.toBlock,
+            "address": self.address
+        }
+        return valfilter(lambda x: x is not None, params)
+
+    def deploy(self, web3):
+        if web3 is None:
+            raise ValueError("Missing required argument 'web3'")
```
pipermerriam: Maybe shift this to:

```python
if not isinstance(web3, Web3):
    raise TypeError("Invalid web3 argument: got: {0}".format(repr(web3)))
```

Raises a more informative error message.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+    def filter_params(self):
+        params = {
+            "topics": self.topics,
+            "fromBlock": self.fromBlock,
+            "toBlock": self.toBlock,
+            "address": self.address
+        }
+        return valfilter(lambda x: x is not None, params)
+
+    def deploy(self, web3):
```
pipermerriam: Can you use `w3` as the variable name for a web3 instance.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+            return list()
+
+
+@to_dict
+def _build_argument_filters_from_event_abi(event_abi):
+    for item in event_abi['inputs']:
+        if item['indexed'] is True:
+            argument_filter_class = TopicArgumentFilter
+        else:
+            argument_filter_class = DataArgumentFilter
```
pipermerriam: I think I find this easier to reason about as:

```python
key = item["name"]
if item['indexed'] is True:
    value = TopicArgumentFilter(arg_type=item['type']
else:
    value = DataArgumentFilter(arg_type=item['type']
yield key, value
```

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+        self.arg_type = arg_type
+
+    def match_single(self, value):
+        self._match_values = self._normalize_match_values((value,))
+
+    def match_any(self, *values):
+        self._match_values = self._normalize_match_values(values)
+
+    @to_tuple
+    def _normalize_match_values(self, match_values):
```
pipermerriam: since this doesn't use `self` can you either decorate it as `staticmethod` or (my preference) elevate it to a module level private function?

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+        for value in match_values:
+            yield array_to_tuple(value)
+
+    @property
+    @abstractmethod
+    def match_values(self):
+        raise NotImplementedError("ArgumentFilter classes must implement match_values method")
```
pipermerriam: I think I'd like to move away from using `NotImplementedError` here and just put `pass` statements since the `abstractmethod` decorator raises an exception for us on instantiation if this method has not been implemented.

web3/utils/events.py
```diff
@@ -238,3 +253,168 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+class DataArgumentFilter(BaseArgumentFilter):
+    @property
+    def match_values(self):
+        return (self.arg_type, self._match_values)
+
+
+class TopicArgumentFilter(BaseArgumentFilter):
+    @to_tuple
+    def _get_match_values(self):
+            yield from (self._encode(value) for value in self._match_values)
```
pipermerriam: extra indentation?

web3/utils/filters.py
pipermerriam: Also, this function probably deserves some small comments explaining what is happening and why.

web3/utils/filters.py
```diff
@@ -166,16 +161,51 @@ def format_entry(self, entry):
...
+    _type = parse_type_string(type_string)
+    if _type.base == "string":
+        if _type.arrlist is not None:
+            return tuple((bytes_to_str(value) for value in data_value))
+        else:
+            return bytes_to_str(data_value)
+    return data_value
+
+
+def bytes_to_str(value):
```
pipermerriam: maybe rename to `decode_utf8_bytes`

web3/utils/filters.py
```diff
@@ -166,16 +161,51 @@ def format_entry(self, entry):
...
+    return value.decode("utf-8")
+
+
+@curry
+def match_fn(match_values_and_abi, data):
```
pipermerriam: I think this function deserves a docstring and some comments.  There's a lot going on here and it is quite dense.

web3/utils/toolz/__init__.py
```diff
@@ -4,6 +4,7 @@
         complement,
         compose,
         concat,
```
pipermerriam: Maybe switch to using the `eth_utils.toolz` namespace if you are able to upgrade to a newer `eth-utils`?  Not really part of this PR but just pointing it out.

web3/utils/toolz/__init__.py
```diff
@@ -4,6 +4,7 @@
...
+        cons,
```
carver: Hm, maybe it's time to bump the eth-utils requirement up and start using `eth_utils.toolz`

web3/utils/encoding.py
carver: and it will have to survive longer anyway, until we can require eth-abi v2+ in web3 v5

setup.py
```diff
@@ -19,7 +19,7 @@
     install_requires=[
         "toolz>=0.9.0,<1.0.0;implementation_name=='pypy'",
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
-        "eth-abi>=1.1.1,<2",
+        "eth-abi>=2.0.0a1,<3.0.0",
```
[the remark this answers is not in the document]
carver: I think it's a must to support old versions of eth-abi. Once we get these new filters into the wild and we're happy with the results after seeing people try to use them for a bit, maybe it will be time to focus on getting v5 out the door.

tests/core/filtering/test_filters_against_many_blocks.py
dylanjw: Just making a note to myself to fix this.

tests/core/contracts/test_contract_events_buildFilter.py
```diff
@@ -0,0 +1,17 @@
...
+    assert _filter.data_filter_set == (('uint256', (100, 200, 300)),)
+    filter_builder.args['value'].match_single(400)
```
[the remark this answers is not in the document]
dylanjw: Ive gone back and forth on this.  Because the `deploy` method generates a new filter from the builder's settings, it seemed not such a risk to allow the builder to be mutable.  But it is probably better to have the first release take the least mutable approach because taking away some mutability would be a breaking change whereas adding it would not.

pipermerriam:
I don't see a problem with raising exceptions in cases where previously failures were silent.

carver:
> I'll summarize with a table

Thanks, that helped a lot! Getting support back for 5 is great :+1: 

> what do you think about leaving the exceptions for the 2 cases that had been failing, but could be fixed.

Totally, if the code was already broken (or failing silently), then raising exceptions is a good option. :+1:

dylanjw:
Hypothesis has found a failing example, when the event array argument has 38 or more elements.

dylanjw:
@pipermerriam @carver I made some changes to limit value reset, and limit value setting after a filter has been deployed from a filter builder instance.  Since there are objects for the argument options a did the following to share the deployment state between the objects: https://github.com/ethereum/web3.py/pull/953/commits/cf3f79552111fa56ad414c3856de97ddecbe51e3  Is an anti-pattern?  It feels like an anti-pattern.

web3/utils/events.py
```diff
@@ -238,3 +254,206 @@ def remove_trailing_from_seq(seq, remove_value=None):
...
+    _toBlock = None
+    _address = None
+
+    def __init__(self, event_abi, formatter=None):
+        self._deployed_flag = Flag()
+        self.event_abi = event_abi
+        self.formatter = formatter
+        self.event_topic = initialize_event_topics(self.event_abi)
+        self.args = AttributeDict(
+            _build_argument_filters_from_event_abi(event_abi, self._deployed_flag))
```
carver: Re:

>  Is this an anti-pattern? It feels like an anti-pattern.

Yeah, I think so too. Maybe instead of this shared flag, replace the `args` attribute in `deploy()`. Something like (pseudocode):

```diff
-        self._deployed_flag(True)
+        old_args = self.args
+        self.args = ImmutableArgs(old_args)
+        old_args.clear()  # make it clear that people should not use an old saved copy of args
```

dylanjw:
@pipermerriam @carver Do you want to give this a final review?  Im just about ready to merge.

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -45,52 +50,41 @@ def fixed_values(draw):
 
 @st.composite
 def array_values(draw):
-    matching = draw(st.lists(elements=st.binary(min_size=2, max_size=2)))
+    matching = draw(st.lists(elements=st.binary(min_size=2, max_size=2), max_size=35, min_size=1))
```
carver: > Hypothesis has found a failing example, when the event array argument has 38 or more elements. Ive set the max number of elements to 35. This is probably a eth-tester issue.

Did you find out any more about this? Does a filter with 38 elements work on, say, `geth --dev`?

I hate to have this constraint on the test, until we know what's going on.

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -151,48 +144,51 @@ def test_data_filters_with_fixed_arguments(
...
+            wait_for_transaction(web3, txn_hash)
+
+        log_entries = event_filter.get_new_entries()
+        assert len(log_entries) == 2
+        assert log_entries[0]['transactionHash'] == txn_hashes[0]
+        assert log_entries[1]['transactionHash'] == txn_hashes[2]
     else:
-        event_filter = create_filter(Emitter, [
-            'LogListArgs', {
-                'filter': {"arg1": matching}}])
```
carver: So this is # 2 from: https://github.com/ethereum/web3.py/pull/953#issuecomment-416095677 right? Can you remind me *what* was failing silently (since this was a working test before).

dylanjw: There was an issue with the way the regex would be built when array arguments were included.  The regex sized each item in the encoded data to be 32 bytes.

dylanjw: the test was marked XFAIL before.

carver: Perfect, thanks.

PR #954, opened 2018-07-11
[pm-api] Add example packages for testing

njgheorghita:
branched from #949 

### What was wrong?
Improved tests
Added an integration test with `StandardToken`
Add packages folder to store packages for testing

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/pm-module/packages/owned/1.0.0.json
+++ b/tests/core/pm-module/packages/owned/1.0.0.json
@@ -1,1 +1,21 @@
-{"manifest_version":"2","meta":{"authors":["Piper Merriam <pipermerriam@gmail.com>"],"description":"Reusable contracts which implement a privileged 'owner' model for authorization.","keywords":["authorization"],"license":"MIT","links":{"documentation":"ipfs://QmUYcVzTfSwJoigggMxeo2g5STWAgJdisQsqcXHws7b1FW"}},"package_name":"owned","sources":{"./contracts/Owned.sol":"ipfs://Qme4otpS88NV8yQi8TfTP89EsQC5bko3F5N1yhRoi6cwGV"},"version":"1.0.0"}
+{
+  "manifest_version": "2",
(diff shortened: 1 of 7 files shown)


tests/core/pm-module/packages/owned/1.0.0.json
```diff
@@ -0,0 +1,21 @@
...
+    "description": "Reusable contracts which implement a privileged 'owner' model for authorization.",
+    "keywords": [
+      "authorization"
+    ],
+    "links": {
+      "documentation": "ipfs://QmUYcVzTfSwJoigggMxeo2g5STWAgJdisQsqcXHws7b1FW"
+    }
+  },
+  "sources": {
+    "./contracts/Owned.sol": "ipfs://Qme4otpS88NV8yQi8TfTP89EsQC5bko3F5N1yhRoi6cwGV"
```
pipermerriam: Should these be represented in their *compact* form.  I don't think I've seen a PR in ethpm to enforce that data format, so it's probably something that needs to have an issue created for, but I suspect we'll run into an error in this repository at some point in the future if we use non-canonical representations of the manifests here.

njgheorghita: Good point, I'll add that issue to `py-ethpm`. And once it gets implemented there it'll force us to update the contracts here.

davesque: Should I understand "compact" to mean `ipfs:Qme...`?

njgheorghita: @davesque I don't think it has anything to do with the contents of the uri - just to enforce the guidelines [here in ethpm-spec](http://ethpm.github.io/ethpm-spec/package-spec.html#document-format)

carver:
If anyone is waiting on me, please go ahead and merge this and other PRs into the `pm-abi` branch at will :+1:

PR #955, opened 2018-07-13
Raise helpful exception in pm module

davesque:
### What was wrong?

If someone tries to use the alpha PM features in web3.py, they will get some confusing errors since the ethpm imports just fail silently.

### How was it fixed?

Re-introduced exception raising for when ethpm is not installed.

#### Cute Animal Picture

### DIFF
--- a/tests/core/pm-module/test_pm_init.py
+++ b/tests/core/pm-module/test_pm_init.py
@@ -5,15 +5,11 @@ from pathlib import (
 import pytest
 
 from web3 import Web3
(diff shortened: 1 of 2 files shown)


davesque:
@njgheorghita I think this is a reasonable solution to having an exception on importing ethpm but still being able to run the tests with or without.

Maybe the problem you were running into earlier was that the exception was being re-raised as a base `Exception` and so you couldn't catch it as an `ImportError`.

PR #967, opened 2018-07-23
Increment parity version to latest stable release for integration testing

dylanjw:
1.10.4 >> 1.11.7

#### Cute Animal Picture

image

### DIFF
--- a/tests/integration/parity/common.py
+++ b/tests/integration/parity/common.py
@@ -30,6 +30,10 @@ class ParityWeb3ModuleTest(Web3ModuleTest):
 
 
 class ParityEthModuleTest(EthModuleTest):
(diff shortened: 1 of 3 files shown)


tests/integration/parity/common.py
```diff
@@ -30,6 +30,10 @@ def _check_web3_clientVersion(self, client_version):
 
 
 class ParityEthModuleTest(EthModuleTest):
+    def test_eth_getBlockByNumber_pending(self, web3):
+        pytest.xfail('Parity deprecated "pending" option in 1.11.1')
```
carver: I think this comment should say they removed (or dropped) the option instead of deprecating it. (If I understand correctly)

If it's just deprecated, I would expect the call to continue working.

PR #969, opened 2018-07-24
Added eth_getCode to ethereum_tester_middleware to appease eth-tester

voith:
### What was wrong?
While porting `populus` for `compatibilty` with `web3 4.x`, I found the following problem with `eth-tester`:
```python
web3.eth.getCode(emitter_contract.address, block_identifier=web3.eth.blockNumber)
```
The above lines gives the following error 
```
.tox/py36-integration-ethtester/lib/python3.6/site-packages/eth_tester/main.py:272: in get_code
    self.validator.validate_inbound_block_number(block_number)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

value = '0x1'

    def validate_block_number(value):
        error_message = (
            "Block number must be a positive integer or one of the strings "
            "'latest', 'earliest', or 'pending'.  Got: {0}".format(value)
        )
        if is_string(value):
            validate_text(value)
            if value not in BLOCK_NUMBER_META_VALUES:
>               raise ValidationError(error_message)
E               eth_tester.exceptions.ValidationError: Block number must be a positive integer or one of the strings 'latest', 'earliest', or 'pending'.  Got: 0x1

error_message = "Block number must be a positive integer or one of the strings 'latest', 'earliest', or 'pending'.  Got: 0x1"
value      = '0x1'

```
Related to Issue #

### How was it fixed?

added `eth_getCode` to `ethereum_tester_middleware` to appease eth-tester

### DIFF
--- a/web3/utils/module_testing/eth_module.py
+++ b/web3/utils/module_testing/eth_module.py
@@ -152,12 +152,17 @@ class EthModuleTest:
 
     def test_eth_getCode(self, web3, math_contract_address):
         code = web3.eth.getCode(math_contract_address)
-        assert is_string(code)
-        assert len(code) > 2
+        assert isinstance(code, HexBytes)
+        assert len(code) > 0
 
     def test_eth_getCode_invalid_address(self, web3, math_contract):
         with pytest.raises(InvalidAddress):
             web3.eth.getCode(math_contract.address.lower())
+
+    def test_eth_getCode_with_block_identifier(self, web3, emitter_contract):
+        code = web3.eth.getCode(emitter_contract.address, block_identifier=web3.eth.blockNumber)
+        assert isinstance(code, HexBytes)
+        assert len(code) > 0
 
     def test_eth_sign(self, web3, unlocked_account_dual_type):
         signature = web3.eth.sign(
--- a/web3/providers/eth_tester/middleware.py
+++ b/web3/providers/eth_tester/middleware.py
@@ -201,6 +201,10 @@ ethereum_tester_middleware = construct_formatting_middleware(
             apply_formatter_if(is_not_named_block, to_integer_if_hex),
         ),
         'eth_uninstallFilter': apply_formatters_to_args(hex_to_integer),
+        'eth_getCode': apply_formatters_to_args(
+            identity,
+            apply_formatter_if(is_not_named_block, to_integer_if_hex),
+        ),
         # EVM
         'evm_revert': apply_formatters_to_args(hex_to_integer),
         # Personal


web3/utils/module_testing/eth_module.py
```diff
@@ -159,6 +159,11 @@ def test_eth_getCode_invalid_address(self, web3, math_contract):
         with pytest.raises(InvalidAddress):
             web3.eth.getCode(math_contract.address.lower())
 
+    def test_eth_getCode_with_block_identifier(self, web3, emitter_contract):
+        code = web3.eth.getCode(emitter_contract.address, block_identifier=web3.eth.blockNumber)
```
voith: > `block_identifier=web3.eth.blockNumber`

Don't know if this is the right thing to do. The ideal way would be to deploy a contract and get the `blockNumber` of the transaction in which it was mined and then call this method. 
However, I inspected the tests and found that only block was added. I wanted to fix this quick, so didn't spend much time. Let me know what guys think about it!

pipermerriam: This is an acceptable solution for now.  :+1:

web3/utils/module_testing/eth_module.py
```diff
@@ -159,6 +159,11 @@ def test_eth_getCode_invalid_address(self, web3, math_contract):
...
+        assert is_string(code)
+        assert len(code) > 2
```
carver: These tests read as if the code is a hex string, but it should actually be a `HexBytes` (because of https://github.com/ethereum/web3.py/blob/master/web3/middleware/pythonic.py#L307 ). It passes anyway, because `is_string` accepts `bytes` and the code is longer than two bytes. But I think it could be confusing.

Some alternative options:
1. assert what the code is equal to exactly: `assert code == Web3.toBytes(hexstr='0x`...
2. assert that the code is bytes, and is exactly the expected length:
```py
assert isinstance(code, bytes)
assert len(code) == ...
```

voith: @carver I had the same thought in mind when I started writing this, but then I had a look at the surrounding code and found:
https://github.com/ethereum/web3.py/blob/910443d8a17d73841341309f8497893a96ddb3b6/web3/utils/module_testing/eth_module.py#L153-L156
But Looks like that test was written before `Hexbytes` was introduced. I'll change it at both places.

voith: > assert code == Web3.toBytes(hexstr='0x')
assert len(code) == ...

I think from the above code you mean to compare deployed code and bytecode runtime. The two need not be the same. I can say this from my previous work on [compare_bytecode](https://github.com/ethereum/populus/pull/471/files#diff-8fa659c73d2daba076463f01c4b4d3e9R170). However, you would know better what is going on internally. The check above won't work.
The lengths don't match too. 
```
(Pdb) len(add_0x_prefix(EMITTER_BYTECODE))
2434
(Pdb) len(code)
2398
```
Or do you mean to say that I should hardcode the length check to 2398 in the test? 

I have changed the `is_string` check to `isinstance(code, HexBytes)`. 
But I kept the length check as is for reasons explained below.  `len > 2` ensures that the bytecode is not empty (`0x`) .

I added a commit to speed up the process if you agree. If you have anything else in mind then I'll revert the last commit.

carver: > len > 2 ensures that the bytecode is not empty (0x).

The `0x` is not counted in the length of a `HexBytes` object. If we just want to check non-empty, which is fine, then let's just test `len(code) > 0`

voith: Ah, I was fooled by `HexBytes`'s `__repr__`. I didn't actually check the length.

voith: @carver Done. squashed all commits.

carver:
Thanks!

PR #974, opened 2018-07-30
Remove utf-8 string decoding from normalizers

dylanjw:
### What was wrong?
The eth-abi v2 alpha release adds utf-8 string decoding to the string type decoder.  String decoding should therefore be removed from web3 return data normalizers.

#### Cute Animal Picture

image


dylanjw:
@pipermerriam Updated to add support for eth-abi >=2 without removing support for <2.

web3/utils/normalizers.py
```diff
@@ -140,11 +144,14 @@ def abi_ens_resolver(w3, abi_type, val):
 
 
 BASE_RETURN_NORMALIZERS = [
-    addresses_checksummed,
-    decode_abi_strings,
+    addresses_checksummed
```
carver: *style nitpick* leave the trailing comma.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -184,6 +188,9 @@ def test_call_get_string_value(string_contract, call):
     assert result == "Caqalai"
 
 
+@pytest.mark.skipif(
+    LooseVersion(eth_abi.__version__) >= LooseVersion("2"),
+    reason="eth-abi >=2 does utf-8 string decoding")
 def test_call_read_string_variable(string_contract, call):
     result = call(contract=string_contract,
                   contract_function='constValue')
```
carver: Maybe add a second test that shows what happens on an ABI decoding error. We should probably do something like capture the eth-abi error and re-raise a generic `ValueError`.

dylanjw: It raises a 'BadFunctionCallOutput' error

dylanjw:
@pipermerriam @carver Ready for another review

setup.py
```diff
@@ -19,7 +19,7 @@
     install_requires=[
         "toolz>=0.9.0,<1.0.0;implementation_name=='pypy'",
         "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
-        "eth-abi>=1.1.1,<2",
+        "eth-abi==2.0.0a1",
```
carver: We shouldn't require people to upgrade. We especially shouldn't require people to install an alpha dependency in a stable release. `eth-abi>=1.1.1,<3` seems like a reasonable choice

dylanjw:
@pipermerriam @carver Is this good to merge?

Issue #975, opened 2018-07-30
Merging topic lists when creating filters not possible in current api.  Yields unexpected results.


dylanjw:
### What was wrong?

The `contract.events.myEvent.createFilter` and `contract.eventFilter` both allow entering topics directly along with "argument filters" via `filter` param dict.  The implementation is built on a false assumption about the topic list.  The constructed topics are merged with the raw topics in a simple list addition:

```
    if topics is None:
        topic_set = construct_event_topic_set(event_abi, argument_filters)
    else:
        topic_set = [topics] + construct_event_topic_set(event_abi, argument_filters)
```

Merging two topic lists with additive results is impossible with a single filter.  Merging two topic lists by using the available OR logic within single topic positions would yield all topic combinations.

Example:
Given the following "logs": AA BB AB
Topic list 1 ["A", "A"] would match AA
Topic list 2 ["B", "B"] would match BB
merged list [["A", "B"], ["A","B"]] would match AA BB AB

### How can it be fixed?
Option 1. Remove option of entering raw topics along with argument filters that generate topics.
Option 2. Provide the functionality of merging topic lists, by extending the Filter object to manage multiple node filters.

something like:

```python
FilterAggregator(Filter):
    def __init__(self, web3, filter_ids):
        self.web3 = web3
        self.filter_ids = filter_ids
    
    def get_new_entries(self):
        log_entries = mapcat(
            compose(self._filter_valid_entries, self.web3.eth.getFilterChanges),
            filter_ids)
        return self._format_log_entries(log_entries)

    def get_all_entries(self):
            log_entries = mapcat(
                compose(self._filter_valid_entries, self.web3.eth.getFilterLogs),
                self.filter_ids)
        return self._format_log_entries(log_entries)
```

dylanjw:
There isnt a test that proves this, but Im 99% sure that this is broken.  I'll write the broken test before starting on either option to fix.

carver:
I think Option 1 is better. Clearly it's subtle enough that reasonable people can get confused by what merging is supposed to do. So let's just remove the opportunity for mistake. People can still manually merge topics themselves first, if they like.

PR #976, opened 2018-07-30
Disallow configuring filters with both manual and generated topic lists

dylanjw:
### What was wrong?

Related to Issue #975 

### How was it fixed?
Only allow configuring log topic filters via the `topics` or `argument_filters` parameter and not both.

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/utilities/test_construct_event_topics.py
+++ b/tests/core/utilities/test_construct_event_topics.py
@@ -31,48 +31,46 @@ def hex_and_pad(i):
         (
             EVENT_1_ABI,
             {},
-            [[EVENT_1_TOPIC]],
+            [EVENT_1_TOPIC],
         ),
         (
             EVENT_1_ABI,
             {'arg0': 1},
-            [[EVENT_1_TOPIC]],
+            [EVENT_1_TOPIC],
         ),
         (
             EVENT_1_ABI,
             {'arg0': 1, 'arg3': [1, 2]},
-            [[EVENT_1_TOPIC]],
+            [EVENT_1_TOPIC],
         ),
         (
             EVENT_1_ABI,
             {'arg1': 1},
             [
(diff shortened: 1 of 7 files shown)


tests/core/utilities/test_construct_event_topics.py
```diff
@@ -31,48 +31,46 @@ def hex_and_pad(i):
         (
             EVENT_1_ABI,
             {},
-            [[EVENT_1_TOPIC]],
```
carver: This API change doesn't have any downstream effects in the public API?

dylanjw: There shouldn't be a bad effect. Because the topic list permutations were removed, `construct_event_topics` only needs to return a nested list when there are multiple options for a single argument filter.

There was a bug in how the topic list was being generated.  This is an example of a correctly formated topic list to match logs with either of two values for the first indexed event arg:

```
[ event_sig_topic, [option1_for_arg1, option2_for_arg1], arg2_topic ]
```

whereas this is what we were generating to be equivalent to the above; we would pass full topic list permutation against each option:
```
[
    [ event_sig_topic, option1_for_arg1, arg2_topic ],
    [ event_sig_topic, option1_for_arg1, arg2_topic ]
]
```
This topic topic list actual means: any of `[event_sig_topic, option1_for_arg1, arg2_topic ]` in the first position and any of `[event_sig_topic, option1_for_arg1, arg2_topic]` in the second position.  Node filters don't understand nested lists as full topic lists.  The spec defines a single topic list that can have nested lists of options for each positional element.

dylanjw: I can break this part out into another PR, and add a failing test.

This issue has been fixed in the new api: https://github.com/ethereum/web3.py/pull/953.  Im fixing this function to make it easier to integrate the new filter builder api.

PR #977, opened 2018-07-30
Replace eth_abi.abi.process_type with eth_abi.grammar.parse

dylanjw:
### What was wrong?

Related to Issue https://github.com/ethereum/eth-abi/issues/85

### How was it fixed?
Replace process_type with grammar.parse

#### Cute Animal Picture
image


carver:
Maybe better to replace with a local `process_type` utility function. Then it can detect which version of eth-abi is in use and use the appropriate option.

web3/utils/abi.py
```diff
@@ -111,6 +116,52 @@ def filter_by_argument_name(argument_names, contract_abi):
...
+
+    def collapse_type(base, sub, arrlist):
+        return base + str(sub) + ''.join(map(repr, arrlist))
+else:
+    from eth_abi.abi import (
+        process_type,
+        collapse_type,
+    )
+
+
```
dylanjw: @carver Something like this?

carver: Yup, although I think it should be >= v1 in the `if` test. It highlights the breaking change at the major version. (Also, 0.5.0 is already released, and we could theoretically release an 0.6.0 on the v0/stable branch).

web3/utils/abi.py
```diff
@@ -1,15 +1,20 @@
...
 import re
 
+import eth_abi
 from eth_abi import (
     is_encodable as eth_abi_is_encodable,
 )
-from eth_abi.abi import (
-    collapse_type,
-    process_type,
+from eth_abi.grammar import (
```
carver: This would have to move into the `if` test.

.circleci/config.yml
```diff
@@ -356,7 +385,8 @@ workflows:
       - py35-integration-parity-ws
       - py35-integration-ethtester-pyethereum
       - py35-integration-ethtester-pyevm
-      - py36-core
+      - py36-core-eth_abi0
+      - py36-core-eth_abi1
```
carver: Can we start testing against eth-abi v2 yet, or is the alpha broken?

dylanjw:
@carver eth_abi < v1 is failing due to incompatibility with the other dependencies. Because the latest web3 is incompatible with eth_abi < 1, should I remove the conditional import of the collapse_type and process_type functions?

carver:
> Because the latest web3 is incompatible with eth_abi < 1, should I remove the conditional import of the collapse_type and process_type functions?

Actually, I can't see why we'd test <1 at all. The web3.py requirement is >1.1.1: https://github.com/ethereum/web3.py/blob/master/setup.py#L22

But we still need to conditionally import, because they won't be there anymore by the time v2 leaves alpha.

dylanjw:
@carver Because eth-abi 2.0.0a1 has some failing tests I will put it in a separate pull request, where the failing tests can be sorted out.

dylanjw:
Will need to rebase once https://github.com/ethereum/web3.py/pull/974 is merged

carver:
Both core tests are running against eth ABI 1.1.1 - it probably makes sense to release the next eth abi v2 alpha with process_type removed, and then pin one of these tests to >= that v2 version.

dylanjw:
Got a error with the caching middleware:

```
=================================== FAILURES ===================================
_____________ test_latest_block_based_cache_middleware_busts_cache _____________

w3 = <web3.main.Web3 object at 0x7fab1353eda0>
mocker = <pytest_mock.MockFixture object at 0x7fab1353e160>

    def test_latest_block_based_cache_middleware_busts_cache(w3, mocker):
        result = w3.manager.request_blocking('fake_endpoint', [])
    
        assert w3.manager.request_blocking('fake_endpoint', []) == result
        w3.testing.mine()
    
        # should still be cached for at least 1 second.  This also verifies that
        # the middleware caches the latest block based on the block time.
>       assert w3.manager.request_blocking('fake_endpoint', []) == result
E       AssertionError: assert '2adffae3-7f2...-5411065fd33d' == 'e2b2f6df-7c20...-c35710a47abe'
E         - 2adffae3-7f2a-4e24-b04c-5411065fd33d
E         + e2b2f6df-7c20-44d7-a12b-c35710a47abe

mocker     = <pytest_mock.MockFixture object at 0x7fab1353e160>
result     = 'e2b2f6df-7c20-44d7-a12b-c35710a47abe'
w3         = <web3.main.Web3 object at 0x7fab1353eda0>

tests/core/middleware/test_latest_block_based_cache_middleware.py:189: AssertionError
```

carver:
> Got a error with the caching middleware:

Was this resolved and then rebased? I don't see any failing CI run.

.circleci/config.yml
```diff
@@ -111,19 +111,12 @@ jobs:
...
   #
-  py35-core-eth_abi0:
+  py35-core:
     <<: *common
     docker:
       - image: circleci/python:3.5
     environment:
-      TOXENV: py35-core-eth_abi0
-
-  py35-core-eth_abi1:
```
carver: I think we should keep running the eth-abi v1 tests.

web3/utils/abi.py
```diff
@@ -111,6 +107,58 @@ def filter_by_argument_name(argument_names, contract_abi):
...
+    def process_type(type_str):
+        normalized_type_str = normalize_type_string(type_str)
+        abi_type = parse_type_string(normalized_type_str)
+
+        type_str_repr = repr(type_str)
+        if type_str != normalized_type_str:
+            type_str_repr = '{} (normalized to {})'.format(
+                type_str_repr,
+                repr(normalized_type_str),
+            )
```
carver: Looks like all this type_str work can go inside the `TupleType` check.

PR #983, opened 2018-07-31
Percentile function to handle out of range percentiles

### LINKED ISSUE #979: Crash in time based gas price strategies
* Version: 4.5.0
* Python: 3.6
* OS: osx

### What was wrong?

My script to check the gas price calculation crashes after the update to web3 4.5:
```
python test_gas_price.py                                                                     0:13:21
RPC: 252000000
  File "test_gas_price.py", line 14, in <module>
    print('Slow', w3.eth.generateGasPrice())
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/web3/eth.py", line 385, in generateGasPrice
    return self.gasPriceStrategy(self.web3, transaction_params)
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/web3/gas_strategies/time_based.py", line 160, in time_based_gas_price_strategy
    sample_size=sample_size,
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/eth_utils/functional.py", line 22, in inner
    return callback(fn(*args, **kwargs))
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/web3/gas_strategies/time_based.py", line 79, in _compute_probabilities
    reverse=True,
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/web3/gas_strategies/time_based.py", line 67, in _aggregate_miner_data
    percentile(gas_prices, percentile=15))
  File "/Users/paul/Work/venv-raiden/lib/python3.6/site-packages/web3/utils/math.py", line 5, in percentile
    raise ValueError("Expected a sequence of at least 3 integers, got {0}".format(values))
ValueError: Expected a sequence of at least 3 integers, got {41000000000, 26000000000}
```

The code is the following:
```py
from web3.auto import w3

from web3 import middleware
from web3.gas_strategies.time_based import medium_gas_price_strategy, fast_gas_price_strategy, slow_gas_price_strategy
from web3.gas_strategies.rpc import rpc_gas_price_strategy

w3.middleware_stack.add(middleware.time_based_cache_middleware)
w3.middleware_stack.add(middleware.latest_block_based_cache_middleware)
w3.middleware_stack.add(middleware.simple_cache_middleware)

w3.eth.setGasPriceStrategy(rpc_gas_price_strategy)
print('RPC:', w3.eth.generateGasPrice())
w3.eth.setGasPriceStrategy(slow_gas_price_strategy)
print('Slow', w3.eth.generateGasPrice())
w3.eth.setGasPriceStrategy(medium_gas_price_strategy)
print('Medium:', w3.eth.generateGasPrice())
w3.eth.setGasPriceStrategy(fast_gas_price_strategy)
print('Fast:', w3.eth.generateGasPrice())
```

### How can it be fixed?

It should crash and return a proper gas price.

dylanjw:
### What was wrong?

Fixes #979

### How was it fixed?
The percentile function requires at least 3 data points to make a calculation. The percentile now throws a custom exception that is caught and `min()` is used as an alternative in these cases.

#### Cute Animal Picture

image


web3/gas_strategies/time_based.py
```diff
@@ -60,11 +61,15 @@ def _aggregate_miner_data(raw_data):
 
     for miner, miner_data in data_by_miner.items():
         _, block_hashes, gas_prices = map(set, zip(*miner_data))
+        try:
+            price_percentile = percentile(gas_prices, percentile=15)
+        except InsufficientData:
+            price_percentile = min(gas_prices)
```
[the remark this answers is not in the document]
dylanjw: Just because `min()` is what was being calculated before the percentile function.  In a private network with just a few transactions per block, wouldn't using the last block `max(prices)` potentially send the price on a continual upward trend?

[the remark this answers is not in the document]
carver: I suppose, but it could also trend down with min. Also maybe it doesn't matter on a private network? I don't feel strongly about it.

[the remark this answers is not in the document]
palango: Couldn't you also include transactions from the last few blocks when there is not enough data available?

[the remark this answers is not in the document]
carver: It's already looking across 120 blocks by default, but this is grouped by miner. So a few scenarios could cause a small data set:

- we are near the beginning of the chain, with only a few blocks to sample
- even with many blocks and transactions, maybe this particular miner has only mined a few blocks/transactions

dylanjw:
This is ready for a review.  I noticed that the issue with too few datapoints when making the percentile calculation had to do with an error when a percentile falls outside the range of numbers.  For example, the 15th percentile would break for <7 data points because 1/6 ~ 17%.  Now when the percentile choice falls below the lowest value, the lowest value is returned.

web3/utils/math.py
```diff
@@ -1,8 +1,13 @@
...
+)
+
+
 def percentile(values=None, percentile=None):
     """Calculates a simplified weighted average percentile
     """
-    if values in [None, tuple(), []] or len(values) < 3:
-        raise ValueError("Expected a sequence of at least 3 integers, got {0}".format(values))
+    if values in [None, tuple(), []] or len(values) < 1:
+        raise InsufficientData("Expected a sequence of at least 1 integers, got {0}".format(values))
```
carver: Maybe `got {0!r}` would help show exactly what weird value was sent.

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
```diff
@@ -108,21 +117,21 @@ def _get_block_by_something(method, params):
...
+        (dict(max_wait_seconds=40, sample_size=5, probability=98), 100),
+        (dict(max_wait_seconds=40, sample_size=5, probability=90), 81),
+        (dict(max_wait_seconds=40, sample_size=5, probability=50), 7),
         # 20 second wait times
-        (dict(max_wait_seconds=20, sample_size=5, probability=98), 63),
-        (dict(max_wait_seconds=20, sample_size=5, probability=90), 62),
-        (dict(max_wait_seconds=20, sample_size=5, probability=50), 50),
+        (dict(max_wait_seconds=20, sample_size=5, probability=98), 100),
+        (dict(max_wait_seconds=20, sample_size=5, probability=90), 100),
+        (dict(max_wait_seconds=20, sample_size=5, probability=50), 34),
```
carver: I'm not reviewing these numbers, just trusting that you double-checked them. :)

dylanjw: It might be a good idea to at least once check how accurate the target times are when compared to actual times on ropsten or even mainnet.

carver: :+1: 

Also, as a quicker test, we could compare against https://ethgasstation.info/.

dylanjw: Ooh yeah.  Then I could compare to main net without spending any gas.

dylanjw: The estimates are comparable to ethgasstation.info, but I noticed that with longer acceptable wait times we were a bit lower, so I ran some tests by sending a transaction back and forth on on mainnet using our estimate.  I tried raising the percentile.  Given the results, I am going to bump up the percentile to 20%, since it looks the best.

15% percentile
| Expected        | Actual           |
| ------------- |:-------------:|
| <2m      | 2m |
| <5m      | 10m, 4m      |
| <30m | 13m      |


20% percentile
| Expected        | Actual           |
| ------------- |:-------------:|
| <2m      |1.5m  |
| <5m      | 2m, 4m, 3m      |
| <30m | 15m      |

25%
| Expected        | Actual           |
| ------------- |:-------------:|
| <2m      | 2m |
| <5m      | 2m      |
| <30m | 3m      |

40%
| Expected        | Actual           |
| ------------- |:-------------:|
| <30m | 7m      |

carver: Solid :+1:

PR #985, opened 2018-08-01
Add test to verify client consistency with None topic args

dylanjw:
### What was wrong?
Want to verify client handling of None topic parameters yields consistent match results.

### How was it fixed?
Extended the getLogs test to try retrieving log matches using None values for topics.

#### Cute Animal Picture

image

### DIFF
--- a/tests/integration/parity/common.py
+++ b/tests/integration/parity/common.py
@@ -97,6 +97,10 @@ class ParityEthModuleTest(EthModuleTest):
         assert receipt is not None
         assert receipt['blockHash'] is None
 
+    def test_eth_getLogs_with_logs_none_topic_args(self, web3):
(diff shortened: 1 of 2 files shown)


tests/integration/parity/common.py
```diff
@@ -97,6 +97,10 @@ def test_eth_getTransactionReceipt_unmined(self, web3, unlocked_account):
         assert receipt is not None
         assert receipt['blockHash'] is None
 
+    def test_eth_getLogs_with_logs_none_topic_args(self, web3):
+        with pytest.raises(AssertionError):
```
carver: In all the other places I can think of, we do `@pytest.mark.xfail` instead of this approach. Nothing wrong with this that I can think of, it's just nice to have a little more stylistic consistency.

Issue #671, opened 2018-02-23
Test newBlockFilter in Parity 1.10


carver:
Pending the release of Parity 1.10...

### What was wrong?

Parity 1.10 is scheduled to fix the bug mentioned in #381 (new block filter returns the current latest block).

### How can it be fixed?

Add a Parity 1.10 integration tests, and remove the `xfail` in `test_eth_newBlockFilter`.

PR #992, opened 2018-08-06
Remove xfail for newBlockFilter parity integration test

dylanjw:
### What was wrong?

Related to Issue #671 

#### Cute Animal Picture
image


carver:
> ...wondering why xfail_strict=true in our pytest.ini file didn't cause the XPASS to be marked as a fail before.

Hmm, that seems like a good thing to investigate.

PR #994, opened 2018-08-07
Fix typo

akhavr:
### What was wrong?

s/trensactions/transactions/

### How was it fixed?

By substituting a correct character :) 

#### Cute Animal Picture


dylanjw:
Thanks for catching that!

PR #995, opened 2018-08-08
Add ERC20 example

### LINKED ISSUE #951: Add an ERC20 balance check example to the docs
### What was wrong?

ERC20 is one of the most common use cases, and there are no explicit examples for interacting with them.

### How can it be fixed?

Add a section to the [Examples section of the docs](http://web3py.readthedocs.io/en/stable/examples.html) that shows how to check the balance of a given token. The new section can be added at the end of this source: https://github.com/ethereum/web3.py/blob/master/docs/examples.rst

This section should show someone how to check their Unicorn token balance.

Rough approach:

1. Show how to create a python contract object bound to the [ERC20 Unicorn token](https://etherscan.io/token/0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7) (you can use the ABI from `ethtoken` linked below)
2. Retrieve the token balance for an address [`0xd1220a0cf47c7b9be7a2e6ba89f429762e7b9adb`](https://etherscan.io/token/0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7?a=0xd1220a0cf47c7b9be7a2e6ba89f429762e7b9adb) appears to be the first address to have received a unicorn, so could be a good candidate.
3. Show how to retrieve the `decimals()`, and how to adjust the token balance to get the token count: dividing by `10^decimals` plus a brief explainer of the difference between the token balance and token count (token balance is always an integer, token count may be a fraction)

This repository has some example code that might be helpful if you're not sure how to connect to an ERC20 token: https://github.com/carver/ethtoken.py (You can footnote the repo if you want, but it's best to copy in anything that the reader might need).

ERC20 probably deserves its own whole page with different use case examples, but that is out of scope.

njgheorghita:
### What was wrong?
No ERC20 example in docs.

fixes #951 

### How was it fixed?
Added an example showing how to interact with an erc20 contract.

#### Cute Animal Picture
image

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -306,3 +306,56 @@ Output:
      'logs': [AttributeDict({'type': 'mined', 'logIndex': 0, 'transactionIndex': 0, 'transactionHash': HexBytes('0x3ac3518cc59d1698aa03a0bab7fb8191a4ef017aeda7429b11e8c6462b20a62a'), 'blockHash': HexBytes('0x94e07b0b88667da284e914fa44b87d4e7fec39761be51245ef94632a3b5ab9f0'), 'blockNumber': 2, 'address': '0xF2E246BB76DF876Cef8b38ae84130F4F55De395b', 'data': '0x', 'topics': [HexBytes('0x6c2b4666ba8da5a95717621d879a77de725f3d816709b9cbe9f059b8f875e284'), HexBytes('0x00000000000000000000000000000000000000000000000000000000000000ff')]})],
      'transactionHash': HexBytes('0x3ac3518cc59d1698aa03a0bab7fb8191a4ef017aeda7429b11e8c6462b20a62a'),
      'transactionIndex': 0}
+
+
+Interacting with an ERC20 Contract
+----------------------------------
+
+The ERC20 (formally `EIP20 <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md>`_) token standard is the most widely used standard in Ethereum. Here's how to check the current state of an ERC20 token contract.
+
+First, create your Python object representing the ERC20 contract.
+For this example, we'll be inspecting the `Unicorns <https://etherscan.io/token/0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7>`_ token contract.
+
+.. code-block:: python
+
+    import json
+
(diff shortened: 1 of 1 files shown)


njgheorghita:
Should the ethpm example eventually go here? or under the `pm-module` doc page?

docs/examples.rst
```diff
@@ -306,3 +306,49 @@ Output:
...
+
+    # Address field should be the checksum address at which the ERC20 contract was deployed
+    erc20 = w3.eth.contract(address='0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7', abi=ERC20_ABI)
+
+Fetch various data about the current state of the ERC20 contract.
+
+.. code-block:: python
+
+    name = erc20.functions.symbol().name()
+    # -> 🦄
```
carver: Name should be `Unicorns`

docs/examples.rst
```diff
@@ -306,3 +306,49 @@ Output:
...
+.. code-block:: python
+
+    balance = erc20.functions.balanceOf('0xD1220A0cf47c7B9Be7A2E6BA89F429762e7b9aDb').call()
+    # -> 1
+
+Calculate the token count from the token balance of an account address. (token count *may* be a fraction)
+
+.. code-block:: python
+    
+    count = balance / (10 ** decimals)
```
carver: Try `Decimal()` instead of `float` here, with a high enough precision to avoid any rounding error. Most applications expect exact results, even if fractional.

docs/examples.rst
```diff
@@ -306,3 +306,56 @@ Output:
...
+
+    # Address field should be the checksum address at which the ERC20 contract was deployed
+    erc20 = w3.eth.contract(address='0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7', abi=ERC20_ABI)
+
+Fetch various data about the current state of the ERC20 contract.
+
+.. code-block:: python
```
pipermerriam: *if* we're willing to do this over an infura HTTP provider these could actually be doctested quite easily

docs/examples.rst
```diff
@@ -306,3 +306,56 @@ Output:
...
+    from decimal import Decimal
+    
+    # Most applications expect *exact* results, so using the Decimal library,
+    # with default precision of 28, is usually sufficient to avoid any rounding error.
+    count = balance / Decimal(10 ** decimals)
+    # -> 1
+
+Balance is *always* an integer in the currency's smallest natural unit (i.e. `wei` for ether). Token balance is typically used only for backend calculations.
+
+Token count *may* be a integer or fraction in the currency's primary unit (i.e. `eth` for ether). Token count is typically displayed to the user on the front-end since it is more readable.
```
carver: Instead of `i.e.` (which means "in other words"). Since ether isn't an ERC20 token, this could be confusing. I would say something like `(equivalent to 'eth' for ether)`

njgheorghita:
@carver I added doctest to the erc20 example, not sure if there's a better solution for using the Infura API key? Open to any suggestions or leaving as is.

carver:
@njgheorghita Yeah, I assume you generated a fresh API key for this?

I can't really think of a better solution. Somehow the key will have to be in the source. But we can keep it from showing in the docs by moving it to a test setup block: http://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testsetup

pipermerriam:
@njgheorghita you could also use an encrypted environment variable on circle-ci **but** that won't really save us since we allow CI to run on forks and thus, anyone can extract the environment variable simply by exporting it with custom code during a CI run....

docs/examples.rst
```diff
@@ -306,3 +306,64 @@ Output:
...
+First, create your Python object representing the ERC20 contract.
+For this example, we'll be inspecting the `Unicorns <https://etherscan.io/token/0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7>`_ token contract.
+
+.. doctest::
+
+    >>> import json
+    >>> import os
+    
+    # Please play nice and don't use this key for any shenanigans, thanks!
+    >>> os.environ['INFURA_API_KEY'] = "b2679bb624354d1b9a2586154651b51f"
```
pipermerriam: :+1: to @carver suggestion to move this into a setup block so it doesn't show up in the actual documentation.

njgheorghita:
@carver Yup, generated a fresh key for this, happy to keep an eye on it for any abuse and regenerate if needed. Thanks for the setup tip, moved the key to a hidden block. Let me know if you'd like any other changes

carver:
LGTM

Issue #836, opened 2018-05-10
Notes on probable bugs in filters


pipermerriam:
### What is wrong?

There are *probably* bugs lurking in contract event filters that filter on data values.

> None of these have actually been verified, this is strictly from me looking over the code.

#### 1. Variable length data elements.

suppose the log data has an ABI with a variable length value *before* other values. `MyEvent(string,bytes32)`

The code which constructs the regex here: https://github.com/ethereum/web3.py/blob/master/web3/utils/filters.py#L138 assumes fixed length placeholders for all variable length data elements which I don't believe is correct.

#### 2. List types

suppose my event has a list type in it like `MyEvent(bytes2[])`

To filter on this value, one might pass in arguments like `filters={'arg_a': [b'aa', b'xx']}`.

However, when the data filters are constructed: https://github.com/ethereum/web3.py/blob/master/web3/utils/events.py#L99

With *normal* single value arguments, if you want to filter against *multiple* options, you pass those options in as a list.  For single values, they should be passed in *singularly*.  For example, to filter against a single `uint` you would do `filters={'arg_X': 1234}` and to filter against multiple options for field `arg_X`, `filters={'arg_X', [1234, 4321]}`.

In the linked code in the `events.py` utils, it assumes that anything that `is_list_like` has already been wrapped in a list, which will result in an error either during execution, or an incorrectly constructed regex.

#### 3. dynamic sized types.

This one I'm **not very sure of at all**.  however,

The data filter regex is constructed using `encode_single`, and then all of the values are joined together to form the regex.  However, this is not how ABI encoding works... as dynamic sized things are actually shoved to a later offset in the encoded data.  **if** events encode their data field the same way, the regex won't match.

### How can we fix this.

All three of these issues deal with regex issues....

https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/

So, lets get rid of the regex.  Rather than regexing the data, lets instead just *try* decoding it using the event ABI, and then matching against the provided arguments in their native python form.

PR #997, opened 2018-08-09
Extend log filter tests

dylanjw:
### What was wrong?

Related to Issue #836 

### How was it fixed?
Adding to the suite of filter tests to show cases that are currently failing.  I am separating this work from #953, by marking tests with xfail.

### Test cases to cover

- [x] Argument filters of dynamic types
- [x] Argument filters of fixed types
- [x] Argument filters of array types
- [ ] Topic filters of dynamic types
- [ ] Topic filters of fixed types
- [ ] Topic filters of array types
- [ ] Block filters over 50+ blocks
- [ ] Log filters over 50+ blocks
- [ ] Transaction filters over 50+ blocks
- [ ] Transaction filters match transactions in unmined "pending" block
- [ ] Log filters matching a few events against many contracts, many emitted logs, many blocks.
- [ ] Contract with invalid return data

#### Cute Animal Picture
image


dylanjw:
Issues found so far:

Block filter only returns partial results when a lot of blocks have been mined (~50).
Pending transaction filter returns partial results when eth-tester auto-mining has been turned off.
Getting an error deploying and transacting with contracts on circleci (works in locally run tests.)

dylanjw:
Im getting errors deploying a contract with 2d arrays on eth-tester.  I will work on adding that test in a different PR.

This is ready for review. @carver @pipermerriam

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -0,0 +1,198 @@
...
+
+    for txn_hash in txn_hashes:
+        wait_for_transaction(web3, txn_hash)
+
+    log_entries = event_filter.get_new_entries()
+    assert len(log_entries) == 1
+    assert log_entries[0]['transactionHash'] == txn_hashes[0]
+
+
+@pytest.mark.xfail(reason="The regex based data filters do not work with dynamic sized types")
```
carver: What kind of failure do we get? Do we just not get the events we were looking for? It seems like we should explicitly raise an exception if we're incapable of generating the filter accurately. Then the test could be a `pytest.raises()` on that exception.

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -0,0 +1,158 @@
+import pytest
+import random
+
+from eth_utils import (
+    to_tuple,
+)
+
+SEED = 50505050505
+random.seed(SEED)
```
carver: Was this left in from some debugging?

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -0,0 +1,158 @@
...
+        which=1).transact
+    non_matching_transact = emitter.functions.logNoArgs(
+        which=0).transact
+
+    event_filter = emitter.events.LogNoArguments().createFilter(fromBlock='latest')
+
+    expected_match_counter = 0
+
+    tx_padder = pad_with_transactions(web3)
+    tx_padder.send(None)
```
carver: Since we're not sending anything, maybe `next(tx_padder)` is clearer here.

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -0,0 +1,158 @@
...
+    accounts = w3.eth.accounts
+    while True:
+        transact = yield
+        tx_hash = transact()
+        for tx_count in range(random.randint(0, 10)):
+            _from = accounts[random.randint(0, len(accounts) - 1)]
+            _to = accounts[random.randint(0, len(accounts) - 1)]
+            value = 50 + tx_count
+            w3.eth.sendTransaction({'from': _from, 'to': _to, 'value': value})
+        yield tx_hash
```
carver: I'm not seeing the benefit of this being a generator. It seems it could be a function you call when you want to pad, like:

```py
def pad_with_transactions(w3):
        accounts = w3.eth.accounts
        for tx_count in range(random.randint(0, 10)):
            _from = accounts[random.randint(0, len(accounts) - 1)]
            _to = accounts[random.randint(0, len(accounts) - 1)]
            value = 50 + tx_count
            w3.eth.sendTransaction({'from': _from, 'to': _to, 'value': value})
```

Then later, instead of invoking the generator and calling `next()`, you call `pad_with_transactions(w3)` again.

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def single_transaction(w3):
+    accounts = w3.eth.accounts
+    while True:
+        _from = accounts[random.randint(0, len(accounts) - 1)]
+        _to = accounts[random.randint(0, len(accounts) - 1)]
+        value = 50
+        tx_hash = w3.eth.sendTransaction({'from': _from, 'to': _to, 'value': value})
+        yield tx_hash
```
carver: Same here, doesn't seem to need to be a generator.

tests/core/filtering/test_filters_against_many_blocks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    transact_once = single_transaction(web3)
+    while transaction_counter < 100:
+        next(transact_once)
+        transaction_counter += 1
+
+    assert len(transaction_filter.get_new_entries()) == transaction_counter
+
+
+@pytest.mark.xfail
```
carver: Maybe add the reason that this is a suspected eth-tester bug?

PR #1008, opened 2018-08-19
 FreeBSD initial support

rayrapetyan:
### What was wrong?

FreeBSD OS wasn't supported

### How was it fixed?

Added necessary checks and provided dev installation instructions in readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Please see OS-specific instructions for:
 - [Linux](docs/README-linux.md#Developer-Setup)
 - [Mac](docs/README-osx.md#Developer-Setup)
 - [Windows](docs/README-windows.md#Developer-Setup)
+- [FreeBSD](docs/README-freebsd.md#Developer-Setup)
 
 Then run these install commands:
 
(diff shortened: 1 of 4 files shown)


rayrapetyan:
@carver, thanks! Do you mean tox -e smth? Any idea how to pass
```
--global-option=build_ext --global-option='-I/usr/local/include'
```
into tox properly?

voith:
@carver Doesn't seem like circleCI supports free bsd: https://hub.docker.com/r/circleci/python/tags/

> Do you mean tox -e smth? Any idea how to pass

@rayrapetyan I think @carver wants to have the test suite on a FreeBSD container in circleCI:
for eg: we have some code running on jessie:
https://github.com/ethereum/web3.py/blob/f1e8e91503153c9942db570ac88752fd63d0a278/.circleci/config.yml#L179

carver:
> Doesn't seem like circleCI supports free bsd

Thanks for checking that @voith ! That's exactly what I had in mind. Nothing more to add to this PR, then.

PR #1018, opened 2018-08-26
Fix #962

### LINKED ISSUE #962: Convert `setup.py` to use preferred style for dependencies
### What was wrong?

The `setup.py` file is not setup in the way most of our other projects are these days.

### How can it be fixed?

Move all of the extras required to a single data structure [like this](https://github.com/carver/ethereum-python-project-template/blob/master/setup.py#L8-L29).  Make sure to add a `[dev]` extras require for easy development installation.

Also update any docs to match this new pattern.

Mukundan314:


pipermerriam:
@carver I can't figure out why lint fails when run within `tox` and not when I run it directly.  You have a bit more experience with `isort`.  Any thoughts?

carver:
Yeah, I think it happens when you are missing a third party lib inside tox. So for example, `isort` seems to think hypothesis isn't an installed library, and then groups it with the locally installed software. So either we can explicitly add it as a dependency when running `flake8`, or we can add it to the isort section of `tox.ini`, in the `known_third_party=` value. Then `isort` knows it's a 3rd-party-lib even though it's not installed.

dylanjw:
closed in favor of #1030

Issue #722, opened 2018-03-27
Planned v5 changes


carver:
v5 has no planned release date, but it would be nice to have some ideas in mind for what we would want to change.

### Potential v5 backwards-incompatible changes

Each of these line items should have its own issue. This is just a place to link them all (and a reminder to create the issue later).

*Edit*: I'll try to update this list to keep it in sync with the comments below:

- [x] drop py3.5 support #1163 
- [x] rename w3.middleware_stack so it's not called stack #1210
- [x] switch to a single provider instead of a list (maybe) #1200 
- [x] rename `Web3.sha3` to `Web3.keccak` #1207
- [x] drop the transact/estimateGas support in `ConciseContract`, and rename to something like `ContractReader` - #1227 
- [x] ~`w3.async` things~ *shouldn't be API-breaking changes, or delay the v5 release*
- [x] only support standard json-rpc methods in the global Web3 namespace, with geth/parity extensions separated into sandboxed namespaces (for example, some `personal_*` calls are not globally [defined](https://github.com/ethereum/wiki/wiki/JSON-RPC) or [supported](https://wiki.parity.io/JSONRPC-personal-module.html) and should probably be moved, like `personal_importRawKey`. Might make sense to have those on something like: `w3.geth.personal.importRawKey`)
  - [x] #1198 
  - [x] #1211
- [x] Stop automatically looking up a testnet connection in `IPCProvider`. #1206 
- [x] Stop inferring `.eth` as a suffix on ENS names. #1205
- [x] remove deprecated methods: #1232 
  - [x] removal of `contract.call()`
  - [x] removal of `contract.transact()`
  - [x] removal of `contract.eventFilter()`
  - [x] removal of `contract.deploy()`
  - [x] removal of `contract.estimateGas()`
  - [x] removal of `contract.buildTransaction()`
- [x] drop `web3.providers.tester.EthereumTesterProvider` and `web3.providers.tester.TestRPCProvider` #1199 
- [x] drop `contract.eventFilter` #1032 
- [x] have `web3.eth.getBlock` and `web3.eth.getTransaction` raise an exception rather than return `None` #1218 
- [x] drop `web3.utils.*` in favor of `web3._utils.*` - #1282 

EDIT: I went through this thread and added all the breaking changes I saw so that they are all in one place. I think once we get through this list and do some testing, we can get a beta release out. - KC

pipermerriam:
Deprecate `Web3.sha3` and rename to `Web3.keccak`

[...]

PR #1020, opened 2018-08-27
Add warnings for web3.sha3, middleware_stack, and python 3.5

### LINKED ISSUE #1022: Rename middleware_stack to middleware_onion
### How can it be fixed?

see #1020

### LINKED ISSUE #1023: Rename sha3 to keccak
### How can it be fixed?

see #1020

dylanjw:
### What was wrong?

Related to Issue #722 

### How was it fixed?
I lumped together a few of the straight forward deprecation warnings:
- renaming sha3 to keccak (fixes #1023)
- renaming middleware_stack to middleware_onion (fixes #1022)
- adding a warning that support for python 3.5 will be dropped in web3.py v5

#### Cute Animal Picture

image


web3/__init__.py
```diff
@@ -1,9 +1,16 @@
...
 
 if sys.version_info < (3, 5):
     raise EnvironmentError("Python 3.5 or above is required")
 
+if sys.version_info < (3, 6):
+    warnings.warn(
+        "Support for Python 3.5 will be removed in web3.py v5",
+        category=DeprecationWarning,
+        stacklevel=2)
+
```
dylanjw: I opted to have both warnings print in the case that the python version is less than 3.5, so that the user doesnt have to update to 3.5 to learn that support for 3.5 will be dropped soon.

pipermerriam: The first one is an exception and will exit the python process before hitting this new warning.... so I don't think this behaves as you expect.

web3/main.py
```diff
@@ -132,11 +138,12 @@ def providers(self, providers):
         self.manager.providers = providers
 
     @staticmethod
+    @deprecated_for("This method has been renamed to web3.keccak")
     @apply_to_return_value(HexBytes)
     def sha3(primitive=None, text=None, hexstr=None):
```
carver: I'm worried that in the future we're going to change something in one implementation or the other, instead of both. Can we just make the deprecated version an alias to the new one, like:

```py
def sha3(...):
   return Web3.keccak(...)
```

I think it's okay to have the error message reference keccak instead of sha3 during deprecation time.

dylanjw:
Referencing the conversation at the end of https://github.com/ethereum/web3.py/issues/1039.  The middleware_stack deprecation warning may need to change or be broken out of this PR.

dylanjw:
I remove the commit renaming middleware_stack to middleware_onion, and added a few tests to ensure the deprecated method, Web3.sha3 works as expected.

Issue #794, opened 2018-04-26
Compatibility with rlp v1.*


carver:
### What was wrong?

pyrlp went through a major upgrade, web3.py should support both v0 and the new v1.

### How can it be fixed?

_Maybe_ the only requirement is to upgrade eth-tester after https://github.com/ethereum/eth-tester/pull/84 is merged and released. Will have to wait and test it out.

dylanjw:
> Maybe the only requirement is to upgrade eth-tester after ethereum/eth-tester#84 is merged and released. Will have to wait and test it out.

This appears to be the case...

carver:
Cool, how did you test it?

dylanjw:
circleci has been running with rlp==1.0.2.  Is that sufficient, or is there a more intentional test I could do?

carver:
Ah, well that tests v1, but not v0. I don't necessarily think we have to add an rlp v0 and v1 test to CI, but at least try it with a local checkout of rlp v0.

dylanjw:
Almost all tests fail with rlp=0.6.0 because py-evm requires rlp>=1.

All of the failures Ive looked at are because the 0.6.0 `eth.rlp.accounts.Account` does not have a copy method.

dylanjw:
Im sort of confused why we need to have support for rlp v0 when rlp is only an eth-tester/py-evm dependency.  eth-tester and py-evm are not backwards compatible with rlp v0.  

Do all web3 changes that rely on tests that rely on the latest py-evm need to be removed and put in the next major version bump?

carver:
> Im sort of confused why we need to have support for rlp v0 when rlp is only an eth-tester/py-evm dependency. eth-tester and py-evm are not backwards compatible with rlp v0.

Ah, I didn't remember that. No need to support rlp v0, then. Thanks for diving into it.

PR #1030, opened 2018-08-29
Convert setup

lastmjs:
### What was wrong?

Related to Issue #962 

### How was it fixed?

I got rid of the requirements .txt files, updated setup.py, tox.ini, and Dockerfile

#### Cute Animal Picture


lastmjs:
If all of my changes work, I believe I'll still need to update some docs.

lastmjs:
Something simple is wrong in `tox.ini`, I'm not sure the format based on https://github.com/ethereum/py-evm/blob/master/tox.ini#L45

dylanjw:
The circleci error is connected to what is being discussed in #794 

You are getting errors because pyethereum depends on rlp<1, and rlp>=1 is being installed.  This PR exposes that issue by installing eth-testrpc as part of of the `dev` extra_requires.

setup.py
```diff
@@ -5,6 +5,57 @@
...
+        "pytest-xdist==1.*",
+        "tox>=1.8.0",
+        "tqdm",
+        "when-changed"
+    ]
+}
+
+extras_require['dev'] = (
+    extras_require['tester'] +
+    extras_require['testrpc'] +
```
dylanjw: Removing this line should fix the failing tests.

lastmjs:
Removing the installation of eth-testrpc did not work, and I'm not sure now why pytest isn't being installed correctly

lastmjs:
Everything seems to be working fine now, I'm really not sure what the imports being incorrectly sorted problem is with the linter.

dylanjw:
Move the `dev` dependencies back to deps, instead of "extra".  The lint tox job needs web3 and its dependencies installed for isort to understand proper import ordering.

PR #1032, opened 2018-08-29
Removes depracated Contract.eventFilter. Fixes #1028.

### LINKED ISSUE #1028: Remove `contract.eventFilter`
* Version: 4.6

### What was wrong?

contract.eventFilter was deprecated in v4, and can be removed in v5.

Exef:
### What was wrong?

Contract.eventFilter() method marked as deprecated and it should be removed in v5.
Related to Issue #1028 

### How was it fixed?

I removed the deprecated function, the test that used it and documentation about it.

#### Cute Animal Picture

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -282,51 +282,6 @@ Each Contract Factory exposes the following methods.
     ``argument_filters``, optional. Expects a dictionary of argument names and values. When provided event logs are filtered for the event argument values. Event arguments can be both indexed or unindexed. Indexed values with be translated to their corresponding topic arguments. Unindexed arguments will be filtered using a regular expression.
     ``topics`` optional, accepts the standard JSON-RPC topics argument.  See the JSON-RPC documentation for `eth_newFilter <https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newfilter>`_ more information on the ``topics`` parameters.
 
-.. py:classmethod:: Contract.eventFilter(event_name, filter_params=None)
-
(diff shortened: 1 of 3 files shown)


Exef:
Thank you, @dylanjw for fixing errors. 

Can you tell me, what have I done wrong with test fixture?

dylanjw:
Thanks for contributing!

pytest fixtures can be parametrized, so that everytime a test uses that fixture, it is run with each parameter.

The fixture decorator `@pytest.fixture(params=['v3', 'v4'])` was passing both 'v3' and 'v4' values to `return_filter_by_api`.  Because the 'v3' conditional was removed, in the case when 'v3' was passed the ValueError on this line https://github.com/ethereum/web3.py/pull/1032/commits/523f4938317282a60ea71688ec50ad400304a7cd#diff-e3a6abe3e92df49f9a94069e9a05f9e7L147 was being raised.

pipermerriam:
@dylanjw before this is merged we'll need to get a legacy `v4` branch created to allow merging of these new `v5` only features into `master` (and then anything that is a bugfix needs to be backported to `v4`)

dylanjw:
The https://github.com/ethereum/web3.py/tree/v4 branch is up to date with 4.7.1

docs/contracts.rst
```diff
@@ -484,42 +439,6 @@ and the arguments are ambiguous.
         1
 
 
-
-.. _event-log-object:
```
veox: Slug is referenced elsewhere in the docs (via `:ref:`), which produces a warning.

Section below also referenced.

PR #1033, opened 2018-08-29
Move web3.utils to web3._utils

dylanjw:
### What was wrong?

`web3.utils` contains code for internal use, and that could be made more clear.

### How was it fixed?

`web3.utils` was moved to `web3._utils`

#### Cute Animal Picture

image


pipermerriam:
> Can AttributeDict be in different module then? Any project that uses python type hints will be hurt.

:+1: would you be up for opening a pull request which moves it to a new `web3.datastructures` module?

pipermerriam:
cc @carver regarding attrdict.  I see `eth-account` uses the `attrdict` library.  Should we converge usage to one or the other?  IIRC the `attrdict` library provided more functionality than we needed but there wasn't anything explicitely wrong with it.  I do see that we wrap it to make it immutable which is desirable.

carver:
> Can AttributeDict be in different module then? Any project that uses python type hints will be hurt.

Yeah, this sounds like a good reason to wait to move `utils` until v5. In case no one noticed, I really hate breaking downstream dependencies in a minor release... :D

carver:
>  I see eth-account uses the attrdict library. Should we converge usage to one or the other?

Yeah, I think a small wrapper around it could go in an eth-utils minor release :+1: and then both web3 and eth-account can use the same implementation.

pipermerriam:
Maybe we can reliably alias it?  I'd be fine special casing this move and adding a deprecation warning to the `web3.utils` module.

Exef:
@pipermerriam ok, I'll create PR with the data structures this weekend.

web3/utils/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+    module_testing,
+    normalizers,
+    request,
+    rpc_abi,
+    six,
+    threads,
+    toolz,
+    transactions,
+    validation,
+)
```
dylanjw: @carver @pipermerriam Is this the right way to alias `utils` -> `_utils`?

carver: I don't have a strong opinion on it. Seems okay to me.

Issue #887, opened 2018-06-05
Only try to resolve ENS names on mainnet (in the default middleware)


carver:
### What was wrong?

The ENS middleware will try to resolve names no matter what your network ID. Test chains shouldn't try to resolve these names, and give confusing errors when you accidentally supply a name-like value to an address field.

### How can it be fixed?

- [ ] Modify the ens middleware to initialize with a network ID and only attempt to resolve names if connected to that network ID.
- [ ] When validating whether a value is acceptable input for an address, only accept an ENS name if connected to the correct network ID.
- [x] Set up the default middleware to use network ID 1 (for mainnet).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


1. [ankitchiplunkar](https://gitcoin.co/profile/ankitchiplunkar) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/web3.py/887/1055?mutate_worker_action=approve&worker=ankitchiplunkar) | [reject worker](https://gitcoin.co/issue/ethereum/web3.py/887/1055?mutate_worker_action=reject&worker=ankitchiplunkar))_. 
    
    Would update the ens middleware.
Would add relevant tests and docs.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/887/1055).

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 180.0 DAI (180.0 USD @ $1.0/DAI) has been submitted by__:

 1. [@ankitchiplunkar](https://gitcoin.co/profile/ankitchiplunkar)



@vs77bb please take a look at the submitted work:
* [PR](https://github.com/ethereum/web3.py/pull/1037) by @ankitchiplunkar

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/887/1055)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/web3.py/887/1055).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
[...]

PR #1037, opened 2018-08-31
Only try to resolve ENS names on mainnet

ankitchiplunkar:
### What was wrong?

Related to Issue # https://github.com/ethereum/web3.py/issues/887

### How was it fixed?
1. Added a check in `abi_name_resolver` to check if the chain is mainnet.
2. Added another check in the same resolver to check if the ens instance is `StaticENS` to enable passing of old tests.
3. Added tests which verify the change
4. Updated the docs

#### Cute Animal Picture

### DIFF
--- a/tests/core/middleware/test_name_to_address_middleware.py
+++ b/tests/core/middleware/test_name_to_address_middleware.py
@@ -1,0 +1,56 @@
+import pytest
+
+from web3 import Web3
+from web3.exceptions import (
+    InvalidAddress,
+)
+from web3.middleware import (  # noqa: F401
+    construct_fixture_middleware,
+    name_to_address_middleware,
+)
+from web3.providers.base import (
+    BaseProvider,
+)
+
+NAME = "dump.eth"
+ADDRESS = "0x0000000000000000000000000000000000000000"
+BALANCE = 0
+
+
+class TempENS():
+    def __init__(self, name_addr_pairs):
+        self.registry = dict(name_addr_pairs)
+
+    def address(self, name, guess_tld=True):
+        # no automated web3 usages should be guessing the TLD
+        assert not guess_tld
+        return self.registry.get(name, None)
+
+
+@pytest.fixture
+def w3():
+    w3 = Web3(providers=[BaseProvider()], middlewares=[])
+    w3.ens = TempENS({NAME: ADDRESS})
+    w3.middleware_stack.add(name_to_address_middleware(w3))
+    return w3
+
+
+def test_pass_name_resolver(w3):
+    return_chain_on_mainnet = construct_fixture_middleware({
+        'net_version': '1',
+    })
+    return_balance = construct_fixture_middleware({
+        'eth_getBalance': BALANCE
+    })
+    w3.middleware_stack.inject(return_chain_on_mainnet, layer=0)
+    w3.middleware_stack.inject(return_balance, layer=0)
+    assert w3.eth.getBalance(NAME) == BALANCE
+
+
+def test_fail_name_resolver(w3):
+    return_chain_on_mainnet = construct_fixture_middleware({
+        'net_version': '2',
+    })
+    w3.middleware_stack.inject(return_chain_on_mainnet, layer=0)
+    with pytest.raises(InvalidAddress, match='.*ethereum\.eth.*'):
+        w3.eth.getBalance("ethereum.eth")
--- a/web3/utils/normalizers.py
+++ b/web3/utils/normalizers.py
@@ -31,6 +31,7 @@ from web3.utils.encoding import (
     to_text,
 )
 from web3.utils.ens import (
+    StaticENS,
     is_ens_name,
     validate_name_has_address,
 )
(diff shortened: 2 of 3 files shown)


ankitchiplunkar:
@carver @pipermerriam would love any inputs on this PR.

web3/utils/normalizers.py
```diff
@@ -137,6 +138,8 @@ def abi_ens_resolver(w3, abi_type, val):
             raise InvalidAddress("Could not look up name, because no web3 connection available")
         elif w3.ens is None:
             raise InvalidAddress("Could not look up name, because ENS is set to None")
+        elif int(w3.net.version) is not 1 and not isinstance(w3.ens, StaticENS):
+            raise InvalidAddress("Could not look up name, because web3 is not connected to mainnet")
```
carver: Would you add the name `val` to this message and the ones above it? Something like:

`"Could not look up name %r, because web3 is not connected to mainnet" % val`

tests/core/middleware/test_name_to_address_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+    w3 = Web3(providers=[HTTPProvider(url)], middlewares=[])
+    w3.ens = ENS.fromWeb3(w3)
+    w3.middleware_stack.add(name_to_address_middleware(w3))
+    return w3
+
+
+def test_pass_name_resolver_mainnet():
+    url = "https://mainnet.infura.io/"
+    w3 = w3_name_resolver(url)
+    w3.eth.getBalance("ethereum.eth")
```
carver: This looks like it's actually hitting Infura for the test, right? Tests should be able to run while disconnected from the Internet.

One option is to use `construct_fixture_middleware` to return the network id (probably injected at layer 0).

tests/core/middleware/test_name_to_address_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_fail_name_resolver_kovan():
+    url = "https://kovan.infura.io/"
+    w3 = w3_name_resolver(url)
+    with pytest.raises(InvalidAddress):
+        w3.eth.getBalance("ethereum.eth")
```
carver: same here

tests/core/middleware/test_name_to_address_middleware.py
ankitchiplunkar: Pushed the test offline but the BaseProvider was missing several methods which had to be implemented eg `eth_getBlockByNumber`, `eth_call` and `eth_getBalance`.
Tried to enable these methods using `construct_fixture_middleware` at layer 0 but the command `w3.eth.getBalance(name)` would call `eth_getBalance` --> `eth_call` --> `eth_getBlockByNumber` (which would again throw method not implemented error).

Maybe after 2-3 recusrive calls the method `construct_fixture_middleware` fails.

Have bypassed this issue by letting pytest catch the `NotImplementedError` do tell me if this is not an acceptable solution.

tests/core/middleware/test_name_to_address_middleware.py
```diff
@@ -0,0 +1,40 @@
...
+    with pytest.raises(NotImplementedError):
+        w3.eth.getBalance('ethereum.eth')
+
+
+def test_fail_name_resolver(w3):
+    return_chain_on_mainnet = construct_fixture_middleware({
+        'net_version': '2',
+    })
+    w3.middleware_stack.inject(return_chain_on_mainnet, layer=0)
+    with pytest.raises(InvalidAddress):
```
carver: This test can now confirm that the name shows up in the exception with something like:

```py
with pytest.raises(InvalidAddress, match='.*ethereum\.eth.*'):
```

tests/core/middleware/test_name_to_address_middleware.py
pipermerriam: @carver drive-by-idea.

- `fixture_recorder_middleware`

Lets you connect whatever provider you like, make some calls, and then export the request/response data in a format that could then be passed into `construct_fixture_middleware` to emulate the request/responses.

carver: Ooh, neat idea.

tests/core/middleware/test_name_to_address_middleware.py
ankitchiplunkar: Done

web3/utils/normalizers.py
```diff
@@ -134,9 +135,14 @@ def abi_address_to_hex(abi_type, data):
...
-            raise InvalidAddress("Could not look up name, because no web3 connection available")
+            raise InvalidAddress("Could not look up name %r, \
+                because no web3 connection available" % val)
         elif w3.ens is None:
-            raise InvalidAddress("Could not look up name, because ENS is set to None")
+            raise InvalidAddress("Could not look up name %r, \
+                because ENS is set to None" % val)
+        elif int(w3.net.version) is not 1 and not isinstance(w3.ens, StaticENS):
+            raise InvalidAddress("Could not look up name %r, \
+                because web3 is not connected to mainnet" % val)
```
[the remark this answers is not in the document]
ankitchiplunkar: Accepted the changes

tests/core/middleware/test_name_to_address_middleware.py
```diff
@@ -0,0 +1,40 @@
...
+    return w3
+
+
+def test_pass_name_resolver(w3):
+    return_chain_on_mainnet = construct_fixture_middleware({
+        'net_version': '1',
+    })
+    w3.middleware_stack.inject(return_chain_on_mainnet, layer=0)
+    with pytest.raises(NotImplementedError):
+        w3.eth.getBalance('ethereum.eth')
```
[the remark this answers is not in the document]
ankitchiplunkar: Have created a new ENS class in the test for resolving this middleware

PR #1038, opened 2018-09-02
Move all used datastructures from utils module

Exef:
### What was wrong?

As discussed in #1033, `web3.utils.datastructures` are used to type checks and shouldn't be treated as internal.

### How was it fixed?

I moved the datastructures.py from internal utils module to the public one. I wonder if I should be contained in a separate module instead of the top one.

#### Cute Animal Picture

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -343,7 +343,7 @@ Fetch various data about the current state of the ERC20 contract.
     >>> erc20.functions.decimals().call()
     0
     >>> erc20.functions.totalSupply().call()
-    2038
+    2039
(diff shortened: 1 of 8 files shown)


pipermerriam:
The failing test is actually due to us querying the live token contract for the Unicorn tokens.  Wondering if you can go and update that line to be 2039 for now until we come up with a better solution to still retain the example since I think it's a useful one, but to get things working for now.

pipermerriam:
I went ahead and did the update.  good to go when tests :heavy_check_mark:

Issue #1039, opened 2018-09-02
AttributeError when running contract method


amirbaer:
* Version: 4.6.0
* Python: 3.6
* OS: osx

### What was wrong?

Please include any of the following that are applicable:

* The code which produced the error
```
# a.py
from ethtoken.abi import EIP20_ABI

from web3 import Web3
from web3.auto import w3
from web3.middleware import geth_poa_middleware

NODE_HTTP_ADDRESS = "http://127.0.0.1:8545"
MTN_CONTRACT_ADDRESS = "0x6160c9d6943f7c1040bb6e247cb100c19566185d"

MTN1_PUB_KEY = "0xf9a5c72412c0200b0801f581c6008e852c963bfe"


w3 = Web3(Web3.HTTPProvider(NODE_HTTP_ADDRESS, request_kwargs={'timeout': 60}), middlewares=[geth_poa_middleware])
print("connected")

w3.eth.enable_unaudited_features()
mtn = w3.eth.contract(address=Web3.toChecksumAddress(MTN_CONTRACT_ADDRESS), abi=EIP20_ABI)
balance = int(mtn.functions.balanceOf(Web3.toChecksumAddress(MTN1_PUB_KEY)).call())
```

* The full output of the error
```
[amirbaer@beth] tmp > python3 a.py
connected
Traceback (most recent call last):
  File "a.py", line 20, in <module>
    balance = int(mtn.functions.balanceOf(Web3.toChecksumAddress(MTN1_PUB_KEY)).call())
  File "/usr/local/lib/python3.6/site-packages/web3/contract.py", line 1108, in call
    block_id = parse_block_identifier(self.web3, block_identifier)
  File "/usr/local/lib/python3.6/site-packages/web3/contract.py", line 1414, in parse_block_identifier
    last_block = web3.eth.getBlock('latest').number
AttributeError: 'dict' object has no attribute 'number'
```

* What type of node you were connecting to.
geth @ rinkeby

### How can it be fixed?

As far as I can tell, by simply replacing:
`web3.eth.getBlock('latest').number`
with:
`web3.eth.getBlock('latest')['number']`

carver:
@amirbaer I highly recommend you add the poa middleware as [described in the docs](https://web3py.readthedocs.io/en/stable/middleware.html#geth-style-proof-of-authority):

```py
# inject the poa compatibility middleware to the innermost layer
>>> w3.middleware_stack.inject(geth_poa_middleware, layer=0)
```

[...]

PR #1040, opened 2018-09-02
fixed faulty dict access syntax which causes an attribute error

amirbaer:
### What was wrong?

Related to Issue #1039 

### How was it fixed?

By fixing the syntax (as detailed in issue #1039).
I could not think of any other possible implications.

#### Cute Animal Picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1416,7 +1416,7 @@ def parse_block_identifier(web3, block_identifier):
     elif block_identifier in ['latest', 'earliest', 'pending']:
         return block_identifier
     elif isinstance(block_identifier, bytes) or is_hex_encoded_block_hash(block_identifier):
-        return web3.eth.getBlock(block_identifier).number
+        return web3.eth.getBlock(block_identifier)['number']
     else:
         raise BlockNumberOutofRange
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
doctest failure is unrelated and was fixed in #1038

web3/contract.py
```diff
@@ -1416,7 +1416,7 @@ def parse_block_identifier(web3, block_identifier):
     elif block_identifier in ['latest', 'earliest', 'pending']:
         return block_identifier
     elif isinstance(block_identifier, bytes) or is_hex_encoded_block_hash(block_identifier):
-        return web3.eth.getBlock(block_identifier).number
```
pipermerriam: cc @dylanjw @carver @voith just in case you don't see this.  This is a good example of why we should be as diligent as possible to not assume the `AttributeDict` API is usable within our own web3 apis.  Forgive me for preaching to the choir if you're already aware.

carver: FWIW, replacing *all* the middlewares as the user did in #1039 is almost definitely not what they wanted to do. There will be a lot of things that work incorrectly, like: fields won't be converted to and from python native types. I'm not suggesting that we should *encourage* attribute syntax, but this is only one of many things that will work in unexpected ways if the user drops all the middleware.

PR #1041, opened 2018-09-03
Add 'xyz' and 'luxe' to reconised TLDs

Arachnid:
web3 presently assumes all ENS names end with '.eth'. We're about to launch two new TLDs, with many more to follow.

This patch adds support for the two new TLDs, but ideally web3 should be modified to not care what TLD you enter (or to 'guess' and add extensions if you don't provide them).

### DIFF
--- a/ens/constants.py
+++ b/ens/constants.py
@@ -9,6 +9,6 @@ EMPTY_ADDR_HEX = '0x' + '00' * 20
 
 MIN_ETH_LABEL_LENGTH = 7
 
-RECOGNIZED_TLDS = ['eth', 'reverse', 'test']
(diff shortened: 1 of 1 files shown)


carver:
Cool thanks.

> ideally web3 should be modified to not care what TLD you enter  (or to 'guess' and add extensions if you don't provide them).

Right, the user experience will get a little bit worse when we make the full switchover to be completely agnostic to TLDs, but it seems inevitable.

The downside is mostly that we can't give a different error message if the thing *looks* like a name that can't be resolved, or if it doesn't look at all like an ENS address. I won't be sad to lose the "TLD guessing".

---

I rebased and pushed to trigger a CI re-run against the latest fixes.

Issue #1011, opened 2018-08-22
No way to specify block number for eth_estimateGas


Arachnid:
At present the library doesn't provide any way to specify block number when calling estimateGas (the second parameter to that function). It'd be useful if this was provided, and especially useful if it can be made accessible when automatically estimating gas via `build_transaction` or `transact` (in order to be able to specify 'pending').

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 170.0 DAI (170.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/1011/1013).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/web3.py/1011/1013).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $17,099.45 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 170.0 DAI (170.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/1011/1013).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/web3.py/1011/1013).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $17,099.45 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #1046, opened 2018-09-06
WIP: No way to specify block number for eth_estimateGas 

ankitchiplunkar:
### What was wrong?

No way to specify block number for eth_estimateGas 

Related to Issue # https://github.com/ethereum/web3.py/issues/1011

### How was it fixed?
Added block_identifier as an input parameter in the estimateGas method(similar to web3.eth.call), this opens a can of worms because:
1. The `estimate_gas` method in the eth_tester library takes only 2 inputs while we now have an extra input (block_identifier). To enable the passing of all tests we need to also change the https://github.com/ethereum/eth-tester library
 ```
E       TypeError: estimate_gas() takes 2 positional arguments but 3 were given
 ```

2. A geth node does not accept block_identifier as a parameter and throws an error, this has been documented before https://github.com/ethereum/go-ethereum/issues/2586 

**Parity**
 ```
>>> w3 = Web3(Web3.HTTPProvider("https://kovan.infura.io/"))
>>> w3.eth.estimateGas(transaction)
method: eth_estimateGas, parameter: [{'data': '0xacabb9ed000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000039746869732d69732d7468652d66697273742d737472696e672d77686963682d657863656564732d33322d62797465732d696e2d6c656e67746800000000000000000000000000000000000000000000000000000000000000000000000000003a746869732d69732d7468652d7365636f6e642d737472696e672d77686963682d657863656564732d33322d62797465732d696e2d6c656e677468000000000000', 'from': '0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf', 'to': '0xF2E246BB76DF876Cef8b38ae84130F4F55De395b'}, 'latest']
response: {'jsonrpc': '2.0', 'id': 0, 'result': 29912}
29912
```

**Geth**
```
>>> w3 = Web3(Web3.HTTPProvider("https://mainnet.infura.io/"))
>>> w3.eth.estimateGas(transaction)
method: eth_estimateGas, parameter: [{'data': '0xacabb9ed000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000039746869732d69732d7468652d66697273742d737472696e672d77686963682d657863656564732d33322d62797465732d696e2d6c656e67746800000000000000000000000000000000000000000000000000000000000000000000000000003a746869732d69732d7468652d7365636f6e642d737472696e672d77686963682d657863656564732d33322d62797465732d696e2d6c656e677468000000000000', 'from': '0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf', 'to': '0xF2E246BB76DF876Cef8b38ae84130F4F55De395b'}, 'latest']
response: {'jsonrpc': '2.0', 'id': 0, 'error': {'code': -32602, 'message': 'too many arguments, want at most 1'}}
  File "/home/ankit/projects/web3.py/web3/eth.py", line 307, in estimateGas
    [transaction, block_identifier],
  File "/home/ankit/projects/web3.py/web3/manager.py", line 114, in request_blocking
    return response['result']
ValueError: {'code': -32602, 'message': 'too many arguments, want at most 1'}
```

Changing the two libraries would be a HUGE task. 
How do you (@Arachnid , @carver) propose that we should proceed?

#### Cute Animal Picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -294,15 +294,22 @@ class Eth(Module):
             [transaction, block_identifier],
         )
 
-    def estimateGas(self, transaction):
+    def estimateGas(self, transaction, block_identifier=None):
         # TODO: move to middleware
         if 'from' not in transaction and is_checksum_address(self.defaultAccount):
             transaction = assoc(transaction, 'from', self.defaultAccount)
 
-        return self.web3.manager.request_blocking(
-            "eth_estimateGas",
-            [transaction],
-        )
+        # TODO: move to middleware
+        if block_identifier is None:
+            return self.web3.manager.request_blocking(
+                "eth_estimateGas",
+                [transaction],
+            )
+        else:
+            return self.web3.manager.request_blocking(
+                "eth_estimateGas",
+                [transaction, block_identifier],
+            )
 
     def filter(self, filter_params=None, filter_id=None):
         if filter_id and filter_params:
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -299,17 +299,15 @@ class Eth(Module):
         if 'from' not in transaction and is_checksum_address(self.defaultAccount):
             transaction = assoc(transaction, 'from', self.defaultAccount)
 
-        # TODO: move to middleware
         if block_identifier is None:
-            return self.web3.manager.request_blocking(
-                "eth_estimateGas",
-                [transaction],
-            )
+            params = [transaction]
         else:
-            return self.web3.manager.request_blocking(
-                "eth_estimateGas",
-                [transaction, block_identifier],
-            )
+            params = [transaction, block_identifier]
+
+        return self.web3.manager.request_blocking(
+            "eth_estimateGas",
+            params,
+        )
 
     def filter(self, filter_params=None, filter_id=None):
         if filter_id and filter_params:
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -632,7 +632,7 @@ The following methods are available on the ``web3.eth`` namespace.
     In most cases it is better to make contract function call through the :py:class:`web3.contract.Contract` interface.
 
 
-.. py:method:: Eth.estimateGas(transaction)
+.. py:method:: Eth.estimateGas(transaction, block_identifier=None)
 
     * Delegates to ``eth_estimateGas`` RPC Method
 
@@ -640,13 +640,19 @@ The following methods are available on the ``web3.eth`` namespace.
     on the blockchain.  Returns amount of gas consumed by execution which can
     be used as a gas estimate.
 
-    The ``transaction`` parameter is handled in the same manner as the
-    :meth:`~web3.eth.Eth.sendTransaction()` method.
+    The ``transaction`` and ``block_identifier`` parameters are handled in the
+    same manner as the :meth:`~web3.eth.call()` method.
 
     .. code-block:: python
 
         >>> web3.eth.estimateGas({'to': '0xd3cda913deb6f67967b99d67acdfa1712c293601', 'from': web3.eth.coinbase, 'value': 12345})
         21000
+
+    .. note::
+        The parameter ``block_identifier`` is not enabled in geth nodes,
+        hence passing a value of ``block_identifier`` when connected to a geth
+        nodes would result in an error like:  ``ValueError: {'code': -32602, 'message': 'too many arguments, want at most 1'}``
+
 
 .. py:method:: Eth.generateGasPrice(transaction_params=None)
 
--- a/tests/integration/go_ethereum/common.py
+++ b/tests/integration/go_ethereum/common.py
@@ -60,6 +60,14 @@ class GoEthereumEthModuleTest(EthModuleTest):
         pytest.xfail('Needs ability to efficiently control mining')
         super().test_eth_modifyTransaction(web3, unlocked_account)
 
+    def test_eth_estimateGas_with_block(self,
+                                        web3,
+                                        unlocked_account_dual_type):
+        pytest.xfail('Block identifier has not been implemented in geth')
+        super().test_eth_estimateGas_with_block(
+            web3, unlocked_account_dual_type
+        )
+
 
 class GoEthereumVersionModuleTest(VersionModuleTest):
     pass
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by carver** [comment]
Changing `geth` to accept the block identifier is surely out of scope. py-evm already supports estimating gas at a given header, so it shouldn't be too much work to add to eth-tester.

So I would just mark any `geth` tests as `xfail` and get it working otherwise working on eth-tester and parity.

**Comment by ankitchiplunkar** [comment]
One more point is that geth accepts only one input (transaction) while parity/ py-evm accept two (transaction & block_identifier)

This means we might have to either:
1. Check the node which is being used and modify the inputs accordingly, or
2. Push the method `estimateGas` in respective geth or parity modules.

Which solution do you prefer?

**Comment by carver** [comment]
> Which solution do you prefer?

I prefer that the json-rpc get standardized in an EIP and that they all conform to it :laughing: 

---

In the meantime...

Parity must work if only one argument is passed, I guess. So, if no block identifier is set in web3, instead of defaulting to latest, we could leave the argument out altogether in the RPC call. This has the weird effect that the number of arguments in the RPC call can be different, which I don't think happens anywhere else. But I'd be curious to see if we could make that work.

Then we don't have to do any node detection, and it would work in both geth and parity. Then if you try with an explicit block-identifier, geth will blow up (which we can't do anything about) and parity will work fine.

**Comment by ankitchiplunkar** [comment]
This is a wayyy better solution, because then we would not have to tag several geth tests as xfail.

Will work on implementing this and modifying the eth-tester now.

**Comment by vs77bb** [comment]
Hi @ankitchiplunkar any progress to show here? Hope you're doing well 🙂

**Comment by ankitchiplunkar** [comment]
@vs77bb yes was busy with EthBerlin last week, getting back to the task.

**Comment by ankitchiplunkar** [comment]
Have updated the estimateGas function as discussed above and added tests for parity and geth.
But, cannot implement middleware for the `block_identifier`, since we are supplying variable number of inputs to the final JSON-RPC call.

Implemented a `block_identifier` middleware by adding the following command to web3.middleware.pythonic.pythonic_middleware

```
'eth_estimateGas': combine_argument_formatters(
            transaction_param_formatter,
            block_number_formatter,
        ),
```

But got the following errors, when only one input was provided to method `estimateGas`

```
    @return_arg_type(2)
    def apply_formatter_at_index(formatter: Callable[..., Any],
                                 at_index: int,
                                 value: List[Any]) -> Generator[List[Any], None, None]:
        if at_index + 1 > len(value):
            raise IndexError(
                "Not enough values in iterable to apply formatter.  Got: {0}. "
>               "Need: {1}".format(len(value), at_index + 1)
            )
E           IndexError: Not enough values in iterable to apply formatter.  Got: 1. Need: 2
```

Have not implemented any middleware for the block_identifier as of now, mostly would need to implement a function (similar to `combine_argument_formatters`) that accepts variable number of inputs in library `eth_utils`. Do you think we should implement this middleware?

Still need to update:
1. eth_tester to take multiple inputs for estimateGas
2. Possible eth_utils to accept variable number of inputs in combine_arguement_formatters
3. Update docs for the estimateGas

**Comment by carver** [comment]
I think if we get creative, we can do this with existing formatters, something like:

```py
@curry
def is_length(target_length, value):
  return len(value) == target_length

estimate_gas_without_block_id = apply_formatter_at_index(0, transaction_param_formatter)
estimate_gas_with_block_id = apply_formatters_to_sequence((
  transaction_param_formatter,
  block_number_formatter,
))

...

'eth_estimateGas': apply_one_of_formatters((
  (is_length(1), estimate_gas_without_block_id),
  (is_length(2), estimate_gas_with_block_id),
)
```
The key part is choosing which formatter to apply with [`apply_one_of_formatters`](https://eth-utils.readthedocs.io/en/latest/utilities.html#apply-one-of-formatters-condition-formatter-pairs-value-new-value). It uses the [newer preferred `apply_formatters_to_sequence`](https://eth-utils.readthedocs.io/en/latest/utilities.html#combine-argument-formatters-formatters-lambda-list-like-new-list-like).

carver:
Can you do a rebase on master instead of merging master in? Makes it cleaner to read the PR.

dylanjw:
I just merged #1033 which will make this rebase a bit complicated. @ankitchiplunkar Let me know if you want help.

ankitchiplunkar:
Thanks all, for helping with squashing and rebasing.

@carver I found that `apply_formatters_to_sequence` was not working but `combine_argument_formatters` worked fine. 

The `parity-http` test is falling for some unknown reason.

Is this approach is acceptable and tests cover everything?

voith:
> Can you amend the commit and write a different summary (first) line of the squashed commit? 

Big 👍 for pushing towards cleaner commits and for a better commit message. With clean commit history it becomes really easy to dig history and we can reliably use `git bisect`.

web3/eth.py
```diff
@@ -294,15 +294,22 @@ def call(self, transaction, block_identifier=None):
...
         # TODO: move to middleware
         if 'from' not in transaction and is_checksum_address(self.defaultAccount):
             transaction = assoc(transaction, 'from', self.defaultAccount)
 
-        return self.web3.manager.request_blocking(
-            "eth_estimateGas",
-            [transaction],
-        )
+        # TODO: move to middleware
+        if block_identifier is None:
```
voith: This code can be reduced to:
```python
params = [transaction]
if block_identifier:
    params.append(block_identifier)
    "eth_estimateGas",
    params
)
```

carver: Rather than mutating the params list, I would go with:

```py
        if block_identifier is None:
            params = [transaction]
        else:
            params = [transaction, block_identifier]

            "eth_estimateGas",
            params,
        )
```

Also, we don't need the second `move to middleware` TODO.

carver:
> I found that apply_formatters_to_sequence was not working but combine_argument_formatters worked fine.

Ok, not a big deal. Just a nice-to-have.

> The parity-http test is falling for some unknown reason.

Yeah, surely not a problem with this PR. parity tests have been flaky.

> Is this approach is acceptable and tests cover everything?

Yup, approach seems good, and tests cover enough for merge, I think.  Just that one last inline comment to address, and should be good to go.

carver:
Thanks for the doc update!

Issue #925, opened 2018-06-22
Home Directory Path Shortcut doesn't work in Linux


fubuloubu:
* Version: 4.4.0
* Python: 3.6.5 **(P.S. this field is old!)**
* OS: linux

### What was wrong?
In linux, the home directory path shortcut `~` doesn't work

```python
w3 = Web3(Web3.IPCProvider("/home/bryant/.ethereum/testnet/geth.ipc")); w3.isConnected()
True

w3 = Web3(Web3.IPCProvider("~/.ethereum/testnet/geth.ipc")); w3.isConnected()
False
```
### How can it be fixed?
![IDK](https://secure.i.telegraph.co.uk/multimedia/archive/03246/shrug_3246314b.jpg)

pipermerriam:
I think we recently merged https://github.com/ethereum/web3.py/pull/921 which allows for `pathlib.Path` objects.  I think what we need to do now, is invert that and do the following.

```python
if isinstance(ipc_path, str):
    ipc_path = pathlib.Path(ipc_path)
elif not isinstance(ipc_path, pathlib.Path):
    raise TypeError("helpful error message about what types are accepted")

ipc_path_str = ipc_path.expanduser().resolve()
```

This should handle conversion to an absolute path as well as handling `~`.

PR #1049, opened 2018-09-10
Fix for #925, added support for tilda in provider ipc path 

pjryan93:
### What was wrong?
Tilda's are not being replaced with the users home directory on Linux and Mac.  
```
w3 = Web3(Web3.IPCProvider("/Users/pjryan/.ethereum/testnet/geth.ipc")); 
w3.isConnected()  #returns True

w3 = Web3(Web3.IPCProvider("~/.ethereum/testnet/geth.ipc")); 
w3.isConnected() #returns False
```

Related to Issue #
Fix for #925.

### How was it fixed?
 Call `expanduser()` and `resolve()` to make sure the Tilda is replaced with the user's home directory.  

#### Cute Animal Picture

happy cat!


carver:
Actually, since this is only breaking in python 3.5, this PR can just wait till 3.5 is dropped in web3.py v5 and merge then. I'm doing a backport to web3 v4 currently.

carver:
You're going to rebase right @kclowes ? :)

kclowes:
Thanks @pjryan93!

PR #1063, opened 2018-09-18
Test against parity v2

dylanjw:
### What was wrong?

Parity 2.0.5 stable has been released.

#### Cute Animal Picture

image


dylanjw:
Appears that the parity test fixtures need to be updated for v2.  The past strategy of setting up the fixtures with go-ethereum and exporting, then importing the chain data into parity is hitting an error:

```
2018-09-18 17:40:30  Stage 5 block verification failed for #1 (0xf1b0…1efb)
Error: Error(Block(InvalidStateRoot(Mismatch { expected: 0x9993ce1966f02d1401e5472a3814a11fe55d53712550ca6c947819c3759b3c33, found: 0xbae0c4f770633c7657d6c1dc387482007619d62fd6751f0a6cc0afa858ee6301 })), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })
```

Next task is changing the parity fixture generation script to set up the chain state in parity directly with a miner.

carver:
> The past strategy of setting up the fixtures with go-ethereum and exporting, then importing the chain data into parity is hitting an error

Hm, that's unexpected. The blocks should all be valid.

dylanjw:
> Hm, that's unexpected. The blocks should all be valid.

Figuring out a private chain config that is cross compatible with newer versions of go-ethereum and parity v2 has been challenging. For previous versions of parity I used a modified version of this chain spec: https://github.com/5chdn/crossclient-chainspec.  This is no longer working, and reading through parity and go-ethereum to figure out the differences in default EIP transitions has been tedious and unfruitful.  The next thing I will try is setting up a private chain in parity along with ethminer.

kclowes:
Closing since this has been taken care of in #1292!

PR #1067, opened 2018-09-20
Fix raw transaction signing middleware with byte addresses

koirikivi:
### What was wrong?

Using `web3.middleware.construct_sign_and_send_raw_middleware` with `bytes` typed addresses  caused improper behaviour in at least the following cases:

1. Passing a `bytes` typed address as the `from` address, in which case the middleware got never executed because the check `transaction.get('from') not in accounts` always evaluated as `True`
2. Passing a `bytes` typed address as the `to` address (for an instance, when doing `web3.eth.contract(address=bytes(...), ...)`), which resulted in an exception.

### How was it fixed?

The transaction is now formatted according to the transaction abi types in `sign_and_send_raw_middleware` (similar to what's done with `abi_middleware`).

I'm not totally sure about the fix, since I don't know web3.py internals that well. I discussed this issue with @voith and he encouraged me to submit a fix though. Any feedback is appreciated :) .

#### Cute Animal Picture

### DIFF
--- a/tests/core/middleware/test_transaction_signing.py
+++ b/tests/core/middleware/test_transaction_signing.py
@@ -95,6 +95,10 @@ def w3_dummy(w3_base, result_generator_middleware):
     return w3_base
 
 
+def hex_to_bytes(s):
+    return to_bytes(hexstr=s)
+
+
 @pytest.mark.parametrize(
     'method,key_object,from_,expected',
     (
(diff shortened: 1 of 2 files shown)


koirikivi:
Thanks for the quick merge!

Is it possible to also get this to V4. This is very much something I'd like to use in our soon-to-hit production code, without forking :) .

carver:
Yeah, by the looks of it, it would be a single bugfix release: 4.7.2 -- hopefully sometime early next week.

PR #1070, opened 2018-09-21
Raise external timeout

carver:
### What was wrong?

We raise `Timeout` exception from `w3.eth.waitForTransactionReceipt()`, but the exception is in a private `_utils` module.

### How was it fixed?

- Add new `TimeExhausted` timeout.
- Update a few usages of `waitForTransactionReceipt`
- Explicitly test the timeout exception raised

#### Cute Animal Picture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -386,7 +386,8 @@ The following methods are available on the ``web3.eth`` namespace.
     returns its transaction receipt.
 
     Optionally, specify a ``timeout`` in seconds. If timeout elapses before the transaction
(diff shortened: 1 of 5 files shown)


carver:
This shouldn't be backported to v4, because the change in exception is not backwards compatible. Linking to #722 for future documentation needs.

pipermerriam:
@carver could we backport to `v4` if we adjusted the utils timeout error to inherit from the built-in?

carver:
The only problem being that we would have a utils exception import from a non-utils package.

pipermerriam:
Ignore me, I didn't actually look at the pull request code.  Doesn't seem worth the complexity.

PR #1073, opened 2018-09-25
PM_API - Integration tests for all standalone use cases

njgheorghita:
### What was wrong?
Expanded integration tests to include all "standalone" packages (i.e. no `build_dependencies`).

### How was it fixed?
Added manifests and wrote tests

#### Cute Animal Picture
image


tests/core/pm-module/test_pm_init.py
```diff
@@ -1,20 +1,21 @@
...
-from solc import compile_source
+from eth_utils import (
+    to_canonical_address,
+)
 
 from web3 import Web3
 
 try:
+    from ethpm import Package  # noqa: E402
+    from ethpm.assets import get_manifest
```
pipermerriam: Seeing this import path has me thinking the `get_manifest` should potentially live under `ethpm.tools` somewhere.  It's a relatively new convention I've been using but the idea is:

- `library.tools` contains *useful* things that are not necessarily part of the core library
- At no place within `library` does it import from `library.tools` (i.e. `library.tools` could in theory be deleted from the codebase and the library code would remain functional)

Previously, these things normally lived as utility modules under `tests/` ' or in  `library.utils`, both of which felt wrong.

tests/core/pm-module/test_pm_init.py
```diff
@@ -37,32 +38,73 @@ def web3():
...
-
-
-@pytest.fixture
-def standard_token_manifest():
-    with open(str(V2_PACKAGES_DIR / 'standard-token' / '1.0.0.json')) as file_obj:
-        return json.load(file_obj)
+def get_ethpm_manifest():
+    def _get_ethpm_manifest(name, filename):
+        return get_manifest(name, filename)
+    return _get_ethpm_manifest
```
pipermerriam: You shouldn't need the inner function and should be able to just return the `get_manifest` function.

njgheorghita: This is the pattern i've been using when I want to pass arguments into a fixture. I tried 
```python
def get_ethpm_manifest(name, filename):
    return get_manifest(name, filename)
```
But that doesn't fly since it expects `name` and `filename` to be available as fixtures. Is that what you were suggesting? Or something different?

pipermerriam: I'm suggesting you write it as:

```python
@pytest.fixture
def get_ethpm_manifest():
    return get_manifest
```

Though this is really just exposing a function that *can* be imported as a fixture instead.  I don't necessarily think this is a good pattern as you could instead, in your testing module just write:

```python
from ethpm.assets import get_manifest as get_ethpm_manifest

...
```

Doesn't seem like the fixture adds a lot of value....

PR #1076, opened 2018-09-25
Quiet deprecation warnings in ens tests

dylanjw:
### What was wrong?

Web3.sha3 was deprecated, but references to it remain in `ens.utils`.

#### Cute Animal Picture

image


carver:
No strong feelings, except that it should go through the deprecation cycle.

PR #1079, opened 2018-09-28
add encodeABI to the docs

AgeManning:
### What was wrong?

The encodeABI function wasn't in the docs. 

### How was it fixed?

Added it to the docs.

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -341,6 +341,17 @@ Each Contract Factory exposes the following methods.
     Returns the transaction hash for the deploy transaction.
 
 
+.. py:classmethod:: Contract.encodeABI(fn_name, args=None, kwargs=None, data=None)
+
+   Encodes the arguments using the Ethereum ABI for the contract function that
+   matches the given `fn_name` and arguements `args`. The `data` parameter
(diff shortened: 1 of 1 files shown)


docs/contracts.rst
```diff
@@ -341,6 +341,17 @@ Each Contract Factory exposes the following methods.
...
 
+.. py:classmethod:: Contract.encodeABI(fn_name, args=None, kwargs=None, data=None)
+
+   Encodes the arguments using the Ethereum ABI for the contract function that
+   matches the given `fn_name` and arguements `args`. The `data` parameter
+   defaults to the function selector
+
+   .. code-block:: python
+
+   >>> Contract.encodeABI(fn_name="register", args=["string",10])
```
pipermerriam: can you use a string like `"rainbows"` to remove any ambiguity/confusion in case someone thinks that's an ABI type.  Also, space after the `,`.

AgeManning:
Yep, corrected the changes. I was on a computer that didn't have proper python format checking :) 

I've made it a single commit, to make it cleaner.

carver:
Thanks!

PR #1083, opened 2018-10-04
Adding examples for specific operating systems

tpmccallum:
This has been tested on fresh instance of Ubuntu 16.04

### What was wrong?
There were a few errors when installing web3 py on Ubuntu

### How was it fixed?
I just installed the dependencies and re-ran the installation a bunch of times until the installation worked from start to finish. The steps have all been placed in a linear fashion (so that the steps can be run as a bash script)


carver:
Thanks for improving the docs!

Issue #1055, opened 2018-09-12
Async versions of web3 apis


pipermerriam:
### What is it?

Implement `w3.async.<something>` which is an async mirror of APIs available on `w3.<something>`.

This will be exploratory work to figure out how we introduce an `async/await` friendly API into web3.

carver:
Note previous conversations in #657 and #856 

I wanted to think through alternatives for co-locating the logic for sync and async inputs. I was wondering if we could still do something functional. This isn't a working example, just a rough sketch:

```py
class Async_Eth(Module):                                                                                                       
    getUncleByBlock = async_caller(uncle_block_input_formatter, identity)            
    ...                                  
                                                                                                                       
class Eth(Module):                                                                                                             
    getUncleByBlock = blocking_caller(uncle_block_input_formatter, identity)                                                                                                   
    ...

#################                                                                                        
                                                                                                                       
@curry                                                                                  
async def async_caller(input_formatter, output_formatter, module, *args, **kwargs):                                    
    method, params = input_formatter(module, *args, **kwargs)                                                          
    result = await module.web3.manager.request_async(method, params)                                                   
    return output_formatter(result)                                                                                    
                                                                                                                                                                                    
                                                                                                                       
@curry                                                                                                                 
def blocking_caller(input_formatter, output_formatter, module, *args, **kwargs):                                 
    method, params = input_formatter(module, *args, **kwargs)                                                          
    result = module.web3.manager.request_blocking(method, params)                                                      
    return output_formatter(result)        

#################    

def uncle_block_input_formatter(eth, block_identifier, uncle_index):                                                   
    method = select_method_for_block_identifier(                                                                       
        block_identifier,                                                                                              
        if_predefined='eth_getUncleByBlockNumberAndIndex',                                                             
        if_hash='eth_getUncleByBlockHashAndIndex',                                                                     
        if_number='eth_getUncleByBlockNumberAndIndex',                                                                 
    )                                                                                                                  
    return (method, [block_identifier, uncle_index])                             
```

So it might be possible to do something like that without introducing a lot of new classes or lines of code.

pipermerriam:
Just realized we still need to figure out how to do middlewares....

dylanjw:
@carver In your example, could we co-locate the modules if we allow the caller function to be configured? Is there anything else that would need to vary between the async and blocking modules?

I have been wondering about the manager request function, the middlewares and the provider request functions.

[...]

PR #1088, opened 2018-10-06
Asynchronous Web3

dylanjw:
### What was wrong?

Related to Issue #1055 

### How was it fixed?
Expanded on @carver 's pseudocode.  I wrestled for a while with how to give the caller function access to the manager request methods.  I haven't found a way around using a descriptor.

#### Cute Animal Picture

image


web3/main.py
```diff
@@ -128,7 +128,7 @@ class Web3:
     toChecksumAddress = staticmethod(to_checksum_address)
 
     def __init__(self, providers=empty, middlewares=None, modules=None, ens=empty):
-        self.manager = RequestManager(self, providers, middlewares)
+        self.manager = self.RequestManager(self, providers, middlewares)
```
carver: Ah, interesting long-standing bug because of the name reuse.

tests/core/async_module/test_async_module.py
```diff
@@ -0,0 +1,66 @@
...
+        return blocking_dummy_request_function(method, params)
+
+
+@pytest.fixture()
+def w3():
+    DummyWeb3 = type(
+        'TestWeb3',
+        (Web3,),
+        {'RequestManager': DummyRequestManager})
+    w3 = DummyWeb3()
```
carver: Could also try `DummyWeb3(modules=[BlockingModule, AsyncModule])` here instead of attaching below, if you prefer. I haven't used it, but looks like it should theoretically work.

tests/core/async_module/test_async_module.py
```diff
@@ -0,0 +1,66 @@
...
+)
+from web3._utils.toolz import (
+    curry,
+    identity,
+)
+from web3.module import Module
+
+
+async def async_dummy_request_function(method, params):
+    await asyncio.sleep(0.1)
```
pipermerriam: `0.1` is a long time during tests.  I'd just make this `asyncio.sleep(0)` which I believe will still achieve your goal.

tests/core/async_module/test_demo_middleware_conversion.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+    return 'eth_sendTransaction', [transaction]
+
+
+class Eth(Module):
+    defaultAccount = ''
+    accounts = PropertyBinder(lambda _: ("eth_accounts", []), identity)
+    sendTransaction = MethodBinder(
+        compose(pythonic_input_formatter, send_transaction_input_formatter),
+        lookup_pythonic_output_formatter('eth_sendTransaction'))
```
dylanjw: There are 3 things happening here and maybe this would be cleaner if they were separate arguments.

1. formatting the module method parameters and selecting the rpc method
2. formatting the rpc method params
3. formatting the results.

Would a middleware-esq structure for 2 & 3 be better here?  Or is it nice to have all of the formatting visible up front in the method.  Im slightly in favor of having the formatters up front like this.  It could be made even more explicit without the lookup functions:

```python
sendTransaction = MethodBinder(
   method_formatter=send_transaction_formatter,
   parameter_formatter=identity,
   results_formatter=to_hexbytes(32))

Comment (carver) [reply tests/core/async_module/test_demo_middleware_conversion.py]: First impression: when selecting the method, it may be more natural to depend on the pre-formatted parameters, or the post-formatted parameters. This approach locks is into one or the other. So I'm slightly against, but am very convincable, if I see some examples that make code reuse cleaner.

**Comment by dylanjw** [comment]
Im currently working through an approach where the RequestManager (and all providers and middlewares) are natively asynchronous; without providing synchronous variants.  Synchronous functionality is provided by running the request function to completion in a loop.

**Comment by dylanjw** [comment]
#### Providers

- [ ] HTTPProvider
- [ ] EthTester
- [ ] IPCProvider
- [ ] WebSocket

#### Middlewares

- [ ] construct_simple_cache_middleware
- [ ] construct_time_based_cache_middleware
- [ ] construct_latest_block_based_cache_middleware
- [ ] _simple_cache_middleware as simple_cache_middleware
- [ ] _time_based_cache_middleware as time_based_cache_middleware
- [ ] _latest_block_based_cache_middleware as latest_block_based_cache_middleware
- [ ] construct_exception_handler_middleware
- [ ] local_filter_middleware
- [ ] construct_fixture_middleware
- [ ] construct_result_generator_middleware
- [ ] construct_error_generator_middleware
- [ ] construct_formatting_middleware
- [ ] gas_price_strategy_middleware
- [ ] name_to_address_middleware
- [ ] normalize_errors_middleware
- [ ] request_parameter_normalizer
- [ ] pythonic_middleware
- [ ] make_stalecheck_middleware
- [ ] http_retry_request_middleware
- [ ] geth_poa_middleware
- [ ] validation_middleware
- [ ] construct_sign_and_send_raw_middleware

File: web3/_utils/async_tools.py [hunk 5]
```diff
@@ -0,0 +1,37 @@
...
+def sync(co: Any):
+    raise TypeError('Called with unsupported argument: {}'.format(co))
+
+
+@sync.register(asyncio.Future)
+@sync.register(types.GeneratorType)
+def sync_co(co: Generator[Any, None, Any]) -> Any:
+    if not _is_awaitable(co):
+        raise TypeError('Called with unsupported argument: {}'.format(co))
+    return asyncio.get_event_loop().run_until_complete(co)
```
pipermerriam: It might be good to create a new event loop each time as I *think* that if the default event loop is already running this will fail (but I'm not sure about that).

web3/_utils/async_tools.py
```diff
@@ -0,0 +1,37 @@
+from functools import singledispatch, wraps
+import asyncio
+import inspect
+import types
+from typing import Any, Callable, Generator
+
+
+def _is_awaitable(co: Generator[Any, None, Any]) -> bool:
```
pipermerriam: Since this is a direct passthrough I think you can drop the utility function and just directly use `inspect.isawaitable`

web3/_utils/async_tools.py
```diff
@@ -0,0 +1,37 @@
...
+    if not asyncio.iscoroutinefunction(f):
+        raise TypeError('Called with unsupported argument: {}'.format(f))
+
+    @wraps(f)
+    def run(*args, **kwargs):
+        return asyncio.get_event_loop().run_until_complete(f(*args, **kwargs))
+    return run
+
+
+sync.register(types.CoroutineType)(sync_co)
```
pipermerriam: Is there a reason why this isn't part of the `@sync.register` decorators for the `sync_co` function above?

web3/_utils/method_callers.py
```diff
@@ -0,0 +1,40 @@
...
+            self.output_formatter)
+
+
+class PropertyBinder:
+    def __init__(self, input_formatter, output_formatter):
+        self.input_formatter = input_formatter
+        self.output_formatter = output_formatter
+
+    def __get__(self, instance, owner):
+        return instance.method_caller(
```
pipermerriam: I think `instance` can be `None` which will result in an error if this is accessed on a class.  Probably not a big problem but it will result in an ugly stack track.  Maybe we can add a check and raise a nicer error.

web3/_utils/async_tools.py
carver: Sounds like a test we should run. Something like:

```py
async def test_async_calling_sync(event_loop, w3_sync):
  asyncio.set_event_loop(event_loop)  # I think this would set the test loop to the default event loop (which would catch the error piper mentioned), but it's worth dropping into pdb to be sure.
  actual_accounts = w3_sync.eth.accounts
  assert actual_accounts == expected_accounts
```

dylanjw: It does fail.  sync() raises an error when Ive mistakenly had it called it in async tests.

dylanjw:
Ive removed the work on async modules (the caller functions and descriptor), and will be moving that to a different PR.

setup.py
```diff
@@ -79,6 +80,7 @@
         "requests>=2.16.0,<3.0.0",
```
pipermerriam: Is this dependency still needed?

web3/_utils/async_tools.py
```diff
@@ -0,0 +1,84 @@
...
+@singledispatch
+def sync(co: Any):
+    raise TypeError('Called with unsupported argument: {}'.format(co))
+
+
+@sync.register(asyncio.Future)
+@sync.register(types.GeneratorType)
+@sync.register(types.CoroutineType)
+def sync_co(co: Generator[Any, None, Any]) -> Any:
+    if not inspect.isawaitable(co):
```
pipermerriam: It'd be good to have some docstrings in this module explaining the "what" and "why" for each of these different pieces of functionality because it isn't intuitive why we need these different pieces of logic.

web3/providers/websocket.py
```diff
@@ -18,57 +16,43 @@
...
         self.websocket_kwargs = websocket_kwargs
 
     async def __aenter__(self):
-        if self.ws is None:
-            self.ws = await websockets.connect(
-                uri=self.endpoint_uri, loop=self.loop, **self.websocket_kwargs
+        thread_id = threading.get_ident()
+        if self.ws.get(thread_id) is None:
+            loop = asyncio.get_event_loop()
+            self.ws[thread_id] = await websockets.connect(
```
pipermerriam: You might consider using a [`threading.local`](https://docs.python.org/3.6/library/threading.html#thread-local-data) for this as I think it is the idiomatic way of doing per-thread data, handling the `thread_id` stuff for you under the hood.

web3/providers/websocket.py
```diff
@@ -80,25 +63,19 @@ def __init__(self, endpoint_uri=None, websocket_kwargs=None):
                     '{0} are not allowed in websocket_kwargs, '
                     'found: {1}'.format(RESTRICTED_WEBSOCKET_KWARGS, found_restricted_keys)
                 )
-        self.conn = PersistentWebSocket(
-            self.endpoint_uri, WebsocketProvider._loop, websocket_kwargs
+        WebsocketProvider.conn = WebSocketConnection(
```
voith: Any reason why `conn` was made a class variable? Every `WebsocketProvider` should have its own connection object, so `self` seems to make more sense to me.

web3/providers/websocket.py
```diff
@@ -18,57 +16,42 @@
...
+class WebSocketConnection:
+    local = threading.local()
 
-    def __init__(self, endpoint_uri, loop, websocket_kwargs):
-        self.ws = None
+    def __init__(self, endpoint_uri, websocket_kwargs):
         self.endpoint_uri = endpoint_uri
-        self.loop = loop
         self.websocket_kwargs = websocket_kwargs
+        self.local.ws = None
```
voith: Nice improvement here!
Curious to know if this was giving an error without `threading.local`

web3/providers/ipc.py
```diff
@@ -198,7 +199,8 @@ def __init__(self, ipc_path=None, testnet=False, timeout=10, *args, **kwargs):
         self._socket = PersistantSocket(self.ipc_path)
         super().__init__(*args, **kwargs)
 
-    def make_request(self, method, params):
+    async def make_request(self, method, params):
+        await asyncio.sleep(0)
```
voith: Perhaps a comment explaining why this sleep is needed. 
Wondering if this code is blocking? Does IPCProvider return control to mainloop when it is busy on a IO intensive task? Asking because you got by with just adding a sleep.

web3/_utils/async_tools.py
```diff
@@ -0,0 +1,95 @@
...
+@run_in_loop_in_thread.register(types.CoroutineType)
+@threadpool
+def run_co_in_new_loop_in_thread(co: Generator[Any, None, Any]) -> Any:
+    try:
+        #  in get_event_loog, self._local._set_called == True, even
+        #  though there is no event_loop.
+        loop = asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+        #  I get an error regarding what seems like a stale loop.
```
voith: I noticed this error when another thread is running something in the background. eg `loop.run_forever`. When another thread tries to run something using the same event loop it gives a `RuntimeError`.

web3/providers/ipc.py
dylanjw: This is a placeholder.  I will be working on making the ipc socket IO non-blocking in this PR next.

voith:
@dylanjw Looking at the test failures, I'm not very sure about using a new event loop every time you want to run something async. I'll just share my experience with twisted here. In twisted or in frameworks that are built using twisted(eg. scrapy), there's just one event loop which drives the main driver class, everything else is chained through using [Defferred's](https://twistedmatrix.com/documents/16.2.0/api/twisted.internet.defer.Deferred.html). 

Also I'm not sure about the consequences of using threadpool here. Can you explain why a ThreadPool is needed here? What are we trying to parallelize here?

dylanjw:
> I'm not very sure about using a new event loop every time you want to run something async.

I took out the creation of a new loop on every call to sync().

> Also I'm not sure about the consequences of using threadpool here. Can you explain why a ThreadPool is needed here? What are we trying to parallelize here?

This is my solution to allow multiple layers of `sync()` calls to be made.  Before I added the threadpool, calling sync() inside sync() raised a RuntimeError, because the loop is taken up by the current execution.

It seems sort of silly to have to run the child sync() loop in a ThreadPoolExecutor.  The outer thread loops get blocked waiting for child threads; and nothing is really happening in parallel.  But I dont know how else to get around the 1 loop per thread limitation.

voith:
@dylanjw Thanks for the explanation!
> Before I added the threadpool, calling sync() inside sync()  raised a RuntimeError

Can you please point me where exactly you used sync inside sync. Wondering why that would even be needed? 
I'm not well versed with asyncio, so I'm trying to understand the PR by asking questions.

dylanjw:
In this PR I have converted the request pipeline to coroutines.  To provide a blocking request function I added sync(), which is a utility function that "converts" asynchronous code to synchronous by running the coroutine to completion.  So `request_blocking()` is just `sync(request_async())`.

```python
115     def request_blocking(self, method, params):
116         """
117         Make a synchronous request using the provider
118         """
119         return sync(self.request_async(method, params))
```

Anywhere in the code that another rpc method call is made (almost entirely inside the middlewares) another sync() wrapped request will be made.  When this occurs, the next level of sync() wrapped code is run to completion in a new loop in a new thread.

This is not the most efficient strategy, but ended up being workable, with not a lot of refactoring.  The websocket connections took some figuring out.  In certain cases, I was getting an error in "recursive" calls to sync() where `get_event_loop` would fail with a RuntimeError because there was no loop, but a loop had previously been set in that thread, [ref](https://github.com/python/cpython/blob/e3d1455fe472589c92b9e57b7d35addbb579edba/Lib/asyncio/events.py#L630).  So in this case, I create and set a new loop in the thread.

Creating a new loop would cause an issue with the persistent websocket connection.  The websocket is associated with a particular loop, so I needed to add a check for when either there is no websocket connection for a particular thread, or the loop last used in the websocket connection is no longer the running loop.

web3/providers/websocket.py

```python
 38     async def __aenter__(self):
 39         loop = asyncio.get_event_loop()
 40         if self.local.ws is None or self.local.loop != loop:
 41             self.local.loop = loop
 42             self.local.ws = await websockets.connect(
 43                 uri=self.endpoint_uri, loop=loop, **self.websocket_kwargs
 44             )
 45         return self.local.ws
```
I havent figured out why the loop sometimes disappears from the thread.

Im still getting a feel for asyncio, and there is a lot experimenting happening in this PR.  Im going to work on an async ipc provider next. I am going to revisit the `sync()` wrapping strategy again.  Im not satisfied with it.  There must be a better way. Im looking at replacing the sync() internals with [Janus](https://github.com/aio-libs/janus/).

dylanjw:
> If you look at the old code, the loop inside WebsocketProvider is a singleton.

If all the threading can be removed from sync(), than i can remove the complexity I added to WebsocketProvider, to support multiple connections on separate threads/loops.  With multithreaded sync(), the websocket provider needs to handle making a connection for each thread.  Or else I would need to add multiplexing for all the threads sharing one websocket connection.

dylanjw:
This PR was mostly experimentation and can be closed now.

PR #1098, opened 2018-10-10
[WIP] pm-api: Write Registry class

njgheorghita:
### What was wrong?
`web3.pm` needs a way to interact with an on-chain registry (i.e. release, fetch, transfer_owner)

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/pm-module/test_ens_integration.py
+++ b/tests/core/pm-module/test_ens_integration.py
@@ -1,0 +1,167 @@
+import pytest
+
+from eth_utils import (
+    to_canonical_address,
+)
+
+from ens import ENS
+from web3 import Web3
+from web3.exceptions import (
+    InvalidAddress,
+)
+
+try:
+    from ethpm import (
+        ASSETS_DIR,
+    )
+    from web3.pm import (
+        PM,
+    )
+except ImportError:
+    ethpm_installed = False
+else:
+    ethpm_installed = True
+
+
+def bytes32(val):
+    if isinstance(val, int):
+        result = Web3.toBytes(val)
+    else:
+        raise TypeError('val %r could not be converted to bytes')
+    if len(result) < 32:
+        return result.rjust(32, b'\0')
+    else:
+        return result
+
+
+@pytest.fixture
+def ens_setup(solc_deployer):
+    # todo: move to module level once ethpm alpha stable
+    ENS_MANIFEST = ASSETS_DIR / 'ens' / '1.0.1.json'
+    ens_deployer = solc_deployer(ENS_MANIFEST)
+    w3 = ens_deployer.package.w3
+
+    # ** Set up ENS contracts **
+
+    # remove account that creates ENS, so test transactions don't have write access
+    accounts = w3.eth.accounts
+    ens_key = accounts.pop()
+
+    # create ENS contract
+    eth_labelhash = w3.keccak(text='eth')
+    eth_namehash = bytes32(0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae)
+    resolver_namehash = bytes32(0xfdd5d5de6dd63db72bbc2d487944ba13bf775b50a80805fe6fcaba9b0fba88f5)
+    ens_package = ens_deployer.deploy("ENSRegistry", transaction={"from": ens_key})
+    ens_contract = ens_package.deployments.get_instance("ENSRegistry")
+
+    # create public resolver
+    public_resolver_package = ens_deployer.deploy(
+        "PublicResolver",
+        ens_contract.address,
+        transaction={"from": ens_key}
+    )
+    public_resolver = public_resolver_package.deployments.get_instance("PublicResolver")
+
+    # set 'resolver.eth' to resolve to public resolver
+    ens_contract.functions.setSubnodeOwner(
+        b'\0' * 32,
+        eth_labelhash,
+        ens_key
+    ).transact({'from': ens_key})
+
+    ens_contract.functions.setSubnodeOwner(
+        eth_namehash,
+        w3.keccak(text='resolver'),
+        ens_key
+    ).transact({'from': ens_key})
+
+    ens_contract.functions.setResolver(
+        resolver_namehash,
+        public_resolver.address
+    ).transact({'from': ens_key})
+
+    public_resolver.functions.setAddr(
+        resolver_namehash,
+        public_resolver.address
+    ).transact({'from': ens_key})
+
+    # create .eth auction registrar
+    eth_registrar_package = ens_deployer.deploy(
+        "FIFSRegistrar",
+        ens_contract.address,
+        eth_namehash,
+        transaction={"from": ens_key}
+    )
+    eth_registrar = eth_registrar_package.deployments.get_instance("FIFSRegistrar")
+
+    # set '.eth' to resolve to the registrar
+    ens_contract.functions.setResolver(
+        eth_namehash,
+        public_resolver.address
+    ).transact({'from': ens_key})
+
+    public_resolver.functions.setAddr(
+        eth_namehash,
+        eth_registrar.address
+    ).transact({'from': ens_key})
+
+    # set owner of tester.eth to an account controlled by tests
+    ens_contract.functions.setSubnodeOwner(
+        eth_namehash,
+        w3.keccak(text='tester'),
+        w3.eth.accounts[2]  # note that this does not have to be the default, only in the list
+    ).transact({'from': ens_key})
+
+    # make the registrar the owner of the 'eth' name
+    ens_contract.functions.setSubnodeOwner(
+        b'\0' * 32,
+        eth_labelhash,
+        eth_registrar.address
+    ).transact({'from': ens_key})
+    return ENS.fromWeb3(w3, ens_contract.address)
+
+
+@pytest.fixture
+def ens(ens_setup, mocker):
+    mocker.patch('web3.middleware.stalecheck._isfresh', return_value=True)
+    ens_setup.web3.eth.defaultAccount = ens_setup.web3.eth.coinbase
+    return ens_setup
+
+
+@pytest.mark.skipif(ethpm_installed is False, reason="ethpm is not installed")
+def test_ens_must_be_set_before_ens_methods_can_be_used(ens):
+    w3 = ens.web3
+    PM.attach(w3, 'pm')
+    with pytest.raises(InvalidAddress):
+        w3.pm.set_registry("tester.eth")
+
+
+@pytest.mark.skipif(ethpm_installed is False, reason="ethpm is not installed")
+def test_web3_ens(ens):
+    w3 = ens.web3
+    PM.attach(w3, 'pm')
+    ns = ENS.fromWeb3(w3, ens.ens.address)
+    w3.ens = ns
+    w3.pm.deploy_and_set_registry()
+    assert w3.pm.registry
+    w3.ens.setup_address('tester.eth', w3.pm.registry.address)
+    actual_addr = ens.address('tester.eth')
+    assert w3.pm.registry.address == to_canonical_address(actual_addr)
+    w3.pm.release_package(b'pkg', b'v1', b'1.com')
+    pkg_name, version, manifest_uri = w3.pm.get_release_data(b'pkg', b'v1')
+    # trim trailing bytes returned from vyper bytes32 type
+    assert pkg_name.rstrip(b'\x00') == b'pkg'
+    assert version.rstrip(b'\x00') == b'v1'
+    assert manifest_uri.rstrip(b'\x00') == b'1.com'
+
+
+@pytest.mark.skipif(ethpm_installed is False, reason="ethpm is not installed")
+def test_registry_init_with_ens_name(ens):
+    w3 = ens.web3
+    PM.attach(w3, 'pm')
+    ns = ENS.fromWeb3(w3, ens.ens.address)
+    w3.ens = ns
+    w3.pm.deploy_and_set_registry()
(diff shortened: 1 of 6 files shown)


web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+
     def get_package_from_manifest(self, manifest):
         pkg = Package(manifest, self.web3)
         return pkg
 
     def get_package_from_uri(self, uri):
         pkg = Package.from_uri(uri, self.web3)
         return pkg
+
+    def release_package(self, name, version, uri):
```
pipermerriam: might be more intuitive if we used `manifest_uri` in these methods.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
 # It is not automatically available on a web3 object.
 # To use the `PM` module, attach it to your web3 object
 # i.e. PM.attach(web3, 'pm')
+
+
 class PM(Module):
+
+    def set_registry(self, address=None):
+        if not address:
+            self.registry = Registry.deploy_new_instance(self.web3)
```
pipermerriam: This is not intuitive behavior imho.  Maybe this should be two different APIs, one that sets it, `set_registry` and one that creates a new one `deploy_and_set_registry`.  Good user guides/examples and documentation are probably the most important part for this since we will have to teach people about the workflows for managing their registry.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+        if not self.registry:
+            raise AttributeError("Registry has not been set for web3.pm")
+        release_uri = self.registry.get_release_data(name, version)
+        return self.get_package_from_uri(release_uri)
+
+
+class Registry(Contract):
+    def __init__(self, address, w3):
+        # only works with v.py registry in ethpm/assets
+        # abi verification for compatibility?
```
pipermerriam: How about checking the `runtime_bytecode`.  Should be a known value right?

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+        registry_package = Package(manifest, w3)
+        registry_deployer = Deployer(registry_package)
+        deployed_registry_package = registry_deployer.deploy("registry")
+        deployed_registry_address = deployed_registry_package.deployments.get_contract_instance("registry").address
+        return cls(deployed_registry_address, deployed_registry_package.w3)
+
+
+    def release(self, name, version, uri):
+        tx_receipt = self.registry.functions.release(name, version, uri).transact()
+        self.w3.eth.waitForTransactionReceipt(tx_receipt)
```
pipermerriam: probably good to return either the receipt or the transaction hash.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+    def owner(self):
+        return self.registry.functions.owner().call()
+
+    @to_tuple
+    def get_all_packages(self):
+        package_count = self.registry.functions.package_count().call()
+        for index in range(package_count):
+            package_id = self.registry.functions.get_package_id(index).call()
+            package_data = self.registry.functions.get_package_data_by_id(package_id).call()
+            yield (package_data[0].rstrip(b'\x00'), package_data[1])
```
pipermerriam: I think it might be nice for future maintainers if you decomposed `package_data` into it's two parts.  I am pretty sure `package_data[0]` is the package name.  It isn't clear to me what `package_data[1]` is.

Also, commenting on the reason for the `rstrip` call would be good.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+        
+
+    @to_tuple
+    def get_all_package_versions(self, name): # -> (version, uri):
```
pipermerriam: Instead of the commented out type strings, consider adding a docstring which explains what this returns.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+        release_count = package_data[1]
+        for index in range(release_count):
+            release_id = self.registry.functions.get_release_id_by_package_and_count(name, (index + 1)).call()
+            release_data = self.registry.functions.get_release_data_by_id(release_id).call()
+            yield (release_data[1].rstrip(b'\x00'), release_data[2].rstrip(b'\x00'))
+
+
+    def transfer_owner(self, new_address):
+        tx_receipt = self.registry.functions.transfer_owner(new_address).transact()
+        self.w3.eth.waitForTransactionReceipt(tx_receipt)
```
pipermerriam: Should probably return the txn hash or receipt.

web3/pm.py
```diff
@@ -17,11 +24,98 @@
...
+        
+
+    @to_tuple
+    def get_all_package_versions(self, name): # -> (version, uri):
+        package_data = self.registry.functions.get_package_data(name).call()
+        release_count = package_data[1]
+        for index in range(release_count):
+            release_id = self.registry.functions.get_release_id_by_package_and_count(name, (index + 1)).call()
+            release_data = self.registry.functions.get_release_data_by_id(release_id).call()
+            yield (release_data[1].rstrip(b'\x00'), release_data[2].rstrip(b'\x00'))
```
pipermerriam: Same here with respect to readability/future-maintenance.  I'd suggest decomposing `release_data` into its parts and leaving some comments which signal why some of the normalization via `rstrip` is being performed.

web3/pm.py
njgheorghita: Yup, but wouldn't that require anybody who wants to use this class to use the `registry.v.py` implementation I wrote? Not sure if that's ideal. Ideally, I'd like to couple this class to the ERC1319 standard rather than a particular implementation in a particular language.

njgheorghita:
@pipermerriam could use another :eyes: on this - added `web3.pm.tie_registry_to_ens`
Also shoutout @carver - the `ens` fixture in the ens test suite was a huge help. If you don't mind :eyes:  the ens stuff in this pr, just in case i'm missing something/doing something wrong, I'd appreciate it. Thanks!

pipermerriam:
slipped through cracks... looking at this now

tests/core/pm-module/test_ens_integration.py
```diff
@@ -0,0 +1,156 @@
...
+        raise TypeError('val %r could not be converted to bytes')
+    if len(result) < 32:
+        return result.rjust(32, b'\0')
+    else:
+        return result
+
+
+@pytest.fixture
+def ens_setup(solc_deployer):
+    ENS_MANIFEST = ASSETS_DIR / 'ens' / '1.0.1.json'
```
pipermerriam: I started to pester you to move this up to module level but realized why it's here (conditional import failures).  Worth a comment anytime we do something abnormal like this.

tests/core/pm-module/test_ens_integration.py
```diff
@@ -0,0 +1,156 @@
...
+    ens_package = ens_deployer.deploy("ENSRegistry", transaction={"from": ens_key})
+    ens_contract = ens_package.deployments.get_instance("ENSRegistry")
+
+    # create public resolver
+    public_resolver_package = ens_deployer.deploy(
+        "PublicResolver",
+        ens_contract.address,
+        transaction={"from": ens_key}
+    )
+    public_resolver = public_resolver_package.deployments.get_instance("PublicResolver")
```
pipermerriam: These seem like it would be beneficial to split them up into more granular fixtures, but I don't see that as required (just that it might make this easier to read/understand if it splits up well into smaller pieces.

tests/core/pm-module/test_registry_integration.py
```diff
@@ -0,0 +1,200 @@
...
+def test_registry_transfer_owner(vy_registry, w3):
+    vy_registry.transfer_owner(w3.eth.accounts[1])
+    assert vy_registry.registry.functions.owner().call() == w3.eth.accounts[1]
+
+
+# web3.pm integration
+@pytest.mark.skipif(ethpm_installed is False, reason="ethpm is not installed")
+def test_pm_deploy_and_set_registry(w3):
+    w3.pm.deploy_and_set_registry()
+    assert isinstance(w3.pm.registry, Registry)
```
pipermerriam: should probably have a pre-check before the deployment to sanity check it wasn't somehow already present.

tests/core/pm-module/test_registry_integration.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@pytest.mark.skipif(ethpm_installed is False, reason="ethpm is not installed")
+def test_pm_set_registry(vy_registry, w3):
+    w3.pm.set_registry(address=to_canonical_address(vy_registry.address))
+    assert isinstance(w3.pm.registry, Registry)
```
pipermerriam: should probably have a pre-check before the deployment to sanity check it wasn't somehow already present.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+        Sets the current registry used in ``web3.pm`` functions that read/write to an
+        onchain registry.
+
+        :address: Address of the on-chain registry - Accepts ENS name
+        :ens_address: If using an ens, but not on the mainnet, this is the address of the central
+        ENS registry on whatever chain being used.
+        """
+        if is_canonical_address(address):
+            self.registry = Registry(address, self.web3)
+        elif isinstance(address, str):
```
pipermerriam: Why `isinstance(v, str)`?  Seems like we should require either a bytes address or a checksum'd address here to be consistent with the rest of the library.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+        """
+        return self.registry.functions.owner().call()
+
+    @to_tuple
+    def get_all_package_ids(self):
+        """
+        Returns (package_name, release_count) for every package on registry.
+        """
+        package_count = self.registry.functions.packageCount().call()
+        for index in range(0, package_count, 4):
```
pipermerriam: using the term `offset` here instead of `index` will be more inline with the vyper implementation.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+            package_ids = self.registry.functions.getAllPackageIds(index, 5).call()
+            for pkg_id in package_ids:
+                if pkg_id != b'\x00' * 32:
+                    pkg_name = self.registry.functions.getPackageName(pkg_id).call()
+                    _, _, release_count = self.registry.functions.getPackageData(pkg_name).call()
+                    # rstrip used to trim trailing bytes returned in package_name: bytes32
+                    yield (pkg_name.rstrip(b'\x00'), release_count)
```
pipermerriam: I was surprised that this returns a tuple containing the release count instead of just the names (or objects).

Maybe `get_all_package_data`?

Alternatively, `get_all_package_names` is probably fine and you could have a `get_release_count` function to lookup the release counts separately.  Not sure it's good to bundle this functionality.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+        """
+        Sets the current registry used in ``web3.pm`` functions that read/write to an
+        onchain registry.
+
+        :address: Address of the on-chain registry - Accepts ENS name
+        :ens_address: If using an ens, but not on the mainnet, this is the address of the central
+        ENS registry on whatever chain being used.
+        """
+        if is_canonical_address(address):
+            self.registry = Registry(address, self.web3)
```
carver: I'd suggest either:

1.
```py
        if is_canonical_address(address) or is_checksum_address(address):
            self.registry = Registry(address, self.web3)  # and update Registry to work with checksum addresses
```
2.
```py
        if is_canonical_address(address):
            self.registry = Registry(address, self.web3)  # and update Registry to work with checksum addresses
        elif is_checksum_address(address):
            self.registry = Registry(to_bytes(hexstr=address), self.web3)
```


It's reasonable for the `Registry` library to be public-facing (used directly by devs) so I'd recommend option 1 for their convenience.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+        self.registry = Registry.deploy_new_instance(self.web3)
+
+    def tie_registry_to_ens(self, ens_name, ens_address=None):
+        """
+        Sets ``web3.PM`` current registry to the given ``ens_name``. Provide an ``ens_address``
+        if you're using an ENS not on the mainnet. ``web3.eth.defaultAccount`` must be an
+        owner of the domain name for this method to work.
+        """
+        self._validate_set_registry()
+        ns = ENS.fromWeb3(self.web3, ens_address)
```
carver: Use `w3.ens` again

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+        pkg = Package.from_uri(manifest_uri, self.web3)
         return pkg
+
+    def set_registry(self, address, ens_address=None):
+        """
+        Sets the current registry used in ``web3.pm`` functions that read/write to an
+        onchain registry.
+
+        :address: Address of the on-chain registry - Accepts ENS name
+        :ens_address: If using an ens, but not on the mainnet, this is the address of the central
```
carver: I think we should just have people set up their `w3.ens` to be the one they want to use. Then it doesn't require these ens configuration parameters on every function.

web3/pm.py
```diff
@@ -1,27 +1,220 @@
...
+
+    def tie_registry_to_ens(self, ens_name, ens_address=None):
+        """
+        Sets ``web3.PM`` current registry to the given ``ens_name``. Provide an ``ens_address``
+        if you're using an ENS not on the mainnet. ``web3.eth.defaultAccount`` must be an
+        owner of the domain name for this method to work.
+        """
+        self._validate_set_registry()
+        ns = ENS.fromWeb3(self.web3, ens_address)
+        ns.setup_address(ens_name, self.registry.address)
```
carver: This seems like a simple enough operation that the whole method doesn't need to exist. It's a one-liner: `w3.ens.setup_address(ens_name, w3.pm.registry.address)`

njgheorghita:
Thanks for the review @carver & @pipermerriam - gearing up to stabilize `py-ethpm` soon & pr to merge `pm-api` into master sometime (hopefully early) next week.

Issue #1102, opened 2018-10-11
Update minimum `eth-utils` requirement.


pipermerriam:
### What was wrong?

The current lower bound for `eth-utils` is `1.0.1`.  The `eth_utils.toolz` module is only fully available in `eth-utils>=1.2.0`

### How can it be fixed?

Bump the `eth-utils` minimum requirement to be `eth-utils>=1.2.0`

carver:
This was included in v4 in #1085

PR #1104, opened 2018-10-11
Upgrade eth-utils to 1.2.0

katakonst:
### What was wrong?

Related to Issue #1102

### How was it fixed?

Upgraded minimum eth-utils  version to 1.2.0

#### Cute Animal Picture
image


pipermerriam:
This should also get backported to the `v4` branch.  cc @carver for release cutting (but maybe next week since it's so close to the weekend).

PR #1017, opened 2018-08-24
raise ValidationError if ether is sent to non payable functions

ankit-maverick:
### What was wrong?

Related to Issue #902 

### How was it fixed?
Raised `ValidationError` in `call`, `transact`, `estimateGas`, `buildTransaction` if ether is sent to functions that are not payable.

### TODOs:
- [x] Get the approach and code reviewed by @carver 
- [x] Write tests

PR #1115, opened 2018-10-18
prevent sending ether to non-payable functions

carver:
Backport #1017 to v4

#### Cute Animal Picture


carver:
cc @ccolorado looks like it will be in the next v4 release.

PR #1121, opened 2018-10-22
Add default trinity locations to IPC path guesser

carver:
### What was wrong?

web3py wasn't auto-detecting trinity path

### How was it fixed?

Added the trinity paths for mac & linux.

#### Cute Animal Picture

### DIFF
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -1,6 +1,8 @@
 import logging
 import os
-import pathlib
+from pathlib import (
+    Path,
+)
 import socket
 import sys
 import threading
@@ -91,6 +93,12 @@ def get_default_ipc_path(testnet=False):
         if os.path.exists(ipc_path):
             return ipc_path
 
+        base_trinity_path = Path('~').expanduser() / '.local' / 'share' / 'trinity'
+        if not testnet:
+            ipc_path = base_trinity_path / 'mainnet' / 'jsonrpc.ipc'
+            if ipc_path.exists():
+                return str(ipc_path)
+
     elif sys.platform.startswith('linux') or sys.platform.startswith('freebsd'):
         ipc_path = os.path.expanduser(os.path.join(
             "~",
@@ -110,6 +118,12 @@ def get_default_ipc_path(testnet=False):
         ))
         if os.path.exists(ipc_path):
             return ipc_path
+
+        base_trinity_path = Path('~').expanduser() / '.local' / 'share' / 'trinity'
+        if not testnet:
+            ipc_path = base_trinity_path / 'mainnet' / 'jsonrpc.ipc'
+            if ipc_path.exists():
+                return str(ipc_path)
 
     elif sys.platform == 'win32':
         ipc_path = os.path.join(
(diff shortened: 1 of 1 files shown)


web3/providers/ipc.py
```diff
@@ -91,6 +93,14 @@ def get_default_ipc_path(testnet=False):
         if os.path.exists(ipc_path):
             return ipc_path
 
+        base_trinity_path = Path('~').expanduser() / 'Library' / 'Application Support' / 'trinity'
+        if testnet:
+            ipc_path = base_trinity_path / 'ropsten' / 'jsonrpc.ipc'
```
voith: This seems problematic. What happens when trinity adds supports for other networks like `rinkeby`?. A boolean `testnet` won't be sufficient to guess the path.

voith: You can, of course, try all possible paths and check if the path exists?
But what happens when a machine is running both `ropsten` and `rinkeby`? :D 
It will always pick the first path it comes across!

web3/providers/ipc.py
```diff
@@ -91,6 +93,14 @@ def get_default_ipc_path(testnet=False):
         if os.path.exists(ipc_path):
             return ipc_path
 
+        base_trinity_path = Path('~').expanduser() / 'Library' / 'Application Support' / 'trinity'
```
pipermerriam: Is this really where `XDG` directories end up for osx users?

web3/providers/ipc.py
pipermerriam: Yeah, that's problematic.  Wondering if we should do something like:

1. mainnet remains special (and we can also implement some checks to validate that we are indeed connected to the mainnet
2. use a name-based pattern for all the other networks (`rinkeby/kovan/ropsten`) and somehow namespace the imports as `from web3.auto.ropsten import ...`

carver: Yup, I am pretty happy with the idea of dropping support for implicitly detecting the testnet IPC (where `from web3.auto.ropsten import w3` is the next-best option for users).

web3/providers/ipc.py
carver: Not sure, I was cribbing off of Parity since I don't have a Mac to test with.

According to [this stackoverflow post](https://stackoverflow.com/a/5084892/8412986) it might actually be `~/Library/trinity` instead.

We might want to use a more unique namespace on the trinity side, actually, like `~/Library/org.ethereum.trinity`. For Mac, Linux, everything.

voith: @carver This is the path on my machine(mac):
`/Users/voith/.local/share/trinity/mainnet/logs/trinity.log`

voith: @carver Below is the path on my mac machine:
`/Users/voith/.local/share/trinity/mainnet/jsonrpc.ipc`

voith: However for parity this is my path: 
`/Users/voith/Library/Application Support/io.parity.ethereum/jsonrpc.ipc`

carver: Weird, I think I might call that a trinity bug.

pipermerriam: @carver I'm ok considering the move, but only in the face of an objective reason to do so.  I really like the simplicity of XDG and it seems like an easier to find location for most developers.

carver: > but only in the face of an objective reason to do so

Fair enough, and I don't think I can speak for our Mac users preferences here, so I'll drop it for now. I'll make the change in this PR to look in the `~/.local` location on Mac.

PR #1123, opened 2018-10-23
Support Ropsten on Infura

carver:
### What was wrong?

No quick way to load Ropsten on Infura.

### How was it fixed?

Added support for:

```py
from web3.auto.infura.ropsten import w3
```

#### Cute Animal Picture


web3/auto/infura/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+)
+
+from .loader import (
+    load_infura_url,
+    INFURA_MAINNET_BASE_URL,
+)
+
+_infura_url = load_infura_url(INFURA_MAINNET_BASE_URL)
+
+w3 = Web3(HTTPProvider(_infura_url))
```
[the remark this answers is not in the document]
carver: Thanks for the reminder. The old one was using HTTP, and I didn't review that choice. I can't think of a reason that we would prefer HTTP over Websockets...

web3/auto/infura/loader.py
```diff
@@ -1,24 +1,17 @@
...
-    HTTPProvider,
-    Web3,
-)
-
 INFURA_MAINNET_BASE_URL = 'https://mainnet.infura.io'
+INFURA_ROPSTEN_BASE_URL = 'https://ropsten.infura.io'
 
 
-def load_infura_url():
+def load_infura_url(base_url):
```
pipermerriam: *nitpick*

`load` feels out of place here.  I would expect load to be for things like loading from a file.  Maybe `get_` or `construct_` or `build_`?  Not convinced these are much better though.

web3/auto/infura/__init__.py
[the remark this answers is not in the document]
pipermerriam: Maybe make it switchable by environment variable.  Something like `WEB3_INFURA_PROVIDER` set to either `https` or `wss`?

web3/auto/infura/loader.py
```diff
@@ -1,24 +1,17 @@
 import logging
 import os
 
-from web3 import (
-    HTTPProvider,
-    Web3,
-)
-
 INFURA_MAINNET_BASE_URL = 'https://mainnet.infura.io'
+INFURA_ROPSTEN_BASE_URL = 'https://ropsten.infura.io'
```
pipermerriam: Orthogonal but realized these are namespaced with `WEB3_` in front of them which is a convention I think we should stick to.  Otherwise we're likely to have collissions or leakage across library boundaries.

carver: I assume you mean for the environment variable instead of these constants, ie~ `INFURA_API_KEY` -> `WEB3_INFURA_API_KEY`

pipermerriam: Correct.

PR #1125, opened 2018-10-25
Stabilize PM module

njgheorghita:
### What was wrong?
Stabillize pm module. 

### How was it fixed?
`web3.pm.PM` - Main class for interacting with on-chain registries.
`web3.pm.ERCRegistry` - Base class for subclasses to inherit from that target specific implementations of an ERC-compliant registry.
`web3.pm.VyperReferenceRegistry` - Class to interact with deployments of the Vyper Reference Registry implementation
`web3.pm.SolidityReferenceRegistry` - Class to interact with deployments of the Solidity Reference Registry implementation.

### Cute animal picture
image


tests/core/pm-module/test_pm_init.py
```diff
@@ -1,24 +1,114 @@
...
+from web3 import Web3
+
+try:
+    from ethpm import Package  # noqa: E402
+    from ethpm.tools import get_manifest as get_ethpm_manifest
+    from ethpm.exceptions import (
+        InsufficientAssetsError,
+    )
+except ImportError:
+    ethpm_installed = False
```
pipermerriam: You should use this instaed.

https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency

njgheorghita: my plan is to get rid of all these now that ethpm is more or less in stable alpha

njgheorghita: unless you think they should be left in regardless?

tests/core/pm-module/test_registry_integration.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+@pytest.fixture
+def w3():
+    w3 = Web3(Web3.EthereumTesterProvider())
+    w3.eth.defaultAccount = w3.eth.accounts[0]
+    w3.eth.defaultContractFactory = LinkableContract
+    try:
+        from web3.pm import PM  # noqa: F811
+    except ModuleNotFoundError as exc:
```
pipermerriam: Why would this happen?

tests/core/pm-module/test_pm_init.py
pipermerriam: No, if `ethpm` is part of the core requirements then these should go away.  Just suggesting a better way to do conditional skips on import failures that I learned about recently.

tests/core/pm-module/test_registry_integration.py
njgheorghita: off the top of my head - not sure - this still very much a dirty/wip pr - going through cleanup right now - might be worth waiting for a "ping" to review to avoid some of the weirdness from dev-ing this module on a side branch, but more review is always better

njgheorghita: mmmm on that note, i just realized that we've dropped py35 in `ethpm` but it's still around in `web3`. so ... it seems as though a merge into master isn't possible until py35 is dropped in `web3`?

njgheorghita: thinking about the workshop, it adds a little complexity, but is it unreasonable to have the attendees / starter repo work off this `pm-api` branch? (seems unlikely that py35 is gettin dropped in master before devcon)

pipermerriam: cc @carver web3 master is v5 right?  can we drop 3.5 there?

carver: Yup to both! :+1:

pipermerriam: @njgheorghita I'll leave it to you and @carver to make sure you have a working alpha/beta release of web3-v5 to work with.  That seams reasonable.  If there are any holdups, it should be reasonably easy if not ideal to have them install from source.

carver: Yup, I need to do a v4 release, too. I'm flying in about 20 hours, so probably makes sense to wait until I'm there. It seems like a reasonable exception to the weekend release restriction, because I'll be able to dedicate time to checking and fixing anything that pops up. So probably a release of v4 and v5 would go out sometime on Saturday-ish.

carver: It would be a v5 alpha, so of course it would have to be `pip install --pre --upgrade web3` to get it. @njgheorghita

njgheorghita: @carver sounds great! i'm gettin into prague midday on sunday, but will definitely be online saturday

njgheorghita:
@carver @pipermerriam ping for final review - sorry about the massive pr, though the vast majority of this pr has already been approved in prior pr chunks. 

@carver, i'm not sure the best way to resolve `core-eth_abi2` failing builds failing in circleci.
it seems as though force reinstalling `eth-abi-v2` in that tox env is causing problems with the `pytest-ethereum` plugin (which depends on `web3` with `eth-abi>=1.2,<2` installed) any thoughts on how to resolve this? the error resolves when I remove all `web3` dependencies from `pytest-ethereum` but of course then, `pytest-ethereum` is fairly useless and breaks.

carver:
> tox env is causing problems with the pytest-ethereum plugin (which depends on web3 with eth-abi>=1.2,<2 installed) any thoughts on how to resolve this?

pytest-ethereum should probably add support for eth-abi v2-alpha.

How much is `pytest-ethereum` being used? If it's only a couple places, maybe it's fastest to just drop it and locally replace any missing utilities until it has support for eth-abi v2-alpha.

docs/web3.pm.rst
```diff
@@ -1,41 +1,37 @@
 Package Manager API
 ===================
 
-.. py:module:: web3.pm
-.. py:class:: PM
 
 The ``web3.pm`` object exposes methods to interact with Packages as defined by `ERC 1123 <https://github.com/ethereum/EIPs/issues/1123>`_.
 
+To learn more about the EthPM spec, visit the `documentation <http://ethpm.github.io/ethpm-spec/>`__.
```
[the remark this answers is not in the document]
njgheorghita: The docs for the EthPM spec are hosted on Github docs rather than RTD. Also added a link to the Py-EthPM docs.

tests/core/pm-module/conftest.py
```diff
@@ -0,0 +1,260 @@
...
+LOADED_SOL_REGISTRY = release_packages(sol_registry(W3))
+VY_PKG_IDS = (VY_PACKAGE_ID_1, VY_PACKAGE_ID_2, VY_PACKAGE_ID_3)
+SOL_PKG_IDS = (SOL_PACKAGE_ID_1, SOL_PACKAGE_ID_2, SOL_PACKAGE_ID_3)
+VY_RLS_IDS = (VY_RELEASE_ID_1, VY_RELEASE_ID_2, VY_RELEASE_ID_3, VY_RELEASE_ID_4)
+SOL_RLS_IDS = (SOL_RELEASE_ID_1, SOL_RELEASE_ID_2, SOL_RELEASE_ID_3, SOL_RELEASE_ID_4)
+
+
+@pytest.fixture
+def w3():
+    return W3
```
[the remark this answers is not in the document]
njgheorghita: I wend with this structure simply to speed up the tests. 
- module level variables ~ 100seconds
- fixtures as you suggested ~ 1000seconds
I also tried scoping via `@pytest.fixture(scope="session")` - but that requires that I similarly scope `vy_deployer` and `sol_deployer` in `pytest-ethereum`. I didn't think this was a great idea, since those plugins are going to be mostly hidden when used in other libraries, and scoping them didn't seem like a great idea.
I designed the tests in a way that they all expect leakage from the other tests sharing the module-level variables, so maybe it's worth including a note on this setup? Or if you think it's worth the inter-test separation as a trade-off for the speed of running the tests, I'm also open to that.

njgheorghita:
@carver It seems that I'm blocked on updating `pytest-ethereum` and `ethpm` to work with `eth-abiV2` until a `web3` version is released which is dependent on `eth-abiV2`  (currently `web3`'s `eth-abi` dependency is `"eth-abi>=1.2.0,<2.0.0",` - Any idea of when I can expect a `web3` release with an `eth-abiV2` dependency?

docs/web3.pm.rst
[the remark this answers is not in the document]
pipermerriam: Also, I totally mis-read the original link and thought it was just linking to the actual repository.

tests/core/pm-module/conftest.py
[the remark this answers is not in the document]
pipermerriam: I think it's worth a comment at minimum denoting that the design is intentional.

carver:
> Any idea of when I can expect a web3 release with an eth-abiV2 dependency?

It first glance, I think that change is backwards compatible (a project that depends on web3 v4 might also depend on eth-abi v1). How about a v5-alpha.0 release?

njgheorghita:
> How about a v5-alpha.0 release?

@carver That'd be great! It's no huge rush though, whenever it's ready is fine

carver:
Ok, there are still a lot of things on the v5 wishlist. With most of my focus mostly on trinity, that's not likely to change really fast. How long before it becomes a blocker for you, to have v5 unreleased?

njgheorghita:
@carver Sounds good, doesn't look like a blocker right now as far as I can see, but I'll ping you here if it starts to appear on the horizon eventually

njgheorghita:
@carver It's appearing that this might be a blocker for getting the registry explorer deployed (via heroku) for the world to see. I've tried some workarounds, but haven't found a working solution. . .
In `requirements.txt`...
`git+git://github.com/ethereum/web3.py@pm-api#egg=web3.py` - however, this gets overridden by `ethpm`'s dependency on `web3`. 
I tried running the `--upgrade --force-reinstall .. ` command directly into heroku, but that failed due to some compilation error on `cytoolz`
I haven't found many other options that seem to suggest a reasonable solution yet. 

Just wanted to give the heads up. The last thing i'm going to try is using a vps rather than heroku, which will hopefully give me some more control over the build. If you have any ideas as to how I can force reinstall this branch via `requirements.txt` or `setup.py` that would be great!

njgheorghita:
@carver @pipermerriam I'm having trouble remembering if requiring a user to "attach" the `PM` module to a `web3` instance before it's available for use as `web3.pm` was a consequence of `py-ethpm`'s earlier instability, or a permanent design decision? or... now that `py-ethpm` is stable should the `PM` module be automatically available on a `web3` instance, or should users continue to "attach" it?

carver:
> I'm having trouble remembering if requiring a user to "attach" the `PM` module to a `web3` instance before it's available for use as `web3.pm` was a consequence of `py-ethpm`'s earlier instability, or a permanent design decision?

I think `PM` could be a default module, when stable.

> or... now that `py-ethpm` is stable should the `PM` module be automatically available on a `web3` instance, or should users continue to "attach" it?

We must be defining `stable` differently. I mean a pypi stable release (currently alpha), which means you're committing to no backwards incompatible changes in the API on that version series (`v0.n` is a little special, that incrementing `n` is enough for a backwards incompatible change, and the web3 setup file should reflect that).

Note that if web3 v5 uses v0.1 of `PM` as a default module, and there is a backwards incompatible change in py-ethpm to v0.2, that upgrade should wait until web3 v6. If we mark `PM` as experimental in the docs and don't include it as a default module, then I think we have more freedom to upgrade py-ethpm in backwards incompatible ways, without waiting until web3 v6.

njgheorghita:
@carver re: comment above, I'm torn between the two
- adding `PM` as a default module on `web3` - making it easier for the user to use, but being restricted to no backwards incompatible changes until `web3 v6`
- leaving `PM` off `web3` by default - forcing the user to `attach` `PM` to their `web3` instance, but having more flexibility with the api while still in `web3 v5`
But, I'm leaning towards the latter option for now as it seems possible the API might change in the near future (especially as collaboration with the other pm teams is picking up). And the `attach` line is more just annoying than an actual impediment to use.

pipermerriam:
As discussed in this morning's stand-up, I'd advocate for requiring explicitly enabling this API for now before we commit to stability.

```python
w3 = Web3(...)
w3.enable_unstable_package_management_api()  # handles the attaching
```

web3/main.py
```diff
@@ -206,3 +206,18 @@ def ens(self):
...
+        if self._pm is not None:
+            return self._pm
+        else:
+            raise AttributeError(
+                "The Package Management feature is disabled by default until "
+                "its API stabilizes. To use these features, please enable them by running "
+                "`w3.enable_unstable_package_management_api()` and try again."
+            )
+
+    def enable_unstable_package_management_api(self):
```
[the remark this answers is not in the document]
pipermerriam: I'm not inclined to implement a `disable` API as they could just re-instantiate their Web3 instance.

docs/web3.pm.rst
```diff
@@ -1,41 +1,53 @@
...
 
 
 Methods
 -------
+The following methods are available on the ``web3.pm`` namespace.
+
+.. autoclass:: web3.pm.PM
+   :members:
+
+.. note:: If you want to implement your own registry and use it with ``web3.pm``, you must create a subclass that inherits from ``ERCRegistry``, implements all the methods defined in ``ERCRegistry``, and manually set it as the ```registry`` attribute on ``web3.pm``.
```
pipermerriam: This seems like something that would be worth putting in it's own section and 1) explaining *why* someone would want to do this and 2) how they would do it with a code example.

If the code example isn't something you want to maintain or you don't see it as very useful then maybe just state that replacing the registry is beyond the scope of this guide.

tests/core/pm-module/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+    Deployer,
+)
+
+from web3 import Web3
+from web3.pm import (
+    SolidityReferenceRegistry,
+    VyperReferenceRegistry,
+)
+
+VY_PACKAGE_ID_1 = b'\xd0Y\xe8\xa6\xeaZ\x8b\xbf\x8d\xd0\x97\xa7\xb8\x92#\x16\xdc\xb7\xf0$\xe8"\x0bV\xd3\xc9\xe7\x18\x8ajv@'  # noqa: E501
```
pipermerriam: It *might* be nice to represent these as their hex values like:

```python
VY_PACKAGE_ID_1 = to_bytes(hexstr='0xd0...')  # noqa: E501
...
```

As those values are easier to visually parse.

tests/core/pm-module/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+SOL_PACKAGE_ID_1 = b"`\xc5\x11+a\x15\x9ekB\xd5M\x94Px9N\x9d_\xc9\xc6\xff\x0f=\xf7\x89w\x00o\x8b\xbc\x06\xd4"  # noqa: E501
+SOL_PACKAGE_ID_2 = b"\xdb\xcf\xb0\xbdq\x15\xbfe\x93P\xd7{\xb2+\xb8\x89\xca\x82\x94\xf6\x1b\x0c\xa4\x80\xf8\xa4{\xb8\xfc\x90L\xc9"  # noqa: E501
+SOL_PACKAGE_ID_3 = b"\xf3\xe4\x00,H\xa7\xf8\xf3H]b\x98\x83\x17\x84\x9c\x17S@\xb6e\x17\xc3\xb2\x99?rVC\xeb\xa8K"  # noqa: E501
+SOL_RELEASE_ID_1 = b"s\x83Vh\xf7\x1cz\xe8\\\xbd\xcd\xbbZ\x99\x05\xfaB\x0f\xfe\x85\xa8G\xd2\x83\xfa\x9b\xee\xfc\xd5l\xac\xc4"  # noqa: E501
+SOL_RELEASE_ID_2 = b"\xe5\xef\x02\x92\xa3\xb3kj\xc2\xbe\x07\xee\x92\xdfa\xbe\x15\xe0\xb9\xf1\x02\xdf2\xcb\xe3\xf0\xc0\x12\xefi\xd4b"  # noqa: E501
+SOL_RELEASE_ID_3 = b"\x12\x80\x14\x8e\n\xf5\xc4~\x95\xb4\x1d\xf1W4\xd0rls \xfcL\xf2\x1e\xfe9#\xfb\x04{S\x89\x9d"  # noqa: E501
+SOL_RELEASE_ID_4 = b"p\x82\xe9T\xe4\xfdj\xdf\x8a%\xc6\xce\xfe!\x8f2\xecf\xd8\xa1\x97\x19\x7f\x1d\x05\xaag\xa6\\\xafQ\x11"  # noqa: E501
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO, format="%(message)s")
```
pipermerriam: Instead of this you can just use `pytest tests/test_things.py --log-cli-level=INFO` to get logging messages to show up in your terminal.  This approach will unconditionally dump them to stdout which probably isn't globally desired.

setup.py
```diff
@@ -69,10 +69,12 @@
     install_requires=[
         "eth-abi>=2.0.0b4,<3.0.0",
         "eth-account>=0.2.1,<0.4.0",
+        "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
         "eth-utils>=1.3.0,<2.0.0",
+        "ethpm>=0.1.4a10,<1",
         "hexbytes>=0.1.0,<1.0.0",
         "lru-dict>=1.1.6,<2.0.0",
-        "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
+        "pytest-ethereum>=0.1.3a6,<1",
```
carver: Looks like this should be moved to the `tester` extra, instead of as a core dependency.

tests/core/pm-module/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+)
+from ethpm import (
+    ASSETS_DIR,
+    Package,
+)
+from ethpm.contract import (
+    LinkableContract,
+)
+from pytest_ethereum import (
+    linker as l,
```
carver: One-letter variable? Maybe don't rename the import at all.

web3/pm.py
```diff
@@ -1,23 +1,597 @@
...
+# To use the `PM` module, attach it to your web3 object using the `attach` facility
+# PM.attach(web3, 'pm')
+
+
+class ERCRegistry(ABC):
+    """
+    The ERCRegistry class is a base class for all registry implementations to inherit from. It
+    defines the methods specified in `ERC 1319 <https://github.com/ethereum/EIPs/issues/1319>`__.
+    All of these methods are prefixed with an underscore, since they are not intended to be
+    accessed directly, but rather through the methods on ``web3.pm``.
```
carver: One important reason that we usually have for methods that are "not intended to be accessed directly" is that the API might change without a major version bump. Would be nice to add a note about whether these method signatures subject to change. That's not super relevant during alpha, but we might as well document for stable now.

setup.py
njgheorghita: Hmmm, maybe I'm missing something, but we use pytest-ethereum's `Deployer` in the `web3.pm` module. But, I would say that this use case of using the `Deployer` is not that critical at all, since it's a relatively straightforward deployment. So, I can definitely pull it out, in which case `pytest-ethereum` would only be needed in the tests and therefore the `tester` extra.

carver: Ah, I just did a Ctrl-f in the github diff, and the diff on `pm` was hidden.

At least semantically, I find it pretty surprising to use anything called `pytest-*` in the main deployed package. If `Deployer` is to be used in production code like this, then maybe it could be moved into a web3 `_utils` module.

PR #1135, opened 2018-11-09
removed _function_names and _event_names

MichaelRazum:
### What was wrong?
ContractEvents and ContractFunctions have not used lists _event_names / _function_names.
Each time a contract is created through the factory method, the _event_names list and the _function_names list are extended.
Creating many contract's results in high memory usage.

### How was it fixed?
Removed unused class level lists. If they are really needed for some reason, a set would prevent the high memory usage. Especially in the case that a contract (with the same ABI) is created multiple times.

#### Cute Animal Picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -105,14 +105,11 @@ class ContractFunctions:
     """Class containing contract function objects
     """
 
-    _function_names = []
-
     def __init__(self, abi, web3, address=None):
         if abi:
             self.abi = abi
             self._functions = filter_by_type('function', self.abi)
             for func in self._functions:
-                self._function_names.append(func['name'])
                 setattr(
                     self,
                     func['name'],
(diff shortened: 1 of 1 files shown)


pipermerriam:
This should be backported to v4

MichaelRazum:
> This should be backported to v4
Looking forward to it : ) How do you do the backport? Just new v4.8.2 or do you want to include it in all the existing tags. Just curious.

pipermerriam:
There is a `v4` branch that a pull request needs to be opened against.

carver:
Recommendation:
- check out `v4` locally
- create a new branch `git branch -b memory-leak-fix-v4 v4`
- cherry pick the commit from the PR
- push up your new branch
- open a PR

MichaelRazum:
Just created the PR!

PR #1145, opened 2018-11-22
Add parity-specific listStorageKeys RPC.

palkeo:
### What was wrong?

There was no support for the parity_listStorageKeys RPC:
https://wiki.parity.io/JSONRPC-parity-module#parity_liststoragekeys

### How was it fixed?

I added a function to implement it in the parity module.

### How was it tested?

I tested locally, and also ran the parity integration tests.
I added a test, but because the current integration tests doesn't seem to use "--fat-db", parity returns None. I make sure it does.

However, the test is run in two variants, and one fails:
`tests/integration/parity/test_parity_http.py::TestParityTraceModuleTest::test_list_storage_keys_no_support[<lambda>]`

Where this one passes:
`tests/integration/parity/test_parity_http.py::TestParityTraceModuleTest::test_list_storage_keys_no_support[address_conversion_func1]`

The failure is because the address seems to be passed as raw bytes, and cannot be serialized to json. That's supposed to be allowed? How to handle that? I don't see code to handle it in the other RPCs.

#### Cute Animal Picture

```
   \
    \ /\
    ( )
  .( o ).
```
(that's an ASCII Bunny viewed from behind)

### DIFF
--- a/web3/_utils/rpc_abi.py
+++ b/web3/_utils/rpc_abi.py
@@ -57,6 +57,8 @@ RPC_ABIS = {
     'personal_unlockAccount': ['address', None, None],
     'personal_sign': [None, 'address', None],
     'trace_call': TRACE_PARAMS_ABIS,
+    # parity
+    'parity_listStorageKeys': ['address', None, None, None],
 }
 
 
--- a/tests/integration/parity/common.py
+++ b/tests/integration/parity/common.py
@@ -169,6 +169,9 @@ class ParityPersonalModuleTest(PersonalModuleTest):
 
 
 class ParityTraceModuleTest(TraceModuleTest):
+    def test_list_storage_keys_no_support(self, web3, emitter_contract_address):
+        super().test_list_storage_keys_no_support(web3, emitter_contract_address)
+
     def test_trace_replay_transaction(self, web3, parity_fixture_data):
         super().test_trace_replay_transaction(web3, parity_fixture_data)
 
--- a/web3/_utils/module_testing/parity_module.py
+++ b/web3/_utils/module_testing/parity_module.py
@@ -11,6 +11,10 @@ from web3._utils.formatters import (
 
 class ParityModuleTest:
 
+    def test_list_storage_keys_no_support(self, web3, emitter_contract_address):
+        keys = web3.parity.listStorageKeys(emitter_contract_address, 10, None)
+        assert keys is None
+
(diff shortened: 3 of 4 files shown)


carver:
Thanks!

> The failure is because the address seems to be passed as raw bytes, and cannot be serialized to json. That's supposed to be allowed?

Yes, we allow three formats:
- EIP-55 checksum address (0x + 40 hex mixed-case characters)
- 20 bytes
- ENS name

> How to handle that? I don't see code to handle it in the other RPCs.

We have a middleware that looks for address inputs, which we have defined here:
https://github.com/ethereum/web3.py/blob/0cbb1e4d3bf5f5fe1e7493f374d824b112d389d8/web3/_utils/rpc_abi.py#L35-L39

See `eth_getBalance` for example.

---

Also, note that another PR should be opened as a backport to the `v4` branch, unless you're okay waiting a while before this feature is out in a stable web3 release. (I'd recommend waiting until this PR his merged before doing the backport, to minimize duplicate work)

palkeo:
I registered my function in the middleware you pointed me to. It seems to work now :)
And I'm totally fine waiting until it's in the stable release.

web3/_utils/rpc_abi.py
```diff
@@ -57,6 +57,8 @@
     'personal_unlockAccount': ['address', None, None],
     'personal_sign': [None, 'address', None],
     'trace_call': TRACE_PARAMS_ABIS,
+    # parity
+    'parity_listStorageKeys': ['address', None, None, None],
```
carver: This seems fine for now, but in general, we probably want a different solution. It seems weird to have this included in this default middleware, for a non-default module. (Comment mostly intended for maintainers)

carver:
_I love the ascii bunny, too XD_

web3/_utils/rpc_abi.py
carver: Ah yes, that's familiar. I guess in that framework this is a method normalizer, really, since it depends on argument type and order.

Contract middleware is one I don't remember talking about. Do you have an example? Are you thinking of attaching a middleware to a specific contract instance, or across all, or what?

pipermerriam: > attaching a middleware to a specific contract instance

This, though thinking through it there isn't a *lot* that this gives us that couldn't be accomplished by pushing users towards using separate `w3` instances in cases where they want different behavior across different contract instances (think in-flight signing, or gas estimations).

Issue #829, opened 2018-05-09
Tuple ABI support


carver:
### What was wrong?

Tuple types are not supported yet in Web3.py

### How can it be fixed?

- Add tests for passing in & returning tuple values with contract functions
- Figure out how to handle `web3.utils.abi.is_encodable` when elements of the tuple are handled as special cases, like what if the function accepts this tuple as an argument: `(address, string, (bytes, bytes32))` and someone passes in `('myname.eth', b'already-encoded', ('0x0123', '0x4567'))`

One option to consider: write custom encoders for address, bytes*, and string, replacing the default ones. Then `is_encodable` would be replaced entirely with `eth_abi.is_encodable`

pipermerriam:
We can *probably* special case the tuple case for `address` and `string` so that our `is_encodable` function knows how to walk tuple types.  Any 3rd party types may just be out of luck for now if they want ENS names or fancy string handling.

pipermerriam:
Lets define this as two phases.

1. support passing in the values only as actual tuples like your example.
2. support passing in richer objects like a `dict` or any object with all of the necessary attributes for the tuple so that fields of the tuple can have names.
3. support for returning richer versions of tuple return types, maybe via `namedtuple` or `dict`

carver:
> so that fields of the tuple can have names.

TIL that [tuple components are named in the ABI](http://solidity.readthedocs.io/en/v0.4.23/abi-spec.html#handling-tuple-types). Nice

nandod:
I would also love to see this fixed as I have functions that return tuples; decode_abi is failing and BTW the exception handling code in call_contract_function ends up calling web3.eth.getCode which is a huge hit on the network. I'd have a look at this myself weren't my python skills so minimal.

stefanmendoza:
I'd be interested in tackling this! May need some guidance, but I'll start digging into the contract ABI spec:
https://solidity.readthedocs.io/en/latest/abi-spec.html

Let me know if there's any other docs that you think would help give me more context!

stefanmendoza:
@carver following up on this - was on vacation, going to start looking into this again this week. 👍

stefanmendoza:
@fabioberger still working on this! Was finishing looking over the contract ABI spec last night to get a better understanding how the tuples / structs work. Was gonna post out any questions I have tonight or tomorrow to clarify what exactly we need / fill in any gaps in my understanding and then get to digging into the implementation tomorrow. 👍

[...]

PR #1147, opened 2018-11-26
Naive tuple support

feuGeneA:
### What was wrong?

Tuples are not supported in contract interfaces (Issue #829).  This PR offers a partial solution to that problem.

I've got a very specific use case that I need to get working.  It's a call to [an existing contract method, which both accepts and returns a `struct`](https://0xproject.com/docs/contracts#Exchange-getOrderInfo), which of course is encoded as a `tuple` in the ABI.

Thankfully, since my requirements don't demand support for ENS names or any other fancy string handling in the `struct` components, I'm taking [@pipermerriam 's comment](https://github.com/ethereum/web3.py/issues/829#issuecomment-387601644) as license to make just the changes I need in order to get my use case working.  My understanding is that the use of `eth_abi`'s registry will be used for that, but that's a bigger task than I need done, so I'm leaving it for someone else to implement :smile:

The problem is that, while the current code just blindly copies the type names from the ABI and passes them to `eth_abi`, the type name `'tuple'` isn't supported by `eth_abi`, which raises an exception `ValueError: No matching entries for 'tuple' in encoder registry`.  Instead, `eth_abi` expects you to *describe* the tuple; that is, indicate the type not just as `'tuple'` but rather as a parenthesized list of the types of the tuple components.  So if you had a function that accepts a Solidity `struct { uint256 a; uint256 b; }` as one of its arguments, then you would need to use `eth_abi` like `encode_abi('(uint256, uint256)', [0, 1])`

### How was it fixed?
My solution is simple: Before calling `eth_abi` methods, look in the ABI for arguments of type `tuple`, and if found walk through them to compose type and argument lists as `eth_abi` expects them.

(By "naive" I mean that my solution does not try to solve the problem of processing, for example, ENS names in address fields of a struct, or any other fancy stuff that would need to be done for the fields of the struct.)

- [x] Support tuples in `is_encodable` and other preparation logic.
- [x] Support tuples in `encode_abi` and other supporting logic
- [x] Support tuples in decoding of value returned by contract method
- [x] Tests!  Run existing and write new!
- [ ] Documentation?

#### Cute Animal Picture


feuGeneA:
Note that at this time there are three separate places that "expand" tuples. There feels like a lot of duplication between them, but they are really three separate cases. One expands just tuple types, another just tuple values, and the third expands both types and values.  Another expansion will be needed for return value decoding, which I expect will re-use the types&values logic.

feuGeneA:
Here is the error output given when running with the current code: https://gist.github.com/feuGeneA/68e714f2cd942f4cfef094b26459b912

web3/_utils/abi.py
```diff
@@ -216,6 +238,32 @@ def check_if_arguments_can_be_encoded(function_abi, args, kwargs):
...
+        if types[i] == "tuple":
+            tuple_type = "("
+            tuple_value = tuple()
+            for j in range(0, len(function_abi['inputs'][i]['components'])):
+                component_type = function_abi['inputs'][i]['components'][j]['type']
+                tuple_type += component_type + ","
+                if isinstance(arguments[i], dict):
+                    component_name = function_abi['inputs'][i]['components'][j]['name']
+                    tuple_value += (arguments[i][component_name],)
+                elif isinstance(arguments[i], tuple):
```
feuGeneA: In my experiments, this function is called once to follow through this `is...tuple` branch, and another through the parallel `is...dict` branch.

feuGeneA:
At this point this is fully functional.  In addition to the CI runs being clean, my own (external) test case of calling into [the contract I'm trying to support](https://0xproject.com/docs/contracts#Exchange-getOrderInfo) is also working.  (Though, I should say, it's working only when running with my [eth-utils PR#141](https://github.com/ethereum/eth-utils/pull/141) in place.)

I would appreciate an initial review before I spend time polishing and whatnot.  @carver ? @pipermerriam ?  Thank you! :pray:

feuGeneA:
I'm not sure why the most recent CI run failed the `py36-integration-parity-ws` check, on the `install parity if needed` step, but it seems unrelated to my changes, since all checks passed on the previous commit, and only a docstring was changed in the last commit.

web3/_utils/abi.py
```diff
@@ -163,6 +175,28 @@ def is_encodable(_type, value):
...
+
+            raise ValueError(
+                "Tuple type {} does not have corresponding data in value {}"
+                .format(_type, value)
+            )
+
+        all_are_encodable = True  # assume until proven otherwise
+        for i in range(0, len(components)):
+            all_are_encodable = all_are_encodable and is_encodable(
+                components[i], value[i]
```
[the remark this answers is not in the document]
feuGeneA: I've cleaned this up since the revision you commented on.  I happened to achieve both the things you mentioned (laziness and avoiding sequence indexing).

web3/_utils/abi.py
```diff
@@ -163,6 +175,28 @@ def is_encodable(_type, value):
     if not isinstance(_type, str):
         raise ValueError("is_encodable only accepts type strings")
 
+    if _type[0] == "(":  # it's a tuple. check encodability of each component
+        components = _type.strip("()").split(",")
+        if not any(
+            [isinstance(value, collection) for collection in [list, tuple]]
```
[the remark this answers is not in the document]
feuGeneA: Nice, incorporated, thank you.

web3/_utils/abi.py
```diff
@@ -163,6 +193,23 @@ def is_encodable(_type, value):
     if not isinstance(_type, str):
         raise ValueError("is_encodable only accepts type strings")
 
+    if _type[0] == "(":  # it's a tuple. check encodability of each component
```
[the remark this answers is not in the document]
feuGeneA: Yes, it is possible that the collapsing function could yield a nested tuple.  See demonstration in the tests I just added.  And you're right, `is_encodable()` does indeed fail in that case.  Let me see if I can't scheme up a reasonably elegant way to handle it.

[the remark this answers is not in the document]
feuGeneA: @davesque I believe I've addressed your concern in my latest commit, thank you for raising it.  Let me know what you think.

feuGeneA:
That failed `py37-integration-parity-http` check doesn't seem to be my fault.  The output is littered with `ConnectionRefusedError`.  Can someone re-run for me?  Thanks!

web3/_utils/abi.py
```diff
@@ -51,14 +51,44 @@ def get_abi_input_types(abi):
     if 'inputs' not in abi and abi['type'] == 'fallback':
         return []
     else:
-        return [arg['type'] for arg in abi['inputs']]
+        return [collapse_if_tuple(abi_input) for abi_input in abi['inputs']]
+
+
+def collapse_if_tuple(abi_type):
```
feuGeneA: This function is deficient in that it doesn't handle arrays of tuples.  Either it should be updated to match [what was done in eth-utils #141](https://github.com/ethereum/eth-utils/pull/141/commits/227beca013b4167a148e610d238892d9679f0e59).

Or, even better, this function should be removed from here, and it's callers should use the one in `eth-utils` instead.  But, that change in `eth-utils` hasn't been released yet...

web3/_utils/abi.py
```diff
@@ -51,14 +51,44 @@ def get_abi_input_types(abi):
     if 'inputs' not in abi and abi['type'] == 'fallback':
         return []
     else:
-        return [arg['type'] for arg in abi['inputs']]
+        return [collapse_if_tuple(abi_input) for abi_input in abi['inputs']]
+
+
+def collapse_if_tuple(abi):
```
feuGeneA: This function should be removed from here, and it's callers should use the one in `eth-utils` instead.  But, [that change in `eth-utils`](https://github.com/ethereum/eth-utils/pull/141/commits/227beca013b4167a148e610d238892d9679f0e59) hasn't been released yet...

web3/_utils/abi.py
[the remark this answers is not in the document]
feuGeneA: @davesque would you mind taking another look?  I realize we're still deliberating strategy in #829 , but I thought it would be good to get your feedback on this while it's still relatively fresh in our minds.

For quick reference: That test showing that `collapse_if_tuple()` can indeed yield a nested tuple is [here](https://github.com/ethereum/web3.py/pull/1147/commits/506ec40568aa47ec111b99e904c370a50c6bd32d#diff-30c9fd211065dbe2e5e546e745d4f853R228).  And the change to support nested tuples in `is_encodable()`, the code in question in this thread, is [here](https://github.com/ethereum/web3.py/pull/1147/commits/5837fd608e59999f9469c6769c5092705d5a7fc7).

stefanmendoza:
@feuGeneA  From first glance this looks good to me. Do you think it would make sense to move these helper methods to eth-abi and the logic from is_encodable into the default ABICodec as well?

feuGeneA:
@stefanmendoza thank you for looking.

I'm good with moving those helper methods to wherever it makes sense to have them.  If/when there's a specific use for them in eth-abi (or its clients) then sure, let's do that.

And yeah, if ABICodec will provide an `is_encodable()` method, then I think we'll need to have that logic in there.

For me, the big question is this: can we accept what I've got here and then later build out the registry stuff on top of this?

I have clients that have been itching for these changes for several weeks, and I'd really like to see these changes landed so I can make them happy sooner rather than later. :smile:  Are you amenable to an incremental approach?

If there are any further changes I could make here that would facilitate such an approach, please ask!

stefanmendoza:
@feuGeneA I'm sure that'd be fine... I'll have to look into this more this evening to see how we might combine my eth-abi PR and this without having to do a lot of refactoring in the subsequent PRs. I'm still trying to get my bearings with this repo (haven't really played around in web3.py much). I'd like some input from @davesque on this since he has context to what the eth-abi PR was and how they may be compatible.

carver:
> I have clients that have been itching for these changes for several weeks, and I'd really like to see these changes landed so I can make them happy sooner rather than later.

I totally understand the sense of urgency in serving client, I've been there. @feuGeneA 

Please understand that from a maintainers perspective, we have to consider all the future work that every new merge might cause. If we rush to merge something that we expect still needs refactoring, it directly reduces the progress we can make on other great features and projects in the space. Even if we don't have to code the refactor, reviewing also takes time.

I'd like to see the outcome of this before we merge:

> I'll have to look into this more this evening to see how we might combine my eth-abi PR and this without having to do a lot of refactoring in the subsequent PRs.

Of course, you're welcome to take a look at @stefanmendoza 's changes to do the same.

---

Thanks everyone! I know contributing to open source projects takes way more effort than just forking off locally to get something done for yourself. We appreciate that work, and the improvements!

web3/_utils/abi.py
```diff
@@ -159,10 +162,69 @@ def collapse_type(base, sub, arrlist):
...
+    Properly handles the case where some component types are themselves tuples.
+
+    >>> get_tuple_component_types('(uint256,(address,address),uint256)')
+    ['uint256', '(address,address)', 'uint256']
+    """
+    component_types = []
+    components_csv = tuple_type[1:-1]  # strip off expected outer parentheses
+    # consider case of inner parentheses:
+    open_paren_index = components_csv.find("(")
+    close_paren_index = components_csv.rfind(")")
```
stefanmendoza: @feuGeneA  I was looking at this and thought there might be a logic issue here...
```
>>> from web3._utils import abi
>>> tuple = '(int,(address,address),(address,address),int)'
>>> abi.get_tuple_component_types(tuple)
['int', '(address,address),(address,address)', 'int']
```
instead of
```
['int', '(address,address)','(address,address)', 'int']
```
Since this does a reverse lookup of the closing paren, it can't handle a tuple with multiple tuples. This logic should fix that:
```
def get_tuple_component_types(tuple_type):
    components = tuple_type[1:-1]
    
    if '(' not in components and ')' not in components:
        # There were no tuples in the list, so we can just split on commas
        return components.split(',')
    
    split_components = re.split('([\.^(^)])', components)    
    reduced_components = list(filter(lambda val: val not in ['', ',', '(', ')'], split_components))
    clean_components = list(map(lambda val: val.strip(','), reduced_components))
    return list(map(lambda maybeTup: '(' + maybeTup + ')' if ',' in maybeTup else maybeTup, clean_components)) 
```
This works on the following cases:
```
(int,int) # Tuple with no inner tuples
['int', 'int']

(int,(address,address),(address,address),int) # Tuple with consecutive inner tuples
['int', '(address,address)', '(address,address)', 'int']

((address,address),int,(address,address),int,(address,address)) # Tuple with tuples as the first and last types
['(address,address)', 'int', '(address,address)', 'int', '(address,address)']
```
this still falls short in the event that we have inner nested tuples:
```
(int,(address,(int,int)),int) # Tuple with nested inner tuple
['int', 'address', '(int,int)', 'int']
```
and may be inefficient due to the creation of so many lists, but I think it's a good starting point / a step closer.

cc @davesque / @carver

stefanmendoza: @davesque welp... That's a lot simpler than what either of us were doing lol
```
>>> tuple1
'(int,int)'
>>> [str(t) for t in parse(tuple1).components]
['int', 'int']

>>> tuple2
'(int,(address,address),(address,address),int)'
>>> [str(t) for t in parse(tuple2).components]
['int', '(address,address)', '(address,address)', 'int']

>>> tuple3
'((address,address),int,(address,address),int,(address,address))'
>>> [str(t) for t in parse(tuple3).components]
['(address,address)', 'int', '(address,address)', 'int', '(address,address)']

>>> tuple4
'(int,(address,(int,int)),int)'
>>> [str(t) for t in parse(tuple4).components]
['int', '(address,(int,int))', 'int']

>>> tuple5
'(((address,address),int,int),int)'
>>> [str(t) for t in parse(tuple5).components]
['((address,address),int,int)', 'int']
```

I need to do a better job of digging through the helper libraries / repos for web3 to see if there are already functions implemented to do what I'm trying to do :P

Thanks, David!

feuGeneA: Thanks @davesque !  Much simpler indeed!

@stefanmendoza I just removed the method entirely, since there was only one caller.  And I migrated our test cases over to `tests/core/utilities/test_abi_is_encodable.py`.

davesque: @feuGeneA @stefanmendoza No problem!  I'm still sort of on holiday but am starting to get back in the groove of things during this week.  I think there might be some other places where the functionality in eth-abi could be used.  Part of my plan for merging this PR is to go through and make sure there aren't still any places where we could be doing that.

davesque:
@feuGeneA Thanks for all your work on this.  Is it possible to squash some of the commits in this branch?  Also, if any of the resulting squashed commits correspond to fixes for a particular error, can you omit the error message since it takes up quite a bit of space in the log?  I'm finding it a bit hard to read and follow the commit history.

feuGeneA:
Yes indeed. I usually am quite meticulous about making a nice commit
history, but the hack-first, polish-later approach that I took here didn't
facilitate that. And I love to squash, but never after review, but since
you asked I'll be happy to clean it up. Thank you for taking the time to
review!

davesque:
@feuGeneA Thanks for that, Eugene!

feuGeneA:
Okay @carver , I've gotten this to a place I'm happy with.  In the interest of having commits that are completely functional and independently verifiable, nearly all of the changes ended up in a single commit (though I was able to tease a few things out).  Let me know what you think, and in particular let me know if there's anything else I can do to facilitate.  Thanks again!

feuGeneA:
Okay @davesque , I've gotten this to a place I'm happy with. In the interest of having commits that are completely functional and independently verifiable, nearly all of the changes ended up in a single commit (though I was able to tease a few things out). Let me know what you think, and in particular let me know if there's anything else I can do to facilitate. Thanks again!

tests/core/utilities/test_abi.py
```diff
@@ -29,9 +34,45 @@ def test_abi_data_tree(types, data, expected):
...
+)
+def test_data_tree_map(func, data_tree, expected):
+    assert data_tree_map(func, data_tree) == expected
+
+
 @pytest.mark.parametrize(
     'types, data, funcs, expected',
     [
+        (  # like web3._utils.rpc_abi.RPC_ABIS['eth_getCode']
+            ['address', None],
```
carver: This has come up a few times, so let's all be on the lookout for it: recursive structures must be handled. For one example, what happens when we try to run a map function over ABI data that has an array inside a tuple?

Something like:
```py
        (
            ['(string,address[])'],
            [(b'a string', [b'\xf2\xe2F\xbbv\xdf\x87l\xef\x8b8\xae\x84\x13\x0fOU\xde9['],
            [addresses_checksummed, abi_string_to_text],
            ['a string', ['0xF2E246BB76DF876Cef8b38ae84130F4F55De395b']],
        ),

File: web3/_utils/abi.py [hunk 9]
```diff
@@ -550,6 +653,13 @@ def __new__(cls, iterable):
 
 
 def abi_sub_tree(data_type, data_value):
+    if (
+        isinstance(data_type, str) and
+        data_type[0] == "(" and
+        isinstance(data_value, tuple)
+    ):
+        return ABITypedData([data_type, data_value])
```
carver: Note below how the array-type ABI values have `abi_sub_tree` applied recursively? I was expecting something like that for a tuple. Otherwise, I suspect `data_tree_map` will fail to apply to the types within the tuple (or arrays within the tuple).

davesque:
@feuGeneA @stefanmendoza Thanks for all your work on this guys!  We'll take things from here to do the work of finalizing things and merging them.

kclowes:
This is in as of #1235 :tada: Thanks everyone!

PR #1153, opened 2018-12-04
Fix contract doc typos

kclowes:
### What was wrong?
Just a few typos I noticed while reading docs. `it` should be `is`

#### Cute Animal Picture


pipermerriam:
Test failure is due to the doctests actually reading *live* data from the original "Unicorn Token" contract.  You can bump the 2039 to 2040 and this will pass.

PR #1156, opened 2018-12-07
Freebsd support: cleanup and tox fix

rayrapetyan:
### What was wrong?
- tox fails to run because of additional include directories dependency
- unnecessary file manipulations and downloads

### How was it fixed?
- put paths into .pydistutils.cfg
- removed extra steps

### DIFF
--- a/docs/README-freebsd.md
+++ b/docs/README-freebsd.md
@@ -4,27 +4,46 @@
 
 ### Prerequisites
 
-Make sure you've UTF-8 defined for charset and lang in your [~/.login_conf](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/using-localization.html),
-otherwise almost every module will fail to install. Python 3 is very sensitive to that.
+Make sure you've UTF-8 defined for charset and lang in your [~/.login_conf](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/using-localization.html), 
+otherwise almost every Python 3 module will fail to install.
+
+`~/.login_conf`:
+```
+me:\
+	:charset=UTF-8:\
+	:lang=en_US.UTF-8:
+```
+
+Also make sure you've defined valid include and library paths in `~/.pydistutils.cfg`, otherwise native compilations fail.
+
+`~/.pydistutils.cfg`:
+```
+[build_ext]
+include_dirs=/usr/local/include
+library_dirs=/usr/local/lib
+```
 
 ```
-sudo pkg install python3 py36-virtualenv leveldb libxml2 libxslt pkgconf secp256k1 wget git
-sudo wget https://raw.githubusercontent.com/bitcoin/bitcoin/master/src/secp256k1/include/secp256k1_recovery.h -O /usr/local/include/secp256k1_recovery.h
-echo '#include "/usr/include/stdlib.h"' | sudo tee -a /usr/include/alloca.h > /dev/null
+sudo pkg install python3 py36-virtualenv git leveldb libxml2 libxslt pkgconf gmake secp256k1
+
+# hack around https://github.com/ethereum/ethash/pull/107#issuecomment-445072692
+sudo touch /usr/local/include/alloca.h
+
 mkdir -p /tmp/venv_python
 virtualenv-3.6 /tmp/venv_python/python3
 source /tmp/venv_python/python3/bin/activate.csh
 
-# these two modules fail to install natively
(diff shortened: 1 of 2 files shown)


docs/README-freebsd.md
```diff
@@ -4,27 +4,44 @@
...
 
-Make sure you've UTF-8 defined for charset and lang in your [~/.login_conf](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/using-localization.html),
-otherwise almost every module will fail to install. Python 3 is very sensitive to that.
+Make sure you've UTF-8 defined for charset and lang in your [~/.login_conf](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/using-localization.html), 
+otherwise almost every Python 3 module will fail to install.
 
+`~/.login_conf`:
 ```
-sudo pkg install python3 py36-virtualenv leveldb libxml2 libxslt pkgconf secp256k1 wget git
-sudo wget https://raw.githubusercontent.com/bitcoin/bitcoin/master/src/secp256k1/include/secp256k1_recovery.h -O /usr/local/include/secp256k1_recovery.h
```
carver: Can you explain a little more why this was here before, and not necessary anymore?

docs/README-freebsd.md
```diff
@@ -44,7 +61,13 @@ BROKEN (build crashes on FreeBSD 11.2)
 ```
 
 ```
+cd web3.py
+tox
```
carver: Maybe `tox -e py36-core` and others, since it's unreasonable to run the whole suite at once right now.

docs/README-freebsd.md
rayrapetyan: @carver, for pkg install - all packages are still there. For secp256k - I've submitted a fix into secp256k port (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=233456). Basically now secp256k1_recovery.h file will be copied during secp256k package installation as it should be so we don't need to download it manually. Thanks.

docs/README-freebsd.md
rayrapetyan: Well, right... I've just followed "It is important to understand that each Pull Request must pass the full test suite as part of the CI check, hence it is often convenient to have tox installed locally as well." rule from https://py-evm.readthedocs.io/en/latest/contributing.html. So do you want me to update that?

carver: To clarify, I'm suggesting something like:
```
tox -e py36-core
tox -e py36-ens
etc
```

Hm, I don't disagree with anything in that statement (also it's in py-evm docs instead of web3.py). How are you suggesting the text should be updated?

docs/README-freebsd.md
rayrapetyan: Hm, weird... I don't see couincurve as a dependency anywhere in this project. Probably from the very beginning everything related to coincurve\secp256k1 were related to py-evm project only.

So for web3.py secp256k1 installation is not needed.

I've performed all steps on a freshly installed OS image and they passed with no errors (I've double-checked it now again).

docs/README-freebsd.md
rayrapetyan: I've messes up py-evm with web3.py again, sorry... Yes, your suggestion looks ok, I'll update pull request.

docs/README-freebsd.md
carver: For anything involving local signing (`w3.eth.account`), coincurve will make things run notably faster, IIRC. So it's probably still worth having.

docs/examples.rst
```diff
@@ -337,7 +337,7 @@ Fetch various data about the current state of the ERC20 contract.
     >>> erc20.functions.decimals().call()
     0
     >>> erc20.functions.totalSupply().call()
-    2040
+    2044
```
rayrapetyan: @carver, I did this change just to pass doctest checks, not sure what caused that. Please check.

pipermerriam:
fixed for good here: https://github.com/ethereum/web3.py/pull/1159

docs/examples.rst
pipermerriam: @rayrapetyan it's the contract reading from live data (and new unicorn tokens being minted)

carver:
Can you rebase on master @rayrapetyan ?

rayrapetyan:
Not sure this is in desired state, please check. Thanks.

carver:
Thanks!

Issue #1025, opened 2018-08-28
Replace ConciseContract with ContractCaller


dylanjw:
`ConciseContract` shouldn't be a factory, it should be it's own calling API. meta-tracking in #722 

### How can it be fixed?

Add deprecation warning to ConciseContract and ImplicitContract.  Add a new api: ContractReader.

one proposed api: 

```
contract = w3.eth.contract(address, abi=abi)
reader = contract.reader({'from': caller_addr, ...})  # essentially ConciseContract(contract) with sugar
reader.balanceOf(address)  # or whatever call you want to make
```

Note that there will be internal breakage when removing `ConciseContract`, notably in the `ens` module. So part of this change will be updating ENS to use the new reader API. Brief discussion at https://github.com/ethereum/web3.py/pull/1118#discussion_r226800528

Exef:
Should ContractReader have access to ContractEvents? In my opinion - yes, because events are also part of __reading__ contract data from blockchain.

carver:
Hm, I'm not sure we'd be making the events API any easier to call. Maybe name it `ContractCaller` for semantic alignment?

dylanjw:
I like ContractCaller.  So the idea is that ContractCaller can only make calls, and removing the ability to use modifiers to make transactions.

For example, this would not be possible:

```
contract.withdraw(amount, transact={'from': eth.accounts[1], 'gas': 100000, ...})
```

I was confused by the name ContractReader, which I thought could be an api to read public variables.

pipermerriam:
for @kclowes

`contract.reader.getThing()`

carver:
Note that's a slight change from the API in the OP: `contract.reader(transaction_dict).getThing()`. (where `transaction_dict` has fields like `from`, `gas`, `gasPrice`, etc that can affect the function call.

I think I like the OP API better, but highly open to persuasion or alternatives

pipermerriam:
@carver I'm inclined to do something like this.

- `contract.reader.getThing()`
  - equivalent to `contract.functions.getThing().call()`
- contract.reader(transaction_dict).getThing()
  - equivalent to `contract.functions.getThing().call(transaction_dict)`

And maybe to alias `contract.r` to `contract.reader` for convenience.

carver:
The non-parenthetical API option looks good to me. :+1: 

---

One-letter variables would be super out-of-character for the tool suite.

[...]

PR #1157, opened 2018-12-11
Contract caller

kclowes:
### What was wrong?
Want to deprecate ConciseContract and ImplicitContract in favor of a ContractCaller. There are more details in the issue linked below.

Todo:
- [x] better testing
- [x] ENS module breakage
- [x] Update docs

Related to Issue #
[#1025](https://github.com/ethereum/web3.py/issues/1025)

### How was it fixed?
Added a new ContractCaller class and added deprecation methods to ConciseContract and ImplicitContract. 

#### Cute Animal Picture


web3/contract.py
```diff
@@ -958,6 +961,7 @@ class ImplicitContract(ConciseContract):
 
     > contract.functions.withdraw(amount).transact({})
     '''
+    @deprecated_for("Please use the verbose contract syntax. Ex: contract.functions.withdraw(amount).transact({})")
```
carver: The deprecation warning will probably read a little weird here, something like "this method was deprecated for: Please use the verbose...".

Maybe the `str` should just start "verbose contract syntax".

web3/contract.py
```diff
@@ -1328,6 +1332,39 @@ def factory(cls, class_name, **kwargs):
         return PropertyCheckingFactory(class_name, (cls,), kwargs)
 
 
+class CallerMethod:
+    def __init__(self, function, normalizers=None):
+        self._function = function
+
+    def __call__(self, *args, **kwargs):
+        # TODO: Remove this and consolidate into ContractCaller.
+        # I don't know why this is working but it is.
```
carver: :rofl:

web3/contract.py
```diff
@@ -1328,6 +1332,39 @@ def factory(cls, class_name, **kwargs):
         return PropertyCheckingFactory(class_name, (cls,), kwargs)
 
 
+class CallerMethod:
+    def __init__(self, function, normalizers=None):
+        self._function = function
+
+    def __call__(self, *args, **kwargs):
+        # TODO: Remove this and consolidate into ContractCaller.
```
carver: Maybe you'd want it to be a staticmethod in ContractCaller or something. Or it could make sense to keep it as a separate method (doesn't need to be a class). Something like:

```py
def call_contract_function(fn, *args, transaction_dict=None, **kwargs):
    if transaction_dict is None:
        transaction_dict = {}
    return fn(*args, **kwargs).call(transaction_dict)
```

Then later in the `ContractCaller`, something like:
```py
for func in self._functions:
    ...
    caller_method = functools.partial(call_contract_function, fn, transaction_dict=transaction_dict)

File: tests/core/contracts/test_contract_call_interface.py [hunk 4]
```diff
@@ -611,3 +614,23 @@ def test_invalid_fixed_value_reflections(web3, fixed_reflection_contract, functi
...
+    assert result == 13
+
+
+def test_caller_with_parens(math_contract):
+    result = math_contract.caller().return13()
+    assert result == 13
+
+
+def test_caller_with_parens_and_transaction_dict(math_contract):
+    result = math_contract.caller({'from': 'notarealaddress.eth'}).add(2, 3)
```
carver: Another potentially interesting API would be to support sending these transaction fields in by keyword. `from` is problematic since it's a keyword, but we could go the pyethereum route and use `sender`.

Here that would mean:
```py
math_contract.caller(sender='notarealaddress.eth').add(2, 3)
```
which is kind of nice.

web3/contract.py
```diff
@@ -1328,6 +1332,39 @@ def factory(cls, class_name, **kwargs):
...
+        # I don't know why this is working but it is.
+        return self._function(*args, **kwargs).call()
+
+
+class ContractCaller:
+    def __init__(self, abi, web3, address, *args, **kwargs):
```
carver: Instead of `*args, **kwargs` I think it would be sufficient to have a single `transaction_dict=None`.

web3/contract.py
```diff
@@ -1328,6 +1332,39 @@ def factory(cls, class_name, **kwargs):
...
+                            web3=self.web3,
+                            contract_abi=self.abi,
+                            address=self.address,
+                            function_identifier=func['name'])
+
+                caller_method = CallerMethod(fn)
+                setattr(self, func['name'], caller_method)
+
+    def __call__(self, *args, **kwargs):
+        return type(self)(self.abi, self.web3, self.address, *args, **kwargs)
```
carver: This is where the kwargs could be consolidated into `transaction_dict`. I'm not sure what `args` would be here, besides the transaction dict.

Something like this could work:
```py
    def __call__(self, transaction_dict=None, **kwargs):
        assert transaction_dict is None or len(kwargs) == 0  # but not actually an assert
        return type(self)(self.abi, self.web3, self.address, transaction_dict=(transaction_dict or kwargs))

File: tests/core/contracts/test_contract_call_interface.py (hunk 4 above)
Comment (kclowes) [reply tests/core/contracts/test_contract_call_interface.py]: Sounds good to me. So we would support both the `sender` kwarg and a transaction dict with `from` as a key, right?

Comment (pipermerriam) [reply tests/core/contracts/test_contract_call_interface.py]: @carver I'd like to introduce that in a subsequent PR since it would make sense to make it available globally.

```python
web3.eth.sendTransaction(sender=..., gas=...)
contract.functions.transfer(...).transact(sender=...)
contract.caller(sender=...).add(...)
```

web3/contract.py
pipermerriam: IIUC the `**kwargs` portion is to support keyword based transaction parameters?  If so my previous comment stands about holding off on introducing this option into the API.

kclowes: It looks like you can use keyword args for the transaction parameters and `block_identifier`

web3/contract.py
kclowes: I was thinking that you could optionally pass in a plain old dictionary without the `transaction` keyword, in which case I'd need to keep *args. Here is an example: https://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.ContractFunction.call. Am I reading those docs right or does a transaction dict always need the `transaction` keyword in front of it?

carver: I can't dive in right this second to make sure it works, but what I was thinking in general is that you can use the fact that positional args can fill in keyword args, like:

```py
def fn(kw1=None, **kwargs):
  return kw1

assert fn(1) == 1
```

web3/contract.py
carver: I'm not sure I'm keen on having the `block_identifier` be something that you set once in the caller and keep around. At the very least, it makes me less excited about the earlier idea (break up the transaction dict into keywords), because then you have a mess of transaction fields and other stuff like the `block_identifier`.

But... if we don't have the `block_identifier` in the caller, then is it just impossible to set the block identifier using the `ContractCaller`? (That might be okay, maybe the "long form" call is required to do that kind of request)

web3/contract.py
```diff
@@ -1328,6 +1337,45 @@ def factory(cls, class_name, **kwargs):
...
+                    func['name'],
+                    web3=self.web3,
+                    contract_abi=self.abi,
+                    address=self.address,
+                    function_identifier=func['name'])
+
+                caller_method = partial(self.call_function, fn, transaction_dict=transaction_dict)
+                setattr(self, func['name'], caller_method)
+        else:
+            raise NoABIFunctionsFound(
```
pipermerriam: I think this is the culprit.  This needs to be lazy.  See how `ContractFunctions` implements a `__getattr__` which gives it a chance to raise the exception anytime a user tries to access a function on the object.

kclowes: Aha! Thanks. Yeah, this just started causing problems.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -611,3 +611,34 @@ def test_invalid_fixed_value_reflections(web3, fixed_reflection_contract, functi
...
+
+
+def test_caller_with_no_abi(web3):
+    with pytest.raises(NoABIFunctionsFound):
+        contract = web3.eth.contract()
+        contract.caller().thisFunctionDoesNotExist()
+
+
+def test_caller_with_block_identifier(math_contract):
+    result = math_contract.caller(block_identifier=10).return13()
```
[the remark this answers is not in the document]
kclowes: Yeah, that makes sense to me. Is there a place that I can see the fixture contract code? I know I can see the bytecode, but it would be nice to see human-readable code somewhere.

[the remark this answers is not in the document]
pipermerriam: It appears not, but you *should* be able to infer the necessary methods from the existing tests and ABI and write a new one.  When/if you do, lets be sure to include the source code (or better yet, an EthPM package of it.

web3/contract.py
```diff
@@ -1328,6 +1337,54 @@ def factory(cls, class_name, **kwargs):
         return PropertyCheckingFactory(class_name, (cls,), kwargs)
 
 
+class ContractCaller:
+    def __init__(self, abi, web3, address, *args, transaction_dict=None, **kwargs):
+        if abi:
+            self.web3 = web3
```
carver: I guess `ContractCaller` doesn't have much to do if ABI is `None`. Still, it seems pretty surprising to skip assigning `web3` and `address`, any time `abi` is missing.

tests/core/contracts/conftest.py
```diff
@@ -527,6 +527,52 @@ def FallballFunctionContract(web3, FALLBACK_FUNCTION_CONTRACT):
     return web3.eth.contract(**FALLBACK_FUNCTION_CONTRACT)
 
 
+CONTRACT_RETURN_ARGS_SOURCE = """
+contract CallerTester {
+  function returnMeta() public payable returns (address, bytes memory, uint256, uint256) {
```
pipermerriam: you might return block number too which would make it easy to test calls at specific block numbers.

kclowes:
@pipermerriam @carver I think this is pretty close if you want to give it another once-over 🙏. The only outstanding question is: do we want to allow `block_identifier` to be passed in as a kwarg? I don't have a strong opinion. The functionality is there now, but I'm happy to take it out if someone has strong feelings. Also, let me know if you have any feedback on the docs. Thanks!

carver:
Can you rebase away the merge from master so those changes don't show up in the diff?

kclowes:
@carver Since this branch was open so long and a bunch of people were touching the same files, git history got really ugly. I couldn't figure out a graceful and somewhat quick way to rebase away the merge from master. So, I grabbed the relevant commits from here and opened a new branch off of the current master – see #1227. I'm much happier with how the history looks over there. Closing this PR in favor of #1227.

pipermerriam:
Lemme know if you'd like a crash course in git rebasing

PR #1159, opened 2018-12-12
fix unicorn token example flakyness

pipermerriam:
### What was wrong?

The example docs had a fixed value for the total supply of the unicorn tokens.  The doctests end up actually running the code examples.  Anytime more unicorn tokens get minted, the doctests fail.

### How was it fixed?

Changed the assertion be `>` check.

#### Cute Animal Picture

cat_bread_4


docs/examples.rst
```diff
@@ -336,8 +336,8 @@ Fetch various data about the current state of the ERC20 contract.
     '🦄'
     >>> erc20.functions.decimals().call()
     0
-    >>> erc20.functions.totalSupply().call()
-    2040
+    >>> total_supply = assert erc20.functions.totalSupply().call()
+    >>> assert total_supply > 2040
```
kclowes: Ah, brilliant! I think this should be `>=` though? And I think there's an extra `assert` on line 339?

pipermerriam: It's 2044 now I think so `>` should be fine.

kclowes: Ha just saw that :) Looks good to me!

PR #1160, opened 2018-12-13
Remove unreachable code

davesque:
Also, remove unnecessary else clause that tends to obscure that problem.

### What was wrong?

There was an `assert False` statement which was unreachable.  It was clearly expected that this statement was not supposed to be reached in a "normal" situation.  However, I don't think it was realized that the statement could never be reach in *any* situation.

### How was it fixed?

Deleted it.

#### Cute Animal Picture

### DIFF
--- a/web3/providers/base.py
+++ b/web3/providers/base.py
@@ -77,8 +77,8 @@ class JSONBaseProvider(BaseProvider):
             response = self.make_request('web3_clientVersion', [])
         except IOError:
             return False
-        else:
-            assert response['jsonrpc'] == '2.0'
-            assert 'error' not in response
-            return True
-        assert False
+
+        assert response['jsonrpc'] == '2.0'
+        assert 'error' not in response
+
+        return True


carver:
> @carver So is the unreachable assert meant to hint that the function should terminate in one of the if branches? If so, that makes more sense.

Exactly, that's how I read the unreachable assert.

> However, I still do kinda feel like that pattern is confusing. It seems like you'd need to explain it to anyone who encounters it, whereas returning false in the except block or true in the normal function body is clear enough at a glance.

Yeah, this is why I don't feel strongly about it.

> Although it does kinda bother me if the pattern is used elsewhere except here...hmm.

It's not used *a whole lot*, but has become a bit of a pattern. I searched for "unreachable" (which didn't even find this usage, of course), and only found two. I think the explicit message is much better than the blank assert. For example:

https://github.com/ethereum/web3.py/blob/3d14361f451bdbe7fa0537a6ea2073569dd746e1/tests/core/eth-module/test_accounts.py#L81-L85

and even better:

https://github.com/ethereum/web3.py/blob/3d14361f451bdbe7fa0537a6ea2073569dd746e1/web3/_utils/transactions.py#L126-L127

davesque:
Mkay, concerns noted.  Since this is only a few lines of code, I'll go ahead and merge this.

PR #1162, opened 2018-12-13
Add a yaml file for Read the Docs

kclowes:
### What was wrong?
Read the Docs build is failing because it's running Python 3.5.2 and we specify that we need Python 3.5.3 or higher.

### How was it fixed?
Added a .readthedocs.yml file

#### Cute Animal Picture

image7

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,2 @@
+python:
+  version: 3.6
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,2 +1,2 @@
-python:
-  version: 3.6
+build:
+  image: latest


carver:
Ooh right, I forgot about that yaml. I think there's some way to trigger a build on the branch (esp if the branch gets pushed upstream) to test it out.

For whatever reason, we used a slightly different yaml in v4: https://github.com/ethereum/web3.py/blob/df03ed098f32054fecc6bb4d2cfd7cf1a524e4cc/.readthedocs.yaml#L1-L2

I don't remember why we used that approach. You might try to blame it and see if there's any more information in the old PR about what was tried.

carver:
> Do you have a preference between cherry picking that commit or updating the file here?

No need to cherry-pick it, update in place is fine. :+1:

pipermerriam:
This might belong in https://github.com/ethereum/ethereum-python-project-template

Issue #1161, opened 2018-12-13
web3 async discussion continued


dylanjw:
This issue supersedes #1055. 

Ive summarized our discussion re the introduction of an async web3 api below, and expanded it a bit.  After learning from experiments in #1088 we discussed pursuing the following design:

(1) Replace api methods with instances of a Method class, with a middleware like layer for formatting/normalizing/validating method input/output.
  This will allow the procedural formatting code to be reused between async and sync api methods.

(2) Reduce middlewares to only those that make calls to the web3 api, moving the formatting middlewares to the Method input/output formatters.
  Having fewer middlewares reduces maintenance overhead of duplicate async variants.

(3) Separate async and sync execution pathway through RequestManagers -> middlewares -> providers.
  Async code needs to be purely async from the top api entry-point, all the way to the request. Sync code similarly needs to be purely sync. Converting async to sync code by running in a event loop incurred to much overhead.

### First milestone: Minimum async call stack that gets the node version. No middlewares or other special features.

Any comments on this api?

```
import asyncio
from web3.providers import coro

aw3 = web3.Web3(coro.IPCProvider(), async=False) <-- Raises exception 
#  "TypeError, cant use async provider in a synchronous Web3 instance.")

aw3 = web3.Web3(coro.IPCProvider(), async=True)
asyncio.run(aw3.version.node())
```

Im leaning toward sharing the Web3 and RequestManager classes and having an `async` setting that will set the manager request method to the async execution path, and load the default async web3 modules. The providers (and eventually middlewares) will be set by the user, and will be checked to be async compatible.

TODO:
- [x] Add a single async provider (coro.EthereumTesterProvider?).
- [x] add async RequestManager execution pathway: request function -> middleware stack -> provider.
- [x] Design and implement Method classes. 

    I propose extracting the method configuration and method selection into a "MethodManager" object that will return either the async or sync callable. ~A modified \_\_getattr\_\_ will hand over control to the method_manager.~ Modules will simply compose the right pieces:

```
class AsyncVersion(Module, Meta=MetaModule):
    method_caller = AsyncMethod
    method_manager = MethodManager(VERSION_METHOD_CONFIG)
```

I may have to use a meta class to change \_\_getattr\_\_ for the class.

```
class MetaModule(type):
    @property
    def method_caller():
        raise NotImplementedError
    @property
    def method_manager():
        raise NotImplementedError
    def __getattr__(cls, attr):
        if attr not in cls.method_config:
            raise AttributeError
[...]

PR #1166, opened 2018-12-15
outline for async web3 api: implement async Version module

dylanjw:
Related to Issue #1161.

Minimal web3 async api.  Includes:
  - `AsyncEthereumTesterProvider` was added, which has eth-tester related middlewares removed, with a request function that doesnt await.
  - `request_async` was updated in RequestManager to be a coroutine. code copy of the blocking request. Im preferring simple over dry for this push.
- Added a framework for new web3 module apis which comprises of:  
    - `AsyncMethod` and `BlockingMethod` callable classes (can be configured for property access)
    - Rough method configuration format (json for now).
    - ~`web3.Module` manages the "attaching" of method callables to the module, via a `method_config` and `method_class`.~ EDIT: methods are attached via a descriptor now.
    - Demo of the changes using web3.version.Version module.
    - Not built-in to the web3 object. e.g.:

```python
from web3 import Web3
from web3.providers.tester.main import AsyncEthereumTesterProvider
from web3.version import AsyncVersion

w3 = Web3(
    AsyncEthereumTesterProvider, 
    middlewares=[], 
    modules={"async_version": AsyncVersion})
```

#### Cute Animal Picture
image

### DIFF
--- a/web3/manager.py
+++ b/web3/manager.py
@@ -1,14 +1,14 @@
 import logging
 import uuid
 
+from web3._utils.decorators import (
+    deprecated_for,
+)
 from web3._utils.threads import (
     spawn,
 )
 from web3.datastructures import (
     NamedElementOnion,
-)
-from web3.exceptions import (
-    UnhandledRequest,
 )
 from web3.middleware import (
     abi_middleware,
@@ -74,19 +74,18 @@ class RequestManager:
     # Provider requests and response
     #
     def _make_request(self, method, params):
-        if self.provider:
-            request_func = self.provider.request_func(self.web3, tuple(self.middleware_onion))
-            self.logger.debug("Making request. Method: %s", method)
-            return request_func(method, params)
-        else:
-            raise UnhandledRequest(
-                "No provider available to respond to the RPC request:\n"
-                "method:{0}\n"
-                "params:{1}\n".format(
-                    method,
-                    params,
-                )
-            )
+        request_func = self.provider.request_func(
+            self.web3,
+            tuple(self.middleware_onion))
+        self.logger.debug("Making request. Method: %s", method)
+        return request_func(method, params)
+
+    async def _coro_make_request(self, method, params):
+        request_func = self.provider.request_func(
+            self.web3,
+            tuple(self.middleware_onion))
+        self.logger.debug("Making request. Method: %s", method)
+        return await request_func(method, params)
 
     def request_blocking(self, method, params):
         """
@@ -99,6 +98,18 @@ class RequestManager:
 
         return response['result']
 
+    async def coro_request(self, method, params):
+        """
+        Couroutine for making a request using the provider
+        """
+        response = await self._coro_make_request(method, params)
+
+        if "error" in response:
+            raise ValueError(response["error"])
+
+        return response['result']
+
+    @deprecated_for("coro_request")
     def request_async(self, raw_method, raw_params):
         request_id = uuid.uuid4()
         self.pending_requests[request_id] = spawn(
--- a/web3/method.py
+++ b/web3/method.py
@@ -1,0 +1,148 @@
+import functools
+
+from eth_utils import (
+    to_tuple,
+)
+from eth_utils.toolz import (
+    identity,
+    pipe,
+)
+
+
+def _munger_star_apply(fn):
+    @functools.wraps(fn)
+    def inner(args):
+        return fn(*args)
+    return inner
+
+
+def get_default_formatters(*args, **kwargs):
+    return ([identity], [identity],)
+
+
+def default_munger(module, *args, **kwargs):
+    if not args and not kwargs:
+        return tuple()
+    else:
+        raise TypeError("Parameters passed to method without parameter mungers defined.")
+
+
+class Method:
+    """Method object for web3 module methods
+
+    Calls to the Method go through these steps:
+
+    1. input munging - includes normalization, parameter checking, early parameter
+    formatting.  Any processing on the input parameters that need to happen before
+    json_rpc method string selection occurs.
+
+            A note about mungers: The first (root) munger should reflect the desired
+        api function arguments. In other words, if the api function wants to
+        behave as: `getBalance(account, block_identifier=None)`, the root munger
+        should accept these same arguments, with the addition of the module as
+        the first argument e.g.:
+
+        ```
+        def getBalance_root_munger(module, account, block_identifier=None):
+            if block_identifier is None:
+                block_identifier = DEFAULT_BLOCK
+            return module, [account, block_identifier]
+        ```
+
+        all mungers should return an argument list.
+
+        if no munger is provided, a default munger expecting no method arguments
+        will be used.
+
+    2. method selection - The json_rpc_method argument can be method string or a
+    function that returns a method string. If a callable is provided the processed
+    method inputs are passed to the method selection function, and the returned
+    method string is used.
+
+    3. request and response formatters are retrieved - formatters are retrieved
+    using the json rpc method string. The lookup function provided by the
+    formatter_lookup_fn configuration is passed the method string and is
+    expected to return a 2-tuple of lists containing the
+    request_formatters and response_formatters in that order.
+    e.g. ([*request_formatters], [*response_formatters]).
+
+    4. After the parameter processing from steps 1-3 the request is made using
+    the calling function returned by the module attribute ``retrieve_caller_fn``
+    and the reponse formatters are applied to the output.
+    """
+    def __init__(
+            self,
+            json_rpc_method=None,
+            mungers=None,
+            formatter_lookup_fn=None,
+            web3=None):
+
+        self.json_rpc_method = json_rpc_method
+        self.mungers = mungers or [default_munger]
+        self.formatter_lookup_fn = formatter_lookup_fn or get_default_formatters
+
+    def __get__(self, obj=None, obj_type=None):
+        if obj is None:
+            raise TypeError(
+                "Direct calls to methods are not supported. "
+                "Methods must be called from an module instance, "
+                "usually attached to a web3 instance.")
+        return obj.retrieve_caller_fn(self)
+
+    @property
+    def method_selector_fn(self):
+        """Gets the method selector from the config.
+        """
+        if callable(self.json_rpc_method):
+            return self.json_rpc_method
+        elif isinstance(self.json_rpc_method, (str,)):
+            return lambda *_: self.json_rpc_method
+        raise ValueError("``json_rpc_method`` config invalid.  May be a string or function")
+
+    def get_formatters(self, method_string):
+        """Lookup the request formatters for the rpc_method
+
+        The lookup_fn output is expected to be a 2 length tuple of lists of
+        the request and output formatters, respectively.
+        """
+        formatters = self.formatter_lookup_fn(method_string)
+        return formatters or get_default_formatters()
+
+    def input_munger(self, val):
+        try:
+            module, args, kwargs = val
+        except TypeError:
+            raise ValueError("input_munger expects a 3-tuple")
+
+        # TODO: Create friendly error output.
+        mungers_iter = iter(self.mungers)
+        root_munger = next(mungers_iter)
+        munged_inputs = pipe(
+            root_munger(module, *args, **kwargs),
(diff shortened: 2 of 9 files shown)


web3/method.py
```diff
@@ -0,0 +1,144 @@
...
+
+from web3.exceptions import (
+    UndefinedMethodError,
+)
+
+TEST_METHOD_CONFIG = {
+    'name': "signInBlood",
+    'mungers': [],
+    'json_rpc_method': "eth_signTransactionInBlood",
+}
```
dylanjw: meant to move this and the test below.  Im designating the functions that the python method arguments get piped through, that should output a param dict as the "mungers".  Havent found a name I like better.  Because these functions can be validators, normalizers, formatters, occur in any order, with probably a huge variation arity, etc., "munger" seems appropriate.  These are distinct from the request parameter and result formatters, taken from the middleware formatters.

web3/method.py
```diff
@@ -0,0 +1,144 @@
...
+
+    e.g.:
+
+        >>> pipe_appends([lambda x: x**2, lambda x: x*10], 5)
+        (25, 250)
+
+    """
+    for fn in fns:
+        val = fn(val)
+        yield val
```
dylanjw: move to utils

web3/module.py
```diff
@@ -1,8 +1,38 @@
...
+
+    def __getattr__(self, attr):
+        #  Method lookup magic
+        try:
+            method = self.lookup_method_fn(
+                self.module_config,
+                self.method_class,
+                attr)
+        except UndefinedMethodError:
+            raise AttributeError
```
dylanjw: Im wondering if it is worth caching the method object after the first lookup.

web3/manager.py
```diff
@@ -102,6 +98,26 @@ def _make_request(self, method, params):
                 )
             )
 
+    async def _make_async_request(self, method, params):
+        """TODO: Can this be made dry? Going for dumb solution for now.
+        """
+        for provider in self.providers:
```
pipermerriam: We *should* be able to drop this loop in `v5` by removing the multi-provider API.  cc @kclowes

web3/manager.py
```diff
@@ -102,6 +98,26 @@ def _make_request(self, method, params):
                 )
             )
 
+    async def _make_async_request(self, method, params):
```
pipermerriam: Naming convention for this type of thing has been to use a `coro_` prefix so maybe `_coro_make_request`.

web3/method.py
pipermerriam: I'm good with "mungers" unless something better surfaces.  Maybe good to try and find a good place to write up a nice comment documenting the `validator/normalizer/formatter` concepts it encapsulates

web3/method.py
```diff
@@ -0,0 +1,144 @@
...
+    3. constructing formatter middlewares - takes the rpc_method and looks up the
+    corresponding input/output formatters. these are the middlewares migrated here.
+
+    4. making the request through the middleware (pipeline)? wrapped request
+    function.
+    """
+    def __init__(self, web3, method_config):
+        self.__name__ = method_config.get('name', 'anonymous')
+        self.__doc__ = method_config.get('doc', '')
+        self.is_property = method_config.get('is_property', False)
```
pipermerriam: What do you think about removing this and only implementing the concept of direct `property` lookups at the `Module` level?

```python
class VersionModule:
    get_version = VersionMethod(...)

    @property
    def version(self):
        return self.get_version()
```

Would work similarly for `async`

web3/method.py
```diff
@@ -0,0 +1,144 @@
...
+            [self.input_munger, self.method_selector, self.get_formatters],
+            (args, kwargs,))
+        return pipe((method, params,), *req_formatters), ret_formatters
+
+    def __call__(self):
+        raise NotImplementedError()
+
+
+@to_tuple
+def pipe_appends(fns, val):
```
pipermerriam: Interesting, it's kind of like https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.accumulate combined with `pipe`.  Maybe re-use the name and call this `accumulate_pipe` or `pipe_and_accumulate`?

web3/module.py
```diff
@@ -1,8 +1,38 @@
...
+    method_class = None
+    module_config = None
+    lookup_method_fn = lookup_method
 
     def __init__(self, web3):
         self.web3 = web3
+        if self.module_config is None:
+            self.module_config = dict()
+
+    def __getattr__(self, attr):
```
pipermerriam: I'm :-1: on using `__getattr__`.  It makes introspection and type hinting not work well.

I'd advocate for something like:

```python
class Module:
    get_version: GetVersionMethod  # this is how we get strong-ish type hints.

    def __init__(self, ...):
        self.get_version = GetVersion(...)
```

In theory we can do this by assigning them at the class level and using the descriptor API to get access to the `w3` instance.

dylanjw:
I want to restate the goals for this pull request now that it has been worked through further.  It is basically restating the opening description, but with more clarity on my part.


This PRs primary intention is to implement the Method objects and the strategy for sync and async execution.  More specifically it aims to determine the association between the modules, method definitions and async and sync calling functions.

The following is a description of the structure ive chosen:

The `Method` class is in charge of interpreting a method configuration. Meaning the Method is concerned with collecting the method lookup, mungers, formatters and executing them on the inputs.

In this first version the method configuration is a dict, but I see these becoming a small class that can validate and document the configuration requirements.

The Method class has a secondary concern with directing the calls to the method to a delegated call function. The Method class defines `__get__` method, becoming a descriptor.  This overrides attribute access from the module with `__get__` which is passed the object or the object class.  In this way **the module is able to define the caller** function delegate, rather than in the Method object.  This saves having to instantiate the Method objects for both async and sync modules.

The method does not retain any state from the module object.  The `__get__` method just  passes itself (the Method object) and the module web3 instance to the call function delegate.  The method object provides the input processing and the web3 instance provides the manager request function. 

Using this structure the Method objects can be defined once in a shared base class. To make this safe, I need to find a way to protect the Method class from mutations after its instantiation with the method config.

Items left todo:

- [ ] Make the configured method objects immutable.
- [ ] Cleanup tests for the configuring the Method class and input processing.

dylanjw:
@pipermerriam @kclowes @carver This is ready for review.

For the next PR I plan on working on another module that requires more input processing and formatters taken from middlewares.

carver:
Sorry, with the holidays, it will take a while before it is reviewed. Maybe early January (for me at least).

dylanjw:
NP.  In the meantime I will start on another PR to work out how migrating the middleware formatters will look, and backmerge any changes from this PR.

web3/method.py
```diff
@@ -0,0 +1,143 @@
...
+    _config = None
+
+    def __init__(self, method_config, web3=None):
+        super().__setattr__('_config', dict(method_config))
+
+    def __get__(self, obj=None, objType=None):
+        return obj.retrieve_caller_fn(self)
+
+    def __setattr__(self, key, val):
+        raise TypeError('Method cannot be modified after instantiation')
```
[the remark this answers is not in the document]
carver: As long as it's not trying to make a security guarantee, it's simply nudging users toward ideal behavior, then I'm more open to this kind of thing.

dylanjw:
Ive added changes to give the mungers access to the module instance, to support certain features, like the default account api.

I didnt want to expose the entire web3 instance to the mungers, so I I created a new Module class to eventually replace it (ModuleV2) that isolates the web3 instance to the method calling functions, rather than exposing a web3 instance to the module. The method callers pass the module instance to the method input processing.

Organizing it this way makes it hard to write "re-entrant" calls at the module level. If I come across any web3 calls in the modules as Im working through the conversion, they will be moved to a middleware, where they can be written for both blocking and async execution pathways.

web3/method.py
```diff
@@ -9,20 +9,21 @@
 )
 
 
-def star_apply(fn):
+def _munger_star_apply(fn):
     @functools.wraps(fn)
-    def inner(args):
-        return fn(*args)
+    def inner(module_args):
+        module, args = module_args
```
pipermerriam: The current API of having all the mungers return the `module` and then pass it onto the next seems problematic/error prone.  Instead of requiring the munger to both be accepted as an argument as well I think we should do the following.

1. only pass as argument.
2. use `curry` or `functools.partial` to set the `module` parameter on all the munger functions before passing the values through.

This would make the `map(star_apply, mungers)` into something like:

```python
map(lambda munger: star_apply(functools.partial(munger, module)), mungers)
```

web3/method.py
```diff
@@ -9,20 +9,21 @@
...
 
 def get_default_formatters(*args, **kwargs):
     return ([identity], [identity],)
 
 
-def default_munger(*args, **kwargs):
+def default_munger(module, *args, **kwargs):
     if not args and not kwargs:
-        return list()
+        return module, list()
```
pipermerriam: Is the `list()` intentional here or should this be converted to `tuple`

web3/version.py
```diff
@@ -1,8 +1,46 @@
...
 
 
+class BaseVersion(ModuleV2):
+    retrieve_caller_fn = None
+
+    _get_node_version = Method('web3_clientVersion')
+    _get_protocol_version = Method('eth_protocolVersion')
+
+    @property
+    def api(self):
```
pipermerriam: I think I'm :+1: on having a mixed-bag of coroutine and normal properties.  Forcing simple APIs like this that don't require I/O to be coroutines makes them more complex than necessary **and** adds un-necessary performance overhead.  :+1:

web3/manager.py
```diff
@@ -74,19 +74,18 @@ def default_middlewares(web3):
...
-                "params:{1}\n".format(
-                    method,
-                    params,
-                )
-            )
+        request_func = self.provider.request_func(
+            self.web3,
+            tuple(self.middleware_stack))
+        self.logger.debug("Making request. Method: %s", method)
+        return request_func(method, params)
```
dylanjw: @kclowes @njgheorghita
Before I merge this, it would be good to get a couple sets of eyes on this.  Its the only place in this PR that touches the critical api.

With no longer iterating over the provider list, I think the UnhandledRequest exception is unnecessary and we can let the CannotHandleRequest pass through.   Or maybe this should be catching the "CannotHandleRequest" error like before?

For reference here is the request function when it had the provider list so you dont have to dig through the history:

```
    #   
    # Provider requests and response
    #   
    def _make_request(self, method, params):
        for provider in self.providers:
            request_func = provider.request_func(self.web3, tuple(self.middleware_stack))
            self.logger.debug("Making request. Method: %s", method)
            try:
                return request_func(method, params)
            except CannotHandleRequest:
                continue
        else:
            raise UnhandledRequest(
                "No providers responded to the RPC request:\n"
                "method:{0}\n"
                "params:{1}\n".format(
                    method,
                    params,
                )   
            )   
```

njgheorghita: From what I understand, it looks good to me. I'm thinking it's fine to let the `CannotHandleRequest` exception pass through from the provider rather than catching it here

njgheorghita: Though, maybe beefing up the msg [here](https://github.com/ethereum/web3.py/blob/f7cfd4e0f56500913c14ee50f40034b0f2e61bb4/web3/providers/auto.py#L78) to have more info like in the `UnhandledRequest` msg from before could be useful

pipermerriam: Looks correct.

pipermerriam: If `CannotHandleRequest` is no longer used anywhere it can be removed (didn't look to see if you already did this)

kclowes: Yeah, I agree with @njgheorghita. `CannotHandleRequest` makes sense to me, especially with a better error message!

adichen3798:
Hey, does web3 currently support async? How do I initiate HTTPProvider Web3 client in an async-matter?

kclowes:
@adichen3798 It does not yet support async, but the goal is to get something out by the end of the quarter!

Nighty13:
Hello , still no support for async?

cosmikwolf:
looks like async web3 is in development here: https://github.com/guanqun/async-web3.py

kclowes:
Nice! Async is a work in progress for us but is actively being worked on. [This issue](https://github.com/ethereum/web3.py/issues/1413) is the most up to date if you're interested in following along.

Issue #782, opened 2018-04-23
Convert an AttributeDict with bytes values to a json string


githubcyc:
* Version: 4.2.1

### What was wrong?

`w3.eth.getTransaction()` returns an `AttributeDict`, which I want to send as a string, say over HTTP. I couldn't find a convenient way to make this conversion.

### How can it be fixed?

A new method that converts an `AttributeDict` to a json string:

```py
txn = w3.eth.getTransaction(txn_hash)
txn_json = Web3.toJson(txn)
```

voith:
try this:
```python
import json

from hexbytes import HexBytes

class HexJsonEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, HexBytes):
            return obj.hex()
        return super().default(obj)


tx = web3.eth.getTransaction(transaction_hash)
tx_dict = dict(tx)
json_tx = json.dumps(tx_dict, cls=HexJsonEncoder)
```

carver:
Seems like a custom JSON encoder could reasonably live in Web3.py. If we do it right, we could even replace some of the middleware with it.

Something like:
```py
tx = web3.eth.getTransaction(transaction_hash)
tx_json = json.dumps(tx, cls=Web3.JsonEncoder)
```

or:

```py
tx = web3.eth.getTransaction(transaction_hash)
tx_json = Web3.toJson(tx)
```
which would do something like the first thing under the hood.

pipermerriam:
👍 to `Web3.toJson(...)`

pipermerriam:
@carver we could also add  `AttrDict.to_json()` to the actual `AttrDict` class.

carver:
> we could also add AttrDict.to_json() to the actual AttrDict class.

Eh, I don't like the idea that `AttributeDict` would know anything about `HexBytes`.

vindard:
I came across this issue as well a while back. I solved it a bit differently though:
```
def toDict(dictToParse):
    # convert any 'AttributeDict' type found to 'dict'
    parsedDict = dict(dictToParse)
    for key, val in parsedDict.items():
        # check for nested dict structures to iterate through
        if  'dict' in str(type(val)).lower():
            parsedDict[key] = toDict(val)
        # convert 'HexBytes' type to 'str'
        elif 'HexBytes' in str(type(val)):
            parsedDict[key] = val.hex()
    return parsedDict
```
[...]

PR #1173, opened 2018-12-21
Adds Web3.toJSON per #782

mikeshultz:
### What was wrong?

Related to Issue #782

### How was it fixed?

Added `Web3.toJSON()` and accompanying tests.  Basically, this extends the standard `json.JSONEncoder` to support `HexBytes` and `web3.datastructures.AttributeDict` and adds a static method to the `Web3` object that utilizes it.

This is my first PR for this project, so please tear it apart. 

It would probably be good to get some more test cases, as well.  In what other ways might a user use `Web3.toJSON()`?  Should there be more supported types?  In what ways should it *not* work?

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/encoding.py
+++ b/web3/_utils/encoding.py
@@ -2,6 +2,8 @@
 import json
 import re
 
+from hexbytes import HexBytes
+
 from eth_abi.encoding import (
     BaseArrayEncoder,
 )
@@ -19,6 +21,8 @@ from eth_utils import (
     remove_0x_prefix,
     to_hex,
 )
+
+from web3.datastructures import AttributeDict
 
 from web3._utils.abi import (
     is_address_type,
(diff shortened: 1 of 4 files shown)


web3/_utils/encoding.py
```diff
@@ -309,3 +313,19 @@ def encode_single_packed(_type, value):
         return codecs.encode(value, 'utf8')
     elif abi_type.base == "bytes":
         return value
+
+
+class Web3JsonEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, AttributeDict):
+            return {k: v for k, v in obj.items()}
```
pipermerriam: This seems like it should be a recursive operation, re-encoding each of the `k` and `v` values using `json.JSONEncoder.default(self, v)`

pipermerriam:
Looks like there is also a linting error.  See the CI failure for details.

web3/_utils/encoding.py
mikeshultz: Looks like `json.JSONEncoder` already handles recursion.  `default()` just needs to provide one to one conversion.  I added a test with a nested `AttributeDict` as verification.

mikeshultz:
Think I addressed @pipermerriam's feedback and fixed the linting issues.  Added simple docs as well.

kclowes:
Thank you!

PR #1176, opened 2018-12-28
Wait for transaction receipt does not wait for blockhash to exist

nicolashardy:
### What was wrong?

wait for transaction receipt does not wait for blockhash to exist

### How was it fixed?

added verification that 'blockHash' is not None

#### Cute Animal Picture


nicolashardy:
(Thanks to @pipermerriam for adding an explicit comment)

Issue #1170, opened 2018-12-19
Write "guide" in docs for ERC20 token interactions.


pipermerriam:
### What was wrong?

Interactions with ERC20 tokens are a very common use case.  Examples in the documentation would likely be a valuable resource.

There is an existing minimalistic guide here: https://github.com/ethereum/web3.py/blob/master/docs/examples.rst that only deals with read-only operations.

### How can it be fixed?

Write a "guide" for interacting with an ERC20 token.  It should probably include as much of the following as is viable.

- creation of the contract factory from the ERC20 ABI
- instantiation of a contract instance pointed at a specific token address
- querying account balances
- performing a direct `transfer(...)` call
- querying the approved spending balance for external transfers
- creating an approval for external transfers
- performing a `transferFrom` call with an approved transfer.

And some additional things:

- The guide should *probably* operate on a local test chain and include basic deployment of an simple ERC20 token contract.  
- Guide should ignore key management, assuming that the node is doing the signing.
- Guide should be fully doctested using a web3 instance attached to an in-memory testing chain.

carver:
Also worth including, monitoring transfer events, as requested in #1171.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 250.0 DAI (250.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/1170/2057).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/web3.py/1170/2057).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $51,847.97 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


[...]

PR #1178, opened 2019-01-01
Add addition guides

iamonuwa:
### What was wrong?
Added additional code examples to the docs

Related to Issue #
#1170 

#### Cute Animal Picture


iamonuwa:
@stefanmendoza understood. I'll make changes and update

iamonuwa:
@pipermerriam thanks for the review. I'll update it

docs/examples.rst
```diff
@@ -362,3 +362,74 @@ Calculate the token count from the token balance of an account address.
...
...t":true,"inputs":[{"name":"_owner","type":"address"}],"name":"balanceOf","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"symbol","outputs":[{"name":"","type":"string"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"name":"_to","type":"address"},{"name":"_value","type":"uint256"}],"name":"transfer","outputs":[{"name":"","type":"bool"}],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[{"name":"_owner","type":"address"},{"name":"_spender","type":"address"}],"name":"allowance","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"anonymous":false,"inputs":[{"indexed":true,"name":"_from","type":"address"},{"indexed":true,"name":"_to","type":"address"},{"indexed":false,"name":"_value","type":"uint256"}],"name":"Transfer","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"_owner","type":"address"},{"indexed":true,"name":"_spender","type":"address"},{"indexed":false,"name":"_value","type":"uint256"}],"name":"Approval","type":"event"}]')  # noqa: 501
```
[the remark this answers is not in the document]
iamonuwa: Can you elaborate more on this? An example will be much helpful here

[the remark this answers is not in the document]
pipermerriam: suppose I had an encoded JSON string which represents the ABI.

```python
abi = "<some-encoded-JSON-string-of-the-ABI>"
w3 = Web3(...)

# these two statements are equivalent
contract = w3.eth.contract(abi=abi, ...)
contract = w3.eth.contract(abi=json.loads(abi))
```

When the `w3.eth.contract` API is given a string for the `abi` parameter, it will do the JSON parsing internally.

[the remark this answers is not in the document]
iamonuwa: Do you mean I should not have the long test of ABI? Replace it with randon string?

docs/examples.rst
```diff
@@ -362,3 +362,82 @@ Calculate the token count from the token balance of an account address.
...
 Token count *may* be a integer or fraction in the currency's primary unit (equivalent to 'eth' for ether). Token count is typically displayed to the user on the front-end since it is more readable.
+
+
+Creation of the contract factory from the ERC20 ABI
+---------------------------------------------------
+Using this command, you can create a contract factory from the ERC20 ABI. You need to specify the contract address
+
+.. testsetup::
+    import os
+    os.environ['INFURA_API_KEY'] = "b2679bb624354d1b9a2586154651b51f"
```
pipermerriam: A step in the right direction, but I'd like to push this a little further.

Right now, as I understand it, some of the examples below are reading token balances from the live [Unicorn Token contract](https://etherscan.io/address/0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7).  This is a problem because our tests can/will fail when either the HTTP requests to infura fail, infura's API changes, this API key is revoked, or the data in the contract itself changes.

To fix this, we need to update this code to instead only read data from a local in-memory EVM, aka, the `EthereumTesterProvider`.

```
.. testsetup::

    from web3 import Web3
    w3 = Web3(Web3.EthereumTesterProvider())
    bytecode = "<the-deployment-bytecode-for-a-standard-ERC20-contract>"
    abi = "<the-ERC20-ABI>"
    factory = w3.eth.contract(abi=abi, bytecode=bytecode)
    deploy_txn = factory.constructor(...).transact()
    deploy_receipt = w3.eth.waitForTransactionReceipt(deploy_txn)
    contract_address = deploy_receipt['contractAddress']
```

This will give you an in-memory ERC20 token contract to work with.  You probably need to add a few extra bits to populate some account balances for use in the examples below.  Then, in the examples below, you can just use that contract address and only use the local.

Please take a look at the doctest API here: http://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

These directives have a `[group]` parameter that can be used to allow you to apply a single `testsetup` block to multiple doctests.

iamonuwa: @pipermerriam the EthereumTesterProvider and EthereumTester methods don't populate accounts for use within the blocks. And ideas?

iamonuwa:
@pipermerriam PTAL.

docs/examples.rst
```diff
@@ -362,3 +362,77 @@ Calculate the token count from the token balance of an account address.
...
+    b'\xe9K\xcfz\xfc9*-+\x17|\xd4at.\xe32t\x13\xf4\\\xe8\x8b\x9e6?\x08\xa1\xb4\xe6[\xd3'
+    >>> txn_receipt = web3.eth.waitForTransactionReceipt(tx_hash)
+
+
+Instantiate a contract pointing at a specific address
+-----------------------------------------------------
+
+.. doctest::
+
+    >>> new_contract = web3.eth.contract(address='0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7', abi=json.loads(abi))
```
pipermerriam: @iamonuwa one more, can you update all of these variables from `web3 -> w3` to stick with our convention which helps mitigate confusion between 1) the main module name and 2) an instance of the `Web3` instance.

docs/examples.rst
```diff
@@ -362,3 +362,77 @@ Calculate the token count from the token balance of an account address.
...
+    >>> new_contract = web3.eth.contract(address=web3.eth.accounts[0], abi=json.loads(abi))
+    >>> txn_hash = new_contract.functions.approve(web3.eth.accounts[0], 12345).transact()
+    >>> txn_receipt = web3.eth.waitForTransactionReceipt(txn_hash)
+
+Performing a transferFrom call with an approved transfer.
+---------------------------------------------------------
+
+.. doctest::
+
+    >>> new_contract = web3.eth.contract(address=web3.eth.accounts[0], abi=json.loads(abi))
```
pipermerriam: Maybe instead of using `web3.eth.accounts[0]` we could do something like this.

```python
alice, bob = web3.eth.accounts[:2]
new_contract.functions.transferFrom(alice, bob, 1234)
```

I think it reads a lot cleaner.

docs/examples.rst
```diff
@@ -362,3 +362,77 @@ Calculate the token count from the token balance of an account address.
...
+
+    >>> new_contract = web3.eth.contract(address='0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7', abi=json.loads(abi))
+    >>> balance = new_contract.functions.balanceOf('0x89205A3A3b2A69De6Dbf7f01ED13B2108B2c43e7')
+
+Query account balance
+---------------------
+
+.. doctest::
+
+    >>> new_contract = web3.eth.contract(address=web3.eth.accounts[0], abi=json.loads(abi))
```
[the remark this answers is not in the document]
iamonuwa: Do you suggest using an actual address here?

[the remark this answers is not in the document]
pipermerriam: Yes, the one from the deployment you do in the `.. testsetup` section.

docs/examples.rst
iamonuwa: Because of the infura, w3 returns empty accounts

pipermerriam: Right, but we shouldn't be using infura anymore now that things should be wired up to the `EthereumTesterProvider`

pipermerriam:
@iamonuwa I ended up re-working a lot of this to remove any use of infura from the docs and to clean up your examples a bit.  Just wanted to point it out to you so you can see it with a slightly cleaner structure.

Issue #1179, opened 2019-01-02
hypothesis used in package code but is only a dev extra, not an installation requirement


feuGeneA:
* Version: 5.0.0-alpha.2
* Python: 3.6
* OS: linux

### What was wrong?

Seems that just importing `utils/datatypes` requires you to have the `hypothesis` package:

```
ImportError while importing test module '/home/circleci/repo/python-packages/order_utils/test/test_asset_data_utils.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_asset_data_utils.py:5: in <module>
    from zero_ex.order_utils.asset_data_utils import (
src/zero_ex/order_utils/__init__.py:25: in <module>
    from web3.utils import datatypes
/usr/local/lib/python3.7/site-packages/web3/utils/__init__.py:1: in <module>
    from web3._utils import (  # noqa: F401
/usr/local/lib/python3.7/site-packages/web3/_utils/hypothesis.py:1: in <module>
    from hypothesis import (
E   ModuleNotFoundError: No module named 'hypothesis'
```

But the dependency is only listed in `setup.py` under `extras_require['dev']`.

### How can it be fixed?

Move the `hypothesis` dependency from `extras_require` to `install_requires`.

---

If my analysis is correct here, let me know and I'll be happy to submit a PR.

pipermerriam:
Ah, a side-effect of our move of `web3.utils -> web3._utils`!

I believe the appropriate fix for this would be to move `web3._utils.hypothesis` to live under `web3.tools.hypothesis`.

Ryan-Gordon:
Seen this issue and figured I would take a crack at it. 

Moved web3._utils.hypothesis to live under web3.tools.hypothesis per your request @pipermerriam and also updated the __init__.py to reflect the change.

I can now successfully import `from web3.utils import datatypes`

@feuGeneA how were you building the web3 package ? 
I tried 3 ways from source : 

+ `pip install .` -- gathered from docs, error remains as hypothesis isin't installed

However 
+ `pip install ".[tester]"`
and
+ `pip install ".[dev]"`

Work as they both trigger the `extras_require` which installs hypothesis 

Let me know if theres any issues

Ryan-Gordon:
All CI jobs passing 🙌

kclowes:
Old issue cleanup - I believe this was fixed in v5.0.0a8.

PR #1184, opened 2019-01-05
#1179 Moved hypothesis to web3.tools

Ryan-Gordon:
Related to #1179

Confirmed working when running either
pip install ".[tester]"
or
pip install ".[dev]"

### What was wrong?
hypothesis was giving a module not found error as reported in related issue.
Related to Issue #1179 

@pipermerriam recommended moving the hypothesis file to `web3.tools.hypothesis`
### How was it fixed?
Moved hypothesis package and verified I can import `from web3.utils import datatypes` which was described as what caused the issue 

#### Cute Animal Picture


Ryan-Gordon:
All 26 checks passing 👍 
After a battle with isort I have the lint ci job passing

If this is good to merge, might want to squash the commits into 1

web3/utils/__init__.py
```diff
@@ -25,3 +24,6 @@
     transactions,
     validation,
 )
+from web3.tools import (  # noqa: F401
```
pipermerriam: We've been establishing the `library.tools` module as a place that non-essential library functionality goes, and the general rule is that we should never import from `library.tools` anywhere outside of that module.

Instead of adding this import, can you update all of the places that use this in the tests to import from the new location.

Ryan-Gordon: I think I get you, so I see ~7 places this is imported, ill be updating those to `web3.tool.hypothesis` rather than just `hypothesis` so its no longer needed here, right? 
I can do that 👍

web3/tools/hypothesis.py
```diff
@@ -1,4 +1,4 @@
-from hypothesis import (
+from web3.tools.hypothesis import (
```
pipermerriam: This import looks broken

tests/core/utilities/test_math.py
```diff
@@ -1,16 +1,15 @@
...
-    strategies as st,
-)
-
 from web3._utils.math import (
     percentile,
 )
 from web3.exceptions import (
     InsufficientData,
 )
+from web3.tools.hypothesis import (
```
pipermerriam: This import looks broken.

tests/core/utilities/test_event_filter_builder.py
```diff
@@ -1,15 +1,14 @@
...
-    given,
-    strategies as st,
-)
-
 from web3._utils.events import (
     DataArgumentFilter,
     TopicArgumentFilter,
     normalize_topic_list,
 )
+from web3.tools.hypothesis import (
```
pipermerriam: This import looks broken.

tests/core/utilities/test_encoding.py
```diff
@@ -28,7 +23,10 @@
     JSONBaseProvider,
 )
 from web3.tools.hypothesis import (
+    example,
+    given,
     hexstr_strategy,
+    strategies as st,
```
pipermerriam: `given`, `example` and `strategies` should be imported directly from the `hypothesis` library.

Ryan-Gordon: I think only 4 packages should from hypothesis are used 
So in this case only `hexer_strategy` should be taken from web3.tools.hypothesis

Ryan-Gordon:
Yes, you are right, a find and replace which didn't work out it seems I can fix though 👍

tests/core/utilities/test_event_filter_builder.py
Ryan-Gordon: Broken in the sense it looks off/should be done a different way ? or non functional ? all tests passing

pipermerriam: Ignore this one... commented on the wrong thing.

tests/core/utilities/test_math.py
Ryan-Gordon: I renamed this to be just hypothesis and import from there as the two imports are `given` and `strategies` which you said elsewhere should only be imported from hypothesis

web3/tools/hypothesis.py
Ryan-Gordon: think this is fixed-- shows as outdated

Ryan-Gordon:
Is there a reason sort gives me different results to the CI jobs run of isort? A few times now I have run sort, fixed the recommended imports and then the CI job fails complaining about some of the edits my local sort recommended

Ryan-Gordon:
All passing checks after cleanup 👍

tests/core/caching-utils/test_generate_cache_key.py
```diff
@@ -3,14 +3,14 @@
...
 )
-from hypothesis import (
-    given,
-    strategies as st,
-)
 
 from web3._utils.caching import (
     generate_cache_key,
 )
+from web3.tools.hypothesis import (
```
pipermerriam: This is still importing from `web3` where it should be importing directly from `hytophesis`.

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from hypothesis import (
+from web3.tools.hypothesis import (
```
pipermerriam: Same.  import should come directly from the `hypothesis` module.

tests/core/filtering/test_contract_topic_filters.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from hypothesis import (
+from web3.tools.hypothesis import (
```
pipermerriam: Same, import should come directly from the `hypothesis` module (not proxied through the `web3.tools.hypothesis` imports.

web3/tools/hypothesis.py
```diff
@@ -0,0 +1,10 @@
+from hypothesis import (  # noqa: F401
+    example,
+    given,
```
pipermerriam: This and any of the other imports here that aren't used should be dropped.

kclowes:
Old PR cleanup - I believe this was fixed in v5.0.0a8. Thanks for the PR though @Ryan-Gordon!

PR #1188, opened 2019-01-07
Migrate tester provider results from middleware to defaults

dylanjw:
### What was wrong?
Many of default results for the EthereumTesterProvider come from a set of special middlewares. Eliminating these middlewares will reduce the number of middlewares that need to be duplicated for async support. Having these middlewares eliminated will also help with testing the early stages of the async api, which implements no middlewares.

Before I get too far with this, does this seem like an acceptable way forward? Should I continue eliminating the rest of the eth_tester middlewares? is there any reason that these results and formatters need the middleware pattern?  Removing them does not preclude using custom middlewares to alter the EthereumTesterProvider defaults. 

#### Cute Animal Picture
image

### DIFF
--- a/web3/providers/eth_tester/defaults.py
+++ b/web3/providers/eth_tester/defaults.py
@@ -157,20 +157,20 @@ API_ENDPOINTS = {
         ),
     },
     'net': {
-        'version': not_implemented,
-        'peerCount': not_implemented,
-        'listening': not_implemented,
+        'version': static_return('1'),
+        'peerCount': static_return(0),
+        'listening': static_return(False),
     },
     'eth': {
-        'protocolVersion': not_implemented,
-        'syncing': not_implemented,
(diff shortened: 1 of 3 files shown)


dylanjw:
Great.  I'll continue with the rest.

Not ready to merge until the following are complete:

- [x] Eliminate ethereum_tester_fixture_middleware
- [ ]  Eliminate default_transaction_fields_middleware
- [ ]  Eliminate ethereum_tester_middleware

dylanjw:
This is ready to merge.  I am going to work on the request parameter formatters and the transaction fields middleware in separate pull requests.

pipermerriam:
@dylanjw you should not have merge rights.

dylanjw:
Assuming piper meant 'now' instead of 'not', in case anyone reading this PR thinks Im out of control.

Issue #1177, opened 2018-12-29
Stateless contract log fetcher (for Infura)


miohtama:
* Version: 4.8.2

### What was wrong?

* I could not find a convenience method on the contract to get event logs over  Infura. Looks like there is only `createFilter` method that requires a stateful node (`eth_createFilter` API). 

### How can it be fixed?

I have sketched a `web3.contract.ContractEvent` function that is compatible with Infura and other stateless nodes. It is copy-paste job from `createFilter`

I need guidance on

* What are the input parameters `createFilter` uses, so I can properly document them out (argument_files, address, topics)

* If there is any conflicting work or feedback on the implementation

After this happy to open a PR.

```
def getLogs(self,
    argument_filters=None,
    fromBlock=None,
    toBlock="latest",
    address=None,
    topics=None):
    """Get events using eth_getLogs API.

    This is a stateless method, as opposite to createFilter.
    It can be safely called against nodes which do not provide eth_newFilter API, like Infura.

    :param argument_filters:
    :param fromBlock:
    :param toBlock:
    :param address:
    :param topics:
    :return:
    """

    if fromBlock is None:
        raise TypeError("Missing mandatory keyword argument to getLogs: fromBlock")

    abi = self._get_event_abi()

    argument_filters = dict()

    _filters = dict(**argument_filters)

    # Construct JSON-RPC raw filter presentation based on human readable Python descriptions
    # Namely, convert event names to their keccak signatures
    data_filter_set, event_filter_params = construct_event_filter_params(
        abi,
        contract_address=self.address,
        argument_filters=_filters,
        fromBlock=fromBlock,
        toBlock=toBlock,
        address=address,
        topics=topics,
    )

    # Call JSON-RPC API
    logs = self.web3.eth.getLogs(event_filter_params)

    # Convert raw binary data to Python proxy objects as described by ABI
    for entry in logs:
        yield get_event_data(abi, entry)
```

pipermerriam:
I can get onboard with this.  Implementation looks reasonable with potential exceptions related to `topics` and `address` (see below)

As for the parameters you should be able to infer them from the other filter creation APIs.

Allowing `topics` to be provided could be confusing or problematic since each event has it's own topic and combining those with user provided topics is likely to be hard to reason about.

Allowing `address` to be provided is also likely to result in confusion since contract *instances* have an address and it needs to be clearly defined whether the user can override this or what the behavior is.

In general :+1:

miohtama:
[...]

PR #1192, opened 2019-01-08
contract.events.X.getLogs() support without creating filters

miohtama:
### What was wrong?

Related to Issue #1177

### How was it fixed?

Added a proposed implementation.

Still working on the polish and documentation and test coverage. Opened a PR for early feedback.

#### Cute Animal Picture

```
\|/          (__)    
     `\------(oo)
       ||    (__)
       ||w--||     \|/
   \|/

```

### DIFF
--- a/tests/core/filtering/test_contract_getLogs.py
+++ b/tests/core/filtering/test_contract_getLogs.py
@@ -13,9 +13,10 @@ def test_contract_getLogs_all(
     web3,
     emitter,
     wait_for_transaction,
+    emitter_event_ids,
 ):
     contract = emitter
-    event_id = 1
+    event_id = emitter_event_ids.LogNoArguments
 
     txn_hash = contract.functions.logNoArgs(event_id).transact()
     wait_for_transaction(web3, txn_hash)
@@ -29,9 +30,10 @@ def test_contract_getLogs_range(
     web3,
     emitter,
     wait_for_transaction,
+    emitter_event_ids,
 ):
     contract = emitter
-    event_id = 1
+    event_id = emitter_event_ids.LogNoArguments
 
     assert web3.eth.blockNumber == 2
     txn_hash = contract.functions.logNoArgs(event_id).transact()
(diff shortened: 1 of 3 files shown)


web3/contract.py
```diff
@@ -1323,6 +1350,72 @@ def build_filter(self):
...
+            toBlock=toBlock,
+            address=self.address,
+        )
+
+        # Call JSON-RPC API
+        logs = self.web3.eth.getLogs(event_filter_params)
+
+        # Convert raw binary data to Python proxy objects as described by ABI
+        for entry in logs:
+            yield get_event_data(abi, entry)
```
pipermerriam: I think using a `yield` pattern here is likely to cause confusion.  Maybe go ahead and decorate this function using `to_tuple` or turn this into  `return tuple(get_event_data(abi, entry) for entry in logs)`

miohtama: Alrighty. We are not streaming the response from the node any case, so there is little marginal gain here.

miohtama: Fixed in the latest revision.

miohtama:
Ready to merge if no further comments  in a new review.

tests/core/filtering/test_contract_getLogs.py
```diff
@@ -0,0 +1,95 @@
...
+    txn_hashes.append(
+        emitter.functions.logTriple(event_id, 1, 3, 1).transact()
+    )
+    for txn_hash in txn_hashes:
+        wait_for_transaction(web3, txn_hash)
+
+    all_logs = contract.events.LogTripleWithIndex.getLogs()
+    assert len(all_logs) == 4
+
+    # Filter all entries where arg2 in (1, 2)
```
kclowes: Nitpick

I think this comment is wrong. Should be `arg1`?

web3/contract.py
```diff
@@ -1323,6 +1350,92 @@ def build_filter(self):
...
+                 'blockHash': HexBytes('...'),
+                 'blockNumber': 3
+                }),
+                AttributeDict(...),
+                ...
+            )
+
+        See also: :func:`web3.middleware.filter.local_filter_middleware`.
+
+        :param argument_filters: TODO
```
kclowes: Nitpick

Did you want to fill this in? If not, let's get rid of the `TODO` :)

PR #1200, opened 2019-01-11
Change manager.providers from list to single manager.provider (V5 beta)

### LINKED ISSUE #1024: Switch to a single provider instead of a list
### How can it be fixed?

Replace `manager.providers` with a single provider class abstraction `manager.provider`.

njgheorghita:
### What was wrong?
Fixes #1024 

### How was it fixed?
- Changed `manager.providers` -> `manager.provider`
- And subsequently changed `web3.providers` -> `web3.provider`

#### Cute Animal Picture
image

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -127,8 +127,8 @@ class Web3:
     isChecksumAddress = staticmethod(is_checksum_address)
     toChecksumAddress = staticmethod(to_checksum_address)
 
-    def __init__(self, providers=empty, middlewares=None, modules=None, ens=empty):
-        self.manager = self.RequestManager(self, providers, middlewares)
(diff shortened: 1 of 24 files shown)


web3/main.py
```diff
@@ -204,9 +204,8 @@ def solidityKeccak(cls, abi_types, values):
         return cls.keccak(hexstr=hex_string)
 
     def isConnected(self):
-        for provider in self.providers:
-            if provider.isConnected():
-                return True
+        if self.provider.isConnected():
```
kclowes: This is a nitpick, but can we just say `return self.provider.isConnected()`? Or is it a pythonic thing to have the explicit `return True`?

njgheorghita: Not sure if it's more or less "pythonic", but I do like your suggestion better, will update!

PR #1205, opened 2019-01-14
ENS - Stop inferring .eth TLD 

njgheorghita:
### What was wrong?
From Planned v5 changes
> Stop inferring .eth as a suffix on ENS names.

Related to Issue #722 

### How was it fixed?
- Updated `label_to_name()` util to raise an `InvalidTLD` exception if label was missing a TLD or used an unsupported TLD
- Removed the option to `guess_TLD` in `ens.address()`

#### Cute Animal Picture
image


ens/utils.py
```diff
@@ -108,16 +109,30 @@ def is_valid_name(name):
...
+    recognized_tlds = get_recognized_tlds()
+    return label_to_name(label, recognized_tlds)
+
+
+def get_recognized_tlds():
+    if 'ENS_RECOGNIZED_TLDS' in os.environ:
+        override_tlds = os.environ['ENS_RECOGNIZED_TLDS'].split(':')
+        return set(DEFAULT_RECOGNIZED_TLDS + override_tlds)
+    else:
+        return DEFAULT_RECOGNIZED_TLDS
```
njgheorghita: @pipermerriam How about this? I wasn't sure how much checking of the env variable to do. (eg. even if it's an empty string, everything should still work fine - and the user should be able to tell where they went wrong via the error msg). I went for also including the default TLDs even if the user sets the env variable, not sure if this should be avoided or not?

ens/utils.py
```diff
@@ -108,16 +109,30 @@ def is_valid_name(name):
...
 def dot_eth_name(label):
-    return label_to_name(label, 'eth', RECOGNIZED_TLDS)
+    recognized_tlds = get_recognized_tlds()
+    return label_to_name(label, recognized_tlds)
+
+
+def get_recognized_tlds():
+    if 'ENS_RECOGNIZED_TLDS' in os.environ:
+        override_tlds = os.environ['ENS_RECOGNIZED_TLDS'].split(':')
+        return set(DEFAULT_RECOGNIZED_TLDS + override_tlds)
```
pipermerriam: Hrm, I'm not sure these should be combined.  If the user supplies their own it would be up to them to ensure they included **all** desired TLDs

ens/utils.py
njgheorghita: @pipermerriam Just saw this on the ens docs
> Can I register a TLD of my own in the ENS?
No, TLDs are restricted to only .eth (on mainnet), or .eth and .test (on Ropsten), plus any special purpose TLDs such as those required to permit reverse lookups. There are no immediate plans to invite proposals for additional TLDs. In large part this is to reduce the risk of a namespace collision with the IANA DNS namespace.

Do we still want to allow users to set their own tlds?

carver: For private networks I guess. Also for ENS team to test new TLDs. (I know Nick has used the library at least once, and requested changes) Also, that doc looks out of date. There is at least `.xyz` and I think another available now.

carver:
Actually... it's not clear to me why we need *any* hard-coding of TLDs. They are resolved like any other name (looked up as a subdomain of the root), so they will just work like any other name that can't be found.

The only reason that we kept that list before was so that we could do something like "laptop.jasoncarver" getting resolved to "laptop.jasoncarver.eth", since `jasoncarver` isn't in the list of TLDs. That's of limited use, and I'm happy to see that go away.

In fact, I think one of the key benefits of removing the "guessing" of the TLDs is that we get to remove any logic that treats the TLDs as special, simplifying several places in the code.

tests/ens/test_setup_name.py
```diff
@@ -127,7 +130,7 @@ def test_setup_name_unowned_exception(ens):
 
 def test_setup_name_unauthorized(ens, TEST_ADDRESS):
     with pytest.raises(UnauthorizedError):
-        ens.setup_name('root-owned-tld', TEST_ADDRESS)
+        ens.setup_name('root-owned-tld.eth', TEST_ADDRESS)
```
carver: This test was originally attempting to set up a TLD, which would require owning the root. (Hence the name `root-owned-tld`)

Since the address sending the transaction doesn't own the root, it's supposed to fail.

tests/ens/test_setup_name.py
```diff
@@ -90,6 +78,21 @@ def test_setup_name(ens, name, normalized_name, namehash_hex):
...
 
+@pytest.mark.parametrize(
+    'name',
+    (
+        'tester',
+        'tester.com',
+        'lots.of.subdomains.tester',
+    ),
+)
+def test_cannot_setup_name_with_missing_or_invalid_tld(ens, name):
```
carver: So I think this whole test can go away if we stop treating TLDs special.

tests/ens/test_setup_address.py
```diff
@@ -95,10 +82,24 @@ def test_set_address(ens, name, full_name, namehash_hex, TEST_ADDRESS):
...
+@pytest.mark.parametrize(
+    'name',
+    (
+        'tester',
+        'tester.com',
+        'TESTER',
+        'lots.of.subdomains.tester',
+    ),
+)
+def test_set_address_raises_exception_with_invalid_or_missing_tld(ens, name, TEST_ADDRESS):
```
carver: I think this test goes away, too.

ens/utils.py
```diff
@@ -108,16 +109,30 @@ def is_valid_name(name):
...
+            f"ENS.py by default supports the following TLDs: {recognized_tlds}. "
+            "If you'd like to use an unsupported TLD, please set the environment variable: "
+            "'ENS_RECOGNIZED_TLDS' to a string of desired TLDs separated by a colon (:)."
+        )
     return '.'.join(pieces)
 
 
 def dot_eth_name(label):
-    return label_to_name(label, 'eth', RECOGNIZED_TLDS)
+    recognized_tlds = get_recognized_tlds()
```
carver: I don't think `dot_eth_name` should exist anymore. Looks like it can be replaced by `normalize_name` everywhere. Then this and `label_to_name` can be dropped entirely.

ens/main.py
```diff
@@ -109,9 +104,9 @@ def setup_address(self, name, address=default, transact={}):
         and calls this method with ``sub.parentname.eth``,
         then ``sub`` will be created as part of this call.
 
-        :param str name: ENS name to set up, in checksum format
-        :param str address: name will point to this address. If ``None``, erase the record.
-            If not specified, name will point to the owner's address.
+        :param str name: ENS name to set up
+        :param str address: name will point to this address, in checksum format. If ``None``,
```
carver: lol, good docfix

ens/exceptions.py
```diff
@@ -77,3 +77,10 @@ class UnderfundedBid(ValueError):
     as your intent to bid.
     '''
     pass
+
+
+class InvalidTLD(ValueError):
```
carver: Since we stop treating TLDs special, we can drop this exception.

ens/constants.py
```diff
@@ -9,6 +9,6 @@
 
 MIN_ETH_LABEL_LENGTH = 7
 
-RECOGNIZED_TLDS = ['eth', 'reverse', 'test', 'luxe', 'xyz']
+DEFAULT_RECOGNIZED_TLDS = ['eth', 'reverse', 'test', 'luxe', 'xyz']
```
carver: can be dropped

docs/ens_overview.rst
```diff
@@ -60,10 +60,8 @@ Look up the address for an ENS name
 
     assert eth_address == '0x5B2063246F2191f18F2675ceDB8b28102e957458'
 
-
-    # ens.py will assume you want a .eth name if you don't specify a full name
-
-    assert ns.address('jasoncarver') == eth_address
+    # ens.py only support names using one of these recognized TLDs
+    # ['eth', 'reverse', 'test', 'luxe', 'xyz']
```
carver: can be dropped

njgheorghita:
Thanks @carver & @pipermerriam for the review, made the changes. Ping for a final :eyes:

ens/utils.py
```diff
@@ -200,7 +185,7 @@ def dot_eth_namehash(name):
     :rtype: bytes
     :raises InvalidName: if ``name`` has invalid syntax
     '''
-    expanded_name = dot_eth_name(name)
+    expanded_name = normalize_name(name)
```
carver: The name of `dot_eth_namehash` should also probably change (since it doesn't append `dot_eth` anymore). Maybe `raw_name_to_hash`?

The other similar method, `name_to_hash`, could be changed to `normal_name_to_hash` to help distinguish.

tests/ens/test_setup_address.py
```diff
@@ -63,27 +57,17 @@
...
+    owner = ens.owner('tester.eth')
 
     ens.setup_address(name, TEST_ADDRESS)
     assert is_same_address(ens.address(name), TEST_ADDRESS)
 
-    # check that .eth is only appended if guess_tld is True
     namehash = Web3.toBytes(hexstr=namehash_hex)
     normal_name = ens.nameprep(full_name)
-    if ens.nameprep(name) == normal_name:
-        assert is_same_address(ens.address(name, guess_tld=False), TEST_ADDRESS)
```
carver: I think just this assertion should stay in, omitting `guess_tld` of course.

tests/ens/test_setup_address.py
```diff
@@ -63,27 +57,19 @@
...
-    owner = ens.owner('tester')
+    owner = ens.owner('tester.eth')
 
     ens.setup_address(name, TEST_ADDRESS)
     assert is_same_address(ens.address(name), TEST_ADDRESS)
 
-    # check that .eth is only appended if guess_tld is True
     namehash = Web3.toBytes(hexstr=namehash_hex)
     normal_name = ens.nameprep(full_name)
     if ens.nameprep(name) == normal_name:
```
carver: Can either drop this `if` test, or have an `else: assert False, "all names must be full names"` below.

PR #1210, opened 2019-01-15
Rename middleware_stack to middleware_onion

njgheorghita:
### What was wrong?
`middleware_stack` needed to be renamed to `middleware_onion`

Related to Issue #1022 

### How was it fixed?
Renamed `middleware_stack` to `middleware_onion`

#### Cute Animal Picture
image


fubuloubu:
Is there anyway this can just be "middleware"? I get why call it onion but... Yeah lol

PR #1211, opened 2019-01-15
Relocate personal RPC endpoints to parity and geth class

njgheorghita:
### What was wrong?

Updated `web3.personal` namespace to `web3.geth.personal` and `web3.parity.personal` which each reflect the personal jsonrpc endpoints supported by the respective client. Made the same update to `PersonalModuleTest` -> `GoEthereumPersonalModuleTest` and `ParityPersonalModuleTest` for integration testing.

#### Cute Animal Picture
image

### DIFF
--- a/web3/_utils/module_testing/personal_module.py
+++ b/web3/_utils/module_testing/personal_module.py
@@ -1,3 +1,5 @@
+import pytest
+
 from eth_utils import (
     is_checksum_address,
     is_list_like,
@@ -13,13 +15,13 @@ PRIVATE_KEY_FOR_UNLOCK = '0x392f63a79b1ff8774845f3fa69de4a13800a59e7083f5187f155
 ACCOUNT_FOR_UNLOCK = '0x12efDc31B1a8FA1A1e756DFD8A1601055C971E13'
 
 
-class PersonalModuleTest:
+class GoEthereumPersonalModuleTest:
     def test_personal_importRawKey(self, web3):
         actual = web3.geth.personal.importRawKey(PRIVATE_KEY_HEX, PASSWORD)
         assert actual == ADDRESS
 
     def test_personal_listAccounts(self, web3):
-        accounts = web3.geth.personal.listAccounts
+        accounts = web3.geth.personal.listAccounts()
         assert is_list_like(accounts)
         assert len(accounts) > 0
         assert all((
@@ -36,14 +38,17 @@ class PersonalModuleTest:
                                             web3,
                                             unlockable_account_dual_type,
                                             unlockable_account_pw):
-        result = web3.geth.personal.unlockAccount(unlockable_account_dual_type, unlockable_account_pw)
+        result = web3.geth.personal.unlockAccount(
+            unlockable_account_dual_type,
+            unlockable_account_pw
+        )
         assert result is True
 
     def test_personal_unlockAccount_failure(self,
                                             web3,
                                             unlockable_account_dual_type):
-        result = web3.geth.personal.unlockAccount(unlockable_account_dual_type, 'bad-password')
-        assert result is False
+        with pytest.raises(ValueError):
+            web3.geth.personal.unlockAccount(unlockable_account_dual_type, 'bad-password')
 
     def test_personal_newAccount(self, web3):
         new_account = web3.geth.personal.newAccount(PASSWORD)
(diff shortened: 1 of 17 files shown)


njgheorghita:
@pipermerriam @carver Here's a first stab at relocating the `personal_*` RPC endpoints to `geth` and `parity` modules respectively. I'm not sure if this is the best way to achieve the goal (many of the tests are broken, but i'll go ahead with fixing those after settling on a pattern). 

I'm not sure if I like the pattern in this PR, but I had trouble finding something I liked better since there are going to be shared endpoints b/w geth&parity as well as endpoints unique to either geth&parity. Also, what about other clients? If someone wanted to use the trinity client, it might be confusing to use a `web3.parity` namespace to access a `personal_*` rpc endpoint (though I'm unsure if this is relevant). 

https://docs.google.com/spreadsheets/d/1rNvszb2hbYrzUbZj0v4oPnSwyJezYYHhgcpiqdVf7Bk/edit?usp=sharing
Here's a link to a spreadsheet I made of all the RPC endpoints I could find, and whether/not they are supported by the different clients & web3.py, in case you find this useful.

pipermerriam:
Combined with the new way module methods are being done in #1166 this might be a little cleaner since we'd be able to re-use the method definitions for any commonly shaped endpoints.

web3/geth.py
```diff
@@ -0,0 +1,20 @@
...
+    Personal,
+)
+
+
+class GethPersonal(Personal):
+    """
+    https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal
+    """
+    def __init__(self, w3):
+        self.w3 = w3
```
dylanjw: This will override the inherited Module \_\_init\_\_, which could cause a bug in the future.

web3/geth.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Geth(Module):
+    @property
+    def personal(self):
+        return GethPersonal(self.web3)
```
dylanjw: Im not sure if it is better, but submodules could be configured in Web3 along with the modules.

Something to the effect of this in web3.Web3:

```python
def get_default_modules():
    return [
        {"name": "eth", "module": Eth},
        {"name": "net", "module": Net},
        {"name": "geth", "module": Geth, "submodules": {"personal": Personal}}
        {"name": "personal", "module": Personal},
        {"name": "version", "module": Version},
        {"name": "txpool", "module": TxPool},
        {"name": "miner", "module": Miner},
        {"name": "admin", "module": Admin},
        {"name": "parity", "module": Parity},
        {"name": "testing", "module": Testing},
    ]

class Web3:
   [...]
    def __init__(self, provider=None, middlewares=None, modules=None, ens=empty):
        self.manager = self.RequestManager(self, provider, middlewares)

        if modules is None:
            modules = get_default_modules()

        for module in modules:
            module['module'].attach(self, module.get('name'))
            if 'submodules' in module:
                for subname, submodule in module['submodules'].items():
                    submodule.attach(getattr(self, module['name']), submodule)

        self.ens = ens
```

The Module.attach method appears to work with modules 2 levels deep. Im leaning toward having 1 way modules are attached.

dylanjw:
With #1166 way of defining methods, the personal module methods could all be written in one place and the partiy/geth personal modules could compose the methods they support.  Then the not implemented methods could just be absent.

web3/_utils/module_testing/personal_module.py
```diff
@@ -61,7 +66,87 @@ def test_personal_sendTransaction(self,
...
+        assert result is True
+
+    def test_personal_unlockAccount_failure(self,
+                                            web3,
+                                            unlockable_account_dual_type):
+        result = web3.parity.personal.unlockAccount(
+            unlockable_account_dual_type,
+            'bad-password',
+            None
+        )
```
njgheorghita: I cannot figure this out for the life of me why parity insists on returning `True` if there is a failure to the `personal_unlockAccount` call.

njgheorghita:
also pinging @dylanjw if you don't mind taking another look over this

web3/_utils/module_testing/personal_module.py
dylanjw: You could a comment to the code explaining that this is the actual expected output at the moment, since its pretty weird.  Maybe open an issue with Parity?

web3/personal.py
```diff
@@ -1,62 +1,71 @@
...
-        )
+def default_root_munger(module, *args):
+    return [*args]
+
+
+def importRawKey():
+    return Method(
+        "personal_importRawKey",
+        mungers=[default_root_munger],
+        formatter_lookup_fn=None,
```
dylanjw: Nice!

Seeing the new Method class used, I think I need to make `Method`'s arguments more intuitive.  I prefer omitting `formatter_lookup_fn` here rather than setting it to None.  Setting it to None looks like it would not have formatters applied, but it actually will have the default formatters applied.  I will make changes to `Method` to either remove `formatter_lookup_fn` from the arguments, or have it override the default formatters when set to None or a custom lookup function.

web3/main.py
```diff
@@ -82,17 +84,17 @@
...
-        "testing": Testing,
-    }
+    return [
+        {"name": "eth", "module": Eth},
+        {"name": "net", "module": Net},
+        {"name": "version", "module": Version},
+        {"name": "txpool", "module": TxPool},
+        {"name": "miner", "module": Miner},
+        {"name": "admin", "module": Admin},
+        {"name": "parity", "module": Parity, 'submodules': {'personal': ParityPersonal}},
```
carver: You may like this format better:
```py
return {
  ...
  "admin": Admin,
  "parity": (Parity, {"personal": ParityPersonal}),
  ...
}
```
One nice thing about this is that the submodule format is the same as the supermodule format, which makes for easier extension to multiple layers if desired. It also means less special casing when stitching the modules togother. Plus I think it's a little nicer to read.

Here's an example with 3 layers:
```py
return {
  ...
  "admin": Admin,
  "parity": (Parity, {
    "personal": ParityPersonal,
    "test": (TestModule, {
      "eth": ParityEthTest,
      "bzz": ParitySwarmTest,
      ...
    }),
  }),
  ...
}
```

web3/personal.py
carver: Yeah, I would expect:
- "`formatter_lookup_fn` kwarg missing": default formatting
- "`formatter_lookup_fn=None`": no formatting
- "`formatter_lookup_fn=my_formatter`": only `my_formatter` runs

web3/main.py
```diff
@@ -130,8 +136,12 @@ def __init__(self, provider=None, middlewares=None, modules=None, ens=empty):
...
             modules = get_default_modules()
 
-        for module_name, module_class in modules.items():
-            module_class.attach(self, module_name)
+        for module in modules:
+            module_class, *submodules = modules[module]
+            module_class.attach(self, module)
+            if submodules:
+                for subname, subclasses in submodules[0].items():
+                    subclasses[0].attach(getattr(self, module), subname)
```
njgheorghita: @carver Thoughts on this setup? Maybe there's a cleaner way to achieve the attaching of subclasses?

web3/main.py
```diff
@@ -130,8 +136,12 @@ def __init__(self, provider=None, middlewares=None, modules=None, ens=empty):
         if modules is None:
             modules = get_default_modules()
 
-        for module_name, module_class in modules.items():
-            module_class.attach(self, module_name)
+        for module in modules:
+            module_class, *submodules = modules[module]
+            module_class.attach(self, module)
+            if submodules:
+                for subname, subclasses in submodules[0].items():
```
carver: Yeah, this `submodules[0]` looks surprising, I think. Also, if we're going to go with a recursive definition, might as well implement with a recursive solution, so it can go arbitrarily (well stack-limit) deep.

Something like a new utility method:
```py
    def attach_modules(parent_module, module_definitions)
       for module_name, module_info in module_definitions.items():
            module_class = module_info[0]
            module_class.attach(parent_module, module_name)

            if len(module_info) == 2:
                submodule_definitions = module_info[1]
                module = getattr(parent_module, module_name)
                attach_modules(module, submodule_definitions)
            elif len(module_info) != 1:
                raise ValidationError("Module definition can only have 1 or 2 elements")
```

Then the web3 init looks like:
```py
        Web3.attach_modules(self, modules)
```

PR #1215, opened 2019-01-17
Remove .DS_Store files 

njgheorghita:
### What was wrong?
My fat finger accidentally committed some `.DS_Store` files I only realized in retrospect in pr #1205 
Deleted them in this PR

#### Cute Animal Picture

image


njgheorghita:
@carver 👍 Added `*.DS-Store` and `.mypy_cache/` here (assuming that mypy is on the roadmap for `web3` eventually). Also included the change to your most recent comment in #1205 in this pr

pipermerriam:
@njgheorghita you can have a global gitignore in your home directory where you add these as well which will prevent this from happening to you in the future across all of your repos. (but it's also good to have these in the local project's `.gitignore` as well.

PR #1216, opened 2019-01-17
Disable pm api by default

njgheorghita:
### What was wrong?
Disable `PM`API by default, and mechanism to enable the unstable API.

#### Cute Animal Picture
image

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -206,3 +206,18 @@ class Web3:
     @ens.setter
     def ens(self, new_ens):
         self._ens = new_ens
+
+    @property
+    def pm(self):
+        if self._pm is not None:
+            return self._pm
+        else:
+            raise AttributeError(
+                "The Package Management feature is disabled by default until "
+                "its API stabilizes. To use these features, please enable them by running "
+                "`w3.enable_unstable_package_management_api()` and try again."
+            )
+
+    def enable_unstable_package_management_api(self):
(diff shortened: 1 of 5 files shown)


web3/main.py
```diff
@@ -206,3 +206,17 @@ def ens(self):
...
+        if self._pm is not None:
+            return self._pm
+        else:
+            raise AttributeError("The Package Management feature is disabled by default until "
+                "its API stabilizes. To use these features, please enable them by running "
+                "`w3.enable_unstable_package_management_api()` and try again."
+            )
+
+    def enable_unstable_package_management_api(self):
+        from web3.pm import PM
```
njgheorghita: This was the quick and easy solution to avoiding the circular import this time around, but it seems to me like this won't be suitable long-term when we want to make the PM api available on the `web3` object by default.

carver: I'm missing it: which circular import? Does `web3.pm` import from `web3.main` somehow?

I don't follow how this will be different from any other modules that are attached on `Web3` by default.

carver:
> @carver Sorry, it's not very clear but `web3.pm` imports from `ethpm.package` which imports `Web3` from `web3.main`

At some point, we may have to vendor `ethpm` like we did `ens`. For now, the internal import is okay. It may instead make sense to do the internal import inside `web3.pm`, but I don't have any suggestion on that yet

PR #1218, opened 2019-01-18
Raise exceptions rather than return None in all jsonrpc calls

njgheorghita:
### What was wrong?
https://github.com/ethereum/web3.py/issues/722#issuecomment-438041562

> I propose that we change web3.eth.getBlock, web3.eth.getTransaction and other similar methods to no longer return None, but instead to raise an exception. The current API encourages/requires return value checking which is not a pattern I think we should be promoting.

### How was it fixed?
Raised a `ValueError` if return value from `Manager._make_request()` is `None`

#### Cute Animal Picture
image

### DIFF
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -485,8 +485,8 @@ class EthModuleTest:
         assert block['hash'] == empty_block['hash']
 
     def test_eth_getBlockByHash_not_found(self, web3, empty_block):
-        block = web3.eth.getBlock(UNKNOWN_HASH)
-        assert block is None
+        with pytest.raises(ValueError):
+            web3.eth.getBlock(UNKNOWN_HASH)
 
     def test_eth_getBlockByNumber_with_integer(self, web3, empty_block):
         block = web3.eth.getBlock(empty_block['number'])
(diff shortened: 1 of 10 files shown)


njgheorghita:
@pipermerriam What other `web3.eth` methods are candidates for this update?

njgheorghita:
@pipermerriam ping for review

web3/manager.py
```diff
@@ -107,6 +110,9 @@ def request_blocking(self, method, params):
         if "error" in response:
             raise ValueError(response["error"])
 
+        if response['result'] is None:
+            raise ValueError(f"The call to {method} did not return a value.")
```
njgheorghita: @kclowes @dylanjw I'm not greatly familiar with the async gameplan for web3 - but it occurred to me that if we are raising exceptions for jsonrpc calls that return `None` in synchronous requests, then we should have the same behavior in the async requests.

dylanjw: This looks good to me.

web3/_utils/module_testing/eth_module.py
```diff
@@ -485,8 +485,8 @@ def test_eth_getBlockByHash(self, web3, empty_block):
         assert block['hash'] == empty_block['hash']
 
     def test_eth_getBlockByHash_not_found(self, web3, empty_block):
-        block = web3.eth.getBlock(UNKNOWN_HASH)
-        assert block is None
+        with pytest.raises(ValueError):
```
pipermerriam: I realize that I'm late to the party on this one, but I think that this should be a custom exception for each object type.

- `BlockNotFound`
- `TransactionNotFound` (use this for receipt fetching too)
- anything else?

cc @njgheorghita

PR #1219, opened 2019-01-22
Fix bug when checking _pm attr on w3

njgheorghita:
### What was wrong?
`w3.pm` would raise a generic `AttributeError` rather than the intended `AttributeError` when checking for `web3._pm` if the attribute hadn't been set yet.

#### Cute Animal Picture
image

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -203,7 +203,7 @@ class Web3:
 
     @property
     def pm(self):
-        if self._pm is not None:
+        if hasattr(self, '_pm'):
             return self._pm
         else:
             raise AttributeError(


web3/main.py
```diff
@@ -203,7 +203,7 @@ def ens(self, new_ens):
 
     @property
     def pm(self):
-        if self._pm is not None:
+        if hasattr(self, '_pm'):
```
pipermerriam: I think I'd be in favor of just having `_pm = None` defined on the class so that we don't have to do `hasattr` checks.

njgheorghita: Defining `_pm = None` on the class causes a conflict with the way the `PM` module is eventually attached when running `w3.enable_unaudited_package_management_api()`

> Cannot set <web3.main.Web3 object at 0x1050596d8> module named '_pm'.  The web3 object already has an attribute with that name

It seems as though we can either `hasattr` check or change the way the `PM` module is attached to the `w3` instance - currently ...

```python
def enable_unaudited_package_management_api(self):
    PM.attach(self, '_pm')
```

njgheorghita: @pipermerriam ping

PR #1221, opened 2019-01-22
Add custom registry for encodability checks

davesque:
### What was wrong?

Related to PR #1147.  Web3.py's encodability checking does not support tuple values.

### How was it fixed?

Added a custom eth-abi registry and encoder subclasses to take advantage of eth-abi's encodability checking for tuple values.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
     url='https://github.com/ethereum/web3.py',
     include_package_data=True,
     install_requires=[
-        "eth-abi>=2.0.0b4,<3.0.0",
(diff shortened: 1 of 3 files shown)


davesque:
Left to do: add a dependency on the latest version of eth-abi.  @carver @pipermerriam Would it be proper if web3.py required a beta version of eth-abi?

carver:
> Would it be proper if web3.py required a beta version of eth-abi?

Yes, but only because this is going into the v5 pre-release. All the core dependencies will need to go stable before v5 goes stable.

PR #1223, opened 2019-01-23
Make docstring style consistent with Python PEPs

davesque:
### What was wrong?

The style of some docstrings is not consistent with recommendations in PEP 257 and PEP 287.

### How was it fixed?

Fixed them.

#### Cute Animal Picture


pipermerriam:
@kclowes looks like we can do that with https://github.com/PyCQA/pydocstyle

PR #1224, opened 2019-01-24
Remove unnecessary exception catching

davesque:
### What was wrong?

Exceptions generated by eth-abi are fairly descriptive.  It seems a bit
redundant to wrap encoding exceptions with this extra message and also
filter out the original exception type.

Also, some code wrapped in this try/except block could never throw `EncodingError` exceptions.

I poked around and it doesn't look like any dependent code is expecting a `TypeError` specifically.  Also, the test suite still seems to pass.  So I'm not sure if there's much risk here.

### How was it fixed?

Removed the relevant try/except block.

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/contracts.py
+++ b/web3/_utils/contracts.py
@@ -3,9 +3,6 @@ import functools
 from eth_abi import (
     encode_abi as eth_abi_encode_abi,
 )
-from eth_abi.exceptions import (
-    EncodingError,
-)
 from eth_utils import (
     add_0x_prefix,
     encode_hex,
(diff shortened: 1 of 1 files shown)


davesque:
@carver 

What's the reasoning behind that?  Is it that it's annoying to have to import exceptions to catch them?

I think I noticed that and tried to follow suit, but eventually I got frustrated that there isn't always a built-in exception that seems appropriate for the given situation.  Then, I just broke down and went back to having a custom set of `Exception` subclasses for the library.

One reason I kinda like that is exceptions can more directly correspond to the task that's being done e.g. `EncodingError` getting raised during encoding of values.

carver:
> What's the reasoning behind that? Is it that it's annoying to have to import exceptions to catch them?

Yeah, that seems about right. Roughly that the global exceptions are a nice standard to rally around. When possible, it's better to stick to a global standard than a project-level one. Like I said, not a strong feeling though (at a basic level, someone naturally has to learn your API to work with your library).

> I don't feel like anyone would be confused if something called `EncodingError` was raised when calling something named `encode_abi`.

Sounds good to me :+1:

PR #1227, opened 2019-01-30
Contract Caller Cleaned up

kclowes:
**This should be the same diff as the original Contract Caller PR (#1157), but branched off of the tip of master since the original had been open so long. There were some really nasty merge conflicts that made for ugly history**

### What was wrong?
Want to deprecate ConciseContract and ImplicitContract in favor of a ContractCaller. There are more details in the issue linked below.

Related to Issue #
[#1025](https://github.com/ethereum/web3.py/issues/1025)

### How was it fixed?
Added a new ContractCaller class and added deprecation methods to ConciseContract and ImplicitContract. 

See #1157

#### Cute Animal Picture


web3/contract.py
```diff
@@ -1442,6 +1449,82 @@ def factory(cls, class_name, **kwargs):
...
+    def __getattr__(self, function_name):
+        if '_functions' not in self.__dict__:
+            raise NoABIFunctionsFound(
+                "The abi for this contract contains no function definitions. ",
+                "Are you sure you provided the correct contract abi?"
+            )
+        elif function_name not in self.__dict__['_functions']:
+            raise MismatchedABI(
+                "The function '{}' was not found in this contract's abi. ".format(function_name),
+                "Are you sure you provided the correct contract abi?"
```
pipermerriam: It might be useful to include the names of all of the known functions in this message since it's very likely a user encountering this error has either mispelled a function name, or upon encountering it, will want to know what names are available.

web3/contract.py
```diff
@@ -1442,6 +1449,82 @@ def factory(cls, class_name, **kwargs):
...
+
+                setattr(self, func['name'], caller_method)
+
+    def __getattr__(self, function_name):
+        if '_functions' not in self.__dict__:
+            raise NoABIFunctionsFound(
+                "The abi for this contract contains no function definitions. ",
+                "Are you sure you provided the correct contract abi?"
+            )
+        elif function_name not in self.__dict__['_functions']:
```
pipermerriam: These conditionals can get away from accessing `self.__dict__` if we instead define these attributes to default to `None` on the class which should also have the benefit of being more precise checks.

```
if self.abi is None:
    # no abi
elif len(self._functions) == 0:
    # no functions in the abi
elif function_name not in self._functions:
    # no matching function
```

web3/contract.py
```diff
@@ -1442,6 +1449,82 @@ def factory(cls, class_name, **kwargs):
...
+                "The function '{}' was not found in this contract's abi. ".format(function_name),
+                "Are you sure you provided the correct contract abi?"
+            )
+        else:
+            return super().__getattribute__(function_name)
+
+    def __call__(self, transaction_dict=None, block_identifier='latest'):
+        if transaction_dict is None:
+            transaction_dict = {}
+        return type(self)(self.abi,
```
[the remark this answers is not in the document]
kclowes: It raises the correct `NoABIFound` error. I added some test cases to be sure though!

web3/contract.py
kclowes: I like that. I think it's updated to what you recommended, but let me know if it's not!

web3/contract.py
```diff
@@ -1442,6 +1450,96 @@ def factory(cls, class_name, **kwargs):
...
+                                        block_identifier=block_id)
+
+                setattr(self, func['name'], caller_method)
+
+    def __getattr__(self, function_name):
+        if self.abi is None:
+            raise NoABIFound(
+                "There is no ABI found for this contract.",
+            )
+        if not self._functions or len(self._functions) == 0:
```
pipermerriam: *nitpick*

This should be an `elif` though it is functionally equivalent in the current form

web3/contract.py
```diff
@@ -1442,6 +1450,96 @@ def factory(cls, class_name, **kwargs):
...
+            raise NoABIFunctionsFound(
+                "The ABI for this contract contains no function definitions. ",
+                "Are you sure you provided the correct contract ABI?"
+            )
+        elif function_name not in self._functions:
+            functions_available = ', '.join([fn['name'] for fn in self._functions])
+            raise MismatchedABI(
+                "The function '{}' was not found in this contract's ABI. ".format(function_name),
+                "Here is a list of all of the function names found: {}".format(functions_available),
+                "Did you mean to call one of those functions?"
```
pipermerriam: This will result in a message like 

> `"Here is a list of all of the function names found: fn_a, fn_bDid you mean to call one of those functions?"

Notice the lack of space before the `Did`.

kclowes: Oops, nice catch!

docs/contracts.rst
```diff
@@ -745,3 +744,59 @@ Utils
...
+
+And a transaction dictionary, with or without the ``transaction_dict`` keyword. For example:
+
+.. doctest::
+
+   >>> from_address = w3.eth.accounts[1]
+   >>> twentyone = myContract.caller({'from': from_address}).multiply7(3)
+   >>> twentyone
+   21
+   >>> twentyone = myContract.caller(transaction_dict={'from': from_address}).multiply7(3)
```
carver: Maybe one more example that includes `block_identifier`?

tests/core/contracts/test_contract_caller_interface.py
```diff
@@ -0,0 +1,143 @@
...
+
+def test_caller_with_transaction_keyword(web3, return_args_contract):
+    address = web3.eth.accounts[1]
+    transaction_dict = {
+        'from': address,
+        'gas': 210000,
+        'gasPrice': web3.toWei(.001, 'ether'),
+        'value': 12345,
+    }
+    contract = return_args_contract.caller(transaction_dict=transaction_dict)
```
carver: This `dict` has been called `transaction` in enough other places that I think it's more consistent to just use `transaction` instead of `transaction_dict` as the keyword argument name.

PR #1235, opened 2019-02-05
Add support for tuples

davesque:
### What was wrong?

This is a rebuild of PR #1147 .  Commits retain authorship from original PR where appropriate.

### How was it fixed?

Adds support for tuples by modifying low-level utility functions used for ABI dict processing and encodability checks.

#### Cute Animal Picture


davesque:
I'm adding things bit by bit here.  The most recent chunk of work done was to rewrite the `get_abi_inputs` utility function to be simpler and more robust.

Eventually, I'll give things another once over to make sure I didn't miss anything.

Known issues:
* Updated `get_abi_*_types` functions need to have updated tests
* New `get_abi_inputs` function needs more tests.  Will probably add some or all of the tests written by @feuGeneA for that function.

web3/_utils/abi.py
```diff
@@ -342,6 +347,66 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+    abi_type_str = comp['type']
+
+    if not abi_type_str.startswith(TUPLE_PREFIX):
+        # Component is non-tuple.  Just return value.
+        return arg
+
+    # Component is tuple (possibly tuple list)...
+
+    array_dims = abi_type_str[TUPLE_PREFIX_LEN:]
+    if array_dims:
```
pipermerriam: Can we tighten this up at all?  Match against a regex or something?  This jumps out as a good way to end up with weird errors jumping out in the case that the ABI type string is malformed.

web3/_utils/abi.py
```diff
@@ -342,6 +347,66 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+        # Component is tuple list (i.e. it has array dimensions).  Construct
+        # new ABI component without array dimensions and recurse with it for
+        # each element in arg.
+        new_comp = copy.copy(comp)
+        new_comp['type'] = TUPLE_PREFIX
+
+        # Assume `arg` is iterable
```
pipermerriam: Same here, easy to do a quick runtime check that `isinstance(arg, collections.abc.Iterable)` or maybe tighter as `not isinstance(arg, (bytes, str)) and isinstance(arg, collections.abc.Sequence)`

web3/_utils/abi.py
```diff
@@ -342,6 +347,66 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+        return type(arg)(_convert_abi_input(new_comp, a) for a in arg)
+
+    # Component is non-list tuple...
+
+    sub_comps = comp['components']
+
+    if isinstance(arg, abc.Mapping):
+        # Arg is mapping.  Convert to properly ordered sequence.
+        arg = tuple(arg[c['name']] for c in sub_comps)
```
pipermerriam: It's going to take me a long time to not have a knee-jerk reflex that python dictionaries aren't ordered (since they now are)

web3/_utils/abi.py
davesque: Good idea.  I'll add that.

web3/_utils/abi.py
davesque: Added a runtime type check using `eth_utils.is_list_like`.

web3/_utils/abi.py
davesque: If you meant `sub_comps`, I believe that should always be a `list` instance.

web3/_utils/abi.py
```diff
@@ -302,6 +313,85 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+    (``args``).  Returns a list of canonical type names for the function's
+    inputs and a list of arguments in corresponding order.  The args contained
+    in ``args`` may contain nested mappings or sequences corresponding to
+    tuple-encoded values in ``abi``.
+    """
+    inputs = abi.get('inputs', [])
+
+    if isinstance(args, abc.Mapping):
+        # `args` is mapping, convert to properly ordered sequence
+        args = tuple(args[i['name']] for i in inputs)
```
carver:
```suggestion
        args = tuple(args[input['name']] for input in inputs)
```

*nit* opinion: even in short comprehensions, I think it's easier to read with a full variable name. It prevents your eyes having to dart back and forth, so you can more or less read it straight through in one pass.

web3/_utils/abi.py
```diff
@@ -302,6 +313,85 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+        tuple_dims = match.group(2)
+
+        return tuple_prefix, tuple_dims
+
+    return None
+
+
+def _convert_abi_input(comp, arg):
+    """
+    Converts an argument ``arg`` corresponding to an ABI component ``comp``
```
carver: `component` here seems confusing because each element of a tuple is called `component`. Maybe this is a `arg_type`, or `arg_spec` since it includes a name?

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -612,3 +617,61 @@ def test_invalid_fixed_value_reflections(web3, fixed_reflection_contract, functi
...
+                    '0xfdF1946A9b40245224488F1a36f4A9ed4844a523',
+                ]),
+                (345, [False, False], [
+                    '0xefd1FF70c185A1C0b125939815225199079096Ee',
+                    '0xf35C0784794F3Cd935F5754d3a0EbcE95bEf851e',
+                ]),
+            ]),
+        ),
+    ),
+)
```
carver: It looks like this is already working, but might be nice to add a regression test for nested arrays of tuples. Seems like the kind of thing that could break in a rewrite/refactor.

web3/_utils/abi.py
```diff
@@ -580,11 +670,25 @@ def abi_sub_tree(abi_type: Optional[Union[TypeStr, ABIType]], data_value: Any) -
...
+    # annotated items
     if abi_type.is_array:
-        it = abi_type.item_type
-        return ABITypedData([abi_type.to_type_str(), [abi_sub_tree(it, i) for i in data_value]])
-    else:
-        return ABITypedData([abi_type.to_type_str(), data_value])
+        # If type is array, determine item type and annotate all
+        # items in iterable with that type
+        item_type_str = abi_type.item_type.to_type_str()
+        data_value = [abi_sub_tree(item_type_str, i) for i in data_value]
```
carver: Maybe `item` instead of `i`? Not sure if that's what `i` is short for, here.

web3/_utils/abi.py
```diff
@@ -580,11 +670,25 @@ def abi_sub_tree(abi_type: Optional[Union[TypeStr, ABIType]], data_value: Any) -
...
+    elif isinstance(abi_type, TupleType):
+        # Otherwise, if type is tuple, determine component types and annotate
+        # items in iterable respectively with those types
+        data_value = type(data_value)(
+            abi_sub_tree(comp_type.to_type_str(), i)
+            for comp_type, i in zip(abi_type.components, data_value)
```
carver: Maybe `component` instead of `i` here?

web3/_utils/contracts.py
```diff
@@ -240,6 +241,8 @@ def get_function_info(fn_name, contract_abi=None, fn_abi=None, args=None, kwargs
 
     fn_arguments = merge_args_and_kwargs(fn_abi, args, kwargs)
 
+    _, fn_arguments = get_abi_inputs(fn_abi, fn_arguments)
```
carver: *nit* overwrote `fn_arguments`, but now it has a slightly different meaning. Maybe `typed_arguments` or `annotated_arguments`? (same as the one in `web3/_utils/abi` I commented on)

tests/core/contracts/conftest.py
```diff
@@ -6,6 +6,53 @@
...
+CONTRACT_TUPLE_ABI = json.loads('[{"constant":true,"inputs":[{"components":[{"name":"a","type":"uint256"},{"name":"b","type":"uint256[]"},{"components":[{"name":"x","type":"int256"},{"name":"y","type":"bool[2]"},{"name":"z","type":"address[]"}],"name":"c","type":"tuple[]"}],"name":"s","type":"tuple"}],"name":"method","outputs":[{"components":[{"name":"a","type":"uint256"},{"name":"b","type":"uint256[]"},{"components":[{"name":"x","type":"int256"},{"name":"y","type":"bool[2]"},{"name":"z","type":"address[]"}],"name":"c","type":"tuple[]"}],"name":"","type":"tuple"}],"payable":false,"stateMutability":"pure","type":"function"}]')  # noqa: E501
```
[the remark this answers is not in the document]
davesque: I was just following the style of existing fixture code with that one.

tests/core/contracts/test_contract_method_to_argument_matching.py
```diff
@@ -13,8 +13,82 @@
 
 SINGLE_FN_NO_ARGS = json.loads('[{"constant":false,"inputs":[],"name":"a","outputs":[],"type":"function"}]')  # noqa: E501
 SINGLE_FN_ONE_ARG = json.loads('[{"constant":false,"inputs":[{"name":"","type":"uint256"}],"name":"a","outputs":[],"type":"function"}]')  # noqa: E501
-MULTIPLE_FUNCTIONS = json.loads('[{"constant":false,"inputs":[],"name":"a","outputs":[],"type":"function"},{"constant":false,"inputs":[{"name":"","type":"bytes32"}],"name":"a","outputs":[],"type":"function"},{"constant":false,"inputs":[{"name":"","type":"uint256"}],"name":"a","outputs":[],"type":"function"},{"constant":false,"inputs":[{"name":"","type":"uint8"}],"name":"a","outputs":[],"type":"function"},{"constant":false,"inputs":[{"name":"","type":"int8"}],"name":"a","outputs":[],"type":"function"}]')  # noqa: E501
 FALLBACK_FUNCTION = json.loads('[{"constant": false, "inputs": [], "name": "getData", "outputs": [{"name": "r", "type": "uint256"}], "payable": false, "stateMutability": "nonpayable", "type": "function"}, {"payable": false, "stateMutability": "nonpayable", "type": "fallback"}]')  # noqa: E501
+MULTIPLE_FUNCTIONS = json.loads('''
```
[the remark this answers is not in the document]
davesque: Was following the style of existing code.

davesque:
@carver I sort of feel like avoiding single letter variables can be just as confusing.  In a lot of the examples cited, the cognitive overhead of the short variable names seems pretty minimal.  Also, giving things longer names often leads to a code section taking up two or three lines when it could have taken up one.  I feel like the overhead of reading a statement that runs across multiple lines is as much if not more than the overhead of reading a statement that uses meaningful abbreviations.  Additionally, the typical alternative of using a variable name that simply lacks pluralization is pretty confusing because then two variables only differ by one letter and it's easy to get typos.

Anyway, I'll change it if we really want but I've often felt like I don't understand the aversion to that coding pattern.  I wanted to explain some of my thoughts about that.

davesque:
@carver I went ahead and addressed all of your feedback.  As far as the testing of deeper tuple nesting goes, is [this](https://github.com/ethereum/web3.py/pull/1235/commits/140e32eca501830aa59c63d29d788bb3b0476efb) what you had in mind?

carver:
> As far as the testing of deeper tuple nesting goes, is [this](https://github.com/ethereum/web3.py/pull/1235/commits/140e32eca501830aa59c63d29d788bb3b0476efb) what you had in mind?

Yup, that's about it @davesque ! :) I suspect there are maybe funkier things that could go wrong if someone messes about with this code, but I think this is plenty of coverage for now. :+1: :shipit:

pipermerriam:
@davesque can you confirm this is :+1: to merge (i.e. you don't have anything outstanding that you know needs to be addressed).

@kclowes or maybe @njgheorghita .  IIUC correctly, this gives us support for tuples, but only as raw tuples, and not any *rich* object representations.

I think what comes next is some version of support for those.

1. named tuples
2. dictionaries
3. arbitrary objects with attributes.

It isn't clear to me exactly how these should be implemented but I think doing them in this order probably makes sense.  The `NamedTuple` one is probably easiest since we can generate the class from the ABI definition and return it in place of a plain tuple.  This could be a core part of how web3.py extends the eth-abi API.

davesque:
@pipermerriam Yes it's ready to merge.  I was waiting for your review.

web3/_utils/abi.py
```diff
@@ -302,6 +313,90 @@ def merge_args_and_kwargs(function_abi, args, kwargs):
...
+        # Arg is list tuple.  A non-list version of its abi will be used to
+        # align each element in `arg`.
+        new_abi = copy.copy(arg_abi)
+        new_abi['type'] = tuple_prefix
+
+        sub_abis = itertools.repeat(new_abi)
+
+    if isinstance(arg, abc.Mapping):
+        # Arg is mapping.  Align values according to abi order.
+        arg = tuple(arg[abi['name']] for abi in sub_abis)
```
pipermerriam: We have a convention of avoiding overwriting local variables, preferring to instead use new variable names.  Maybe we can call this `aligned_arg`

```python
if is_mapping(arg):
    aligned_arg = tuple(...)
else:
    aligned_arg = arg
```

> NOTE: the above statement has an implicit `else` clause which is also somewhat against our conventions.  It might should look something like.

```python
if is_mapping(arg):
    ...
elif is_list_like(arg):
    ...
else:
    raise TypeError("Something something invalid")
```

This would simply be moving the `TypeError` from below up to be part of this clause.

pipermerriam: Also noting this relates to #1267 as it appears we already support dictionaries for the input arguments for tuple types.

web3/_utils/abi.py
```diff
@@ -580,11 +675,28 @@ def abi_sub_tree(abi_type: Optional[Union[TypeStr, ABIType]], data_value: Any) -
...
+    # annotated items
     if abi_type.is_array:
-        it = abi_type.item_type
-        return ABITypedData([abi_type.to_type_str(), [abi_sub_tree(it, i) for i in data_value]])
-    else:
-        return ABITypedData([abi_type.to_type_str(), data_value])
+        # If type is array, determine item type and annotate all
+        # items in iterable with that type
+        item_type_str = abi_type.item_type.to_type_str()
+        data_value = [
```
pipermerriam: Same here with convention of not overwriting local variables which if fixed will require this `if/elif` to get a final `else` clause.

davesque:
@carver @pipermerriam Thanks guys for the reviews.  When the tests pass, I'll go ahead and merge this.

PR #1237, opened 2019-02-07
Add manifest and input arg validation to pm.release_package()

njgheorghita:
### What was wrong?
More validation in `pm.release_package()` would be useful to prevent mix-ups between on-registry data and data actually in the manifest, and prevent pushing invalid manifest to your registry

### How was it fixed?
- Added validation to `pm.release_package()` that the manifest uri is a valid content-addressed (IPFS/github) URI.
- Add checks that given `package_name` and `version` match what's stored in the manifest

#### Cute Animal Picture
image

### DIFF
--- a/web3/pm.py
+++ b/web3/pm.py
@@ -33,6 +33,19 @@ from ethpm.typing import (
     Address,
     Manifest,
 )
+from ethpm.utils.backend import (
+    resolve_uri_contents,
+)
+from ethpm.utils.ipfs import (
+    is_ipfs_uri,
+)
+from ethpm.utils.manifest_validation import (
+    validate_manifest_against_schema,
+    validate_raw_manifest_format,
+)
+from ethpm.utils.uri import (
+    is_valid_content_addressed_github_uri,
+)
 from ethpm.validation import (
     validate_package_name,
     validate_package_version,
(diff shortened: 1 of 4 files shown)


web3/pm.py
```diff
@@ -463,12 +476,36 @@ def release_package(
...
-        validate_package_version(version)
+            * ``package_name``: Must be a valid package name, matching the given manifest.
+            * ``version``: Must be a valid package version, matching the given manifest.
+            * ``manifest_uri``: Must be a valid content-addressed URI. Currently, only IPFS
+            and Github content-addressed URIs are supported.
+        """
+        if (
+            not is_ipfs_uri(manifest_uri) and
+            not is_valid_content_addressed_github_uri(manifest_uri)
+        ):
```
pipermerriam: We've avoided multi-line if statements via the following convention.

```python
is_something = (
    <multi-line-statement
)
if is_something:
    ...
```

web3/pm.py
```diff
@@ -463,12 +476,36 @@ def release_package(
...
+            raise PMError(
```
pipermerriam: Seems like this should be some sort of `ValidationError` type error.  The base library error is normally not used directly, but rather designed to be a way for someone to *catch any library error from `py-ethpm`*

PR #1238, opened 2019-02-08
Store registry address as checksummed

njgheorghita:
### What was wrong?
Minor ux improvement

#### Cute Animal Picture
image


njgheorghita:
Merging since minor fix

Issue #1246, opened 2019-02-17
Deprecate Infura with no API key


davesque:
### What was wrong?

Currently, the Infura provider is usable without an API key.

### How can it be fixed?

This needs to be deprecated since Infura will soon cut support for interactions without an API key.

pipermerriam:
@kclowes I assigned this to you.  I don't know exact timeline but I think it's less than two months.  Can you see about getting a deprecation warning in for this ASAP.  I think it needs to get backported to v4 as well.

PR #1247, opened 2019-02-18
Deprecate no infura api key in v5

kclowes:
### What was wrong?
Starting March 27th, Infura will officially stop supporting requests with no API key. Infura calls it a project ID, so I allow users to use either `WEB3_INFURA_API_KEY`, or `WEB3_INFURA_PROJECT_ID` env var. I took out the ability to use `INFURA_API_KEY` for this branch, but will leave it in for the v4 branch.

There is already a warning in place if there is no API key, but I updated the message. After March 27th, we should consider throwing an error if there is no key so the user can have a nice message.

Related to Issue #1246 

### How was it fixed?
See above.

#### Cute Animal Picture
bulldog-puppy-paw-up

### DIFF
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -33,7 +33,7 @@ def test_load_provider_from_env(monkeypatch, uri, expected_type, expected_attrs)
         assert getattr(provider, attr) == val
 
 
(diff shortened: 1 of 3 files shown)


pipermerriam:
> After March 27th, we should consider throwing an error if there is no key so the user can have a nice message.

I think we should probably schedule the exception raising around a week before that date at the very latest to give people time to respond.  While this is technically a breaking change, I do think it should be backported to v4 since users would encounter a less friendly error message if we don't and thus, it's a breaking change either way.

PR #1250, opened 2019-02-20
Send Infura API requests to the correct URL with API key

kclowes:
### What was wrong?
Backport of #1247 to v4

Related to Issue #1246 

### How was it fixed?
See #1247 

#### Cute Animal Picture

bulldog-puppy-paw-up

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -342,8 +342,7 @@ Fetch various data about the current state of the ERC20 contract.
     '🦄'
     >>> erc20.functions.decimals().call()
     0
-    >>> erc20.functions.totalSupply().call()
-    2039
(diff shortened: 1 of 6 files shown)


docs/examples.rst
```diff
@@ -343,7 +343,7 @@ Fetch various data about the current state of the ERC20 contract.
     >>> erc20.functions.decimals().call()
     0
     >>> erc20.functions.totalSupply().call()
-    2039
+    2047
```
pipermerriam: Maybe lets go ahead and change this to be `>>> assert erc20.functions.totalSupply().call() > 2047` so we can be done with this nonsense.

Issue #1244, opened 2019-02-15
local_filter_middleware example doesn't work


jgeary:
* Version: 5.0.0a5

### What was wrong?

I upgraded from 4.8.2 to 5.0.0a5 to use the local filter middleware so I don't have to deal with the node deleting my filters. Unfortunately, I've been unable to use `local_filter_middleware`, which I believe is due to incorrect / lacking documentation.

According to the docs, here is the proper usage of `local_filter_middleware`:
```
from web3.middleware import local_filter_middleware
w3.middleware_onion.add(local_filter_middleware())
```

When I do this, I get the error `local_filter_middleware() missing 2 required positional arguments: 'make_request' and 'w3'`. Sure enough, the middleware does require those 2 arguments ([source](https://github.com/ethereum/web3.py/blob/master/web3/middleware/filter.py#L284)). There are no docs or even tests from the related PR (#732) that ever call `local_filter_middleware` with `w3` or `make_request`. I have no idea what to pass in for `make_request`. I found it odd that I was calling `local_filter_middleware()` at all, but `middleware_onion.add(local_request_middleware)` doesn't work either.

### How can it be fixed?

Please let me know what arguments I'm supposed to pass into `local_filter_middleware()`, and consider editing the relevant documentation for the next release so others don't need to ask.

dylanjw:
The doc example shouldn't have the parentheses.  It should be:

```python
from web3.middleware import local_filter_middleware
w3.middleware_onion.add(local_filter_middleware)
```

dylanjw:
> `middleware_onion.add(local_request_middleware)` doesn't work either.

@jgeary What issue did you have? Im not getting an error adding the middleware this way.

```
Python 3.6.4 (default, Jun  5 2018, 18:04:05) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from web3 import Web3, EthereumTesterProvider
>>> from web3.middleware import local_filter_middleware
>>> w3 = Web3(EthereumTesterProvider())
>>> w3.middleware_onion.add(local_filter_middleware)
>>> 
```

jgeary:
@dylanjw thanks for getting back about this. im using an asyncio event loop to watch for new events from multiple different filters in a continuous loop, and it seems like this middleware isn't compatible with that. here is what im working with: (modified to make it more generic and ignore unnecessary bits):

`
import web3
from web3 import Web3, WebsocketProvider
[...]

PR #1253, opened 2019-02-23
Fix example for adding local_filter_middleware to middlewares

dylanjw:
### What was wrong?

Related to Issue #1244

### How was it fixed?

Removed the parentheses. 

#### Cute Animal Picture

image

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -381,7 +381,7 @@ retrieved using JSON-RPC endpoints that don't rely on server state.
 .. code-block:: python
 
     >>> from web3 import Web3, EthereumTesterProvider
-    >>> w3 = Web3(EthereumTesterProvider)
+    >>> w3 = Web3(EthereumTesterProvider())
     >>> from web3.middleware import local_filter_middleware
     >>> w3.middleware_onion.add(local_filter_middleware)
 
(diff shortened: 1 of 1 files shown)


docs/middleware.rst
```diff
@@ -383,7 +383,7 @@ retrieved using JSON-RPC endpoints that don't rely on server state.
     >>> from web3 import Web3, EthereumTesterProvider
     >>> w3 = Web3(EthereumTesterProvider)
```
kclowes: I think EthereumTesterProvider should have parens too? 

`w3 = Web3(EthereumTesterProvider())`

dylanjw: Doh! Thanks.

docs/middleware.rst
```diff
@@ -381,9 +381,9 @@ retrieved using JSON-RPC endpoints that don't rely on server state.
 .. code-block:: python
```
pipermerriam: Can we update these to be doctests so we catch these types of errors in CI.  Probably good to revert the changes and confirm the doctest failures and then re-apply the fixes to ensure that doctest is doing what it's supposed to.

kclowes:
@dylanjw do you have time to convert this to a doctest? No worries if not :)

dylanjw:
@kclowes I will try to get to it later this week.

PR #1255, opened 2019-02-23
Make local block filter return empty list when when no blocks mined 

dylanjw:
### What was wrong?

Related to Issue #1253

#### Cute Animal Picture

image

### DIFF
--- a/web3/middleware/filter.py
+++ b/web3/middleware/filter.py
@@ -277,7 +277,7 @@ class RequestBlocks:
 def block_hashes_in_range(w3, block_range):
     from_block, to_block = block_range
     if from_block is None or to_block is None:
-        return None
+        return
     for block_number in range(from_block, to_block + 1):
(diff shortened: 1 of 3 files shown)


dylanjw:
@kclowes @njgheorghita Do either of you have a moment to review this?

web3/middleware/filter.py
```diff
@@ -277,6 +277,8 @@ def get_filter_changes(self):
 @to_list
 def block_hashes_in_range(w3, block_range):
     from_block, to_block = block_range
+    if from_block is None or to_block is None:
+        return None
```
pipermerriam: I think this should just be `return`.  The *value* from a `return` statement in a generator doesn't actually get returned.

pipermerriam: If we get `flake8-bugbear` added to the lint run for this repo it will catch things like this (by explicitely enabling [`B901`](https://github.com/PyCQA/flake8-bugbear#opinionated-warnings))

PR #1268, opened 2019-03-06
Add custom exceptions for jsonrpc lookups that return none

njgheorghita:
### What was wrong?
Rather than simply raising a `ValueError` when jsonrpc lookup calls returned `None`, raise a custom exception with better info.

re. [comment here](https://github.com/ethereum/web3.py/pull/1218/files/cf6c556013214a8ab05c08b537107e7fe229c907#r262141215)

#### Cute Animal Picture
image

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -46,7 +46,9 @@ from web3.contract import (
     Contract,
 )
 from web3.exceptions import (
+    BlockNotFound,
     TimeExhausted,
+    TransactionNotFound,
 )
 from web3.iban import (
     Iban,
@@ -142,10 +144,13 @@ class Eth(Module):
             if_number='eth_getBlockByNumber',
         )
 
-        return self.web3.manager.request_blocking(
-            method,
-            [block_identifier, full_transactions],
-        )
+        try:
+            return self.web3.manager.request_blocking(
+                method,
+                [block_identifier, full_transactions],
+            )
+        except ValueError:
+            raise BlockNotFound(f"Block with id: {block_identifier} not found.")
 
     def getBlockTransactionCount(self, block_identifier):
         """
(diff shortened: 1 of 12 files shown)


web3/eth.py
```diff
@@ -158,10 +163,13 @@ def getBlockTransactionCount(self, block_identifier):
...
-        return self.web3.manager.request_blocking(
-            method,
-            [block_identifier],
-        )
+        try:
+            return self.web3.manager.request_blocking(
+                method,
+                [block_identifier],
+            )
+        except ValueError:
```
pipermerriam: Does this `ValueError` come from an *error* type response from the JSON-RPC server?  Is this client specific behavior (like geth returns an error response and parity returns empty)?   Struggling to remember how all of this works.

I ask because it feels a little worrisome to be blindly converting `ValueError -> ThingNotFound` as this would be an easy way for our exception messaging to be wrong in a case where the `ValueError` means something else but we're blindly reporting it as a `ThingNotFound`.

Is there a way we can be more precise with this?

njgheorghita: @pipermerriam The `ValueError` comes from [here](https://github.com/ethereum/web3.py/blob/master/web3/manager.py#L99) which I implemented in the previously-related pr to be raised everytime a jsonrpc call returns `None` - off the top of my head, I don't think it's client-specific behavior meaning that it would probably require some more complexity to have a more precise error message, but if you 👍 I'll dig deeper and see about improving the error reporting

pipermerriam: I think we should drop at internal raising of the `ValueError` and do the `None` detection within each of the individual methods.  The current `ValueError` based approach has a strong potential for false positives, and completely disallowing `None` returns from `JSON-RPC` calls globally probably isn't a tenable API (sorry I didn't catch that previously).

PR #1271, opened 2019-03-11
Add deprecation warning to jsonrpc endpoints dropped in EIP 1474

njgheorghita:
### What was wrong?
Add deprecation warnings for jsonrpc endpoints dropped in [EIP 1474](https://github.com/bitpshr/EIPs/blob/147609158191a701cf02c8556d5e529ffcb31cd2/EIPS/eip-1474.md)

#### Cute Animal Picture
image

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -24,6 +24,7 @@ from web3.utils.blocks import (
 )
 from web3.utils.decorators import (
     deprecated_for,
+    deprecated_in_v5,
 )
 from web3.utils.empty import (
     empty,
@@ -374,6 +375,7 @@ class Eth(Module):
     def setContractFactory(self, contractFactory):
         self.defaultContractFactory = contractFactory
 
+    @deprecated_in_v5
     def getCompilers(self):
         return self.web3.manager.request_blocking("eth_getCompilers", [])
 
--- a/web3/net.py
+++ b/web3/net.py
@@ -1,6 +1,9 @@
 from web3.module import (
     Module,
 )
+from web3.utils.decorators import (
+    deprecated_in_v5,
+)
 
 
 class Net(Module):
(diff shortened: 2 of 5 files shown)


web3/eth.py
```diff
@@ -374,6 +374,7 @@ def contract(self,
     def setContractFactory(self, contractFactory):
         self.defaultContractFactory = contractFactory
 
+    @deprecated_for("Dropped in EIP 1474")
```
pipermerriam: I think the message a user will get is going to be awkward

https://github.com/ethereum/web3.py/blob/e578dd266ecff2e01398407cd6472894ddb59bc6/web3/utils/decorators.py#L54

This probably warrants just manually raising the proper deprecation warning exception.  Similarly, in v5, we should potentially not fully remove the methods, but make them do a `raise DeprecationWarning()` which *I believe* that when raised as an exception is just like normal exception raising.  This way people aren't greeted with an `AttributeError` but a nice explanation of why the method they are accessing is missing.

pipermerriam: So warnings don't break codeflow, they just dump information out so a user who's watching for them knows something is going away.  Using the `raise` syntax breaks code flow so they are two separate mechanisms.  These can be easy to mixup since the exception class is called `DeprecationWarning` but they are to different mechanisms that use the same class.

- `warning.warn(...)` - for `v4`
- `raise DeprecationWarning`  - for `v5`

kclowes: Are we switching to the `raise DeprecationWarning` for all methods we're deprecating in v5? Or just specific ones?

pipermerriam: `raise DeprecationWarning` should only be used if it was *previously* deprecated meaning that users of `v4` would have already been warned.

Otherwise it should use `warning.warn(...)` to begin giving users a warning

PR #1275, opened 2019-03-12
Move txpool to geth namespace

njgheorghita:
### What was wrong?
Since only geth supports `txpool_*` jsonrpc endpoints, and they are also not defined in EIP 1474, `web3.txpool` was moved to `web3.geth.txpool`

#### Cute Animal Picture
image


njgheorghita:
@kclowes Good catch! Also forgot to update the docs 😬

Issue #1281, opened 2019-03-15
Remove web3.utils in favor of web3._utils


kclowes:
* Version: 5a7

### What was wrong?
When installing v5a7, someone got an `ImportError: cannot import name 'compat'`. Commenting out the line  https://github.com/ethereum/web3.py/blob/master/web3/utils/__init__.py#L5 fixed the error. 

### How can it be fixed?
`web3.utils` only existed for backward compatibility, so can be removed in v5.

PR #1282, opened 2019-03-15
Remove web3/utils folder

kclowes:
### What was wrong?
An `importerror` was getting thrown because of the no longer existing `compat` module that was getting imported in `web3/utils/__init__.py`

Related to Issue #1281 

### How was it fixed?
Removed the utils folder since we don't need it for backwards compatibility in v5

#### Cute Animal Picture
download


kclowes:
Ah, I like that. Done!

Issue #1273, opened 2019-03-12
Relocating jsonrpc endpoints


njgheorghita:
### What was wrong?
Web3's api needs updating to match [EIP 1474](https://github.com/bitpshr/EIPs/blob/147609158191a701cf02c8556d5e529ffcb31cd2/EIPS/eip-1474.md), and the [Parity](https://wiki.parity.io/JSONRPC) / [Geth](https://github.com/ethereum/go-ethereum/wiki/Management-APIs) supported jsonrpc endpoints.

#### Configuration.
Only endpoints listed in EIP 1474 should be available under the `web3` namespace
- eg. `eth_signTransaction` -> `web3.eth.signTransaction`

Endpoints supported by Geth but not defined in EIP 1474 should be available under the `web3.geth` namespace
- eg. `txpool_status` -> `web3.geth.txpool.status`

Endpoints supported by Parity but not defined in EIP 1474 should be available under the `web3.parity` namespace
- eg. `shh_post` -> `web3.parity.shh.post`

EIP 1474 defined endpoints currently missing in Web3
- `eth_signTransaction`
- `eth_signTypedData`
- `eth_submitHashrate`
- `eth_submitWork`

What needs moving in Web3?
- `web3.shh` -> `web3.parity.shh` and `web3.geth.shh`
- `web3.version.ethereum` -> `web3.eth.protocolVersion`
- `web3.version.node` -> ??? (jsonrpc method = `web3_clientVersion`)
- deprecate `web3.version` api entirely?
- `web3.txpool` -> `web3.geth.txpool`
- `web3.miner` -> `web3.geth.miner`
- `web3.admin` -> `web3.geth.admin`

@kclowes 

### How can it be fixed?

Fill this section in if you know how this could or should be fixed.

njgheorghita:
@kclowes From what I can tell, we're very close to wrapping this up. One concern I have is that the [reference](https://github.com/ethereum/go-ethereum/wiki/Management-APIs) I've been using for the Geth-implemented endpoints might be outdated (last updated 2017). Have you come across any better resources?

As of now, `web3.version` only has one endpoint left in it, `web3.version.api` which returns the web3 api version number. IMO, it's worth considering moving this endpoint to the `web3` namespace and deprecating the `web3.version` namespace entirely. Thoughts?

[...]

PR #1290, opened 2019-03-19
Relocate web3.api and shuffle docs

njgheorghita:
### What was wrong?
Deprecate `web3.version` and move non-standard jsonrpc endpoints to parity/geth specific documentation pages.

see conversation [here](https://github.com/ethereum/web3.py/issues/1273#issuecomment-473990976)

#### Cute Animal Picture

image


web3/version.py
```diff
@@ -44,8 +44,7 @@ def ethereum(self):
 class Version(Module):
     @property
     def api(self):
-        from web3 import __version__
-        return __version__
+        raise DeprecationWarning("This method has been deprecated as of EIP 1474.")
```
kclowes: I know there was a comment about raising warnings vs not in a past PR, but why are we using one over the other? Also, maybe this warning should say something like: `This method has been deprecated ... Please use web3.api instead`?

njgheorghita: Good call on the warning msg! re other stuff: from [here](https://github.com/ethereum/web3.py/pull/1271#discussion_r264277304)

> So warnings don't break codeflow, they just dump information out so a user who's watching for them knows something is going away. Using the raise syntax breaks code flow so they are two separate mechanisms. These can be easy to mixup since the exception class is called `DeprecationWarning` but they are two different mechanisms that use the same class.

> warning.warn(...) - for v4
> raise DeprecationWarning - for v5

PR #1294, opened 2019-03-20
Add error if no Infura key is found

kclowes:
### What was wrong?
Infura will stop supporting API calls without a key on 3/27. 

Related to Issue #1246 

### How was it fixed?
Now an error is thrown when connecting to Infura if there is no API key found. 

#### Cute Animal Picture

maxresdefault

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 import time
 import warnings
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -1,9 +1,13 @@
 import importlib
-import logging
+import os
 import pytest
 
-from web3.auto import (
-    infura,
+from eth_utils import (
+    ValidationError,
+)
+
+from web3.exceptions import (
+    InfuraKeyNotFound,
 )
 from web3.providers import (
     HTTPProvider,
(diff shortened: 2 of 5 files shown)


conftest.py
```diff
@@ -13,6 +14,9 @@
 )
 
 
+os.environ['WEB3_INFURA_PROJECT_ID'] = 'test-key'
```
kclowes: This seems... hacky. Let me know if there is a better way. Since I'm raising the error now instead of logging a warning, I can't run tests without this line

pipermerriam: You should be using the pytest `monkeypatch` fixture.  It has a mechanism for setting the variable in such a way that it reverts the change after the test/tests run.

pipermerriam: https://docs.pytest.org/en/latest/monkeypatch.html

tests/core/providers/test_auto_provider.py
```diff
@@ -123,3 +113,11 @@ def test_web3_auto_infura_websocket_default(monkeypatch, caplog, environ_name):
     w3 = infura.w3
     assert isinstance(w3.provider, WebsocketProvider)
     assert getattr(w3.provider, 'endpoint_uri') == expected_url
+
+
+def test_web3_auto_infura_raises_error_with_nonexistent_scheme(monkeypatch):
+    monkeypatch.setenv('WEB3_INFURA_API_KEY', 'test')
+    monkeypatch.setenv('WEB3_INFURA_SCHEME', 'not-a-scheme')
+
+    with pytest.raises(ValidationError):
```
pipermerriam: maybe add a `match` keyword to this to ensure the validation error is the expected one (only really relevant if there are multiple paths for `ValidationError` to be raised.

PR #1295, opened 2019-03-21
Implement eth_chainId endpoint

njgheorghita:
### What was wrong?
Implement `eth_chainId` re: @carver 's comment [here](https://twitter.com/pedrouid/status/1107976163312455680?s=09) & on Gitter.

Though, it appears it's not implemented in Geth yet? Or just not in the versions we're testing against. Also, Parity simply returns a `null`, which according to [this](https://wiki.parity.io/JSONRPC-eth-module#eth_chainid) means that it's unavailable.

#### Cute Animal Picture

image

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -153,6 +153,18 @@ The following properties are available on the ``web3.eth`` namespace.
        '63'
 
 
+.. py:attribute:: Eth.chainId
+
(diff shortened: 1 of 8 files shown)


carver:
You should be able to use https://github.com/ethereum/go-ethereum/pull/17617 to backtrack which version is needed. Would be nice to update the geth integration version to test out eth_chainId

PR #1296, opened 2019-03-21
Raise Infura error v4 backport

kclowes:
### What was wrong?
Infura will no longer allow API access without a key after 3/27.

Related to Issue #1246 

### How was it fixed?
Now, an error will be raised when infura is loaded if there is no API key.

#### Cute Animal Picture

download-1

### DIFF
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -1,5 +1,5 @@
 import importlib
-import logging
+import os
 import pytest
 import sys
 from tempfile import (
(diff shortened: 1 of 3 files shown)


tests/core/providers/test_auto_provider.py
```diff
@@ -25,6 +25,13 @@
 TEMP_DIR = gettempdir()
 
 
+# Need to have env var set before import or an error is raised
+os.environ['WEB3_INFURA_API_KEY'] = 'test'
+from web3.auto import (  # noqa E402 isort:skip
+    infura,
+)
+
```
kclowes: This also feels gross. I think it's my favorite option though because it keeps the weirdness contained to this test file. My next favorite option is just to have a normal environment variable set in CI and then make anyone developing have an environment variable set locally. Totally open to other options ¯\_(ツ)_/¯

PR #1298, opened 2019-03-25
Allow infura secret

kclowes:
### What was wrong?
Originally, I thought I read somewhere that the infura-provided secret wouldn't be used for anything right now. Then, I read recently that you can check a box to make it required.

Related to Issue #1246 

### How was it fixed?
Added the ability for people to send in a secret key to the Websocket and HTTP providers.

#### Cute Animal Picture
download

### DIFF
--- a/web3/providers/auto.py
+++ b/web3/providers/auto.py
@@ -25,12 +25,12 @@ def load_provider_from_environment():
     return load_provider_from_uri(uri_string)
 
 
-def load_provider_from_uri(uri_string):
+def load_provider_from_uri(uri_string, headers=None):
     uri = urlparse(uri_string)
     if uri.scheme == 'file':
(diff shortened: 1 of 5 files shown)


web3/providers/auto.py
```diff
@@ -25,12 +25,12 @@ def load_provider_from_environment():
     return load_provider_from_uri(uri_string)
 
 
-def load_provider_from_uri(uri_string):
+def load_provider_from_uri(uri_string, headers=None):
```
pipermerriam: This seems to be fine for now but we *might* want to abstract this in the future to just be `provider_init_kwargs` and have it just be a pass-through, though as @carver has mentioned in the past, these types of passthroughs can end up being really hard to track and reason about, especially in any situation where there is more than a single level of it happening.

PR #1300, opened 2019-03-26
Add Nodesmith auto-provider

sdesmond46:
### What was wrong?

[Nodesmith](https://nodesmith.io) is a new JSON RPC service which provides access to hosted nodes. This change adds a Nodesmith auto-provider which can be used to interact with Ethereum networks.

Related to Issue: N/A

### How was it fixed?

Added a new provider and unit tests for it.
Also added a goerli provider for Infura.

#### Cute Animal Picture

DogChain

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -176,6 +176,28 @@ Then set the environment variable ``WEB3_INFURA_PROJECT_ID`` with your Project I
     >>> w3.isConnected()
     True
 
+Nodesmith
+~~~~~~~~~
+
+To connect to Nodesmith's mainnet or various testnet endpoints, first register
+for a free API Key at https://nodesmith.io.
+
+Then set the environment variable ``NODESMITH_API_KEY`` with your API key::
+
+    $ export NODESMITH_API_KEY=YourApiKey
+
+In your code, import the appropriate module for the network you want to connect to
+with the format ``web3.auto.nodesmith.<network_name>``. For example, to connect to the
+`Goerli <https://goerli.net/>`_ test network:
+
+.. code-block:: python
+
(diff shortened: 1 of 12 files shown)


pipermerriam:
I wasn't aware of your service.  I think we will want to have a discussion about this type of functionality.  For some context.

- Infura was added back when it was largely the only option of it's kind.
- Adding more is likely to set a precedent that I'm hesitant about as more cloud providers show up.
- I still like users having an easy way for quick direct connection like this.
- I am not keen on the maintenance overhead that this will result in as recently seen by infura's change to require API keys.

sdesmond46:
> I wasn't aware of your service. I think we will want to have a discussion about this type of functionality. For some context.
> 
> * Infura was added back when it was largely the only option of it's kind.
> * Adding more is likely to set a precedent that I'm hesitant about as more cloud providers show up.
> * I still like users having an easy way for quick direct connection like this.
> * I am not keen on the maintenance overhead that this will result in as recently seen by infura's change to require API keys.

Thanks for taking a look at the PR! We definitely understand the hesitation to set a precedent which could require future maintenance, but we do think it's important for people to have clear options and alternatives. We're happy to provide give feedback on and help implement any new patterns for getting new cloud providers included in the library.

fubuloubu:
Maybe there's a way to abstract out a "websocket provider" that packages can be built for separately from the web3 library, but can hook in easily as a provider option. Same with Infura (just switch them out e.g. `pip install web3-nodesmith-provider`)

marcgarreau:
Confirming that 3rd-party packages will be the future of auto-providers. Related issue: https://github.com/ethereum/web3.py/issues/1736.

PR #1301, opened 2019-03-26
Update geth version for integration tests

njgheorghita:
### What was wrong?
- update geth version for integration tests from `1.8.1` to `1.8.22` (inspired by [convo here](https://github.com/ethereum/web3.py/pull/1295#issuecomment-475323375))

- update go version used in integration tests

- removed unused geth 1.6 fixture data

- xfail test for `eth_submitHashrate` in `1.8.22` geth tests as it appears it has been dropped in favor of `ethash_submitHashRate`

#### Cute Animal Picture
image

PR #1308, opened 2019-03-29
Update shh module & relocate to parity and geth namespaces

njgheorghita:
### What was wrong?

- Updated `Shh` module to V2
- Exposed supported Whisper endpoints in `GethShh` and `ParityShh`
- Hooked up integration tests for supported endpoints for geth and parity

#### Cute Animal Picture
image


web3/shh.py
```diff
@@ -1,79 +1,179 @@
...
+    )
 
-    def deleteKeyPair(self, id):
-        return self.web3.manager.request_blocking("shh_deleteKeyPair", [id])
 
-    def hasKeyPair(self, id):
-        return self.web3.manager.request_blocking("shh_hasKeyPair", [id])
+def post_munger(module, message):
+    if message and ("payload" in message):
+        return (message,)
```
njgheorghita: @dylanjw I'm not sure if this is a misunderstanding on my end or a bug with the munger, but this was the only way I could successfully use this munger. From what I understand this should return something like `return module, [message]` - but that wouldn't work.

docs/web3.shh.rst
```diff
@@ -7,9 +7,13 @@ SHH API
...
 
+Web3.py exposes select Whisper methods under the ``web3.geth`` and ``web3.parity`` 
+namespaces to interact with the Whisper RPC API based on which endpoints are 
+supported by each respective client.
+
+
 .. warning:: The Whisper protocol is in flux, with incompatible versions supported
-    by different major clients. So it is not currently included by default in the web3
-    instance.
+    by different major clients.
```
njgheorghita: These docs definitely need to be split up & moved - but I was thinking of doing that in a following pr since it's dependent on #1290 which handles splitting up the geth & parity docs. If you think it belongs on this pr, I can include it here once #1290 gets approved & merged.

web3/parity.py
```diff
@@ -18,6 +18,44 @@
...
+    subscribe,
+    unsubscribe,
+)
+
+
+class ParityShh(ModuleV2):
+    """
+    https://wiki.parity.io/JSONRPC-shh-module
+    """
+    info = info()
```
dylanjw: It might be clearer if these methods were just assigned to variable names, rather than the output of these functions.  `info = Method(...)` and then class attribute assignment could lose the parens.  `info = info`.

web3/shh.py
```diff
@@ -1,79 +1,179 @@
-from web3._utils.filters import (
-    ShhFilter,
-)
-from web3.module import (
-    Module,
+from web3.method import (
```
dylanjw: It might be good to move this whole file to `utils` now, since nothing in here works without being assigned to a module. People might think that `web3.ssh.*` is part of the api.

web3/shh.py
dylanjw: mungers only need to return the arguments, and not the module object.  This was changed, and it looks like I didnt update the docstring in the Method class.

So what you have here is correct.

tests/integration/parity/common.py
```diff
@@ -167,36 +170,14 @@ def test_eth_getLogs_without_logs(self, web3, block_with_txn_with_log):
...
-        super().test_eth_call_with_0_result(web3, math_contract, math_contract_address)
+class CommonParityShhModuleTest(ParityShhModuleTest):
+    def test_shh_sync_filter(self, web3):
+        pytest.xfail("Skip until parity filter bug is resolved")
+        super().test_shh_sync_filter(web3)
 
-    def test_trace_filter(self, web3, txn_filter_params, parity_fixture_data):
-        super().test_trace_filter(web3, txn_filter_params, parity_fixture_data)
+    def test_shh_async_filter(self, web3):
+        pytest.xfail("Skip until parity filter bug is resolved")
```
pipermerriam: Should have a comment linking to whatever issue this is referring to.

web3/_utils/module_testing/shh_module.py
```diff
@@ -0,0 +1,389 @@
...
+        sender = web3.parity.shh.newKeyPair()
+        sender_pub = web3.parity.shh.getPublicKey(sender)
+
+        receiver = web3.parity.shh.newKeyPair()
+        receiver_pub = web3.parity.shh.getPublicKey(receiver)
+
+        topic = '0x13370000'
+        payloads = [web3.toHex(text="test message :)"), web3.toHex(text="2nd test message")]
+
+        shh_filter_id = web3.parity.shh.newMessageFilter({
```
pipermerriam: There seems to be a *lot* of similarity between these two implementations (geth and parity) of the tests.  Thoughts on trying to collapse these into something a little more DRY?

web3/geth.py
```diff
@@ -33,6 +33,31 @@
     sign,
     unlockAccount,
 )
+from web3.shh import (
+    addPrivateKey,
```
pipermerriam: I think this large number of imports could be sufficient justification to just import the `ssh` module and to access the methods as properties of the module (if you're inclined to do it that way).

web3/shh.py
pipermerriam: If we're going to do it that way, the `mungers` property on the `Method` class needs to be converted to a tuple as well as auditing it for any other mutable properties to ensure that modifications to a `Method` object don't have unintended side effects.

web3/_utils/module_testing/shh_module.py
njgheorghita: I'd like to, but imo they aren't as similar as they may appear. Besides the different shh modules being tested against, the way parity implements `post`, `newMessageFilter`, `info` and `version` all differ in the params accepted by the jsonrpc endpoint, and the type of data returned. The only tests that are the same are `test_shh_asymmetric_key_pair` and `test_shh_symmetric_key_pair`. Beyond that it seems to me like sharing tests might require a fair amount of config overhead that might end up more confusing

web3/shh.py
njgheorghita: > as well as auditing it for any other mutable properties to ensure that modifications to a `Method` object don't have unintended side effects.

Can you explain this a little further - what are some potential side effects? what are the other mutable properties we should check?

njgheorghita:
@pipermerriam @kclowes ping for final review.

Also, geth v1.7.2 integration tests just broke all of a sudden
```
ci.go:180: You have Go version go1.12.1
ci.go:181: go-ethereum requires at least Go version 1.7 and cannot
ci.go:182: be compiled with an earlier version. Please upgrade your Go installation.
```
Makes no sense to me why this happened, especially since geth v1.7.2 tests have been compiling happily with go1.12 ever since I updated the go version in #1301 

Also raises the question, why are we running integration tests against geth 1.7.2 and 1.8.22? Is it fine to just drop the 1.7.2 tests as they're becoming more and more outdated?

kclowes:
I am not sure why this _just_ broke, but it's due to [this line](https://github.com/ethereum/go-ethereum/blob/release/1.7/build/ci.go#L182) in 1.7. Apparently go1.12 is less than go1.7. I can't find how long they will support geth 1.7 (or if they even still are supporting 1.7?), but the last 1.7 release was Nov. 2017. So I guess I see 3 options: 
1) Downgrade the go version to 1.9 (which is unsupported) here
2) Backport the fix that's in geth 1.8 to 1.7 ([here](https://github.com/ethereum/go-ethereum/blob/release/1.8/build/ci.go#L218))
3) We drop integration tests for 1.7.
Like I said, I don't have a good feel for how long Geth 1.7 will be supported or how many people are still using geth 1.7, so I'm not sure what the right answer is, but I think I would naively lean toward option 1 and/or 2. I could also be persuaded otherwise.

docs/web3.geth.rst
```diff
@@ -447,3 +447,273 @@ The following methods are available on the ``web3.geth.txpool`` namespace.
...
+    * Returns the added key pair's ID
+
+    .. code-block:: python
+
+        >>>web3.geth.shh.addPrivateKey('0x7b8190d96cd061a102e551ee36d08d4f3ca1f56fb0008ef5d70c56271d8c46d0')
+        '86e658cbc6da63120b79b5eec0c67d5dcfb6865a8f983eff08932477282b77bb'
+
+.. py:method:: Shh.deleteKeyPair(self, id)
+
+    * Deletes the specifies key if it exists.
```
kclowes: *specified

njgheorghita:
@kclowes 
>  Apparently go1.12 is less than go1.7. 

where did you see that? I'm wholly unfamiliar with go haha - but that seems weird to me, also when it seems 1.12 is the [latest release].

pipermerriam:
What is the oldest version of go-ethereum that supports the mainnet?

njgheorghita:
@pipermerriam Not sure if I understand your question - but from what I can tell it's v1.0.0

> This is the first major and first live-net version of the go-ethereum client geth (1.0.0) and marks a major milestone in the history of Ethereum and the Go Ethereum codebase.

kclowes:
@njgheorghita I learned yesterday that you can compare strings in go. So in geth v1.7 they were comparing `runtime.Version() < 'go1.7'`, where in our case `runtime.Version() == go1.12`. But because it's not comparing numbers, I think what's happening is that it compares the first 5 characters and `go1.1` is less than `go1.7`. It looks like the geth team changed that line of code  in version 1.8 though. So 1.12 is definitely the latest go release, there is just a bug in how geth1.7 checks the go version.

pipermerriam:
@kclowes you can do this type of string comparison (and make the same mistake) in python as well (I've done it)

@njgheorghita I was intending to mean the *current* mainnet.  Basically, I'm wondering how useful it is for us to support any of the 1.7x line of go-ethereum

pipermerriam:
@njgheorghita ignore me, I now see this is the version of go, not goethereum

kclowes:
@pipermerriam Interesting! I can see where it would be a useful feature.

I think our problem has just fixed itself. So ignore all of the above. Now I really don't know what was going on 😆

njgheorghita:
Pffft that's frustrating, but who cares it works!!

PR #1315, opened 2019-04-05
Remove all remaining guess_tld kwargs

njgheorghita:
### What was wrong?
There were some lingering `guess_tld` kwargs that should have been removed in https://github.com/ethereum/web3.py/pull/1205

#### Cute Animal Picture
image


njgheorghita:
@kclowes - Seems like something strange is happening. Now all the geth1.7.1 tests are passing and only **some** of the 1.8.22 tests are failing

kclowes:
So weird. I just rebuilt and it looks like we're back in business!

.circleci/config.yml
```diff
@@ -81,8 +81,8 @@ geth_steps: &geth_steps
           export GOROOT=/usr/local/go
           export GETH_BINARY="$HOME/.py-geth/geth-$GETH_VERSION/bin/geth"
           if [ ! -e "$GETH_BINARY" ]; then
-            curl -O https://storage.googleapis.com/golang/go1.12.1.linux-amd64.tar.gz
-            tar xvf go1.12.1.linux-amd64.tar.gz
+            curl -O https://storage.googleapis.com/golang/go1.9.1.linux-amd64.tar.gz
```
kclowes: Wait... does this PR pass the test suite with go1.12? If it does, we might as well keep it there. If not, 1.9 is good

njgheorghita: oo good catch!

Issue #1193, opened 2019-01-08
Bug: Missing `constructor` in ABI results in silently ignoring constructor arguments.


pipermerriam:
https://github.com/ethereum/web3.py/blob/2a67ea9f0ab40bb80af2b803dce742d6cad5943e/web3/contract.py#L775-L786

### What was wrong?

In the case where you have a contract factory who's ABI **does not** contain a `constructor` function, then any arguments passed into the `contract_factory.constructor(...)` call are silently ignored.

### How can it be fixed?

In the event that the `else` block is hit indicating that the contract has no formal `constructor` function, a check should be made for whether any arguments were provided.  If either `args` or `kwargs` is non-empty, and exception should be raised.

PR #1316, opened 2019-04-05
1193 bugfix: don't silently ignore constructor args when no constructor function is provided

redshiftzero:
### What was wrong?

Fix bug #1193 (constructor arguments are passed and silently ignored when no constructor function is provided)

### How was it fixed?

Raise exception as suggested in the ticket

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_contract_constructor_encoding.py
+++ b/tests/core/contracts/test_contract_constructor_encoding.py
@@ -21,6 +21,19 @@ def test_contract_constructor_abi_encoding_with_constructor_with_no_args(SimpleC
 
 
 @pytest.mark.parametrize(
+    'args,kwargs',
+    (
+        (None, 'kwarg-is-ignored'),
+        ('arg-is-ignored', None),
+    ),
+)
+def test_contract_error_if_additional_args_are_supplied_with_no_constructor_fn(MathContract,
+                                                                               args, kwargs):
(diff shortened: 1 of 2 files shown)


tests/core/contracts/test_contract_constructor_encoding.py
```diff
@@ -20,6 +20,19 @@ def test_contract_constructor_abi_encoding_with_constructor_with_no_args(SimpleC
...
+@pytest.mark.parametrize(
+    'args,kwargs',
+    (
+        (None, 'kwarg-is-ignored'),
+        ('arg-is-ignored', None),
+    ),
+)
+def test_contract_error_if_additional_args_are_supplied_with_no_constructor_fn(MathContract,
+                                                                               args, kwargs):
+    with pytest.raises(TypeError):
```
pipermerriam: Can you tighten this test a little by adding `match="Constructor args"` to the `pytest.raises` call so that we're sure it's matching the correct error.

redshiftzero: good call, fixed & repushed

kclowes:
Thank you @redshiftzero! If you pull in the latest master, those failures should go away

redshiftzero:
thanks @kclowes @pipermerriam, this should be ready for re-review

kclowes:
Looks good to me! @pipermerriam do you have any other feedback?

Issue #1196, opened 2019-01-09
Convert all `toolz` imports to be from compat module `eth_utils.toolz`


pipermerriam:
Depends on https://github.com/ethereum/web3.py/pull/1195 which updates the `eth-utils` dependency to be `>=1.3.0`

### What was wrong?

We currently have `web3._utils.toolz` which handles the `toolz/cytoolz` import logic.  This compat module was added to `eth-utils` making this code redundant.

### How can it be fixed?

Convert all `toolz` imports to be from `eth_utils.toolz`

jmcph4:
Just for reference, as of cba95384, this **should** be the list of affected files:

    web3.py$ grep -r --include "*.py" toolz . | cut -f1 -d :
    ./web3/module.py
    ./web3/gas_strategies/time_based.py
    ./web3/eth.py
    ./web3/middleware/gas_price_strategy.py
    ./web3/middleware/exception_handling.py
    ./web3/middleware/validation.py
    ./web3/middleware/attrdict.py
    ./web3/middleware/geth_poa.py
    ./web3/middleware/signing.py
    ./web3/middleware/formatting.py
    ./web3/middleware/pythonic.py
    ./web3/middleware/pythonic.py
    ./web3/middleware/filter.py
    ./web3/middleware/normalize_errors.py
    ./web3/method.py
    ./web3/providers/eth_tester/middleware.py
    ./web3/providers/eth_tester/defaults.py
    ./web3/providers/eth_tester/defaults.py
    ./web3/providers/eth_tester/defaults.py
    ./web3/_utils/normalizers.py
    ./web3/_utils/formatters.py
    ./web3/_utils/contracts.py
    ./web3/_utils/abi.py
    ./web3/_utils/toolz/curried.py
    ./web3/_utils/toolz/curried.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/toolz/__init__.py
    ./web3/_utils/validation.py
    ./web3/_utils/encoding.py
    ./web3/_utils/filters.py
    ./web3/_utils/events.py
    ./web3/_utils/datatypes.py
    ./web3/_utils/rpc_abi.py
    ./web3/_utils/transactions.py
[...]

PR #1317, opened 2019-04-08
Replace `web3._utils.toolz` imports

jmcph4:
### What was wrong?

Related to Issue #1196.

### How was it fixed?

Basic text search-and-replace, via (From project root):

    $ vim $(grep -r --include "*.py" "web3._utils.toolz" | cut -f1 -d :)

Then, iterate over each file, running the following substitution once in Vim:

    %s/web3._utils.toolz/eth_utils.toolz/gI

#### Cute Animal Picture

Ragdoll cat


kclowes:
:tada: Thank you! As soon as the linting is fixed I'll merge!

web3/middleware/pythonic.py
```diff
@@ -1,6 +1,10 @@
 import codecs
 import operator
 
+from cytoolz.curried import (
```
kclowes: @pipermerriam It seems like this should either be 
```
from eth-utils.toolz import (
  curried
)
```
and then in the code we say `curried.keymap...` 
-or- 
we pull the logic from [this deleted file](https://github.com/ethereum/web3.py/pull/1317/files#diff-b022462abb470e6f70ffa82d41cd9641L1) into eth-utils. Do you have a preference?

pipermerriam: Yes, your example of curry-ing the function locally using the `eth-utils` import is correct.

Issue #1241, opened 2019-02-12
Implement recently added signing standards from Parity and Geth


tuxxy:
See:
- Geth EIP-191/712 -- https://github.com/ethereum/go-ethereum/pull/17789
- Parity EIP-191 -- https://github.com/paritytech/parity-ethereum/pull/9701
- Parity EIP-712 -- https://github.com/paritytech/parity-ethereum/pull/9631

An API should be implemented in web3py that allows for flexible utility between node backends with regards to the signing API proposed in the above issues. This API should also not require that the developer handle private key material. This API should rely on the account management of the respective backend node software.

- [ ] Geth EIP-191 API
- [ ] Geth EIP-712 API
- [ ] Parity EIP-191 API
- [x] Parity EIP-712 API

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1499.0 DAI (1499.0 USD @ $1.0/DAI)  attached to it as part of the nucypher fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/web3.py/1241/2339).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/web3.py/1241/2339).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $54,519.33 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [iamonuwa](https://gitcoin.co/profile/iamonuwa) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/web3.py/1241/2339?mutate_worker_action=approve&worker=iamonuwa) | [reject worker](https://gitcoin.co/issue/ethereum/web3.py/1241/2339?mutate_worker_action=reject&worker=iamonuwa))_.**

I've been following the development of the library + contributed in the past to the codebase. I would love to work with the team to implement this feature. I'll love to work on this task as it will enhance my understanding of the web3.py project as well.

[...]

PR #1319, opened 2019-04-08
Add support for eth_signTypedData and personal_signTypedData RPC call

Bhargavasomu:
### What was wrong?

Related to Issue #1241 

### How was it fixed?
The `eth_signTypedData` was created by mimicking the `eth_sign` implementation throughout the codebase. 

#### Cute Animal Picture


Bhargavasomu:
@kclowes I am getting the following error in most of the test cases on `circle ci`.
```
ValueError: Unknown RPC Endpoint: eth_signTypedData
```
I even tried by replacing `eth_signTypedData` with `account_signTypedData`, but I am getting the same error. Is there something I am missing? Thankyou.

kclowes:
@Bhargavasomu It doesn't look like `eth_signTypedData` has been released yet for Geth, so we can add an `xfail` to those tests. An `xfail` example can be seen here: https://github.com/ethereum/web3.py/blob/cba95384a367e979da67fe2a1f7cca13dc92eed5/tests/integration/go_ethereum/common.py#L20

For Parity, it looks like we'll have to add the `--jsonrpc-experimental` flag (from [this PR](https://github.com/paritytech/parity-ethereum/pull/9928)) when we start the parity node for the tests.

Bhargavasomu:
@kclowes I have added the methods `eth_signTypedData` and `personal_signTypedData`. Please review this and please let me know how to get the `integration-ethtester-pyevm` tests green.

kclowes:
@Bhargavasomu You are seeing the address error because you define elsewhere that the address is the first argument. So you need to switch your argument order here: https://github.com/ethereum/web3.py/pull/1319/files#diff-7416d0a42ff89713f12010ad00c62e41R247

Then I think we'll need to also add a line like this one for `signTypedData`: https://github.com/ethereum/web3.py/blob/master/tests/integration/test_ethereum_tester.py#L222 to `test_ethereum_tester.py` because it's not implemented in `EthereumTesterProvider`.

fubuloubu:
> Should be done separately but it would be good if our `EthereumTesterProvider` could get support added for these endpoints.

I thought about this, and doesn't this not make sense? Signing should happen outside of the node. At least these endpoints are not implemented here:

https://github.com/ethereum/web3.py/blob/f4a7d8c55238bc6f3afcb697c56a91d80e8cd13a/web3/providers/eth_tester/defaults.py#L204
https://github.com/ethereum/web3.py/blob/f4a7d8c55238bc6f3afcb697c56a91d80e8cd13a/web3/providers/eth_tester/defaults.py#L329

Bhargavasomu:
@kclowes I have made the necessary changes. Please review and let me know.

web3/_utils/rpc_abi.py
```diff
@@ -52,6 +52,7 @@
     'eth_sendTransaction': TRANSACTION_PARAMS_ABIS,
     'eth_signTransaction': TRANSACTION_PARAMS_ABIS,
     'eth_sign': ['address', 'bytes'],
+    'eth_signTypedData': ['address', None],
```
pipermerriam: I think a comment got dropped here.  I *think* validation at this level makes sense, but that is really just to ensure that the user gets a good error message if they pass in invalid data.

web3/_utils/module_testing/personal_module.py
```diff
@@ -169,3 +230,117 @@ def test_personal_sign_and_ecrecover(self,
...
+            "c1bd56fe1d745ffa101e74697108668c401c"
+        )
+        assert signature == expected_signature
+        assert len(signature) == 32 + 32 + 1
+
+    def test_invalid_personal_sign_typed_data(self,
+                                              web3,
+                                              unlockable_account_dual_type,
+                                              unlockable_account_pw):
+        invalid_typed_message = '''
```
pipermerriam: This test looks equivalent to the one above and I don't see anywhere that it asserts the signature being invalid.. am I missing something?

Bhargavasomu: @pipermerriam this was incomplete when I pushed it. It has now been modified so that an invalid test has been produced and made sure that an error is raised.

Bhargavasomu:
@kclowes I have implemented all that has been mentioned except the `Docs`. Please review and let me know. (Will update the docs now.)

/cc @pipermerriam

Bhargavasomu:
@kclowes The docs have been updated now too. Please review and let me know.

Bhargavasomu:
@kclowes yes that is why I didn't put up the validations as part of `web3.py`.
1. It can be used without a connection to the node.
2. Moreover I feel that all such validations should be done on the node side, rather than at `web3` side. (Please correct me if I am wrong.)

I have addressed the other issues. Please review once more. Thankyou.

web3/_utils/module_testing/eth_module.py
```diff
@@ -202,6 +203,107 @@ def test_eth_sign(self, web3, unlocked_account_dual_type):
...
+        # ValueError is caused by the error in the response which is as follows
+        # ``Expected 2 items for array type Person[2], got 1 items``
+        with pytest.raises(ValueError) as e:
+            web3.eth.signTypedData(
+                unlocked_account_dual_type,
+                json.loads(invalid_typed_message)
+            )
+        try:
+            assert "Expected 2 items for array type Person[2], got 1 items" in str(e.value)
+        except AssertionError:
```
kclowes: Since this is getting `xfail`ed anyway, I don't think you need this `try`/`except` here?

Bhargavasomu: @kclowes I thought the same too, but it is still being failed. I wasn't sure why this is happening and hence resorted to this method.

kclowes: In that case, there is a method called `skip_if_testrpc` that I think would be better to use instead.

Bhargavasomu:
@kclowes I have addressed the issue about `skip_if_testrpc`. Please review. Thankyou.

Bhargavasomu:
@pipermerriam is this good to merge with?

tuxxy:
Bump on this PR? We (@nucypher) are currently blocking on this and this would alleviate the pressure a ton.

docs/web3.eth.rst
```diff
@@ -654,6 +654,19 @@ The following methods are available on the ``web3.eth`` namespace.
         '0x1a8bbe6eab8c72a219385681efefe565afd3accee35f516f8edf5ae82208fbd45a58f9f9116d8d88ba40fcd29076d6eada7027a3b412a9db55a0164547810cc401'
 
 
+.. py:method:: Eth.signTypedData(account, jsonMessage)
+
+    * Delegates to ``eth_signTypedData`` RPC Method
+
+    Please note that the ``jsonMessage`` argument is the loaded JSON Object
+    and **NOT** the JSON String itself.
```
pipermerriam: Not a blocker but it does seem somewhat trivial for us to support it in both formats.  Can be done as a follow up.  Easier to add support than remove it.

tests/integration/parity/test_parity_http.py
```diff
@@ -49,6 +49,7 @@ def parity_command_arguments(
         '--password', passwordfile,
         '--jsonrpc-port', rpc_port,
         '--jsonrpc-apis', 'all',
+        '--jsonrpc-experimental',
```
pipermerriam: I'm confused,  Tests above suggest that new APIs are not in parity.  This line suggests it enables them (or enables something...???)

web3/_utils/module_testing/eth_module.py
```diff
@@ -202,6 +203,104 @@ def test_eth_sign(self, web3, unlocked_account_dual_type):
...
+                    "contents": "Hello, Bob!"
+                }
+            }
+        '''
+        with pytest.raises(ValueError) as e:
+            web3.eth.signTypedData(
+                unlocked_account_dual_type,
+                json.loads(invalid_typed_message)
+            )
+        assert "Expected 2 items for array type Person[2], got 1 items" in str(e.value)
```
pipermerriam: This should use the `match` keyword in `pytest.raises` to do it inline using pytest's built-in matching functionality.

tests/integration/parity/common.py
```diff
@@ -168,6 +168,22 @@ def test_eth_getLogs_without_logs(self, web3, block_with_txn_with_log):
...
+                               unlocked_account_dual_type):
+        pytest.xfail('eth_signTypedData JSON RPC call has not been released in parity')
+        super().test_eth_signTypedData(
+            web3, unlocked_account_dual_type
+        )
+
+    def test_invalid_eth_signTypedData(self,
+                                       web3,
+                                       unlocked_account_dual_type):
+        pytest.xfail('eth_signTypedData JSON RPC call has not been released in parity')
```
pipermerriam: Is there a reason these are not using the decorator method of marking xfails?  That approach is superior because it still has pytest run the test and reports un-expected test success as a failure to ensure that all of the `xfail` tests actually fail (rather than this approach which will cause the test to unconditionally fail).

web3/_utils/module_testing/personal_module.py
```diff
@@ -169,3 +231,111 @@ def test_personal_sign_and_ecrecover(self,
...
+                }
+            }
+        '''
+        with pytest.raises(ValueError) as e:
+            web3.parity.personal.signTypedData(
+                json.loads(invalid_typed_message),
+                unlockable_account_dual_type,
+                unlockable_account_pw
+            )
+        assert "Expected 2 items for array type Person[2], got 1 items" in str(e.value)
```
pipermerriam: Same here with using the `match` keyword in `pytest.raises`

tests/integration/parity/test_parity_http.py
kclowes: IIRC, personal_signTypedData is implemented in parity, but eth_signTypedData is not yet. It looks like Geth has them merged into master, but they haven't been released yet.

tests/integration/parity/common.py
kclowes: I changed it to the decorator for these tests, but the rest of the file has the `pytest.xfail` pattern. I added an issue to switch them over throughout the codebase (#1341)

docs/web3.eth.rst
kclowes: Added issue #1342 to track!

PR #1326, opened 2019-04-12
Change Admin RPC method casing

kclowes:
### What was wrong?
We have decided to move JSON RPC method names from camelCase to snake_case.

Related to Issue #722

### How was it fixed?
Changed methods to snake case and added deprecation warnings. 

#### Cute Animal Picture

download


kclowes:
@pipermerriam Before I get too much farther, I just wanted to get some feedback about the general approach. Is there a better way to handle the deprecation warnings?

tests/core/admin-module/test_admin_nodeInfo.py
```diff
@@ -1,7 +1,20 @@
...
+
+
 def test_admin_nodeInfo(web3, skip_if_testrpc):
     skip_if_testrpc(web3)
 
-    node_info = web3.geth.admin.nodeInfo
+    with pytest.warns(DeprecationWarning):
+        node_info = web3.geth.admin.nodeInfo
+
+        assert 'enode' in node_info
```
pipermerriam: Random idea: We end up wanting to talk about enode addresses in web3.py, trinity, and the devp2p sub-project in trinity as well.  Assuming that plugins become a thing in trinity, we'll want enode address information in 3rd party plugins as well as if we implement PSS or SHH protocols as trinity plugins.  Maybe we should have a small `enode` library that implements some basic primatives for working with and validating enodes.

web3/geth.py
```diff
@@ -74,14 +81,21 @@ class GethAdmin(ModuleV2):
     """
     https://github.com/ethereum/go-ethereum/wiki/Management-APIs#admin
     """
+    add_peer = add_peer
```
pipermerriam: Instead of interleaving these, maybe it would be good to keep all the non-deprecated ones together and to group all the deprecated ones below

web3/method.py
```diff
@@ -150,3 +151,16 @@ def _pipe_and_accumulate(val, fns):
...
+
+class DeprecatedMethod():
+    def __init__(self, method, old_name, new_name):
+        self.method = method
+        self.old_name = old_name
+        self.new_name = new_name
+
+    def __get__(self, obj=None, obj_type=None):
+        warnings.warn(
+            "%s is deprecated in favor of %s" % (self.old_name, self.new_name),
```
pipermerriam: We prefer `.format(...)` style string formatting over `%` (and try to stick with a single style per codebase).

web3/providers/eth_tester/defaults.py
```diff
@@ -264,14 +264,21 @@ def personal_send_transaction(eth_tester, params):
         'getMessages': not_implemented,
     },
     'admin': {
+        'add_peer': not_implemented,
```
pipermerriam: Same here with grouping these by active/deprecated.

docs/web3.geth.rst
```diff
@@ -27,8 +27,38 @@ The ``web3.geth.admin`` object exposes methods to interact with the RPC APIs und
...
+            },
+        }
+
+
 .. py:method:: nodeInfo()
 
+    .. warning:: Deprecated: This method is deprecated in favor of
+      :meth:`~web3.geth.admin.node_info()`
+
     * Delegates to ``admin_nodeInfo`` RPC Method
```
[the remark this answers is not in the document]
kclowes: Okay, so just have the deprecation message here and the other documentation above?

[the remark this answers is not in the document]
pipermerriam: That's my inclination.  Duplicating them seems like a problem since they'd almost definitely go out of sync.

tests/core/admin-module/test_admin_nodeInfo.py
kclowes: This sounds good to me. Is an issue in this codebase the best way to track that?

web3/method.py
carver: Actually, since web3 is py3.6+ as of v5, we can now move to f-strings as the standard! :tada:

f-strings are way better than either `%` or `format()`.

carver:
> Looks  quick ping to @carver for a   on the `DeprecatedMethod` descriptor approach.

Yup, approach looks fine to me!

web3/_utils/module_testing/go_ethereum_admin_module.py
```diff
@@ -0,0 +1,108 @@
...
+            'id': '',
+            'name': '',
+            'enode': '',
+            'ip': '',
+            'ports': AttributeDict({}),
+            'listenAddr': '',
+            'protocols': AttributeDict({})
+        })
+        # Test that result gives at least the keys that are listed in `expected`
+        assert len(expected.keys() - result.keys()) == 0
```
kclowes: This feels pretty janky... let me know if there is a better way. Different Geth versions return different `AttributeDict`s

kclowes:
@pipermerriam ping for another round of review. There weren't any integration tests, so I added some, but I'm not sure how much value they actually add. Let me know what feedback you have!

web3/_utils/module_testing/go_ethereum_admin_module.py
pipermerriam: integration tests are janky so seems fine to me.  You can do something like `assert not set(expected.keys())difference(result.keys)` but that's just a slightly different way to assert that they have the same keys.

web3/method.py
```diff
@@ -150,3 +151,16 @@ def _pipe_and_accumulate(val, fns):
...
+class DeprecatedMethod():
+    def __init__(self, method, old_name, new_name):
+        self.method = method
+        self.old_name = old_name
+        self.new_name = new_name
+
+    def __get__(self, obj=None, obj_type=None):
+        warnings.warn(
+            f"{self.old_name} is deprecated in favor of {self.new_name}",
+            category=DeprecationWarning, stacklevel=2)
```
pipermerriam: any chance this is going to suffer from the same issue you just fixed with `stacklevel`?

kclowes: Ah, nice catch!

PR #1333, opened 2019-04-15
Upgrade eth-tester backport

kclowes:
### What was wrong?
Same issue as #1332, just Backported to v4. Invalid opcode in Solidity 0.5.7 but not 0.5.3

Related to py-evm issue [#1748](https://github.com/ethereum/py-evm/issues/1748)

### How was it fixed?
Pinned eth-tester py-evm requirement to 0.1.0-beta.39

#### Cute Animal Picture

cute-baby-animals-2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import (
 
 extras_require = {
     'tester': [
-        "eth-tester[py-evm]==0.1.0-beta.33",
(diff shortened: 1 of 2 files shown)


carver:
At least one of the issues: v4 expects eth-tester to raise a `KeyError` when the sender account is missing, but eth-tester now raises a `ValidationError`. Since eth-tester is still experimental, I think it's okay to just change the API and expect the new exception.

PR #1348, opened 2019-05-06
Upgrade eth-account to v0.4

carver:
Update docs, fix backwards-incompatible import.

#### Cute Animal Picture


carver:
Note that this should not be backported to web3 v4. Also, I realized as I was updating the docs that the keys in returned dicts are still camel case. Not sure what to do with that. I guess it means including duplicates of the keys (camel case and snake case) in the v0.4 of eth-account, and dropping camel-case in v0.5. I can do that in eth-account v0.4.1, but probably won't get around to it for a bit.

Issue #1358, opened 2019-05-21
web3py.ens should not do IDNA processing


Arachnid:
To correctly normalise ENS names, they need to be normalised with UTS46, but not subjected to the IDNA rules beyond that. See [this thread](https://discuss.ens.domains/t/non-compliant-name-on-official-lookup-https-manager-ens-domains-name-rainbow-eth/992/10?u=nickjohnson) for details.

Issue #1357, opened 2019-05-21
ENS: Name normalisation is using `idna.decode` instead of `idna.encode`


joshuawu:
Normalisation function:

https://github.com/ethereum/web3.py/blob/abb3b631fb76eca21f02d420785c32c17ea9e760/ens/utils.py#L74-L91

The line in question:

https://github.com/ethereum/web3.py/blob/abb3b631fb76eca21f02d420785c32c17ea9e760/ens/utils.py#L89

Shouldn't this be `idna.encode(...)`? I don't see how this makes sense trying to decode the original input into Unicode. We want to encode it into ASCII.

pipermerriam:
Do you know if there are any test vectors from the ENS spec that would demonstrate the correct behavior so we can add a test?

joshuawu:
> Do you know if there are any test vectors from the ENS spec that would demonstrate the correct behavior so we can add a test?

There are no test vectors in the EIP #137 spec.

kclowes:
@joshuawu I believe decode is correct here because the purpose is to take any of the ASCII encoded urls `xn--...` and turn them into utf-8. See this testcase: https://github.com/ethereum/web3.py/blob/master/tests/ens/test_nameprep.py#L15

PR #1364, opened 2019-05-24
Use UTS46 domain name processing, not full IDNA processing

kclowes:
### What was wrong?
web3py was doing full IDNA processing, instead of UTS46.

Related to Issue #1358 and #1357 

### How was it fixed?
Added some emoji and other test cases that are valid according to UTS46 but not IDNA2008 to make sure they are `nameprep`ped correctly.

#### Cute Animal Picture

images


kclowes:
@carver From what I can tell, this stills works how it was intended, but I would appreciate your 👀to be sure. And, are there other test cases I should consider?

carver:
Nothing more to add, except there is a lot of room to improve the test suite of nameprep. I wouldn't assume that green means that everything is great. It's worth looking around to find a nice full set of fixtures to test against.

ens/utils.py
```diff
@@ -81,10 +81,12 @@ def normalize_name(name):
         return name
     elif isinstance(name, (bytes, bytearray)):
         name = name.decode('utf-8')
+    elif name[0:4] == 'xn--':
+        name = idna.decode(name, uts46=True, std3_rules=True)
```
carver: I guess this would cause any name encoded as `xn--` to reject the emoji?

ens/utils.py
```diff
@@ -81,10 +81,12 @@ def normalize_name(name):
...
     try:
-        return idna.decode(name, uts46=True, std3_rules=True)
+        return idna.uts46_remap(name, std3_rules=True)
```
carver: This appears to be where the failure about the leading `-` happened. I don't remember where that rule comes from, so it's worth making sure it's something actually required by ENS nameprep (and uts46).

ens/utils.py
kclowes: It would, but the `xn--` prefix is from IDNA encoding AIUI, so the only way to get to an `xn--` prefixed domain name is via `idna.encode` which throws an error if it encounters a character that isn't IDNA compliant (like an emoji).

Issue #1351, opened 2019-05-10
Event ProcessReceipt fails for events with same id but different indexed attributes


Uxio0:
* Version: 4.9.2
* Python: 3.7
* OS: linux

### What was wrong?

When I use `rich_logs = erc721.events.Transfer().processReceipt(tx_receipt)` I expect the relevant logs to be decoded and the others ignored. But if I try to process a tx with mixed `ERC20` and `ERC721` `Transfer` events (`ERC721` has the third parameter of the event indexed, `ERC20` not) it will fail decoding the event.

`ERC20` Transfer Event `event Transfer(address indexed from, address indexed to, uint256 value);` = `3 topics`
`ERC721` Transfer Event `event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);` = `4 topics`

Problem is that signature for `Transfer` is the same, but as `tokenId` is indexed for `ERC721` it will have one more `log topic` than `ERC20`. When `web3.py` tries to decode an `ERC20` event with the `ERC721` contract, it will fail as it's expecting one topic more.

> <ipython-input-16-bedb56c17231> in <module>
> ----> 1 erc721.events.Transfer().processReceipt(tx_receipt)
> 
> ~/.virtualenvs/gnosis-py/lib/python3.7/site-packages/web3/utils/decorators.py in _wrapper(*args, **kwargs)
>      12         def _wrapper(*args, **kwargs):
>      13             if obj is not None:
> ---> 14                 return self.method(obj, *args, **kwargs)
>      15             else:
>      16                 return self.method(objtype, *args, **kwargs)
> 
> ~/.virtualenvs/gnosis-py/lib/python3.7/site-packages/web3/contract.py in processReceipt(self, txn_receipt)
>    1277     @combomethod
>    1278     def processReceipt(self, txn_receipt):
> -> 1279         return self._parse_logs(txn_receipt)
>    1280 
>    1281     @to_tuple
> 
> ~/.virtualenvs/gnosis-py/lib/python3.7/site-packages/eth_utils/functional.py in inner(*args, **kwargs)
>      44         @functools.wraps(fn)
>      45         def inner(*args, **kwargs) -> T:  # type: ignore
> ---> 46             return callback(fn(*args, **kwargs))
>      47 
>      48         return inner
> 
> ~/.virtualenvs/gnosis-py/lib/python3.7/site-packages/web3/contract.py in _parse_logs(self, txn_receipt)
>    1283         for log in txn_receipt['logs']:
[...]

PR #1366, opened 2019-06-05
Event process receipt

### LINKED ISSUE #1330: Warnings during doc build: missing event-log-object
* Version: v5.0.0-alpha.9-37-gf4a7d8c5 (current `master`, commit f4a7d8c55238bc6f3afcb697c56a91d80e8cd13a)
* Python: 3.7.3
* OS: Arch Linux
* `pip freeze` output: MISSING

### What was wrong?

During `make docs`:

```
/home/veox/src/web3.py/docs/contracts.rst:714: WARNING: undefined label: event-log-object (if the link has no caption the label must precede a section header)
/home/veox/src/web3.py/docs/releases.rst:671: WARNING: undefined label: event-log-object (if the link has no caption the label must precede a section header)
```

Direct fallout of PR #1032.

### How can it be fixed?

* Review docs.
* Build docs in CI and fail on warnings.
* "Lint" docs.

kclowes:
### What was wrong?
There was no way to ignore or discard event log errors if they came up during processing.

Related to Issue #1351. Also fixes #1330 

### How was it fixed?
Added a new optional error kwarg to `processReceipt` to handle discarding and ignoring errors. 

I'll rebase before merging.

#### Cute Animal Picture
download


docs/contracts.rst
```diff
@@ -733,6 +732,61 @@ For example:
        >>> rich_logs[0]['args']
        {'myArg': 12345}
 
+   By default, an error will be raised if the log doesn't match the provided ABI.
+   However, errors can be optionally ignored by passing in the ``ignore`` or
+   ``discard`` flag located in the ``EventLogErrorFlags`` class. The ``discard``
+   flag will discard any logs that raise an error during processing, and return
```
pipermerriam: Maybe

> will **silently* discard any logs ...

docs/contracts.rst
```diff
@@ -733,6 +732,61 @@ For example:
...
+   the rest. The ``ignore`` flag will return the raw log that caused the error,
+   without any processing, and will process any logs that do not raise an error.
+   An ``EventLogErrorFlag`` is imported from web3/_utils/events.py.
```
pipermerriam: I'm realizing that we *probably* need to think about the real use cases with these.  Here are my thoughts.

- `discard` might be considered scary but probably useful one.
- `warn` might be a good new one that raises a warning and discards the log
- `ignore` probably needs a mechanism for reliable programatic differentiation which logs are properly processed and which logs threw errors.

docs/contracts.rst
```diff
@@ -733,6 +732,61 @@ For example:
...
+
+   .. code-block:: python
+
+       >>> from web3._utils.events import EventLogErrorFlags
```
pipermerriam: I'm thinking that the `EventLogErrorFlags` should be private and we only expose the individual flags for import.  Also, lets make them capcase since I think that lines up better with APIs elsewhere in the standard library. `DISCARD`, `IGNORE`

web3/_utils/module_testing/event_contract.py
```diff
@@ -0,0 +1,64 @@
+
+EVNT_CONTRACT_CODE = (
+    "6080604052348015600f57600080fd5b5061010b8061001f6000396000f30060806040526004361"
+    "0603f576000357c0100000000000000000000000000000000000000000000000000000000900463"
+    "ffffffff1680635818fad7146044575b600080fd5b348015604f57600080fd5b50606c600480360"
+    "38101908080359060200190929190505050606e565b005b7ff70fe689e290d8ce2b2a388ac28db3"
+    "6fbb0e16a6d89c6804c461f65a1b40bb15816040518082815260200191505060405180910390a17"
+    "f56d2ef3c5228bf5d88573621e325a4672ab50e033749a601e4f4a5e1dce905d481604051808281"
+    "5260200191505060405180910390a1505600a165627a7a72305820ff79430a04cf654d7b46edc52"
+    "9ccaa5d7f77607f54bb58210be0c48455292c810029"
```
pipermerriam: @njgheorghita and @kclowes probably already said this a few times but I think it makes sense to start dog-fooding the ethpm APIs and to package these up as ethpm packages to get these blobs of bytecode out of the tests.

web3/contract.py
```diff
@@ -983,16 +984,23 @@ def _get_event_abi(cls):
...
         for log in txn_receipt['logs']:
             try:
                 decoded_log = get_event_data(self.abi, log)
             except MismatchedABI:
                 continue
+            except ValueError as e:
+                if errors == EventLogErrorFlags.discard:
+                    continue
+                elif errors == EventLogErrorFlags.ignore:
+                    decoded_log = log
```
pipermerriam: wondering about renaming this to `rich_log` or something similar.

docs/contracts.rst
kclowes: That sounds fine to me. I couldn't decide 😆 I left them as lowercase for the future so that if we do end up defining them as methods, we wouldn't have `def STRICT`, but I think uppercase makes more sense for now 👍

docs/contracts.rst
kclowes: Yeah, agreed. I like the `warn` option.

For `ignore`, maybe we would return the log, with an additional errors field that we add? So the returned log might look like: 
```
["logs": {
    "address": "0x00000000000000000000",
    ...,
    "type": "mined",
    "errors": "ValueError: <error message>"
}]
```

pipermerriam: seems reasonable.

kclowes:
@pipermerriam I just have a couple things that could use your :eyes: when you get a chance: 
1) Do you have an opinion on where the `EventLogErrorFlags` class lives? right now it's in `web3/event_log_error_flags.py`. I had it in `web3/_utils/events.py`, but that didn't feel quite right since `_utils` implies private.
2) Does the `IGNORE` branch make sense to you in [this line](https://github.com/ethereum/web3.py/pull/1366/files#diff-2171ba22a13b6793b04c4a1287049a67R1011)? Is there a better way to add a field to an `AttributeDict`?

Thanks!

docs/contracts.rst
```diff
@@ -733,6 +733,63 @@ For example:
...
 
+   The default ``WARN`` flag logs a warning to the console if the log has errors, and the log that raised the warning will be discarded. Any logs that didn't encounter errors during processing will be returned.
+   There are also a few other flags available:
+
+     - ``STRICT`` - stops all processing and raises the error encountered.
+     - ``IGNORE`` - returns any raw logs that raised an error with an added "errors" field, along with any other logs were able to be processed.
+     - ``DISCARD`` - silently discards any logs that have errors, and returns processed logs that don't have errors.
+     - ``WARN`` - logs a warning to the console for the log that has an error, and discards the log. Returns any logs that are able to be processed.
+
+   An ``EventLogErrorFlag`` needs to be imported from web3/event_log_error_flags.py.
```
pipermerriam: Maybe just make them importable from` web3.logs` and don't mention the class `EventLogErrorFlag` anywhere in the docs.

Issue #1293, opened 2019-03-20
Support eth_chainId


carver:
geth and parity both support the endpoint now (admittedly, this is hearsay), so we should to.

Also, we could update the local signing middleware and transaction builder to include the connected chain id when signing.

Huge bonus points for opening a PR to add it to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md

kclowes:
- [x] eth_chainId implemented - #1295 
- [ ] Update local signing middleware and transaction builder to include the connected chainId when signing
- [ ] 🏅Bonus: Add PR to EIP-1474

cygnusv:
There's an inconsistency in the docs wrt `w3.eth.chainId`. They say it returns an [integer value](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.chainId), but in fact it just returns the original hex string defined by [EIP-695](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-695.md#abstract)

carver:
> it just returns the original hex string

Ah, sounds like a bug.

cygnusv:
My bad! It seems you guys already fixed this in the last release.
https://github.com/ethereum/web3.py/issues/1387

carver:
Great! I'll close this issue, then.

> * Update local signing middleware and transaction builder to include the connected chainId when signing

@kclowes IIRC, this already happened, right? ^

kclowes:
@carver Unless I'm missing something (very possible), I don't think we've added chainId to the middleware and transaction builder yet. I'll open a new issue with a better title though so it's less confusing.

carver:
Ok sounds good, thanks!

rbval:
Trying to bump my web3.py version from 4.8.0 to 5.1.0 in order to get support for `poll_latency` in `getTransactionReceipt`.

Seems like this breaks running the stack locally with docker due to `chainId` not being support yet:
```
      File "/usr/local/lib/python3.6/dist-packages/web3/_utils/transactions.py", line 32, in <lambda>
        'chainId': lambda web3, tx: web3.eth.chainId,
      File "/usr/local/lib/python3.6/dist-packages/web3/eth.py", line 106, in chainId
        return self.web3.manager.request_blocking("eth_chainId", [])
      File "/usr/local/lib/python3.6/dist-packages/web3/manager.py", line 97, in request_blocking
        raise ValueError(response["error"])
[...]

PR #1378, opened 2019-07-08
Change net.version to eth.chainId for default transaction params

cducrest:
### What was wrong?

On a chain using a different network id and chain id, the default parameters for a transaction would be incorrectly filled when using Parity / Geth. It would use the `net_version` RPC instead of `eth_chainId`

Parity's net_version RPC: https://wiki.parity.io/JSONRPC-net-module#net_version
Parity's eth_chainId RPC: https://wiki.parity.io/JSONRPC-eth-module#eth_chainid
Geth net_version RPC: https://github.com/ethereumproject/go-ethereum/wiki/JSON-RPC#net_version
Geth eth_chainId RPC: https://github.com/ethereumproject/go-ethereum/wiki/JSON-RPC#eth_chainId

Related to Issue https://github.com/ethereum/web3.py/issues/1293

### How was it fixed?

Change the `TRANSACTION_DEFAULTS` to use web3.eth.chainId
This works for Parity and Geth afaik. It does not comply with EIP1474: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md#description-4
But as noted in the related issue, the EIP1474 should be updated and the EIP is in draft state.

I could not write a proper test for it since I would need a test chain with a different network and chain id, and could not find a relatively quick way to do that.

#### Cute Animal Picture

Edit: I know I messed up and wrote `chaindId` instead of `chainId`, also looking into fixing some tests.

### DIFF
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -58,7 +58,7 @@ def test_build_transaction_not_paying_to_nonpayable_function(
         'data': '0xe4cb8f5c',
         'value': 0,
         'gasPrice': 1,
-        'chainId': 1,
+        'chainId': 61,
     }
 
 
@@ -79,7 +79,7 @@ def test_build_transaction_with_contract_no_arguments(web3, math_contract, build
         'data': '0xd09de08a',
         'value': 0,
         'gasPrice': 1,
-        'chainId': 1,
+        'chainId': 61,
     }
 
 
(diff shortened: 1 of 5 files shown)


cducrest:
I am not sure why in `tests/core/contracts/test_contract_buildTransaction.py` you would expect the `chainId` in the tests to be 1 instead of 61.
After getting more familiar with the tests, I also believe the current test base is sufficient for this change.

kclowes:
Thank you @cducrest!

PR #1381, opened 2019-07-11
Remove xfail from passing test

kclowes:
### What was wrong?
A stray xfail got left behind on a test that was actually passing

### How was it fixed?
Removed it.

#### Cute Animal Picture

download


kclowes:
Yeah, I'm not sure what happened between merges either. I'm going to take a deeper look tomorrow!

PR #1383, opened 2019-07-12
[ethPM] Update Infura strategy

njgheorghita:
### What was wrong?
`ethpm` had its own strategy for connecting to infura via an api key set as an environment variable, which has been updated to use `web3`'s strategy. I'd like to get rid of hardcoding the api key, but since we rely on it for testing I'm not sure the best way to achieve that is / if it's even possible.

#### Cute Animal Picture
image

### DIFF
--- a/tests/ethpm/conftest.py
+++ b/tests/ethpm/conftest.py
@@ -41,12 +41,6 @@ def pytest_addoption(parser):
 
 
 @pytest.fixture
-def infura_env(monkeypatch):
-    # Please play nice and don't use this key for any shenanigans, thanks!
-    monkeypatch.setenv("WEB3_INFURA_PROJECT_ID", '4f1a358967c7474aae6f8f4a7698aefc')
-
-
-@pytest.fixture
 def package_names():
     return PACKAGE_NAMES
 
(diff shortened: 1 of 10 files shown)


kclowes:
Assuming there is nothing special about that particular key, we could pull the API key from the environment. That would require that devs have it set locally, and then we would set the env variable in circle. If we do that, it's probably a good idea to add instructions to the README

tests/ethpm/conftest.py
```diff
@@ -40,6 +40,12 @@ def pytest_addoption(parser):
     parser.addoption("--integration", action="store_true", default=False)
 
 
+@pytest.fixture
+def infura_env(monkeypatch):
+    # Please play nice and don't use this key for any shenanigans, thanks!
+    monkeypatch.setenv("WEB3_INFURA_PROJECT_ID", '4f1a358967c7474aae6f8f4a7698aefc')
```
pipermerriam: I think I'm ok with this, at least as a temporary solution until we figure out a better approach.  You could also use the environment variables and skip+warn when the variable isn't present.

Issue #1367, opened 2019-06-06
Latest docs incorrectly suggest that sendRawTransaction returns a string, actually HexBytes


wjmelements:
Docs say string:
<img width="662" alt="Screen Shot 2019-06-05 at 6 20 40 PM" src="https://user-images.githubusercontent.com/799573/59000403-33425800-87bf-11e9-90cd-87c9b18d4c20.png">
However:
<img width="556" alt="Screen Shot 2019-06-05 at 6 15 04 PM" src="https://user-images.githubusercontent.com/799573/59000411-363d4880-87bf-11e9-972d-f51a85c19f78.png">

kclowes:
You're right! Thanks for pointing that out!

PR #1384, opened 2019-07-13
Update Docs to Show Return of Eth.sendRawTransaction() to be HexBytes, not string.

crawfordleeds:
### What was wrong?

The documentation was not clear that the return type of Eth.sendRawTransaction is a HexBytes object, as originally pointed out by @wjmelements in Issue #1367.

### How was it fixed?

updating web3.eth.rst, line ~638.

#### Cute Animal Picture

IMG_3703

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -236,7 +236,7 @@ The following methods are available on the ``web3.eth`` namespace.
         })
 
     * Merkle proof verification using py-trie.
-    
+
     The following example verifies that the values returned in the AttributeDict are included in the state of given trie ``root``.
 
     .. code-block:: python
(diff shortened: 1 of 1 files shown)


crawfordleeds:
Good catch, that was a mistake, and likely a remnant of some copy and pasting with another project. Sorry about that. I pushed a reversion to the original address. Thanks!

crawfordleeds:
Actually, now that I look into it, don't these addresses need to use the EIP 55 checksum address encoding? I just tried to run the code with the original address, '0xd3cda913deb6f67967b99d67acdfa1712c293601', and receive the following error:

```
TypeError: Transaction had invalid fields: {'to': '0xd3cda913deb6f67967b99d67acdfa1712c293601'}
```

It works fine with the one I originally changed it to, '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55'. 

Am I mistaken about that? If not, perhaps this is deserving of its own issue because there are a few instances of the docs not using EIP 55 checksummed addresses. I'm happy to take a crack at it too.

kclowes:
Yep, you're right. They should be using checksummed addresses. Let's keep this PR as-is and open up a new issue to make sure all addresses in the docs are checksums

crawfordleeds:
Sounds great, I'll keep an eye out for that issue.

kclowes:
@crawfordleeds added the issue! #1388 if you want to take a crack at it. Thanks again!

PR #1392, opened 2019-07-18
[WIP] Move linkable contract to web3.contract

njgheorghita:
### What was wrong?
todo: generalize api so it's compatible with more than just `ethpm.Package`

#### Cute Animal Picture


web3/contract.py
```diff
@@ -315,6 +362,58 @@ def factory(cls, web3, class_name=None, **kwargs):
         contract.fallback = Contract.get_fallback_function(contract.abi, contract.web3)
 
         return contract
+    
+    @classmethod
+    def link_bytecode(cls, attr_dict: Dict[str, str]) -> Type["LinkableContract"]:
```
pipermerriam: Typing wise, I believe this is the better way to do this.

```python
TContract = TypeVar("TContract", bound="Contract")

class Contract:
    @classmethod
    def link_bytecode(cls: Type[TContract], ...) -> Type[TContract]:
        ...
```

I *think* that this results in it working properly for subclasses.

Also, `LinkableContract` is no longer a thing right?

web3/contract.py
```diff
@@ -1516,3 +1624,65 @@ def get_function_by_identifier(fns, identifier):
             'Could not find any function with matching {0}'.format(identifier)
         )
     return fns[0]
+
+
+def apply_all_link_refs(
+    bytecode: bytes, link_refs: List[Dict[str, Any]], attr_dict: Dict[str, str]
```
pipermerriam: I think you could use a `mypy_extensions.TypedDict` for the ref dictionaries.

https://mypy.readthedocs.io/en/latest/more_types.html#typeddict

kclowes:
@njgheorghita I'm doing some housekeeping - is this PR dead? 💀

njgheorghita:
@kclowes thanks for the headsup! I think its safe to kill this for now since it was more of a refactor and not a huge priority to get cleaned up right now

PR #1393, opened 2019-07-18
[ethPM] web3.pm management

njgheorghita:
### What was wrong?
If `web3.pm` is already enabled - trying to re-enable (via `w3.enable_unstable_package_management_api`) raises an error (`cannot attach an attribute that's already been attached`), this can cause problems when using `web3.auto.infura.w3` but now it silently continues about its business.

Also..
- renamed `ethpm`'s `ValidationError` to `EthPMValidationError` to avoid confusion with `web3`'s `ValidationError`
- bypass local validation for large ipfs files, since local validation only works for ipfs files < 256kb. effort to improve validation here https://github.com/ethpm/py-ethpm/pull/165. will port that over once it's working

#### Cute Animal Picture
image

### DIFF
--- a/docs/ethpm.rst
+++ b/docs/ethpm.rst
@@ -358,7 +358,7 @@ By default, the manifest builder does *not* perform any validation that the gene
    ...     validate(),
    ... )
    Traceback (most recent call last):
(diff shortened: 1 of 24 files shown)


njgheorghita:
merging since this is a minor cosmetic pr

PR #1202, opened 2019-01-12
Remove doctest dependency on ethtoken

stefanmendoza:
### What was wrong?

ethtoken has a dependency on web3 and web3 depends on ethtoken for doctest. We should remove this circular dependency and just include the EIP20 ABI in the doctest for now.

See: https://github.com/carver/ethtoken.py/pull/4#discussion_r246949117

### How was it fixed?

Remove ethtoken dependency and added the EIP20 ABI to the relevant test.

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #1395, opened 2019-07-21
Remove doctest dependency on ethtoken

stefanmendoza:
### What was wrong?

** **Reopened PR from https://github.com/ethereum/web3.py/pull/1202** **

ethtoken has a dependency on web3 and web3 depends on ethtoken for doctest. We should remove this circular dependency and just include the EIP20 ABI in the doctest for now.

See: https://github.com/carver/ethtoken.py/pull/4#discussion_r246949117

### How was it fixed?

Remove ethtoken dependency and added the EIP20 ABI to the relevant test.

#### Cute Animal Picture


stefanmendoza:
@pipermerriam / @carver / @kclowes 

Reopened this from #1202. @pipermerriam  - I addressed [this comment](https://github.com/ethereum/web3.py/pull/1202#discussion_r247598374) in the second commit here:
9102db0

README.md
```diff
@@ -109,7 +109,7 @@ pytest -n 4 -f --maxfail=1
 
 2. Execute `tox` for the tests
 
-There are multiple [components](https://github.com/ethereum/web3.py/blob/master/.travis.yml#L53) of the tests. You can run test to against specific component. For example:
+There are multiple [components](https://github.com/ethereum/web3.py/blob/master/.circleci/config.yml#L144) of the tests. You can run test to against specific component. For example:
```
stefanmendoza: Updated this to a similar line to what was linked before as the project no longer uses Travis. Previously pointed to this:
https://github.com/ethereum/web3.py/blob/v4.9.0/.travis.yml#L53

docs/web3.eth.account.rst
```diff
@@ -264,11 +264,15 @@ To sign a transaction locally that will invoke a smart contract:
...
+
+
 .. doctest::
 
-    >>> from ethtoken.abi import EIP20_ABI
```
carver: I still think it's important for people to be able to run the examples locally, which this breaks. At least, we should link to the ABI, as we discussed in: https://github.com/ethereum/web3.py/pull/1202#discussion_r254895974

stefanmendoza: Do you mean put a literal link to the ABI? Like as a comment?

carver: Right, a link so that documentation readers can fill in the `EIP20_ABI` variable when they run it in their own console.

stefanmendoza: Does this look fine? Also, just curious, but is it redundant to set `nonce = 0` in the setup?

https://github.com/ethereum/web3.py/pull/1395/commits/a27a5a45f5cc31ae6edf902654ddbd867be64b95

carver: Yeah, looks redundant to me.

stefanmendoza: I was going to remove it, but it actually seems it's set because we don't actually make the call below:

```
>>> nonce = w3.eth.getTransactionCount('0x5ce9454909639D2D17A3F753ce7d93fa0b9aB12E')  # doctest: +SKIP
```

stefanmendoza:
@carver the nonce assignment isn't actually called in the `doctest`, so I left the `nonce = 0` assignment:

> `>>> nonce = w3.eth.getTransactionCount('0x5ce9454909639D2D17A3F753ce7d93fa0b9aB12E') # doctest: +SKIP`

Should be good to go once the build finishes!

stefanmendoza:
@carver  @pipermerriam bumping this 🎉

docs/web3.eth.account.rst
```diff
@@ -264,11 +264,18 @@ To sign a transaction locally that will invoke a smart contract:
...
+    # When running locally, execute the statements found in the file linked below to load the EIP20_ABI variable.
+    # See: https://github.com/carver/ethtoken.py/blob/v0.0.1-alpha.4/ethtoken/abi.py
```
pipermerriam: cc @njgheorghita you might update this sometime with an *optional* way to load the ABI using an EthPM package.

PR #1400, opened 2019-07-26
Update registry assets to simple solidity registry implementation

njgheorghita:
### What was wrong?
New simplified solidity registry implementation [here] is the default registry implementation moving forward. `ethpm`'s assets folder needs updating to use this implementation. 

#### Cute Animal Picture
image


ethpm/assets/simple-registry/contracts/PackageRegistry.sol
```diff
@@ -0,0 +1,373 @@
...
+            allPackageIds[packageCount] = packageId;
+            packageCount++;
+        } else {
+            package.updatedAt = block.timestamp;
+        }
+        cutRelease(packageId, releaseId, packageName, version, manifestURI);
+        return releaseId;
+    }
+
+    function cutRelease(
```
pipermerriam: maybe name `_cutRelease` to signal privateness.

ethpm/assets/simple-registry/contracts/PackageRegistry.sol
```diff
@@ -0,0 +1,373 @@
...
+
+    // @dev Returns release id that *would* be generated for a name and version pair on `release`.
+    // @param packageName Package name
+    // @param version Version string (ex: '1.0.0')
+    function generateReleaseId(
+        string memory packageName,
+        string memory version
+    )
+        public
+        view
```
pipermerriam: Should this be `pure`?

ethpm/assets/simple-registry/contracts/PackageRegistry.sol
```diff
@@ -0,0 +1,373 @@
...
+    {
+        return keccak256(abi.encodePacked(packageName, version));
+    }
+
+    function generatePackageReleaseId(
+        bytes32 packageId,
+        uint packageReleaseCount
+    )
+        private
+		pure
```
pipermerriam: weird indentation

njgheorghita:
> I'm inclined to have you get 1-2 more people to look at this before merging just because it's smart contract code. They don't necessarily have to be solidity experts but more 👀 might catch something I missed.

Yup, doing that in https://github.com/ethpm/solidity-registry/pull/1. The contract here is `v2.0.0a1` - and will be updated to a stable version once https://github.com/ethpm/solidity-registry/pull/1 is approved/merged.

ethpm/assets/simple-registry/contracts/PackageRegistry.sol
njgheorghita: Yes, it should. But the ERC / Interface define this function as `view`. Which looks to be the case b/c in the complex solidity implementation, it generates these values via the `PackageDB` and `ReleaseDB`. So, I left it as `view` here to maintain compatibility with the erc/ `PackageRegistryInterface`.

PR #1403, opened 2019-07-29
Update dependencies for release

kclowes:
### What was wrong?
I ran into a hiccup with the release. Needed to make some changes to setup.py

#### Cute Animal Picture

image


setup.py
```diff
@@ -35,10 +35,10 @@
         "flaky>=3.3.0",
         "hypothesis>=3.31.2",
         "pytest>=4.4.0,<5.0.0",
-        "pytest-mock==1.*",
+        "pytest-mock>=1.10.4,<2.0.0",
         "pytest-pythonpath>=0.3",
-        "pytest-watch==4.*",
-        "pytest-xdist==1.*",
+        "pytest-watch>=4.2.0,<5.0.0",
+        "pytest-xdist>=1.29.0,<2.0.0",
```
carver: I don't think this is standardized one way or another, but lately I've been leaving off the minor/patch 0's, like: `>=1.29,<2`. Whichever you like better, I really don't have a strong opinion. :+1:

PR #1404, opened 2019-07-29
Eliminate `signTransaction` warning

fubuloubu:
### What was wrong?
I get the following warning when using this middleware:
```python
eth_account/signers/local.py:88: DeprecationWarning: signTransaction is deprecated in favor of sign_transaction
```
It's annoying lol

### How was it fixed?
Removed deprecated usage

#### Cute Animal Picture
Ack


fubuloubu:
I believe this is all the usages that should change. There are more, but this only affects instances of `Account.signTransaction()`

davesque:
Hehe, that picture.

fubuloubu:
The one failed CircleCI task seems pretty unrelated to the functionality here? Not sure what the issue is.

fubuloubu:
I don't have the repo handy, I would appreciate it if you could handle that for me!

Issue #1412, opened 2019-08-07
Missing ValidationError when calling bytes argument with string value


palango:
* Version: 5.0.0
* Python: 3.7
* OS: linux/travis
* `pip freeze` output

See https://travis-ci.org/palango/raiden-contracts/builds/568742398#L398

### What was wrong?

I updated the web3.py, eth-tester and further dependencies in a PR of the raiden contracts: https://github.com/raiden-network/raiden-contracts/pull/1153

So far everything looks good, but I get on test failure that looks like a regression to me: https://travis-ci.org/palango/raiden-contracts/builds/568742398#L494

In that test we check that the web3 contract instance throws a `ValidationError` when the contract is called with an invalid value.

The function in question is here:
https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/data/source/raiden/SecretRegistry.sol#L17

The test code is:
```py
        with pytest.raises(ValidationError):
           secret_registry_contract.functions.registerSecret("")
           Failed: DID NOT RAISE <class 'web3.exceptions.ValidationError'>
```
### How can it be fixed?

This should either raise a `ValidationError` or it should be documented that `strings` can now be passed to `bytes` parameters.

pipermerriam:
It looks like the offending code is here:

https://github.com/ethereum/web3.py/blob/b899050df8614db1d5ec339a920a7aa6b570fa99/web3/_utils/abi.py#L149-L157

If the value is text and can be decoded as hex it gets accepted.

My initial thought is that we should **only** accept `"0x" prefixed hex values.  This is complicated since we can't exactly make that change since it would be backwards incompatible.

We might be able to be more strict about just the empty string and require that `0x` be present if the string is empty....  cc @carver

carver:
> it should be documented that `strings` can now be passed to `bytes` parameters.

~~Yeah, the short-medium term answer is this ^.  Bytes parameters will interpret strings as hex, as Piper pointed out.~~

~~On first look, I think I'm good with requiring an `0x` prefix in order to interpret a string as hex, but we couldn't make that change until v6, even for the empty string. So it just belongs in the v5 migration docs.~~

Nope, this is a straight-up bug. Only a 64-character hex-string should be accepted, since the ABI specifies `bytes32`.

kclowes:
[...]

PR #1415, opened 2019-08-09
Don't allow empty string for hex values

kclowes:
### What was wrong?
Because we allow a bytestring or a hexstring to be passed into functions that require bytes types, an empty string was mistakenly being decoded as hex.

Related to Issue #1412

### How was it fixed?
Added a check for an empty string in the hex string mixin class. Now an empty hexstring requires a '0x'. 

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -252,6 +252,29 @@ def test_call_get_byte_array(arrays_contract, call):
     assert result == expected_byte_arr
 
 
+@pytest.mark.parametrize('args,expected', [([b''], [b'\x00']), (['0x'], [b'\x00'])])
+def test_set_byte_array(arrays_contract, call, transact, args, expected):
+    transact(
+        contract=arrays_contract,
+        contract_function='setByteValue',
+        func_args=[args]
+    )
+    result = call(contract=arrays_contract,
+                  contract_function='getByteValue')
+
+    assert result == expected
+
+
+@pytest.mark.parametrize('args', ([''], ['s']))
+def test_set_byte_array_with_invalid_args(arrays_contract, transact, args):
+    with pytest.raises(ValidationError):
+        transact(
+            contract=arrays_contract,
+            contract_function='setByteValue',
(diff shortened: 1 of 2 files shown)


carver:
Is this about fulfilling this idea:

> But even without the flag, the ValidationError from OP should be raised. +1

It seems strange & difficult to explain in the docs, that these are valid inputs for `bytes32`:
```py
b''
b'X'
"0x"
"58"
"0x58"
```

But then `""` is invalid. Whatever we decide here definitely belongs in a new section about ABI type-coercion, with examples.

What I was clumsily communicating in my earlier quote was the idea that anything shorter than 32 bytes should be invalid. I thought the extra flag would require all `str`s, no matter the length, to start with `"0x"`.

I understand the hesitation to make strings shorter than 32 bytes invalid (as backwards incompatible). So, I think both of those changes (require 0x on all strings and invalidate too-short values) could go behind the same flag.

kclowes:
> Is this about fulfilling this idea:
>> But even without the flag, the ValidationError from OP should be raised. +1

yep, this is the attempt to do that. What I understood from #1412 is that an empty string is an invalid hex value, and therefore passing an empty `string` into a function that accepts a `bytes` value is a bug. But yeah, it's weird that other string values can be accepted without the '0x' prefix, as long as they can be decoded to hex.

> Whatever we decide here definitely belongs in a new section about ABI type-coercion, with examples.
👍 I'll add that

> I understand the hesitation to make strings shorter than 32 bytes invalid (as backwards incompatible). So, I think both of those changes (require 0x on all strings and invalidate too-short values) could go behind the same flag.

So, it seems like the most consistent (and non-breaking) approach here is leaving the default as-is, and adding docs about the validity of different parameters. Then, we can add a flag that, if enabled, requires all hex values to be prefixed with a `0x`, and doesn't allow strings without the correct number of bytes. I started down that path earlier and then changed directions. Then, for v6, we can make strict the default. I will add warnings for values that will throw an error in v6.

kclowes:
Closing in favor of #1419

PR #1418, opened 2019-08-12
[ethPM] Add expression begin/close chars to package name regex

njgheorghita:
### What was wrong?
Package name regex expression needs expression beginning/closing chars (`^`/`$`) to correctly validate package names. 

#### Cute Animal Picture

image


njgheorghita:
@pipermerriam Definitely, a little suspicious about how it went this long without getting a test but it's there now.

PR #1419, opened 2019-08-13
Flag for Stricter ABI Type Checking

kclowes:
#### What was wrong?
web3 v5 will accept a bytestring or a hexstring if a bytes value is specified in an ABI. Because an empty string can be decoded as hex, it wasn't raising a `ValidationError` like v4 did. It was also requested that we not allow bytes values that are shorter than the specified size. 

#### How was it fixed
This PR adds a stricter type checking mechanism for bytes values that is available on the web3 instance. Using `w3.enable_strict_bytes_type_checking()` will disallow strings that don't have a `'0x'`, and will disallow byte strings that are different from the size the ABI specifies. 

I also noticed that we were using eth-abi's codec for events instead of the one registered in web3, so changed that over to use the codec on the web3 instance.

- Added a deprecation warning for strings without a '0x' - in v6 it will be invalid to pass a string without `0x` to an ABI with bytes specified

#### Cute animal picture
Screen Shot 2019-08-19 at 5 03 16 PM

Todos:
- [x] don't allow strings without a '0x' prefix
- [x] don't allow bytestrings that are the wrong size
- [x] cleanup
- [x] docs
- [x] events
- [x] Fix decoder


web3/contract.py
```diff
@@ -351,7 +351,7 @@ def encodeABI(cls, fn_name, args=None, kwargs=None, data=None):
         :param data: defaults to function selector
         """
         fn_abi, fn_selector, fn_arguments = get_function_info(
-            fn_name, contract_abi=cls.abi, args=args, kwargs=kwargs,
+            fn_name, cls.web3.codec, contract_abi=cls.abi, args=args, kwargs=kwargs,
```
pipermerriam: just noticing that we still call it `web3` instead of `w3` here.  Might be a good future cleanup to do.

web3/main.py
```diff
@@ -245,3 +252,6 @@ def enable_unstable_package_management_api(self):
         from web3.pm import PM
         if not hasattr(self, '_pm'):
             PM.attach(self, '_pm')
+
+    def enable_strict_bytes_type_checking(self):
+        return setattr(self, 'codec', ABICodec(build_strict_registry()))
```
pipermerriam: any reason this isn't just `self.codec = ABICodec(...)`?

kclowes: Nope! 👍

docs/overview.rst
```diff
@@ -346,3 +346,170 @@ Modules
...
+   - '1234'   # valid - string that can be decoded as a hex string
+   - '0xabcd' # valid - hex string with exactly 4 bytes
+   - b'1234'  # valid - bytestring with exactly 4 bytes
+
+The following values will raise an error by default:
+
+.. code-block:: python
+
+   - b'abcde'  # invalid - more than 4 bytes
+   - '0xabcde' # invalid - hex string with more than 4 bytes
```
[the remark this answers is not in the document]
kclowes: oops, thanks!

web3/_utils/abi.py
```diff
@@ -146,25 +153,131 @@ def validate_value(cls, value):
 
 
 class AcceptsHexStrMixin:
+    def validate(self):
```
pipermerriam: I *think* that you just need to do this:

```python
    def encode(self, value):
        original_value = value
        if is_text(value):
            try:
                value = decode_hex(value)
            except binascii.Error:
                self.invalidate_value(
                    value,
                    msg=f'{value} is an invalid hex string',
                )
            else:
                self.check_for_0x(original_value)
        # THE VALUE IS NORMALIZED
        super().encode(value)
```

docs/contracts.rst
```diff
@@ -491,6 +491,218 @@ and the arguments are ambiguous.
...
+
+    ... #      function getBytes2Value() public view returns (bytes2[] memory) {
+    ... #          return bytes2Value;
+    ... #      }
+    ... #  }
+
+.. doctest::
+
+    >>> abi = '''
+    ...  [
```
pipermerriam: This is a pretty large block to be in the docs.  You might hide this inside of a [`.. testsetup::`](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testsetup) directive so that it doesn't render for the user.  Then you can do what you did below for the bytecode:

```
>>> # abi = ...
```

docs/examples.rst
```diff
@@ -229,45 +229,45 @@ The following example demonstrates a few things:
...
-    
+
+
     w3 = Web3(EthereumTesterProvider())
-    
+
     contract_source_path = 'contract.sol'
     compiled_sol = compile_source_file('contract.sol')
-    
+
```
pipermerriam: For PR's like this that are growing quite large, pulling these kinds of changes out into their own small PRs is good practice.

kclowes:
@pipermerriam this is ready for another (and hopefully the last) round of real review

web3/_utils/abi.py
```diff
@@ -145,26 +152,168 @@ def validate_value(cls, value):
...
+    is_strict = False
+
+
+class StrictByteStringEncoder(AcceptsHexStrEncoder):
+    subencoder_cls = encoding.ByteStringEncoder
+    is_strict = True
+
+
+class ExactLengthBytesEncoder(encoding.BaseEncoder):
+    # TODO: move this to eth-abi once the api is stabilized
```
pipermerriam: Might be good to open a tracking issue for this so it doesn't get lost.  Probably a "Good First Issue" candidate.

PR #1426, opened 2019-08-21
added next_nonce to parity module

ankitchiplunkar:
### What was wrong?

The parity JSON-RPC method `parity_nextNonce` was not implemented.
https://wiki.parity.io/JSONRPC-parity-module#parity_nextnonce

### How was it fixed?
Added the method and its respective result formatter

### Further direction:
Any ideas on how I can test this?

#### Cute Animal Picture

### DIFF
--- a/web3/middleware/pythonic.py
+++ b/web3/middleware/pythonic.py
@@ -387,5 +387,7 @@ pythonic_middleware = construct_formatting_middleware(
         'evm_snapshot': hex_to_integer,
         # Net
         'net_peerCount': to_integer_if_hex,
+        # Parity
+        'parity_nextNonce': to_integer_if_hex,
     },
 )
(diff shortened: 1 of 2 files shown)


kclowes:
Thanks @ankitchiplunkar!

> Any ideas on how I can test this?

You should be able to add a test to [this file](https://github.com/ethereum/web3.py/blob/master/web3/_utils/module_testing/parity_module.py), and then it will get run in each of the parity integration tests. Let me know if you need more direction!

ankitchiplunkar:
Hmm,
The above method, gives the next possible nonce by for a transaction by an address in both :
1. Pending blocks
2. Mempool transactions

How can I add data in both mempool and pending blocks to test this functionality?

pipermerriam:
@ankitchiplunkar no need to test the full range of possibilities for next nonce.  The tests there are more to test that the integration is right, ensuring that it doesn't throw any errors and all of the return values and input values get correctly formatted for the request/response.  We can just *assume* that parity correctly tracks nonces without testing the full range of possible cases.

PR #1427, opened 2019-08-21
[ethPM] Fix bug in package id and release id fetching strategy

njgheorghita:
### What was wrong?
`w3.pm.get_all_packages()` and `w3.pm.get_all_release_ids()` were wonky when dealing with large amounts of packages / releases. This can be seen [here](http://explorer.ethpm.com/browse/ropsten/0x8C49FA23F023cB5E1561830a984441E4383A3348) and [here](0x9637E6948529f47FCf480fA9F3a2Cd04e1022ee4).

The tests only covered cases with ~6 releases which worked fine. 

#### Cute Animal Picture

image

### DIFF
--- a/web3/pm.py
+++ b/web3/pm.py
@@ -26,9 +26,6 @@ from eth_utils import (
     to_text,
     to_tuple,
 )
-from eth_utils.toolz import (
-    concat,
-)
 
 from ethpm import (
     ASSETS_DIR,
@@ -217,6 +214,9 @@ class ERC1319Registry(ABC):
         pass
 
 
+BATCH_SIZE = 100
+
+
 class SimpleRegistry(ERC1319Registry):
     """
     This class represents an instance of the `Solidity Reference Registry implementation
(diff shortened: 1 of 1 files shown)


web3/pm.py
```diff
@@ -245,11 +246,12 @@ def _get_package_name(self, package_id: bytes) -> str:
     @to_tuple
     def _get_all_package_ids(self) -> Iterable[Tuple[bytes]]:
         num_packages = self._num_package_ids()
+        pointer = 0
+        package_ids = []
+        while pointer < num_packages:
+            new_ids, pointer = self.registry.functions.getAllPackageIds(pointer, (pointer + BATCH_SIZE)).call()
+            package_ids.append(new_ids)
```
njgheorghita: @pipermerriam Not sure how you feel about this - feels a bit mutate-y to me, but also the easiest way to implement it. The `BATCH_SIZE` is also arbitrary, but I tested it locally with values of `1`, `2`, `4`, `10`, and `100` and it works fine. I didn't update the tests in this case b/c releasing 100+ packages just to test this case seems a bit overkill.

pipermerriam: This should be able to be converted to a non-mutative solution reasonably easy.  IIUC this correctly you need to reverse each of the individual chunks and return in order which I think would be equivalent to just replacing this statement with `yield from reversed(new_ids)`

web3/pm.py
```diff
@@ -245,29 +245,28 @@ def _get_package_name(self, package_id: bytes) -> str:
...
-        release_ids = [
-            self.registry.functions.getAllReleaseIds(
-                package_name, index, (index + 4)
-            ).call()[0]
-            for index in range(0, num_releases, 4)
-        ]
-        for release_id in concat([x[::-1] for x in release_ids]):
-            yield release_id
+        pointer = 0
+        while pointer < num_releases:
```
pipermerriam: One more thing.  This is probably fine, however, you could save yourself a headache by enforcing that `pointer` always increases on each loop iteration to ensure you cannot end up in an infinite loop.

PR #1430, opened 2019-08-22
Decorator Removed

### LINKED ISSUE #1423: `ens.utils.dict_copy` only operates on keyword arguments
### What was wrong?

There is a `dict_copy` decorator used in the `ens` module.  It is intended to allow us to write methods like this:

```
@dict_copy
def my_method(self, some_params={}):
    ...
```

Normally using a mutable value like a dictionary as a default argument is problematic because the same mutable object is used across each call.  This decorator institutes a `copy` of this, but **only** if it is passed as a keyword argument.

There are at least some places where `dict_copy` is used but where the parameter that needs to be copied is not enforced to be a keyword argument, meaning that any calls that use it positionally will not benefit from the protection.

### How can it be fixed?

I'm inclined to suggest removing the decorator entirely and changing all places it was used to follow the standard pattern of:

```python
def my_method(self, some_params=None):
    if some_params is None:
        some_params = {}
    ...
```

Patil2099:
Fixes #1423

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -92,8 +92,8 @@ class ENS:
         reversed_domain = address_to_reverse_domain(address)
         return self.resolve(reversed_domain, get='name')
 
-    @dict_copy
-    def setup_address(self, name, address=default, transact={}):
+    
+    def setup_address(self, name, address=default,transact=None):
         """
         Set up the name to point to the supplied address.
         The sender of the transaction must own the name, or
@@ -111,6 +111,8 @@ class ENS:
         :raises InvalidName: if ``name`` has invalid syntax
         :raises UnauthorizedError: if ``'from'`` in `transact` does not own `name`
         """
+        if transact is None:
+            transact={}
         owner = self.setup_owner(name, transact=transact)
         self._assert_control(owner, name)
         if is_none_or_zero_address(address):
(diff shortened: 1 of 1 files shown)


Patil2099:
@pipermerriam

kclowes:
@Patil2099 We'll review once all the tests are passing! Thanks!

Code0x58:
`dict_copy` does more than work around `def f(d={}):` reusing the default parameter object.

From the tests (given that dict comparisons are failing), it looks like it is also being relied on to create [deep copies](https://docs.python.org/3/library/copy.html#copy.deepcopy) of objects so that the code using it can mutate the object without affecting the original object.

It would probably be good to mark `dict_copy` as deprecated if nowhere in the codebase uses it, then make an issue for it and tag it with the next major release.

You will also need to find where copies are needed to avoid mutating dictionaries the function/code doesn't own. I think keeping on this approach would be better for the code in the long run rather than continuing to rely on `dict_copy`.

`dict_copy` has a few issues, including documentation and the name being wrong (they match the intent, but not what it actually does). For example it copies a lot more than just dicts, and _any_ type of mutable object inside them (not just dict/list/tuple/set) will be deep copied. I suspect it didn't copy args just to avoid working around the first parameter of class methods. Killing it with fire sounds good to me.

PR #1431, opened 2019-08-22
Add towncrier

kclowes:
### What was wrong?
Release notes were being written by hand. Now when make release is run, release notes will be generated from the newsfragments files. For every pull request, we'll need to add a new newsfragment file. Instructions can be found in the `newsfragments/README.md` file.

### How was it fixed?
Added towncrier, which should generate release notes automagically ✨ 

#### Cute Animal Picture

image


newsfragments/README.md
```diff
@@ -0,0 +1,24 @@
...
+`<ISSUE>` is an issue number, and `<TYPE>` is one of:
+
+* `feature`
+* `bugfix`
+* `doc`
+* `misc`
+
+ So for example: `123.feature.rst`, `456.bugfix.rst`
+
+ If the PR fixes an issue, use that number here. If there is no issue,
```
cburgdorf: Maybe I already had this wrong in the places where I introduced this but I just noticed that there are whitespace errors in this file here

newsfragments/README.md
```diff
@@ -0,0 +1,24 @@
...
+then open up the PR first and use the PR number for the newsfragment.
+
+ Note that the `towncrier` tool will automatically
```
cburgdorf: And here

newsfragments/validate_files.py
```diff
@@ -0,0 +1,31 @@
...
+
+import os
+import pathlib
+
+ALLOWED_EXTENSIONS = {
+    '.feature.rst',
+    '.bugfix.rst',
+    '.doc.rst',
+    '.removal.rst',
+    '.misc.rst',
```
cburgdorf: I'm not sure how common it is for `web3` to make performance improvements and hence bundle those in their own section. If you think it's worth doing that, then check out this [commit](https://github.com/ethereum/trinity/commit/d7a4b5e87fbf26e70e6fa83c1faf43def598b9f8) for the changes needed to do that.

It's worth pointing out that that sections are generally only included if they have items (meaning, even if you prepare for a perf section, it will only come up in the release notes, if it's actually used)

Makefile
```diff
@@ -43,12 +43,24 @@ build-docs:
 docs: build-docs
 	open docs/_build/html/index.html
 
+validate-docs:
```
cburgdorf: One thing that is missing from this PR (and which is important) is to call this validation in CI. You want that to ensure people don't accidentally commit illegal files which might be hard to spot in a review (e.g. `4711.docs.rst`, `4711.fix.rst`, `4711.feat.rst` all *look*  ok to my tidy reviewer eyes but they are all wrong).   

Looking into the `tox.ini` shows that there's currently a `doctest` job.

https://github.com/ethereum/web3.py/blob/9ec8556ce48d52c858e09e2e2acd42bd71854cfb/tox.ini#L40

Maybe follow what we did with Trinity and turn that into a full `docs` job so that it combines all of this:

1. Build the docs 
2. Validate the doctest snippets
3. Validate the newsfragment files
4. Validate the release notes

There's one extra thing to point out here: `towncrier --draft` is to catch errors in the release notes but is has proven to be a weak solution so far. It doesn't catch actual errors in the RST. So what might happen is that someone adds a newsfragment file and missing a closing backtick or something like that.

We haven't got around fixing that yet but I believe the best case scenario would be to make the doc job be run like this:

1. Validate the doctest snippets
2. Run towncrier using a pseudo version (no `--draft`, right the actual files to disk)
3. Now build the docs
4. Validate the doctest snippets

Using that order, it would be guaranteed that any formatting errors in the release notes are caught by building the docs **after** the release notes where written to disk.

The one downside (which is discussed [here](https://github.com/ethereum/lahja/pull/153#issuecomment-521342852)) is that this command would leave the repository in a dirty state (modified release notes, deleted newsfragments). It's not a big deal in the CI run as nobody cares about the state of the files on a short-lived CI server disk but if someone would run the command locally that might be unexpected. 

To deal with that I'd recommend to do the following:

1. Do not allow the command to run when the repository is in a dirty state
2. Run a `git reset --hard` at the very end to not leave the repository in a dirty state

I'm not saying this PR needs to go that extra mile to do this form of validation, just pointing out what I believe is the best case scenario.

I do think so that the PR needs to ensure the basic validation is triggered in CI.

newsfragments/README.md
kclowes: Yikes! Thanks!

newsfragments/validate_files.py
kclowes: I think we can put performance improvements under features for now, we haven't done very man. But, if we start to do more it would for sure be worth adding!

Makefile
kclowes: Yeah, I noticed that it wasn't being run in CI and tinkered with it for a little bit yesterday, and then gave up. I was running into an error where CircleCI/pip doesn't know what to do with a project that has both a pyproject.toml and a setup.py. I figured that since I am pretty much the only one releasing, I would just run the `validate-docs` command locally before I released, but I'll try again and hopefully get something auto working.

I don't love the idea of a surprise `git reset --hard`, but would be okay with it in the sequence you described. I'll timebox it today and see if I can get the optimal solution working, and if not I'll leave it for later.

Makefile
```diff
@@ -47,6 +47,10 @@ validate-docs:
 	python newsfragments/validate_files.py
 	towncrier --draft
 
+validate-and-test-docs: validate-docs
+	$(MAKE) -C docs doctest
```
cburgdorf: I think instead of this new job, I'd recommend using `build-docs` from within `validate-docs`. That also ensures that the *html*  output is also tested instead of only running the `doctest` job.

tox.ini
```diff
@@ -37,7 +38,7 @@ commands=
     integration-parity-ipc: pytest -s {posargs:tests/integration/parity/test_parity_ipc.py}
     integration-parity-http: pytest -s {posargs:tests/integration/parity/test_parity_http.py}
     integration-parity-ws: pytest -s {posargs:tests/integration/parity/test_parity_ws.py}
-    doctest: make -C {toxinidir}/docs doctest
+    doctest: make -C {toxinidir} validate-and-test-docs
```
cburgdorf: Probably good to rename `doctest` to just `docs` to avoid confusion. The term `doctest` seems loaded with the `doctest` directive to test code examples in the docs. Renaming to `docs` might convey better that this is a job that validates the entire docs from building the html, to running the examples and creating the release notes.

PR #1432, opened 2019-08-23
Update web3.geth.rst

### LINKED ISSUE #1429: Add Snake Case Methods for JSON-RPC calls
### What was wrong?
We'd like to start using pythonic snake_case names for all JSON-RPC methods. The pattern has been laid out in the Geth admin module. 

### How can it be fixed?
Add deprecation warnings, new snake case methods, and tests for all other modules. Ideally, each module would be its own PR. 

- [x] Eth
- [x] GethShh
- [x] GethPersonal
- [x] GethTxPool
- [x] GethMiner
- [x] Net
- [x] ParityShh
- [x] ParityPersonal
- [x] Parity

Patil2099:
Fixes #1429

### DIFF
--- a/docs/web3.geth.rst
+++ b/docs/web3.geth.rst
@@ -525,7 +525,7 @@ Full documentation for Geth-supported endpoints can be found `here <https://gith
         >>>web3.geth.shh.post({'payload': web3.toHex(text="test_payload"), 'pubKey': recipient_public, 'topic': '0x12340000', 'powTarget': 2.5, 'powTime': 2})
         True
 
-.. py:method:: Shh.newMessageFilter(self, criteria)
+.. py:method:: Shh.new_message_filter(self, criteria)
 
     * Create a new filter id. This filter id can be used with ``ShhFilter`` to poll for new messages that match the set of criteria.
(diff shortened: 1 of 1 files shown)


kclowes:
@Patil2099 thanks! These methods don't exist yet, so in addition to updating the docs, you'll need to write the methods. Let me know if you need more guidance. You should be able to follow the pattern in #1326.

Patil2099:
Where should I add these methods?

kclowes:
In the Geth Personal class, using `importRawKey` as an example

In `web3/_utils/personal.py` add:

```python
import_raw_key = Method(
    "personal_importRawKey",
)

importRawKey = DeprecatedMethod(import_raw_key, 'importRawKey', 'import_raw_key')
```

Then in `web3/geth.py`:
```python
from web3._utils.personal import (
    importRawKey,
)

class GethPersonal(ModuleV2):
    import_raw_key = import_raw_key
    ...
```

You'll also need to add a deprecation warning in the docs to the old methods, add tests for the snake case methods, and test to make sure a deprecation warning gets raised.

PR #1445, opened 2019-09-05
Fix LogsBloom formatter error

enjeyw:
### What was wrong?

When `getBlock(block_identifier='pending')` is called, the pending block will return `logsBloom = null`. The current block formatter attempts to convert the null value to HexBytes and fail.

### How was it fixed?
The LogsBloom HexBytes formatter is only applied when logsBloom is not null.

### Todo:

#### Cute Animal Picture

lilly

### DIFF
--- a/newsfragments/1445.bugfix.rst
+++ b/newsfragments/1445.bugfix.rst
@@ -1,0 +1,1 @@
+Add null check to logsbloom formatter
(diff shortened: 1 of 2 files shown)


kclowes:
That parity test is flaky, not something you did. Do you mind adding a test for this functionality? Thanks!

PR #1446, opened 2019-09-05
WIP: Allow jsonschema <3.1

Code0x58:
This expands the range of allowed jsonschema versions as [3.0.0+ versions](https://pypi.org/project/jsonschema/#history) have been around for over 6 months. If the tests pass, then it seems reasonable to increase the upper bound to avoid causing conflicts.


setup.py
```diff
@@ -73,7 +73,7 @@
         "eth-utils>=1.4.0,<2.0.0",
         "hexbytes>=0.1.0,<1.0.0",
         "ipfshttpclient>=0.4.12,<1",
-        "jsonschema>=2.6.0,<3",
+        "jsonschema>=2.6.0,<4.0",
```
kclowes: We should add a tox test suite to make sure that both major versions 2.x and 3.x work here

Code0x58: The behaviour in 2 and 3 is to use a given default (or fallback to global default) jsonschema draft spec if one isn't provided in within the spec (would be [_$schema_](https://json-schema.org/understanding-json-schema/reference/schema.html) in [`package.spec.json`](https://github.com/ethereum/web3.py/blob/master/ethpm/assets/spec/package.spec.json)). This change just makes that gives a default that works in both 2.6 and 3.x.

It looks like the jsonschema maintainer is good with semantic versioning, so given these two things I don't think it would be worth adding additional envs for the different versions (which isn't done at the moment for different package versions).

_p.s. source files for [v2.6.0](https://github.com/Julian/jsonschema/blob/v3.0.2/jsonschema/validators.py) and [v3.0.2](https://github.com/Julian/jsonschema/blob/v3.0.2/jsonschema/validators.py) - `validate(...)` and `validator_for(...)` are the relevant methods_

kclowes: > which isn't done at the moment for different package versions

It's not, but all of our production libraries are pinned to a major version. 

I think probably the better option if it doesn't break anything is to pin to v3. I don't really see a reason to allow v2 any longer, unless there's something I'm missing.

Also want to ping @njgheorghita for any feedback/thoughts!

njgheorghita: > I think probably the better option if it doesn't break anything is to pin to v3. I don't really see a reason to allow v2 any longer, unless there's something I'm missing.

Sounds good with me!

Issue #1408, opened 2019-08-01
Fix Flaky Parity Shh Module integration tests


kclowes:
### What was wrong?

Two intermittent test failures have popped up in the parity shh module.
`tests/integration/parity/test_parity_ipc.py::TestParityShhModuleTest::test_shh_remove_filter`
`tests/integration/parity/test_parity_ipc.py::TestParityShhModuleTest::test_shh_post`

See [this circle run](https://circleci.com/gh/ethereum/web3.py/70708) for the full output.

### How can it be fixed?
Looking at the history of the failures, I suspect it has to do with the refactoring of the parity tests (#1355), but I'm not sure.

PR #1447, opened 2019-09-05
Fix flaky parity whisper test

kclowes:
### What was wrong?
Our parity integration tests were flaking out on the `shh.post` method sometimes.

Related to Issue #1408 

### How was it fixed?
Since we don't care what comes back from parity, just that we're interfacing with it correctly, added a check for an error response, and make sure it's the correct error response. Otherwise, continue with the test as usual.

### Todo:

#### Cute Animal Picture

Screen Shot 2019-09-05 at 11 45 09 AM


web3/_utils/module_testing/shh_module.py
```diff
@@ -320,13 +321,26 @@ def test_shh_remove_filter(self, web3):
...
 
-        assert web3.parity.shh.post({
-            'payload': payload,
-            'topics': [topic],
-            'to': {'public': receiver_pub},
-            'priority': 500,
-            'ttl': 400,
-        })
+        try:
+            web3.parity.shh.post({
```
Code0x58: for both changes: accepting a failure due to falking, then requiring a failure in in the `with pytest.raises(...)` in the next block doesn't seem very smooth. I don't know if the code after the try/except requires this code to be successful for it to pass.

If the the rest of the test can work after this code has failed, then you could do something like this to simplify things and only run once:
```python
try:
   ...
except ValueError as e:
   # :(
   if str(e) != "Pow too low to compete with other messages":
       raise
```

kclowes:
Ah, I like that better. It does feel cleaner. Thanks!

web3/_utils/module_testing/shh_module.py
```diff
@@ -385,6 +395,8 @@ def test_shh_post(self, web3):
             "ttl": 400,
         })
 
+        assert response is True
```
Code0x58: this change looks a little suspect to me, but probably fine

kclowes: oops, don't need that 😆

PR #1451, opened 2019-09-13
Upgrade py-evm to 0.2.0 beta

kclowes:
### What was wrong?
Piper pointed out that Jason did a bunch of work to make py-evm much faster :fire: tests now go so fast 🚗💨 

### How was it fixed?
Added new dependency requirement

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/1451.misc.rst
+++ b/newsfragments/1451.misc.rst
@@ -1,0 +1,1 @@
+Upgrade eth-tester dependency to use 0.2.0-beta.2
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import (
 
 extras_require = {
     'tester': [
-        "eth-tester[py-evm]==0.1.0-beta.39",
+        "eth-tester[py-evm]==v0.2.0-beta.2",
         "py-geth>=2.0.1,<3.0.0",
     ],
     'linter': [


carver:
> tests now go so fast

Glad to hear it. Just curious if you have it available: what kinds of speedups were you seeing? Always curious to see the impact on different workloads.

newsfragments/1451.misc.rst
```diff
@@ -0,0 +1 @@
+Upgrade py-evm to 0.2.0-beta.2
```
carver: This is the eth-tester version. (in this eth-tester, py-evm is upgraded to https://py-evm.readthedocs.io/en/latest/release_notes.html#v0-3-0-alpha-1 )

Maybe worth noting that that version of py-evm is now 3 months old, with ~5 releases since, so another upgrade and bump of py-evm inside eth-tester might be worthwhile.

kclowes: > This is the eth-tester version

Oops, nice catch!

And upgrading eth-tester again seems like a good idea!

kclowes:
I specifically noticed the ‘tests/core/contracts’ specs were faster. I didn't look at the actual numbers but they’d be quick to get if you want!

Issue #1449, opened 2019-09-10
Websocket timeout isn't working properly


fab-spring:
* Version: 5.0.5b
* Python: 3.7
* OS: osx

### What was wrong?
```
In [19]: w = Web3(Web3.WebsocketProvider('WS_URL_THAT_DOESNT_WORK', websocket_timeout=1, websocket_kwargs={'timeout': 1}))
In [20]: w.isConnected()
```

The commands above hangs. It appears that it doesn't timeout properly.

gohuygo:
@kclowes I addressed this issue in https://github.com/ethereum/web3.py/pull/1452. Please take a look.

voith:
@fab-spring will it be possible for you to provide `WS_URL_THAT_DOESNT_WORK` or at least the format of the URL? refer https://github.com/ethereum/web3.py/pull/1452#issuecomment-532589911

kkx:
hi, i am having the same issues here, when i try to connect to an unreachable endpoint(correct format but unreachable). seems that in the establishing socket connection step we are not using websocket_timeout as a configurable option inside of asyncio.wait_for
i tested to change this [line](https://github.com/ethereum/web3.py/blob/aa30b3bf97d67b8434cb399c659d8b6425a9bec1/web3/providers/websocket.py#L67)
to 
```
self.ws = await asyncio.wait_for(
                    websockets.connect(
                    uri=self.endpoint_uri, loop=self.loop, **self.websocket_kwargs
                  ),
                    timeout=websocket_timeout
                )
```
and worked for me.

kclowes:
Closing, if this is still an issue, please reopen. The WebsocketProvider will become LegacyWebsocketProvider in v7, and there is a new WebsocketProvider that will be released in v7 as well. You can try out the new WebsocketProvider in v6 by importing WebsocketProviderV2. Docs [here](https://web3py.readthedocs.io/en/stable/providers.html#websocketproviderv2-beta).

PR #1452, opened 2019-09-18
Throw TimeoutError if future does not resolve in a timely manner

pumpchaser:
### What was wrong?
The web3py documentation states that we can set a timeout kwarg when instantiating a provider but this doesn't throw a Timeout exception if the future does not get resolved - https://web3py.readthedocs.io/en/stable/providers.html#web3.providers.rpc.HTTPProvider

### How was it fixed?
Pass an optional timeout issue to future.result() which will throw a TimeoutError if the future does not resolve. See documentation link:
https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe

### Todo:

#### Cute Animal Picture


voith:
@gohuygo Can you add a test case for this? As per issue #1449, I'd like to see that when `WS_URL_THAT_DOESNT_WORK ` is provided to `WebsocketProvider`, it raises a TimeoutError. I couldn't replicate the issue locally. It's not clear how this incorrect URL looks like. `websockets` library should error if a non-working URL is inputted. 
In any case, I agree that `future.result` should have a timeout but I'm not sure that the fix in this PR closes #1449.

gohuygo:
@voith Yeah sure, I'll look into adding tests over the weekend

voith:
@gohuygo There is one test for WebSocket timeout. This could be a starting point: https://github.com/ethereum/web3.py/blob/6f9889a52e3911390bcccbec1fdd092e653f7f5e/tests/core/providers/test_websocket_provider.py#L24-L57
There are some integration tests too that you can take a look at. eg:
https://github.com/ethereum/web3.py/blob/master/tests/integration/go_ethereum/test_goethereum_ws.py

gohuygo:
@voith @pipermerriam Can you please take another look? I've added a unit test for this and ensures that it fails without my fix in place.

gohuygo:
@voith I've also updated the description to no longer reference the link to issue #1449 since I couldn't produce a test to replicate it. Regardless, this PR should still be relevant since we can potentially run out of memory of futures never get resolved and we keep opening websocket connections.

tests/core/providers/test_websocket_provider.py
```diff
@@ -64,6 +60,13 @@ def test_websocket_provider_timeout(w3):
         w3.eth.accounts
 
 
+@patch('web3.providers.websocket.WebsocketProvider.coro_make_request')
+def test_websocket_provider_future_timeout(coro_make_request_mock, w3):
+    coro_make_request_mock.side_effect = asyncio.coroutine(lambda o: asyncio.sleep(0.02))
```
gohuygo: This needs to me mocked out or else the timeout will be thrown from within coro_make_request and will never reach `future.result()`

tests/core/providers/test_websocket_provider.py
```diff
@@ -18,15 +23,6 @@
     WebsocketProvider,
 )
 
-if sys.version_info >= (3, 8):
```
voith: Hmm. strange. The tests were failing without this change. I had even checked the documentation and I couldn't find `concurrent.futures.TimeoutError`. But I just checked and it seems like it magically appears again.

tests/core/providers/test_websocket_provider.py
voith: I guess what this does is that the request doesn't actually go to the WebSocket server. So this is trying to simulate the behavior where the coroutine that makes a request to the WebSocket server gets stuck for some reason and this can only happen when the WebSocket library doesn't respect the `websocket_timeout `  and keeps waiting for some reason?
https://github.com/ethereum/web3.py/blob/0aa00c8e44ee0be116c8ad2baea952c05ff7ad34/web3/providers/websocket.py#L98-L109
This could be a bug in the WebSocket library and I don't know if anyone has a test case to reproduce it and report it upstream. I'm not opposed to this change, but I'm not fully convinced by what is causing this!

PS: I'm not the maintainer of this project. However, the WebSocket code was initially added by me and I try to stay by its side.

gohuygo: Yes, pretty much. This simulates the situation where we get a response in a timely manner from the websocket server, but the future response takes too long to resolve (i.e. takes longer than self.websocket_timeout to resolve).

Can you help me tag the right person to get this PR merged? Otherwise, I can bug you guys later on gitter. Thanks

tests/core/providers/test_websocket_provider.py
gohuygo: Yeah, not sure what happened here. I just noticed that 3.8 tests were breaking so I looked at it and saw this. Tests are green now though 👍

tests/core/providers/test_websocket_provider.py
kclowes: @gohuygo I will take a look soon! @voith is more familiar with the websocket provider than I am so I was waiting for his questions to get resolved before I jumped in.

voith: > Can you help me tag the right person to get this PR merged?

That'd be @kclowes!

> I was waiting for his questions to get resolved before I jumped in

My only concern here is not knowing what is causing the WebSocket library to not respect the  `websocket_timeout`. As you can see in the testcase below for `websocket_timeout` works fine when a  request gets timed out. 
https://github.com/ethereum/web3.py/blob/0aa00c8e44ee0be116c8ad2baea952c05ff7ad34/tests/core/providers/test_websocket_provider.py#L62-L64

> takes longer than self.websocket_timeout to resolve

A possible problem I see with `future.result(self.websocket_timeout)` is that if there's a large response that is being streamed, it can get canceled if it exceeds the timeout. This can happen with filters when you request a big chunk of data. Thinking about this now, I'm 👎 on this change.

Instead of mocking what I really wanted to see is the case when the websocket request keeps hanging and doesn't respect the `websocket_timeout`.  Since PR initially started out with a fix for https://github.com/ethereum/web3.py/issues/1449, I was looking for a test case that would simulate that behavior.

TLDR; I'm 👍 if someone proves to me that the websocket library is not respecting `websocket_timeout`.Ideally, I would still like to see this fixed upstream.
I'm 👎 as this can cancel a streaming response.

huy-springlabs: Good point. Let's close this PR then. Thanks for the review

fselmo:
> Good point. Let's close this PR then. Thanks for the review

Closing with the above in mind. This has also been in an idle / stale state for some time. `WebsocketProvider` is likely to be replaced by `WebsocketProviderV2` in the very near future.

Issue #1413, opened 2019-08-07
Web3 Async API Tracking Issue


kclowes:
Copy/paste from #1161, but with checkboxes and slightly more granularity. These will likely become their own issues.

- [x] Replace api methods with instances of a Method class, with a middleware like layer for formatting/normalizing/validating method input/output.

- [ ] Reduce middlewares to only those that make calls to the web3 api, moving the formatting middlewares to the Method input/output formatters.

- [ ] Separate async and sync execution pathway through RequestManagers -> middlewares -> providers.
  - Async Providers:
    - [x] HTTPProvider
    - [ ] IPCProvider
    - [x] WebsocketProvider
  - Async Versions of each Module
    - [x] Eth,
    - [x] Net
    - [x] Version
    - [x] ~~Parity Personal~~
    - [x] ~~Parity Shh~~
    - [x] Geth Admin
    - [ ] Geth Miner
    - [x] Geth Personal
    - [x] ~~Geth Shh~~
    - [x] Geth Tx Pool
    - [x] ~~Testing~~

dbfreem:
@kclowes  I was thinking about taking on one of these Async additions if it didn't turn into a mountain of work. I had one question. 

On the Geth one above I see there is Admin, Miner, Personal, Shh, **TxPool**. I looked at the [docs](https://web3py.readthedocs.io/en/latest/web3.geth.html#gethtxpool-api) on how these methods are used. When I try to run inspect or status (see below code) in TxPool module I get `the method txpool_inspect does not exist/is not available'`. At first glance it looks like this is because only the Personal and Admin modules are wired up in [geth.py](https://github.com/ethereum/web3.py/blob/master/web3/geth.py#L140). 

Is the documentation wrong here or am I missing something

```
from web3 import HTTPProvider, Web3

web3 = Web3(HTTPProvider("http://127.0.0.1:8545"))
web3.isConnected()
print(web3.geth.txpool.inspect())
```

kclowes:
@dbfreem that error message is coming back from your node. You’ll need to be attached to a local Geth node to have access to those methods. Sometimes you also need a flag on startup to have access to those modules on the Geth side. Running ‘geth —help’ should show the commands you need. It don’t remember the exact command off the top of my head but it’s something like ‘—http.api=personal’

rencryptofish:
are all of the above outstanding tasks?

dbfreem:
[...]

PR #1209, opened 2019-01-15
Migrate formatting middlewares to ``Method`` formatters

dylanjw:
### What was wrong?

Based on: #1166

Related to Issue #1161 and #1057

TODO:

- [x] pythonic_middleware
- [x] attrdict_middleware middleware
- [x] request_param_normalizer middleware
- [x] ~validation_middleware~
- [x] abi_middleware

#### Cute Animal Picture

image

PR #1461, opened 2019-09-26
Async formatters

kclowes:
### What was wrong?
Updates #1209 to most recent master since it was opened so long ago.

Related to Issue #1413 

### How was it fixed?
Pulled in master and updated tests.

### Todo:

#### Cute Animal Picture
image


web3/_utils/method_formatters.py
```diff
@@ -0,0 +1,471 @@
...
+
+from web3._utils.abi import (
+    is_length,
+)
+from web3._utils.encoding import (
+    hexstr_if_str,
+    to_hex,
+)
+from web3._utils.formatters import (
+    apply_formatter_at_index,
```
pipermerriam: you can use the `eth_utils` versions of these tools

web3/_utils/method_formatters.py
```diff
@@ -0,0 +1,471 @@
...
+)
+from hexbytes import (
+    HexBytes,
+)
+
+from web3._utils.abi import (
+    is_length,
+)
+from web3._utils.encoding import (
+    hexstr_if_str,
```
pipermerriam: this is available from `eth_utils` (this would be a good "Good First Issue" ticket to open)

tests/core/method-class/test_result_formatters.py
```diff
@@ -0,0 +1,64 @@
...
+    if result == 'ok':
+        return 'OKAY'
+    return result
+
+
+def formatters(method):
+    return ((identity,), ((result_formatter,), None))
+
+
+def test_method():
```
pipermerriam: `pytest` is going to pick this up and run it like it's a test.

web3/_utils/method_formatters.py
```diff
@@ -458,10 +453,13 @@ def method_filter(formatter_dict, method):
 
 @to_tuple
 def get_request_formatters(method):
```
kclowes: @pipermerriam before I do the result/error formatters, is this along the lines of what you were thinking?

pipermerriam: Not quite.  This still results in the formatters being a tuple of formatter functions instead of what I think would be a better API of it being a single formatter function.

```python
def get_request_formatter(method):
    formatters = ... # use a rough equivalent of the current logic to grab the right formatter functions from the PYTHONIC and ABI formatters
    return compose(*reversed(formatters))  # I think the `reversed` is needed to have them apply in the right order.
```

This results in a single formatter function for each method.

kclowes: Aha! Okay, that makes sense to me. I like that.

web3/manager.py
```diff
@@ -87,22 +99,24 @@ def _make_request(self, method, params):
...
 
         return response['result']
 
-    async def coro_request(self, method, params):
+    async def coro_request(self, method, params, error_formatters=None):
         """
         Couroutine for making a request using the provider
         """
         response = await self._coro_make_request(method, params)
+        apply_error_formatters(error_formatters, response)
```
pipermerriam: Interesting.  I would have expected this to be inside of the `if "error" in response` block below.  Is there a reason it is here?

web3/method.py
```diff
@@ -104,21 +113,7 @@ def method_selector_fn(self):
...
-    def input_munger(self, val):
-        try:
-            module, args, kwargs = val
-        except TypeError:
-            raise ValueError("input_munger expects a 3-tuple")
-
+    def input_munger(self, module, args, kwargs):
         # TODO: Create friendly error output.
         mungers_iter = iter(self.mungers)
         root_munger = next(mungers_iter)
```
pipermerriam: I know this code wasn't changed, but it strikes me as something that is in need of a nice comment explaining what it is doing as it is quite cryptic at-a-glance.

web3/manager.py
kclowes: Hm, good catch. Not that I can tell. Moved!

PR #1465, opened 2019-10-01
Eth Module 

Patil2099:
#1429

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -70,8 +70,12 @@ class Eth(Module):
         raise NotImplementedError()
 
     @property
+    def protocol_version(self):
+        return self.web3.manager.request_blocking("eth_protocol_version", [])
+
+    @property
     def protocolVersion(self):
-        return self.web3.manager.request_blocking("eth_protocolVersion", [])
+        raise DeprecationWarning("This method has been deprecated")
 
     @property
     def syncing(self):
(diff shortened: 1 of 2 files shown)


Patil2099:
@kclowes Do I need to change the name of Attributes also?
Thanks!

kclowes:
@Patil2099 if you mean attributes that come back in the response, then no. We will leave those as-is. We are just changing the method names. Thank you!

Patil2099:
@kclowes Can you please check my PR and let me know if everything is correct? I haven't changed the test cases that is the reason for all the test failing.

web3/eth.py
```diff
@@ -85,68 +89,88 @@ def coinbase(self):
...
-        return self.web3.manager.request_blocking("eth_blockNumber", [])
+    def block_number(self):
+        return self.web3.manager.request_blocking("eth_block_number", [])
 
     @property
     def chainId(self):
         return self.web3.manager.request_blocking("eth_chainId", [])
 
-    def getBalance(self, account, block_identifier=None):
+    def get_balance(self, account, block_identifier=None):
```
kclowes: We'll want to follow the same pattern that is in place in the Geth module. So we'll need to add a utils file (probably called `web3/_utils/eth.py`, and define variables like `get_balance = Method('eth_getBalance', mungers=[...])`. Then we'll also need to add the deprecated methods like: `getBalance = DeprecatedMethod(...)`. Then we'll need to import those into this file like we do [here](https://github.com/ethereum/web3.py/blob/master/web3/geth.py#L5). We'll also need to make a class like we do [here](https://github.com/ethereum/web3.py/blob/master/web3/geth.py#L80). Hopefully that's enough to get you going, but let me know if anything is unclear!

web3/eth.py
```diff
@@ -85,68 +89,88 @@ def coinbase(self):
...
         if block_identifier is None:
             block_identifier = self.defaultBlock
         return self.web3.manager.request_blocking(
-            "eth_getBalance",
+            "eth_get_balance",
```
kclowes: I believe this will still be `eth_getBalance`. I think it should be the actual JSON-RPC method name.

PR #1471, opened 2019-10-17
Add python3.8 support

voith:
### What was wrong?
web3.py doesn't support python3.8.

### How was it fixed?
Add support for python 3.8. 
### Todo:

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -370,6 +370,113 @@ jobs:
     environment:
       TOXENV: py37-integration-ethtester
       ETHEREUM_TESTER_CHAIN_BACKEND: eth_tester.backends.PyEVMBackend
+  #
+  # Python 3.8
+  #
+  py38-core:
+    <<: *common
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-core
+
+  py38-ens:
+    <<: *common
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-ens
+
+  py38-ethpm:
+    <<: *ethpm_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-ethpm
+      WEB3_INFURA_PROJECT_ID: 4f1a358967c7474aae6f8f4a7698aefc
+
+  py38-integration-goethereum-ipc-1.7.2:
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-ipc
+      GETH_VERSION: v1.7.2
+
+  py38-integration-goethereum-http-1.7.2:
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-http
+      GETH_VERSION: v1.7.2
+
+  py38-integration-goethereum-ws-1.7.2:
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-ws
+      GETH_VERSION: v1.7.2
+
+  py38-integration-goethereum-ipc-1.8.22:
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-ipc
+      GETH_VERSION: v1.8.22
+
+  py38-integration-goethereum-http-1.8.22:
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-http
+      GETH_VERSION: v1.8.22
+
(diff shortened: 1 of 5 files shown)


voith:
Most of Ethereum's python tooling won't be able to update to `python 3.8` unless https://github.com/pytoolz/cytoolz/issues/136 is resolved. I have been actively following the issue, but there haven't been any updates to it.

voith:
A fix for cytoolz has been released. I have opened a [PR](https://github.com/ethereum/eth-utils/pull/174) in `eth-utils` to add support for python3.8.

setup.py
```diff
@@ -78,7 +78,8 @@
         "protobuf>=3.0.0,<4",
         "pypiwin32>=223;platform_system=='Windows'",
         "requests>=2.16.0,<3.0.0",
-        "websockets>=7.0.0,<8.0.0",
+        "websockets>=8.1.0,<9.0.0",
+        "cytoolz>=0.10.1",
```
voith: need to get rid of this once python3.8 support for `eth-utils` is released.

.circleci/config.yml
```diff
@@ -179,7 +179,8 @@ jobs:
       - image: circleci/python:3.6
     environment:
       TOXENV: py36-ethpm
-      WEB3_INFURA_PROJECT_ID: 4f1a358967c7474aae6f8f4a7698aefc
+      WEB3_INFURA_PROJECT_ID: ae13436d29b34850b640af9c80f80871
+      WEB3_INFURA_API_SECRET: 067ddad6021048f0840a38a4e52eb568
```
voith: @kclowes @njgheorghita Tests were failing without `WEB3_INFURA_API_SECRET `. [Here's](https://circleci.com/gh/ethereum/web3.py/79087?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) a failing build. It seems to me that the handshake in websockets is much more strict in this new version(`v8.1`).  I have generated this temporary key pair from my personal account. Please create a new key pair. The infura api is rate limited, so it'd be better to add this environment variables through [circleci's config](https://circleci.com/gh/ethereum/web3.py/79087?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). Else the tests might fail randomly with rate limit errors if someone starts abusing this key.

.circleci/config.yml
```diff
@@ -370,6 +372,114 @@ jobs:
...
+    docker:
+      - image: circleci/python:3.8
+    environment:
+      TOXENV: py38-integration-goethereum-ws
+      GETH_VERSION: v1.8.22
+
+  py38-integration-parity-ipc:
+    <<: *parity_steps
+    docker:
+      - image: circleci/python:3.8
```
voith: I want to point out that it seems like stretch images are no longer needed. Back when [this commit](https://github.com/ethereum/web3.py/commit/8327480b0f1408c59050e243d8b664b065d4c2c4) was added parity would fail to build on other distros.

.circleci/config.yml
kclowes: 🤔 that's interesting. It seems like you shouldn't need the secret key unless something changed with that project. I wonder if the project settings associated with the key changed recently? It's not my key so I can't see it.

> it'd be better to add these environment variables through circleci's config. 

We opted to leave the keys in the code (usually with a comment that says please don't abuse) because anyone who contributes has access to the circle env variables anyway and we thought it's better to just put them out there, rather than have someone assume at some point that the CircleCI env vars are actually secret and expose a real vulnerability. If someone abuses we'll just change them out.

setup.py
pipermerriam: Available in `v1.8.0` now.

.circleci/config.yml
njgheorghita: > Tests were failing without WEB3_INFURA_API_SECRET .

>  It seems like you shouldn't need the secret key unless something changed with that project. 

Yup, nothing's changed with the project settings, so I'm not sure that this was the cause of the error, as it looks like we're not close to hitting the rate limits on that key. For the `4f1....` key, you don't need a secret. I'm guessing it was just a glitchy request with Infura. I just ran the tests with the `4f1...` key and everything seemed to pass just fine.

voith: well, I couldn't replicate the error locally on OSX. However, after entering circleCI in debug mode, the secret key is the only thing that fixed the issue. I was getting a 401 unauthorized error. 
Here's a similar [issue](https://github.com/aaugustin/websockets/issues/575) in websockets.

voith:
argh, looks like the checksum of parity binary has changed. I can take a look at this tomorrow. 
If anyone else wants to take a stab at this then please be my guest.

voith:
> argh, looks like the checksum of parity binary has changed
Tests are green now. The test failures were random and had nothing to do with the change in the checksum.

.circleci/config.yml
kclowes: I couldn't get circle tests to pass without the secret key either ¯\_(ツ)_/¯. Updated to a throw-away key.

PR #1478, opened 2019-10-22
Add Web3Exception to all exception classes

jpic:
Exception mixin inherited by all exceptions of web3.py

This allows:

    try:
        some_call()
    except Web3Exception as e:
        # deal with web3 exception
    except:
        # deal with other exceptions

### What was wrong?

I currently have this code:

```
        try:    
            result = something_that_calls_web3()
        except (    
            web3.exceptions.BadFunctionCallOutput    
            web3.exceptions.BlockNotFound    
            web3.exceptions.BlockNumberOutofRange    
            web3.exceptions.CannotHandleRequest    
            web3.exceptions.FallbackNotFound    
            web3.exceptions.InfuraKeyNotFound    
            web3.exceptions.InsufficientData    
            web3.exceptions.InvalidAddress    
            web3.exceptions.InvalidEventABI    
            web3.exceptions.LogTopicError    
            web3.exceptions.ManifestValidationError    
            web3.exceptions.MismatchedABI    
            web3.exceptions.NameNotFound    
            web3.exceptions.NoABIEventsFound    
            web3.exceptions.NoABIFound    
            web3.exceptions.NoABIFunctionsFound    
            web3.exceptions.PMError    
            web3.exceptions.StaleBlockchain    
            web3.exceptions.TimeExhausted    
            web3.exceptions.TransactionNotFound    
            web3.exceptions.ValidationError    
        ) as e:    
            deal_with_web3_exception()
```

This has readability problem, and also not forward-compatible with new exceptions, and also will break if an exception is removed from web3 package.

### How was it fixed?

With this patch, I could replace the above with:

```
        try:    
            result = something_that_calls_web3()
        except web3.exceptions.Web3Exception as e:    
            deal_with_web3_exception()
```

This fixes readability and forward-compatibility.

#### Cute Animal Picture

### DIFF
--- a/web3/exceptions.py
+++ b/web3/exceptions.py
@@ -201,22 +201,6 @@ class TimeExhausted(Web3Exception):
     pass
 
 
-class PMError(Web3Exception):
-    """
-    Raised when an error occurs in the PM module.
-    """
-
-    pass
-
-
-class ManifestValidationError(PMError):
-    """
-    Raised when a provided manifest cannot be published, since it's invalid.
-    """
-
-    pass
-
-
 class TransactionNotFound(Web3Exception):
     """
     Raised when a tx hash used to lookup a tx in a jsonrpc call cannot be found.
--- a/web3/exceptions.py
+++ b/web3/exceptions.py
@@ -5,6 +5,10 @@ from typing import (
     Dict,
 )
 
+from eth_utils import (
+    ValidationError,
+)
+
 from web3.types import (
     BlockData,
 )
@@ -143,7 +147,7 @@ class FallbackNotFound(Web3Exception):
     pass
 
 
-class ValidationError(Web3Exception):
+class Web3ValidationError(Web3Exception, ValidationError):
     """
     Raised when a supplied value is invalid.
     """
@@ -151,7 +155,7 @@ class ValidationError(Web3Exception):
     pass
 
 
-class ExtraDataLengthError(ValidationError):
+class ExtraDataLengthError(Web3ValidationError):
     """
     Raised when an RPC call returns >32 bytes of extraData.
     """
--- a/web3/exceptions.py
+++ b/web3/exceptions.py
@@ -10,7 +10,22 @@ from web3.types import (
 )
 
 
-class BadFunctionCallOutput(Exception):
+class Web3Exception(Exception):
+    """
+    Exception mixin inherited by all exceptions of web3.py
+
+    This allows::
+
+        try:
+            some_call()
+        except Web3Exception:
+            # deal with web3 exception
+        except:
+            # deal with other exceptions
+    """
+
+
+class BadFunctionCallOutput(Web3Exception):
     """
     We failed to decode ABI output.
 
@@ -20,7 +35,7 @@ class BadFunctionCallOutput(Exception):
     pass
 
 
-class BlockNumberOutofRange(Exception):
+class BlockNumberOutofRange(Web3Exception):
     """
     block_identifier passed does not match known block.
     """
@@ -28,7 +43,7 @@ class BlockNumberOutofRange(Exception):
     pass
 
 
-class CannotHandleRequest(Exception):
+class CannotHandleRequest(Web3Exception):
     """
     Raised by a provider to signal that it cannot handle an RPC request and
     that the manager should proceed to the next provider.
@@ -37,7 +52,7 @@ class CannotHandleRequest(Exception):
     pass
 
 
-class TooManyRequests(Exception):
+class TooManyRequests(Web3Exception):
     """
     Raised by a provider to signal that too many requests have been made consecutively.
     """
@@ -45,7 +60,7 @@ class TooManyRequests(Exception):
     pass
 
 
-class MultipleFailedRequests(Exception):
+class MultipleFailedRequests(Web3Exception):
     """
     Raised by a provider to signal that multiple requests to retrieve the same
     (or similar) data have failed.
@@ -54,7 +69,7 @@ class MultipleFailedRequests(Exception):
     pass
 
 
-class InvalidAddress(ValueError):
+class InvalidAddress(Web3Exception):
     """
     The supplied address does not have a valid checksum, as defined in EIP-55
     """
@@ -62,7 +77,7 @@ class InvalidAddress(ValueError):
     pass
 
 
-class NameNotFound(ValueError):
+class NameNotFound(Web3Exception):
     """
     Raised when a caller provides an Ethereum Name Service name that
     does not resolve to an address.
@@ -71,7 +86,7 @@ class NameNotFound(ValueError):
     pass
 
 
-class StaleBlockchain(Exception):
+class StaleBlockchain(Web3Exception):
     """
     Raised by the stalecheck_middleware when the latest block is too old.
     """
@@ -93,7 +108,7 @@ class StaleBlockchain(Exception):
         return self.args[0]
 
 
-class MismatchedABI(Exception):
+class MismatchedABI(Web3Exception):
     """
     Raised when an ABI does not match with supplied parameters, or when an
     attempt is made to access a function/event that does not exist in the ABI.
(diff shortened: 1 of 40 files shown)


web3/exceptions.py
```diff
@@ -2,7 +2,22 @@
...
+        try:
+            some_call()
+        except Web3Exception as e:
+            # deal with web3 exception
+        except:
+            # deal with other exceptions
+    """
+
+
+class BadFunctionCallOutput(Exception, Web3Exception):
```
pipermerriam: I'd advocate for `Web3Exception` to inherit from `Exception` and to keep the inheritance chain for all of these child exception classes to only have a single parent of `Web3Exception`.

web3/exceptions.py
```diff
@@ -58,108 +73,108 @@ def __str__(self):
...
-class NoABIFunctionsFound(AttributeError):
+class NoABIFunctionsFound(AttributeError, Web3Exception):
     """
     Raised when an ABI is present, but doesn't contain any functions.
     """
     pass
 
 
-class NoABIFound(AttributeError):
+class NoABIFound(AttributeError, Web3Exception):
```
pipermerriam: @kclowes can you make a note in your breaking v6 changes tracking issue to remove `AttributeError` from these exception classes bases.

web3/exceptions.py
```diff
@@ -58,108 +73,108 @@ def __str__(self):
...
-class InfuraKeyNotFound(Exception):
+class InfuraKeyNotFound(Exception, Web3Exception):
     """
     Raised when there is no Infura Project Id set.
     """
     pass
 
 
-class LogTopicError(ValueError):
+class LogTopicError(ValueError, Web3Exception):
```
pipermerriam: @kclowes same here.

web3/exceptions.py
jpic: Thanks for permission to break BC, notes added in new v6 migration document.

jpic:
Added v6 migration notes

kclowes:
Thank you for this @jpic! We're finally about ready to pull it in. I just fixed a bunch of merge conflicts that have accumulated, and need to make some updates in the v6 migration guide, but then it looks good to go!

web3/exceptions.py
```diff
@@ -20,15 +35,15 @@ class BadFunctionCallOutput(Exception):
...
+class BlockNumberOutofRange(Web3Exception):
     """
     block_identifier passed does not match known block.
     """
 
     pass
 
 
-class CannotHandleRequest(Exception):
+class CannotHandleRequest(Web3Exception):
```
pacrob, line 46: Are `TooManyRequests` and `MultipleFailedRequests` left inheriting form `Exception` intentionally?

kclowes: 🤔 Nope, not sure how I missed those. Nice catch, thanks!

pacrob (approved):
lgtm!

fselmo (commented):
I'm happy to see this making it in. I just need a bit of clarification on some things.

- `eth_utils` has a `ValidationError` that we use in a few places within `web3.py` while in others we use `web3.exceptions.ValidationError`. It looks like at some point there was an idea to use the `eth_utils.exceptions.ValidationError` [across all of the python libraries](https://github.com/ethereum/eth-utils/issues/114). Indeed this isn't used anywhere within the actual `eth-utils` library. Do we want to keep this design? Are there good arguments for or against it? This seems like a good time to revisit this for `web3.py` and it wouldn't be so hard of a change to make. If we think a better design is for each project to have their own `ValidationError` class, do we want to go toward namespace validation errors, as we did with `EthPMValidationError`?

- nit: I slightly prefer the less lengthy `EthPMException` over `PyEthPMException`. The `Py` seems redundant and the module name is `ethpm`. But I can take or leave it. It does make `EthPMValidationError` seem like it needs a `Py` if we keep it... but I prefer both without it.

---

edit: [link to py-evm PR for this change](https://github.com/ethereum/py-evm/pull/1194)

kclowes:
> - eth_utils has a ValidationError that we use in a few places within web3.py while in others we use web3.exceptions.ValidationError. It looks like at some point there was an idea to use the eth_utils.exceptions.ValidationError https://github.com/ethereum/eth-utils/issues/114. Indeed this isn't used anywhere within the actual eth-utils library. Do we want to keep this design? Are there good arguments for or against it? This seems like a good time to revisit this for web3.py and it wouldn't be so hard of a change to make. If we think a better design is for each project to have their own ValidationError class, do we want to go toward namespace validation errors, as we did with EthPMValidationError?

I lean toward each project having their own `ValidationError`, because the whole point of this PR is to be able to catch the top-level `Web3Exception`. If the `ValidationError` is from `eth-utils`, then you'd have to catch `Web3Exception` and `eth_utils.exceptions.ValidationError`. I guess another route we could go is to have the `ValidationError` from web3 inherit from the eth-utils exception, but I'm not sure what would be gained from that. I guess I should rename to `Web3ValidationError` then? Also curious if you have a different opinion @fselmo. It looks like in the py-evm PR they hit an issue where they were catching the wrong `ValidationError` but that would be remedied if the `ValidationError` class was renamed to `Web3ValidationError`. And if we're going to prefix the `ValidationError` with `Web3` I could see an argument for prefixing all Exceptions with `Web3` but that seems redundant.

> - nit: I slightly prefer the less lengthy EthPMException over PyEthPMException. The Py seems redundant and the module name is ethpm. But I can take or leave it. It does make EthPMValidationError seem like it needs a Py if we keep it... but I prefer both without it.

Yeah, agree the Py seems redundant. Will update!

fselmo:
I personally don't see a scenario where I'd want to catch only web3 validation errors instead of catching them across all web3 libraries that `web3.py` interacts with. `web3.py` hooks up its codec from `eth-abi`, makes calls through `eth-tester` to `py-evm`, uses `eth-account` for the account logic... etc... I think that catching a single `ValidationError` between all these still makes sense.

However... if we want the most fine tuning, I don't see why we can't make `Web3ValidationError`, similarly to `EthPMValidationError` (which you mentioned as well). And if we go this route, I don't think we need to prefix all `web3.py` exception classes with "Web3"... only the ones that are shared across all the libraries (so far only the `ValidationError`).

Without a general `ValidationError`, catching all validation errors across stack calls would look something like:
```py
from web3.exceptions import Web3ValidationError
from eth_tester.exceptions import EthTesterValidationError
from eth.exceptions import PyEVMValidationError
...

ALL_WEB3_VALIDATION_ERRORS = [
    Web3ValidationError, EthTesterValidationError, PyEvmValidationError, ...,
]

try:
    ...
except Exception as e:
    if isinstance(e, ALL_WEB3_VALIDATION_ERRORS):
        ...
```

which looks a bit similar to the original issue filed for web3 exceptions in general.

With a general `ValidationError` within `eth-utils`:
```py
from eth_utils.exceptions import ValidationError

try:
    ...
except ValidationError:
        ...
```

For even more fine tuning, if we still opt for name-spaced validation errors which inherit from the general `eth-utils` class:

```py
from eth_utils.exceptions import ValidationError

class Web3ValidationError(Web3Exception, ValidationError):
    pass
    
 
try:
    ...
    
# catch all web3 exceptions, including Web3ValidationError
except Web3Exception:
    ...
    
# OR catch only web3 validation exceptions
except Web3ValidationError:
    ...
    
# OR catch any validation errors that may happen on calls to other web3 libraries
except ValidationError:
    ...
```

My vote is either using `ValidationError` across all libraries (the old intended way discussed in 2018) or inheriting each library's validation error from that `eth-utils` class. Thoughts?

kclowes:
> I personally don't see a scenario where I'd want to catch only web3 validation errors instead of catching them across all web3 libraries that web3.py interacts with

Yeah, that's a good point. It feels a little like clunky to me to expect a user to import from `eth-utils` exceptions through the web3 library. Let's go with the inheritance, since that offers the most customization and I'll update `ValidationError` -> `Web3ValidationError` here.

fselmo:
Makes sense. We would only need to import `ValidationError` into the `web3.exceptions` file in order to pass it through but that's a pretty roundabout way of doing it and not immediately obvious what's happening to someone reading the code. I dig the inheritance approach 👍🏼

fselmo (approved):
lgtm 👍🏼 

We can definitely benefit from some brief documentation on this hierarchy, perhaps in a separate PR?

kclowes:
Good thought! I'll add it to the migration guide PR that was created off of this branch. As soon as this gets in, I'll PR that.

Issue #1197, opened 2019-01-09
Remove `web3._utils.formatters` in favor of `eth-utils` versions


pipermerriam:
Depends on https://github.com/ethereum/web3.py/pull/1195 which upgrades the `eth-utils` dependency

### What was wrong?

https://github.com/ethereum/web3.py/blob/864bc1d2a93ba7a6ca2a0629839b0d7a4c944e12/web3/_utils/formatters.py

We currently have a module `web3._utils.formatters` which implements a number of *functional* formatter functions.  These functions were elevated to live in `eth-utils` recently and thus are duplicate code.

### How can it be fixed?

Remove any/all functions from `web3._utils.formatters` in favor of the [`eth-utils` versions](https://eth-utils.readthedocs.io/en/latest/utilities.html#applicators).

dylanjw:
Can I include this change in #1209?  It might be the same amount of work to do as it would to rebase this change from master.

pipermerriam:
If they can be done separately that would be ideal since this change should just be code removal and modification of imports.  Mixing that with functionality changes makes for more difficult review.  Feel free to write them together but to segregate your commits so that when you're done you can split them into two PRs.

PR #1479, opened 2019-10-24
Use eth-utils methods instead of custom web3 methods

kclowes:
### What was wrong?
There were a couple duplicate functions in both eth-utils and web3. We want to standardize on eth-utils.

Related to Issue #1197 

### How was it fixed?

Took out the duplicate methods from web3, and added a note to the v6 breaking changes for apply_formatters_to_array which is very close but not quite the same between the two libraries. 

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/1479.misc.rst
+++ b/newsfragments/1479.misc.rst
@@ -1,0 +1,1 @@
+Use formatting functions from eth-utils instead of web3
--- a/tests/core/utilities/test_formatters.py
+++ b/tests/core/utilities/test_formatters.py
@@ -1,8 +1,11 @@
 
 import pytest
 
-from web3._utils.formatters import (
+from eth_utils.curried import (
     apply_formatters_to_dict,
+)
+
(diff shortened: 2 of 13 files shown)


kclowes:
@pipermerriam I think it is a breaking change because we use it in middleware? The web3 `apply_formatters_to_array` converts the result to a list, but the eth-utils version does not. Because we use the web3 `apply_formatters_to_array` in the pythonic formatting middleware, a bunch of the ENS tests started failing because they were expecting a list to be returned from `w3.eth.accounts`, but got a tuple instead. I suspect others may run into the same problem. Maybe it's better to use the eth-utils `apply_formatters_to_array` and convert to a list on the web3 side?

pipermerriam:
> Maybe it's better to use the eth-utils `apply_formatters_to_array` and convert to a list on the web3 side?

That is what I would advocate for (and ideally leave a note to do any cleanup in v6 if you think it's worth removing the shim then)

Issue #1481, opened 2019-10-28
Make Quick Start in the docs easier to follow


jschiarizzi:
* Version: 5.2.2

### What was wrong?

The quick start is not quick enough. Several python developers I recommended it to were stuck on connecting a node. The quick start example requires users to look through docs elsewhere (bad user experience, most will quit at this step) in order to connect to a provider.

### How can it be fixed?

Adding a simple explanation to the quick start docs explaining how to quickly connect to an infura provider using a global variable.

PR #1482, opened 2019-10-28
add quickest infura instructions

jschiarizzi:
### What was wrong?
The quick start was not quick enough. Several python developers I recommended it to were stuck on connecting a node. The quick start example required users to look through docs elsewhere (bad user experience, most will quit at this step) in order to connect to a provider.

Related to Issue #1481

### How was it fixed?
Added explanation of Infura and setting Infura global variable to quick start docs.

### Todo:

#### Cute Animal Picture (my puppy Stella)
1021191535_HDR

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -40,14 +40,16 @@ To use the web3 library you will need to initialize the
     
     $ export WEB3_INFURA_PROJECT_ID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-Use the ``auto`` module to :ref:`guess at common node connection options
-<automatic_provider_detection>`. Or specify to use the Infura node by importing from ``web3.auto.infura``.
+Use the ``web3.auto.infura`` module to connect to the Infura node.
 
 .. code-block:: python
 
     >>> from web3.auto import w3
     >>> w3.eth.blockNumber
     4000000
+    
+.. NOTE:: Use the ``auto`` module to :ref:`guess at common node connection options
+<automatic_provider_detection>`.
 
 This ``w3`` instance will now allow you to interact with the Ethereum
 blockchain.
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -34,25 +34,36 @@ Using Web3
 ----------
 
 To use the web3 library you will need to initialize the
-:class:`~web3.Web3` class and connecting to an Ethereum node. The quickest way to do so for free is by setting up an account on `Infura https://infura.io/>`. On Infura, create a project and copy the Project ID. Then set the environment variable ``WEB3_INFURA_PROJECT_ID``.
+:class:`~web3.Web3` class and connecting to an Ethereum node.
+The quickest way to do so for free is by setting up an account on
+`Infura https://infura.io/>`. On Infura, create a project and copy
+the Project ID. Then set the environment variable ``WEB3_INFURA_PROJECT_ID``.
 
 .. code-block:: shell
-    
+
     $ export WEB3_INFURA_PROJECT_ID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 Use the ``web3.auto.infura`` module to connect to the Infura node.
 
 .. code-block:: python
 
-    >>> from web3.auto import w3
+    >>> from web3.auto.infura import w3
     >>> w3.eth.blockNumber
     4000000
-    
-.. NOTE:: 
-    Use the ``auto`` module to :ref:`guess at common node connection options <automatic_provider_detection>`.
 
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -34,14 +34,18 @@ Using Web3
 ----------
 
 To use the web3 library you will need to initialize the
-:class:`~web3.Web3` class.
+:class:`~web3.Web3` class by connecting to an Ethereum node. The quickest way to do so for free is by setting up an account on `Infura https://infura.io/>`. On Infura, create a project and copy the Project ID. Then set the environment variable ``WEB3_INFURA_PROJECT_ID``.
```
kclowes: I would say here: `class and connect to an Ethereum node... `. It's not quite accurate to say initialize the class _by_ connecting to the node, because you can initialize the class without connecting to a node.

docs/quickstart.rst
```diff
@@ -34,14 +34,18 @@ Using Web3
...
+    
+    $ export WEB3_INFURA_PROJECT_ID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 Use the ``auto`` module to :ref:`guess at common node connection options
-<automatic_provider_detection>`.
+<automatic_provider_detection>`. Or specify to use the Infura node by importing from ``web3.auto.infura``.
 
 .. code-block:: python
 
-    >>> from web3.auto import w3
```
kclowes: I think it would be clearer to keep this line, but maybe move it to after line 43. So something like: 

```
Use the ``auto`` module to :ref:`guess at common node connection options <automatic_provider_detection>`.

.. code-block:: python

    >>> from web3.auto import w3

Or, specify the use of an Infura node by importing from ``web3.auto.infura``:

.. code-block:: python

    >>> from web3.auto.infura import w3
    >>> w3.eth.blockNumber
    4000000
```

docs/quickstart.rst
jschiarizzi: Agree, my wording is incorrect, that makes more sense

docs/quickstart.rst
jschiarizzi: I agree that would be more clear, but I'm trying to keep it as short as possible. Obviously using `eth.auto` was chosen for the quick start, however, I've found `eth.auto.infura` to "just work"  more reliably and I would even remove the `eth.atuo` part. Since `eth.auto` was chosen to use for a reason I think it's fine to leave in but the focus should be on getting it going fastest, not explaining various connecting possibilities, IMHO.  :)

What do you think?

docs/quickstart.rst
jschiarizzi: fixed with commit d9b744a

docs/quickstart.rst
kclowes: I agree that `web3.auto.infura` tends to work more reliably than `web3.auto` (since you have to either be running a node or have an environment variable set), and I think it adds value to have it in the Quickstart. Maybe we could add a note to the section or something: 

```
.. NOTE::

If importing your web3 instance via `web3.auto` doesn't work, try connecting to Infura using `web3.auto.infura`
```

jschiarizzi: My point is that, since this is a "quick start" and not actually a tutorial `web3.auto.infura` should be default, and `web3.auto` should be foot note if that's how we want to do it

kclowes: I'm good with that. I just think it's confusing that it says use the web3 auto module and doesn't show how. We can add that piece to a note though.

jschiarizzi: That's fair. moved the explanation of Auto and the link to other connection methods to a note under the connection block of code.

jschiarizzi:
oof now i'm failing a check... Let's see if I can fix it not.

jschiarizzi:
Great, looks like the checks are passing now. I never made a .rst file before so that syntax took a hot minute to figure out. :)  Should be good for a merge now @kclowes

docs/quickstart.rst
```diff
@@ -34,26 +34,37 @@ Using Web3
...
     >>> w3.eth.blockNumber
     4000000
-    
-.. NOTE:: 
-    Use the ``auto`` module to :ref:`guess at common node connection options <automatic_provider_detection>`.
 
 This ``w3`` instance will now allow you to interact with the Ethereum
 blockchain.
 
+.. NOTE::
```
kclowes: Any objections to the updated wording @jschiarizzi?

kclowes:
Thanks @jschiarizzi! Feel free to open another PR if you don't like my updates :)

jschiarizzi:
Looking good, thanks. Sorry I didn't comment, busy week.

PR #1486, opened 2019-10-29
[ethPM] Update registry URI definition

njgheorghita:
### What was wrong?
I've received a request to improve the registry uri definition to allow specific assets within a package to be specified. Also, I think this is a cleaner way to define the target package version.

Old URI definition
`erc1319://snakecharmers.eth/dai?version=1.0.0`

New URI definition
- to specify a package
`erc1319://snakecharmers.eth/dai@1.0.0`

- to specify any specific asset in a package, just namespace it after the version
`erc1319://snakecharmers.eth/dai@1.0.0/contract_types/DSToken/abi`
`erc1319://snakecharmers.eth/dai@1.0.0/sources/token.sol`

#### Cute Animal Picture
image

### DIFF
--- a/docs/ethpm.rst
+++ b/docs/ethpm.rst
@@ -216,13 +216,13 @@ way through the EIP process)
 
 ::
 
-   scheme://address:chain_id/package-name?version=x.x.x
+   scheme://address:chain_id/package-name@version
 
 -  URI must be a string type
--  ``scheme``: ``erc1319``
--  ``address``: Must be a valid ENS domain or a valid checksum address
+-  ``scheme``: (required) ``ethpm`` or ``erc1319``
+-  ``address``: (required) Must be a valid ENS domain or a valid checksum address
    pointing towards a registry contract.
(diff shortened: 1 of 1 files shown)


njgheorghita:
@pipermerriam Looking for feedback on this update - and then I'll set about implementing the changes

pipermerriam:
In general, :+1: to this idea of more expressiveness, though we want to be absolutely sure that there cannot be any ambiguity in the various components.  For example, there is nothing in the spec for version strings that forbids the `@` symbol from being part of the version string.  Similarly, I don't believe that `/` is disallowed in version strings either, which would mean that given the url `erc1319://snakecharmers.eth/dai@1.0.0/contract_types/DSToken/abi` the version could either be `1.0.0` or `1.0.0/contract_types/DSToken/abi`

Figuring out how to properly specify the scheme such that it eliminates this sort of ambiguity is going to be necessary.

This also suggests that you may want a test suite that uses hypothesis to generate the various values for a url, construct the url, and then re-parse it to see if the values survive the round trip.

PR #1488, opened 2019-11-01
Add type hints to web3.main

njgheorghita:
### What was wrong?
Start adding mypy type hints to `web3` - starting with `web3.main`.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -15,7 +15,7 @@ from eth_utils import (
 from hexbytes import (
     HexBytes,
 )
-from typing import Any, Dict, Iterable, List, TYPE_CHECKING, Union
+from typing import Any, cast, Dict, List, Sequence, TYPE_CHECKING
 
 from eth_typing import HexStr, Primitives
 from eth_typing.abi import TypeStr
@@ -32,7 +32,6 @@ from web3._utils.decorators import (
 )
 from web3._utils.empty import (
     empty,
-    Empty,
 )
 from web3._utils.encoding import (
     hex_encode_abi_type,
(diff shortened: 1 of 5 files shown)


web3/main.py
```diff
@@ -223,38 +241,38 @@ def solidityKeccak(cls, abi_types, values):
...
+    def isConnected(self) -> bool:
         return self.provider.isConnected()
 
-    def is_encodable(self, _type, value):
+    def is_encodable(self, _type: TypeStr, value: Any) -> bool:
         return self.codec.is_encodable(_type, value)
 
     @property
-    def ens(self):
+    def ens(self) -> Union[ENS, Empty]:
```
njgheorghita: I'm not sure if there's a better way to handle this, since it should always be returning only `ENS` - but since `web3.ens` is set to a default `Empty` type in the constructor args, this seems like the only way to make mypy happy.

pipermerriam: I think you could fix this using a `cast` right?

web3/main.py
```diff
@@ -137,7 +151,13 @@ class Web3:
     isChecksumAddress = staticmethod(is_checksum_address)
     toChecksumAddress = staticmethod(to_checksum_address)
 
-    def __init__(self, provider=None, middlewares=None, modules=None, ens=empty):
+    def __init__(
+        self,
+        provider: BaseProvider=None,
+        middlewares: List[Any]=None,
```
pipermerriam: This should probably be `Sequence`

web3/main.py
```diff
@@ -137,7 +151,13 @@ class Web3:
...
+        modules: Dict[str, Iterable[Any]]=None,
```
pipermerriam: This should probably be `Sequence[Any]`.

PR #1489, opened 2019-11-01
[ethpm] Support ENS in Registry URIs

njgheorghita:
### What was wrong?
ENS support was needed in ERC1319 URIs.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/backends/registry.py
+++ b/ethpm/backends/registry.py
@@ -8,7 +8,11 @@ from urllib import (
 from eth_typing import (
     URI,
 )
+from eth_utils import (
+    is_address,
+)
 
+from ens import ENS
 from ethpm._utils.registry import (
     fetch_standard_registry_abi,
 )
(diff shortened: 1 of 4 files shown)


njgheorghita:
ping @pipermerriam for :eyes:

ethpm/backends/registry.py
```diff
@@ -78,11 +82,23 @@ def parse_registry_uri(uri: str) -> RegistryURI:
...
+    from web3.auto.infura import w3
     validate_registry_uri(uri)
     parsed_uri = parse.urlparse(uri)
-    address, chain_id = parsed_uri.netloc.split(":")
+    address_or_ens, chain_id = parsed_uri.netloc.split(":")
+    ns = ENS.fromWeb3(w3)
+    if ns.address(address_or_ens):
+        address = ns.address(address_or_ens)
+        ens = address_or_ens
+    elif is_address(address_or_ens):
```
pipermerriam: Seems like you might want this one to come first in the `if/elif` ordering since it's a simpler check and might be the more common case.

ethpm/backends/registry.py
```diff
@@ -78,11 +82,23 @@ def parse_registry_uri(uri: str) -> RegistryURI:
     """
     Validate and return (authority, pkg name, version) from a valid registry URI
     """
+    from web3.auto.infura import w3
```
pipermerriam: It would be ideal if we could setup a test chain and install the ENS infrastructure on it easily so that we could test this stuff locally without hitting infura.  Maybe that's a project for the Web3.py codebase to support under something like `web3.tools` for doing one-step setup of all of the ENS contracts on a development chain.

PR #1492, opened 2019-11-04
Support type hints in ethpm module

njgheorghita:
### What was wrong?
Type hints in `ethpm` module.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/contract.py
+++ b/ethpm/contract.py
@@ -51,7 +51,7 @@ class LinkableContract(Contract):
                 "Contract cannot be instantiated until its bytecode is linked."
             )
         validate_address(address)
-        super(LinkableContract, self).__init__(address=address, **kwargs)
+        super(LinkableContract, self).__init__(address=address)
 
     @classmethod
     def factory(
--- a/ethpm/contract.py
+++ b/ethpm/contract.py
@@ -51,7 +51,8 @@ class LinkableContract(Contract):
                 "Contract cannot be instantiated until its bytecode is linked."
             )
         validate_address(address)
-        super(LinkableContract, self).__init__(address=address)
+        # type ignored to allow for undefined **kwargs on `Contract` base class __init__
+        super(LinkableContract, self).__init__(address=address, **kwargs)  # type: ignore
 
     @classmethod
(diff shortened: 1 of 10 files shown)


ethpm/contract.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, address: bytes, **kwargs: Any) -> None:
                 "Contract cannot be instantiated until its bytecode is linked."
             )
         validate_address(address)
-        super(LinkableContract, self).__init__(address=address, **kwargs)
```
pipermerriam: Doesn't seem like it would be ok to drop the `**kwargs` here.  That would silently ignore any additional parameters beyond `address` (like `abi`, `bytecode`)?

ethpm/package.py
```diff
@@ -371,7 +374,9 @@ def deployments(self) -> Union["Deployments", Dict[None, None]]:
         return Deployments(deployments, all_contract_instances, self.w3)
 
     @to_dict
-    def _get_all_contract_instances(self, deployments):
+    def _get_all_contract_instances(
+        self, deployments: Dict[str, Any]
```
pipermerriam: Aren't deployments structured?  Can you define a `TypedDict` in place of the `Any`?

ethpm/deployments.py
```diff
@@ -12,6 +18,11 @@
     Contract,
 )
 
+try:
+    from typing import TypedDict  # type: ignore
+except ImportError:
+    from mypy_extensions import TypedDict  # <=3.7
+
```
njgheorghita: @pipermerriam I'm not sure if this is the best way to handle this, since `web3` will be supporting 3.8 soon, where `TypedDict` is automatically included with `typing`? Is it better to just go with the `mypy_extensions` import for now and add `mypy_extensions` to `setup.py['install_requires']`?

pipermerriam: This is generally an acceptable pattern.  If you find yourself copying these import lines across multiple files I'd recommend putting it somewhere like `library._compat` and then importing from there everywhere else.

PR #1495, opened 2019-11-06
Type hints for ENS module

njgheorghita:
### What was wrong?
Type hints needed in `ENS` module.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/ens/constants.py
+++ b/ens/constants.py
@@ -1,10 +1,20 @@
+from typing import (
+    cast,
+)
+
+from eth_typing import (
+    Address,
+)
+from hexbytes import (
+    HexBytes,
+)
 
 ACCEPTABLE_STALE_HOURS = 48
 
 AUCTION_START_GAS_CONSTANT = 25000
 AUCTION_START_GAS_MARGINAL = 39000
 
-EMPTY_SHA3_BYTES = b'\0' * 32
-EMPTY_ADDR_HEX = '0x' + '00' * 20
+EMPTY_SHA3_BYTES = cast(HexBytes, b'\0' * 32)
+EMPTY_ADDR_HEX = cast(Address, '0x' + '00' * 20)
 
 REVERSE_REGISTRAR_DOMAIN = 'addr.reverse'
--- a/ens/constants.py
+++ b/ens/constants.py
@@ -1,5 +1,5 @@
 from eth_typing import (
-    AnyAddress,
+    HexAddress,
 )
 from hexbytes import (
     HexBytes,
@@ -11,6 +11,6 @@ AUCTION_START_GAS_CONSTANT = 25000
 AUCTION_START_GAS_MARGINAL = 39000
 
 EMPTY_SHA3_BYTES = HexBytes(b'\0' * 32)
-EMPTY_ADDR_HEX = AnyAddress('0x' + '00' * 20)
+EMPTY_ADDR_HEX = HexAddress('0x' + '00' * 20)
 
 REVERSE_REGISTRAR_DOMAIN = 'addr.reverse'
--- a/ens/constants.py
+++ b/ens/constants.py
@@ -1,9 +1,5 @@
-from typing import (
-    cast,
-)
-
 from eth_typing import (
-    Address,
+    AnyAddress,
 )
 from hexbytes import (
     HexBytes,
@@ -14,7 +10,7 @@ ACCEPTABLE_STALE_HOURS = 48
 AUCTION_START_GAS_CONSTANT = 25000
 AUCTION_START_GAS_MARGINAL = 39000
 
-EMPTY_SHA3_BYTES = cast(HexBytes, b'\0' * 32)
-EMPTY_ADDR_HEX = cast(Address, '0x' + '00' * 20)
+EMPTY_SHA3_BYTES = HexBytes(b'\0' * 32)
+EMPTY_ADDR_HEX = AnyAddress('0x' + '00' * 20)
 
 REVERSE_REGISTRAR_DOMAIN = 'addr.reverse'
--- a/ens/main.py
+++ b/ens/main.py
@@ -1,8 +1,24 @@
+from typing import (
+    TYPE_CHECKING,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+    cast,
+)
 
+from eth_typing import (
+    Address,
+    ChecksumAddress,
+    Hash32,
+)
 from eth_utils import (
     is_binary_address,
     is_checksum_address,
     to_checksum_address,
+)
+from mypy_extensions import (
+    TypedDict,
 )
 
 from ens import abis
@@ -29,7 +45,28 @@ from ens.utils import (
     raw_name_to_hash,
 )
 
+if TYPE_CHECKING:
+    from web3 import Web3  # noqa: F401
+    from web3.contract import (  # noqa: F401
+        Contract,
+    )
+    from web3.providers import (  # noqa: F401
+        BaseProvider,
+    )
+
+
 ENS_MAINNET_ADDR = '0x314159265dD8dbb310642f98f50C066173C1259b'
+
+
+# relocate / move to eth_typing
+class TxDict(TypedDict):
+    nonce: int
+    gasPrice: int
+    gas: int
+    # _from: Union[ChecksumAddress, str]  # address or ENS name
+    to: Union[ChecksumAddress, str]  # address or ENS name
+    value: int
+    data: Union[bytes, str]
 
 
 class ENS:
@@ -48,7 +85,9 @@ class ENS:
     is_valid_name = staticmethod(is_valid_name)
     reverse_domain = staticmethod(address_to_reverse_domain)
 
-    def __init__(self, provider=default, addr=None):
+    def __init__(
+        self, provider: 'BaseProvider'=cast('BaseProvider', default), addr: Address=None
+    ) -> None:
         """
         :param provider: a single provider used to connect to Ethereum
         :type provider: instance of `web3.providers.base.BaseProvider`
@@ -62,7 +101,7 @@ class ENS:
         self._resolverContract = self.web3.eth.contract(abi=abis.RESOLVER)
 
     @classmethod
-    def fromWeb3(cls, web3, addr=None):
+    def fromWeb3(cls, web3: 'Web3', addr: Address=None) -> 'ENS':
         """
         Generate an ENS instance with web3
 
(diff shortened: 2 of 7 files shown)


ens/main.py
```diff
@@ -29,9 +45,30 @@
...
+
 ENS_MAINNET_ADDR = '0x314159265dD8dbb310642f98f50C066173C1259b'
 
 
+# relocate / move to eth_typing
+class TxDict(TypedDict):
+    nonce: int
+    gasPrice: int
+    gas: int
+    # _from: Union[ChecksumAddress, str]  # address or ENS name
```
njgheorghita: I'm not sure what the best way to handle this is, since `TypedDict`s forbid using reserved keywords (`from`) as keys. Right now, I just `type: ignore` wherever is necessary - (i.e. line 171).

ens/utils.py
```diff
@@ -23,38 +39,46 @@
...
+def ensure_hex(data: HexBytes) -> HexBytes:
     if not isinstance(data, str):
         return Web3().toHex(data)
     return data
 
 
-def init_web3(providers=default):
-    from web3 import Web3
+# this was providers -> provider?
+def init_web3(provider: 'BaseProvider'=cast('BaseProvider', default)) -> 'Web3Type':
```
njgheorghita: I changed the arg `providers` -> a single `provider`. If I understand correctly, initializing a `web3` instance only accepts a single `provider` not some kind of sequence or iterable of providers.

ens/utils.py
```diff
@@ -23,38 +39,46 @@
 default = object()
 
 
-def Web3():
-    from web3 import Web3
-    return Web3
+if TYPE_CHECKING:
+    from web3 import Web3 as Web3Type  # noqa: F401
```
njgheorghita: I'm  not  a fan of using `Web3Type` here, but it was the simplest way I could think of since `Web3` is defined on line 49.

ens/main.py
pipermerriam: I believe there is an alternate syntax for defining `TypedDict` that uses strings as field names.  Does it allow `from` there?

cc @kclowes maybe this is a breaking change we can work on introducing to change `from` to `sender`?

ens/utils.py
pipermerriam: Is the issue just with the name?  You can do `import Web3 as _Web3` maybe?

ens/utils.py
pipermerriam: cc @kclowes

ens/main.py
njgheorghita: Ahh found it, yup works fine now!

ens/constants.py
```diff
@@ -1,10 +1,20 @@
...
+)
 
 ACCEPTABLE_STALE_HOURS = 48
 
 AUCTION_START_GAS_CONSTANT = 25000
 AUCTION_START_GAS_MARGINAL = 39000
 
-EMPTY_SHA3_BYTES = b'\0' * 32
-EMPTY_ADDR_HEX = '0x' + '00' * 20
+EMPTY_SHA3_BYTES = cast(HexBytes, b'\0' * 32)
```
pipermerriam: I think you can just do `HexBytes(b'\x' * 32)`.

ens/constants.py
```diff
@@ -1,10 +1,20 @@
...
+EMPTY_ADDR_HEX = cast(Address, '0x' + '00' * 20)
```
pipermerriam: I don't think this is accurate...

https://github.com/ethereum/eth-typing/blob/e415a1c70d0751283f048d6365165839cdd18540/eth_typing/evm.py

Suggests it should be `HexAddress` or `AnyAddress`

ens/main.py
```diff
@@ -29,9 +44,30 @@
...
+
+
 ENS_MAINNET_ADDR = '0x314159265dD8dbb310642f98f50C066173C1259b'
 
 
+TxDict = TypedDict("TxDict", {
+    "nonce": int,
+    "gasPrice": int,
+    "gas": int,
+    "from": Union[Address, str],
```
pipermerriam: I think all of these need to be `AnyAddress`

ens/main.py
```diff
@@ -111,14 +151,16 @@ def setup_address(self, name, address=default, transact={}):
...
+            transact = cast(TxDict, {})
         owner = self.setup_owner(name, transact=transact)
         self._assert_control(owner, name)
         if is_none_or_zero_address(address):
             address = None
         elif address is default:
             address = owner
         elif is_binary_address(address):
-            address = to_checksum_address(address)
+            address = cast(Address, to_checksum_address(address))
```
pipermerriam: This seems wrong too.  I thought that `eth-utils` returned properly typed values from these.

ens/main.py
```diff
@@ -145,6 +187,9 @@ def setup_name(self, name, address=None, transact={}):
         :raises UnauthorizedError: if ``'from'`` in `transact` does not own `name`
         :raises UnownedName: if no one owns `name`
         """
+        if transact is None:
+            transact = cast(TxDict, {})
```
pipermerriam: This might be the necessary way to go but does `transact: TxDict = {}` work?

ens/main.py
```diff
@@ -210,7 +255,9 @@ def owner(self, name):
         return self.ens.caller.owner(node)
 
     @dict_copy
-    def setup_owner(self, name, new_owner=default, transact={}):
+    def setup_owner(
+        self, name: str, new_owner: Address=cast(Address, default), transact: TxDict=None
```
pipermerriam: We need a better pattern for this where we need to differentiate between `None` and not-provided..... Annoying that it ends up messing up the typing for things.

ens/main.py
```diff
@@ -276,7 +325,16 @@ def _first_owner(self, name):
         return (owner, unowned, name)
 
     @dict_copy
-    def _claim_ownership(self, owner, unowned, owned, old_owner=None, transact={}):
+    def _claim_ownership(
+        self,
+        owner: Address,
+        unowned: Sequence[str],
+        owned: str,
+        old_owner: Optional[Address]=None,
```
pipermerriam: Is the `Optional` necessary here?

ens/utils.py
```diff
@@ -23,38 +39,45 @@
...
+    )
+
+
+def Web3() -> Type['_Web3']:
+    from web3 import Web3 as Web3Main
+    return Web3Main
 
 
-def dict_copy(func):
+def dict_copy(func: Callable[..., Any]) -> Callable[..., Any]:
```
pipermerriam: I think you need to use this pattern to preserve the return type of the function.

```
TFunc = TypeVar("TFunc", bound=Callable[..., Any])
def decorator(func: TFunc) -> TFunc:
    ...
```

ens/utils.py
```diff
@@ -103,27 +126,27 @@ def is_valid_name(name):
...
-def to_utc_datetime(timestamp):
+def to_utc_datetime(timestamp: float) -> Optional[datetime.datetime]:
     if timestamp:
         return datetime.datetime.fromtimestamp(timestamp, datetime.timezone.utc)
     else:
         return None
 
 
-def sha3_text(val):
+def sha3_text(val: Any) -> HexBytes:
```
pipermerriam: Should probably be [`AnyStr`](https://docs.python.org/3/library/typing.html#typing.AnyStr) if you want to include `bytes` types or just `str` if not (but it looks like it supports both)

ens/utils.py
```diff
@@ -155,20 +178,20 @@ def raw_name_to_hash(name):
     return normal_name_to_hash(normalized_name)
 
 
-def address_in(address, addresses):
+def address_in(address: Address, addresses: Sequence[Address]) -> bool:
```
pipermerriam: I'm realizing most places we are currently using `Sequence` can typically use `Collection`.  We only need `Sequence` if we intend to index into data structure which is often not the case.

ens/utils.py
kclowes: Yep, you're correct! We no longer accept multiple providers.

ens/main.py
kclowes: Sounds good to me! Added to the v6 breaking changes issue!

ens/main.py
njgheorghita: @pipermerriam I'm having trouble using `AnyAddress` though I agree it should be the type here. It might be my understanding of how to use a `TypeVar` mypy throws these errors whenever I try to use `AnyAddress`

> ens/main.py:64: error: Type variable "eth_typing.evm.AnyAddress" is unbound
ens/main.py:64: note: (Hint: Use "Generic[AnyAddress]" or "Protocol[AnyAddress]" base class to bind "AnyAddress" inside a class)
ens/main.py:64: note: (Hint: Use "AnyAddress" in function signature to bind "AnyAddress" inside a function)

I've tried adding a bound to the definition in `eth_typing` but mypy doesn't allow defining both values and a bound in a `TypeVar`

Also, mypy has trouble actually comparing a `AnyAddress` to an `Address`, `ChecksumAddress`, `HexAddress` - and complains (i.e. when the type is defined as `AnyAddress`

> ens/main.py:236: error: Argument 1 to "address_to_reverse_domain" has incompatible type "HexAddress"; expected "Address"

njgheorghita:
@pipermerriam ready for a final :eyes: . My bad on losing the commit history, but I had to do a rebase / force push to get around some weird caching error on circle. 

I went with using `ChecksumAddress` > `Address` where appropriate b/c that's what `web3` requires afaiu - and using `AnyAddress` (which is a `TypeVar` not a type) was problematic.

Also pinging @cburgdorf just in case you have some extra time and are dying to review some types :wink:

ens/main.py
njgheorghita: @kclowes Just my two cents - I got things working with `from` as a key, so I'm not sure it requires a breaking change to `sender`. Personally, I prefer `from` b/c of the relationship to the `to` key, but if we want to go ahead and update it to `sender` then maybe we should also update the `to` key to `receiver` or something to keep that relationship

ens/main.py
```diff
@@ -29,8 +46,30 @@
...
+
+
 ENS_MAINNET_ADDR = '0x314159265dD8dbb310642f98f50C066173C1259b'
 
+TxDict = TypedDict("TxDict", {
+    "nonce": int,
+    "gasPrice": int,
+    "gas": int,
+    "from": Union[ChecksumAddress, str],
+    "to": Union[ChecksumAddress, str],
```
pipermerriam: Worth verifying that `str` is accurate (allowed).  What would the value be for `str`?  Also, do we allow raw binary addresses?  If so then `Address` should be part of this too....

newsfragments/1495.misc.rst
```diff
@@ -0,0 +1 @@
+Add type hints to ENS module.
```
pipermerriam: I don't think we need newsfragments for these.  Probably only need a fragment at the point where we enable type hints for 3rd party installers of the web3 library.

ens/main.py
njgheorghita: A `str` could be an ens name - in #1497 I've updated `str` to a `ENS` type to be more specific. Also updated to include `Address` since raw binary addresses are accepted

ens/main.py
kclowes: Sweet, I'm fine with either. I think both have equivalent pros and cons, but I like the idea of keeping the wording relationship with either `from/to` or `sender/receiver`. I'll remove the change from the breaking changes issue, but @pipermerriam if you have a strong opinion otherwise, let me know!

PR #1497, opened 2019-11-11
Initial type hints on web3.contract

njgheorghita:
### What was wrong?
Type hints needed in `web3.contract`

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -3,10 +3,31 @@
 """
 import copy
 import itertools
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Collection,
+    Dict,
+    Generator,
+    Iterable,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 import warnings
 
 from eth_abi.exceptions import (
     DecodingError,
+)
+from eth_typing import (
+    BlockNumber,
+    ChecksumAddress,
+    Hash32,
+    HexStr,
 )
 from eth_utils import (
     add_0x_prefix,
@@ -67,6 +88,7 @@ from web3._utils.events import (
     is_dynamic_sized_type,
 )
 from web3._utils.filters import (
+    LogFilter,
     construct_event_filter_params,
 )
 from web3._utils.function_identifiers import (
@@ -104,6 +126,18 @@ from web3.logs import (
     WARN,
     EventLogErrorFlags,
 )
+from web3.types import (
+    ABI,
+    ABIEvent,
+    ABIFunction,
+    BlockIdentifier,
+    EventData,
+    TxDict,
+    TxReceipt,
+)
+
+if TYPE_CHECKING:
+    from web3 import Web3  # noqa: F401
 
 ACCEPTABLE_EMPTY_STRINGS = ["0x", b"0x", "", b""]
 
@@ -112,7 +146,7 @@ class ContractFunctions:
     """Class containing contract function objects
     """
 
-    def __init__(self, abi, web3, address=None):
+    def __init__(self, abi: ABI, web3: 'Web3', address: ChecksumAddress=None) -> None:
         self.abi = abi
         self.web3 = web3
         self.address = address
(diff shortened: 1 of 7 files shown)


web3/contract.py
```diff
@@ -130,14 +164,15 @@ def __init__(self, abi, web3, address=None):
                         address=self.address,
                         function_identifier=func['name']))
 
-    def __iter__(self):
+    # mypy can't use Iterable here b/c __next__ isn't implemented
+    def __iter__(self) -> Optional[Generator[str, None, None]]:
```
njgheorghita: note: mypy can't use Iterable here b/c __next__ isn't implemented

web3/contract.py
```diff
@@ -637,9 +686,16 @@ class ConciseContract:
...
+    def __init__(
+        self,
+        classic_contract: Contract,
+        method_class: Union[Type['ConciseMethod'], Type['ImplicitMethod']]=ConciseMethod
+    ) -> None:
+        # web3/contract.py:711: error: Incompatible types in assignment (expression has type
+        # "Tuple[Callable[..., Any], Callable[..., Any]]", variable has type
+        # "Tuple[Callable[..., Any]]")
+        # type ignored b/c error inference +=
+        classic_contract._return_data_normalizers += CONCISE_NORMALIZERS  # type: ignore
```
njgheorghita: It seems to me like mypy is having some trouble with the `+=` here: I can't figure out how to resolve the error it throws: `web3/contract.py:711: error: Incompatible types in assignment (expression has type "Tuple[Callable[..., Any], Callable[..., Any]]", variable has type "Tuple[Callable[..., Any]]")` - which is really strange since it's just repeating the same type twice in the first definition.

cburgdorf: When you use a `Tuple` like a collection like e.g. `(1, 2, 3,)` there are multiple types you could use to describe that. 

1. 

```python
Tuple[int, int, int]
```

2.

```python
Tuple[int, ...]
```

What happens in your example is that mypy infers different types for the variable `_return_data_normalizers`  and the expression. If you add a type hint so that mypy knows that it should treat `_return_data_normaliers` as `Tuple[Callable[..., Any], ...]` the error should go away.

njgheorghita: Ahhh, finally got things worked out - thanks @cburgdorf!!

njgheorghita:
pinging @pipermerriam just b/c it says I removed the request to review above, so I'm not sure if you ever actually received one

web3/types.py
```diff
@@ -0,0 +1,122 @@
...
+    "outputs": Optional[Sequence[ABIFunctionParameter]],
+    "stateMutability": Union["pure", "view", "nonpayable", "payable"],
+    "payable": bool,
+    "constant": bool,
+})
+
+
+ABI = Sequence[Union[ABIFunction, ABIEvent]]
+
+BlockIdentifier = Union[str, BlockNumber, bytes, Hash32]
```
pipermerriam: I think `bytes` is out of place since `Hash32` is intended to represent a 32-byte `bytes` value.

web3/contract.py
```diff
@@ -497,11 +538,11 @@ def _encode_constructor_data(cls, args=None, kwargs=None):
 
             deploy_data = to_hex(cls.bytecode)
 
-        return deploy_data
+        return HexStr(deploy_data)
 
 
-def mk_collision_prop(fn_name):
-    def collision_fn():
+def mk_collision_prop(fn_name: str) -> Callable[[], None]:
```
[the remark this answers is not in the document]
njgheorghita: Ahh that's cool, wasn't aware of that type - seems like I should be using that in a lot more places rather than `-> None:`? And using `-> None:` only when `None`  is being explicitly returned?

njgheorghita:
> Can you review the classes like `ContractFunctions` and see what the right generic base class would be? I believe that at minimum it would be `Collection[ABIFunction]` but I think it could be one of the higher order types like `Sequence`.
> 
> https://docs.python.org/3/library/collections.abc.html
> 
> This table is useful for figuring it out.

@pipermerriam Right now, `ContractFunctions` & `ContractEvents` would only be `Iterable` b/c they're missing `__contains__` and `__len__` to become `Collection` and a couple more to be `Sequence`. Were you thinking that you wanted any more of the special methods implemented?

PR #1498, opened 2019-11-11
Fix SyntaxWarning

Fedalto:
### What was wrong?

Python is complaining with a SyntaxWarning when comparing with literals using `is`/`is not`

`is` and `is not` are identity operators, that check if they are the
same objects, which is not what we want when comparing on these cases.

### How was it fixed?

Change to use `==` and `!=` where appropriate

### Todo:

#### Cute Animal Picture

.jpg)

### DIFF
--- a/newsfragments/1498.misc.rst
+++ b/newsfragments/1498.misc.rst
@@ -1,0 +1,1 @@
+Fix Syntax warnings: use `==` instead of `is` where appropriate, and similarly `!=` instead of `is not`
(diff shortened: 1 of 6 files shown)


Fedalto:
The CI is failing with this change, which might mean that it have uncovered a test that should be failing. But I don't have the context to know for sure.

kclowes:
Thanks @Fedalto. It looks like a bug to me too. I'll take a look!

PR #1499, opened 2019-11-13
Add type hints to web3.datastructures

njgheorghita:
### What was wrong?
Type hints in `web3.datastructures`

#### Cute Animal Picture
image

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -3,7 +3,7 @@
 """
 import copy
 import itertools
-from typing import (
+from typing import (  # noqa: F401
     TYPE_CHECKING,
     Any,
     Callable,
@@ -12,6 +12,7 @@ from typing import (
     Generator,
     Iterable,
     List,
+    MutableMapping,
     NoReturn,
     Optional,
     Sequence,
@@ -128,12 +129,13 @@ from web3.logs import (
     WARN,
     EventLogErrorFlags,
 )
-from web3.types import (
+from web3.types import (  # noqa: F401
     ABI,
     ABIEvent,
     ABIFunction,
     BlockIdentifier,
     EventData,
+    LogParams,
     TxParams,
     TxReceipt,
 )
@@ -1083,7 +1085,7 @@ class ContractEvent:
                 if errors == DISCARD:
                     continue
                 elif errors == IGNORE:
-                    new_log = MutableAttributeDict(log)
+                    new_log: LogParams = MutableAttributeDict(log)
                     new_log['errors'] = e
                     rich_log = AttributeDict(cast(Dict[Any, Any], new_log))
                 elif errors == STRICT:
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1085,7 +1085,7 @@ class ContractEvent:
                 elif errors == IGNORE:
                     new_log = MutableAttributeDict(log)
                     new_log['errors'] = e
-                    rich_log = AttributeDict(new_log)
+                    rich_log = AttributeDict(cast(Dict[Any, Any], new_log))
                 elif errors == STRICT:
                     raise e
                 else:
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -3,7 +3,7 @@
 """
 import copy
 import itertools
-from typing import (  # noqa: F401
+from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
@@ -12,7 +12,6 @@ from typing import (  # noqa: F401
     Generator,
     Iterable,
     List,
-    MutableMapping,
     NoReturn,
     Optional,
     Sequence,
@@ -1087,7 +1086,7 @@ class ContractEvent:
                 elif errors == IGNORE:
                     new_log: LogParams = MutableAttributeDict(log)
                     new_log['errors'] = e
-                    rich_log = AttributeDict(cast(Dict[Any, Any], new_log))
+                    rich_log = AttributeDict(new_log)
                 elif errors == STRICT:
                     raise e
                 else:
--- a/web3/datastructures.py
+++ b/web3/datastructures.py
@@ -4,7 +4,7 @@ from collections import (
 from collections.abc import (
     Hashable,
 )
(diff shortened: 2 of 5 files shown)


web3/contract.py
```diff
@@ -1085,7 +1085,7 @@ def _parse_logs(self, txn_receipt: TxReceipt, errors: EventLogErrorFlags) -> Ite
                 elif errors == IGNORE:
                     new_log = MutableAttributeDict(log)
                     new_log['errors'] = e
-                    rich_log = AttributeDict(new_log)
+                    rich_log = AttributeDict(cast(Dict[Any, Any], new_log))
```
pipermerriam: seems like the type here could be tightened since we know roughly what a log looks like...

web3/datastructures.py
```diff
@@ -21,30 +30,34 @@
...
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.__dict__)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return self.__class__.__name__ + "(%r)" % self.__dict__
 
-    def _repr_pretty_(self, builder, cycle):
+    def _repr_pretty_(self, builder: pretty, cycle: bool) -> None:
```
pipermerriam: Not sure I understand the need for this `TypeVar`.  Seems like `Any` would suffice (and probably doesn't matter to us at all since we don't call this API)

web3/datastructures.py
```diff
@@ -54,23 +67,23 @@ def _repr_pretty_(self, builder, cycle):
...
+    def _apply_if_mapping(cls, value: Any) -> Any:
         if isinstance(value, Mapping):
-            return cls(value)
+            return cls(cast(Dict[Any, Any], value))
         else:
             return value
 
     @classmethod
-    def recursive(cls, value):
+    def recursive(cls, value: Any) -> Mapping[Any, Any]:
```
pipermerriam: I suspect that the return type here should be `ReadableAttributeDict`

web3/datastructures.py
```diff
@@ -21,30 +30,34 @@
 # https://docs.python.org/3/reference/datamodel.html#object.__hash__
 
 
-class ReadableAttributeDict(Mapping):
+pretty = TypeVar("pretty", bound=Any)
+
+
+class ReadableAttributeDict(Mapping[Any, Any]):
```
pipermerriam: I think the more correct way to do this would be:

```python
TKey = TypeVar("TKey", bound=Hashable)
TValue = TypeVar("TValue")

class ReadableAttributeDict(Mapping[TKey, TValue]):
    ...
```

This does require that you specify key/value types in most places that you use this but I think that's a good thing.

web3/datastructures.py
```diff
@@ -54,23 +67,23 @@ def _repr_pretty_(self, builder, cycle):
         builder.text(")")
 
     @classmethod
-    def _apply_if_mapping(cls, value):
+    def _apply_if_mapping(cls, value: Any) -> Any:
```
pipermerriam: This might be better typed as:

```python
T = TypeVar("T")
TValue = TypeVar("TValue")

@classmethod
def _apply_if_mapping(cls: Type[T], value: TValue) -> Union[T, TValue]:
    ...
```

web3/datastructures.py
```diff
@@ -79,40 +92,42 @@ class AttributeDict(ReadableAttributeDict, Hashable):
...
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
         if isinstance(other, Mapping):
             return self.__dict__ == dict(other)
         else:
             return False
 
 
-class NamedElementOnion(Mapping):
+class NamedElementOnion(Mapping[Any, Any]):
```
pipermerriam: Same here, probably want to define these as `TypeVar` types so that you can specify more concretely what the types of this mapping are (since it is used to house homogenous data and thus can be strongly typed that way)

njgheorghita:
@pipermerriam :eyes: Not sure if I got it all right, but gave it a shot to make the suggested updates.

web3/contract.py
```diff
@@ -128,12 +129,13 @@
     WARN,
     EventLogErrorFlags,
 )
-from web3.types import (
+from web3.types import (  # noqa: F401
```
pipermerriam: this `noqa` looks suspect (like it probably isn't needed...)

web3/contract.py
pipermerriam: If you can't tighten up these `Any` types, maybe leave a comment indicating why?

web3/datastructures.py
```diff
@@ -54,67 +71,70 @@ def _repr_pretty_(self, builder, cycle):
...
             return value
 
     @classmethod
-    def recursive(cls, value):
+    def recursive(cls, value: TValue) -> 'ReadableAttributeDict[TKey, TValue]':
         return recursive_map(cls._apply_if_mapping, value)
 
 
-class MutableAttributeDict(MutableMapping, ReadableAttributeDict):
+class MutableAttributeDict(MutableMapping[Any, Any], ReadableAttributeDict[TKey, TValue]):
```
pipermerriam: Was there an issue with using `MutableMapping[TKey, TValue` here?

web3/datastructures.py
```diff
@@ -54,67 +71,70 @@ def _repr_pretty_(self, builder, cycle):
...
 
 
-class AttributeDict(ReadableAttributeDict, Hashable):
+class AttributeDict(ReadableAttributeDict[TKey, TValue], Hashable):
     """
     This provides superficial immutability, someone could hack around it
     """
 
-    def __setattr__(self, attr, val):
+    def __setattr__(self, attr: Any, val: Any) -> None:
```
pipermerriam: seems like `attr` should be `str` and `val` should be `TValue`.

web3/contract.py
njgheorghita: It's needed for the `LogParams` type hint on line `1087` - I'm not sure why flake isn't picking up on it

PR #1502, opened 2019-11-13
Add type hints to web3.eth

njgheorghita:
### What was wrong?
Type hints were missing in `web3.eth`

#### Cute Animal Picture
https://i.dailymail.co.uk/i/pix/2016/01/11/20/300DF79400000578-0-image-m-21_1452545856980.jpg


web3/eth.py
```diff
@@ -397,39 +467,39 @@ def filter(self, filter_params=None, filter_id=None):
...
         return self.web3.manager.request_blocking(
             "eth_uninstallFilter", [filter_id],
         )
 
-    def contract(self,
-                 address=None,
-                 **kwargs):
+    def contract(
+        self, address: Union[Address, ChecksumAddress, ENS]=None, **kwargs: Any
+    ) -> Any:
```
njgheorghita: This return type is particularly confusing, AFAIK it should be something like: `Union[Contract, ConciseContract, ContractCaller, Type[Union[Contract, ConciseContract, ContractCaller]]]` - however that breaks about a million things throughout the entire codebase, as does pretty much everything else I've tried. Also, b/c `ContractCaller` doesn't implement a factory, should it be allowed in here / `setContractFactory` below?

pipermerriam: You might be able to make this work with `typing.overload`.  Define one where `address`: None` and have that return the `Type[Contract]`, and another that has `address: Union[Address, ...]` and have it return `Contract`

Alternate approach would be to just *force* the type to be `Contract` and pretend the other types don't exist via either `cast` or `# type: ignore`.

ens/main.py
```diff
@@ -52,11 +52,11 @@
         BaseProvider,
     )
     from web3.types import (  # noqa: F401
-        TxDict,
+        TxParams,
     )
 
 
-ENS_MAINNET_ADDR = '0x314159265dD8dbb310642f98f50C066173C1259b'
+ENS_MAINNET_ADDR = cast(ChecksumAddress, '0x314159265dD8dbb310642f98f50C066173C1259b')
```
pipermerriam: I think the more correct way to do this is `ChecksumAddress('0x31415...')`

web3/eth.py
```diff
@@ -53,91 +75,121 @@
...
     defaultAccount = empty
     defaultBlock = "latest"
-    defaultContractFactory = Contract
+    defaultContractFactory: Type[Union[Contract, ConciseContract, ContractCaller]] = Contract  # noqa: E704,E501
     iban = Iban
     gasPriceStrategy = None
 
-    def namereg(self):
+    def namereg(self) -> NoReturn:
         raise NotImplementedError()
```
pipermerriam: @kclowes these APIs can be removed.  They won't ever be implemented and they aren't needed.  Just carryovers from a very long time ago.

web3/eth.py
```diff
@@ -53,91 +75,121 @@
...
         return self.web3.manager.request_blocking("eth_hashrate", [])
 
     @property
-    def gasPrice(self):
+    def gasPrice(self) -> Wei:
         return self.web3.manager.request_blocking("eth_gasPrice", [])
 
     @property
-    def accounts(self):
+    def accounts(self) -> Collection[ChecksumAddress]:
```
pipermerriam: return type should be concrete: `Tuple[...]`

web3/eth.py
```diff
@@ -53,91 +75,121 @@
...
             [account, block_identifier],
         )
 
-    def getStorageAt(self, account, position, block_identifier=None):
+    def getStorageAt(
+        self,
+        account: Union[Address, ChecksumAddress, ENS],
+        position: int,
+        block_identifier: BlockIdentifier=None
+    ) -> Hash32:
```
pipermerriam: I think the return here should be a `bytes` type

web3/eth.py
```diff
@@ -274,26 +332,30 @@ def getTransactionReceipt(self, transaction_hash):
...
         )
 
-    def replaceTransaction(self, transaction_hash, new_transaction):
+    def replaceTransaction(self, transaction_hash: Hash32, new_transaction: TxParams) -> Hash32:
         current_transaction = get_required_transaction(self.web3, transaction_hash)
         return replace_transaction(self.web3, current_transaction, new_transaction)
 
-    def modifyTransaction(self, transaction_hash, **transaction_params):
+    # todo: Update Any to stricter kwarg checking with TxParams
+    # https://github.com/python/mypy/issues/4441
```
pipermerriam: :smile: can't wait till they implement this.

PR #1503, opened 2019-11-14
Add type hints to web3.manager

njgheorghita:
### What was wrong?
Type hints missing in `web3.manager`

#### Cute Animal Picture
image

### DIFF
--- a/web3/manager.py
+++ b/web3/manager.py
@@ -1,7 +1,8 @@
 import logging
-from typing import (
+from typing import (  # noqa: F401
     TYPE_CHECKING,
     Any,
+    Callable,
     Collection,
     Dict,
     List,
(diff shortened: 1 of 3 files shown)


web3/manager.py
```diff
@@ -26,13 +37,20 @@
...
 
+if TYPE_CHECKING:
+    from web3 import Web3  # noqa: F401
 
-def apply_error_formatters(
-        error_formatters,
-        response):
 
+def apply_error_formatters(
+    error_formatters: Dict[str, Any], response: JsonRpcResponse
```
pipermerriam: type looks wrong for `error_formatters`.  I'd expect it to be a callable since it's being used in `pipe`.

web3/types.py
```diff
@@ -81,6 +82,9 @@
     "blockNumber": int,
 })
 
+JsonRpcResponse = Dict[str, Any]
+
+Middleware = Callable[[Any, Any], Any]
```
pipermerriam: I think that these `Any` types can be tightened some, at least for the arguments to the middleware callables.

PR #1504, opened 2019-11-14
Add type hints to web3.iban

njgheorghita:
### What was wrong?
Type hints missing from `web3.exceptions`, `web3.geth`, `web3.iban`

Wait for review until #1502 is merged

#### Cute Animal Picture
image

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -126,7 +126,7 @@ class ENS:
         self,
         name: str,
         address: Union[Address, ChecksumAddress, HexAddress]=cast(ChecksumAddress, default),
-        transact: 'TxParams'={}
+        transact: "TxParams"={}
     ) -> Hash32:
         """
         Set up the name to point to the supplied address.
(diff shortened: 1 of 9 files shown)


ens/main.py
```diff
@@ -52,7 +52,7 @@
         BaseProvider,
     )
     from web3.types import (  # noqa: F401
```
pipermerriam: I think it would be better to have `web3.types` use the `if TYPECHECKING` pattern and allow direct imports from this module...

web3/contract.py
```diff
@@ -1350,7 +1350,7 @@ def __getattr__(self, function_name: str) -> ContractFunction:
             return super().__getattribute__(function_name)
 
     def __call__(
-        self, transaction: TxDict=None, block_identifier: BlockIdentifier='latest'
+        self, transaction: TxParams=None, block_identifier: BlockIdentifier='latest'
```
pipermerriam: I just saw that python 3.8 introduces [`typing.Literal`](https://docs.python.org/3/library/typing.html#typing.Literal) which will be a nice improvement to the `BlockIdentifier` and other *magic* string based values.  Seems like the kind of thing that we may get a backport available for as well.

ens/main.py
njgheorghita: Agreed, but it seems to me like it might be necessary in the `ens` module, since the docs build breaks when trying to import `ENS` and I haven't been able to diagnose why. In `web3` we can directly import from `web3.types`

PR #1505, opened 2019-11-14
Add type hints to web3.module

njgheorghita:
### What was wrong?
Type hints needed in `web3.module`

#### Cute Animal Picture

image

### DIFF
--- a/web3/module.py
+++ b/web3/module.py
@@ -3,7 +3,6 @@ from typing import (
     Any,
     Callable,
     Coroutine,
-    Dict,
     Union,
 )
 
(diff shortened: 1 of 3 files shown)


web3/module.py
```diff
@@ -1,11 +1,32 @@
...
+)
+
+if TYPE_CHECKING:
+    from web3 import Web3  # noqa: F401
+
 
 @curry
-def apply_result_formatters(result_formatters, result):
+def apply_result_formatters(
+    result_formatters: Dict[str, Any], result: JsonRpcResponse
```
pipermerriam: The type for `result_formatters` looks wrong here.  `pipe` expects a value as the first argument and then all the rest are expected to be callables...

web3/types.py
```diff
@@ -81,6 +81,7 @@
     "blockNumber": int,
 })
 
+JsonRpcResponse = Dict[str, Any]
```
pipermerriam: Can we do better with a `TypedDict` since we generally know what the fields are for valid `JSON-RPC` responses?

PR #1506, opened 2019-11-14
Add type hints to web3.parity

njgheorghita:
### What was wrong?
Types were missing from `web3.net` and `web3.parity`

#### Cute Animal Picture
image

### DIFF
--- a/web3/parity.py
+++ b/web3/parity.py
@@ -1,6 +1,7 @@
 from typing import (
     Any,
     Dict,
+    List,
     NewType,
     Sequence,
     Union,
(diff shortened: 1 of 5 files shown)


web3/parity.py
```diff
@@ -77,66 +123,81 @@ class Parity(Module):
...
         )
 
-    def listStorageKeys(self, address, quantity, hash_, block_identifier=None):
+    def listStorageKeys(
+        self,
+        address: Union[Address, ChecksumAddress, ENS],
+        quantity: int,
+        hash_: Hash32,
+        block_identifier: BlockIdentifier=None,
+    ) -> Sequence[Hash32]:
```
pipermerriam: Return types should be concrete so likely `Tuple` or `List`

web3/parity.py
```diff
@@ -77,66 +123,81 @@ class Parity(Module):
...
             [transaction_hash]
         )
 
-    def traceCall(self, transaction, mode=['trace'], block_identifier=None):
+    def traceCall(
+        self,
+        transaction: TxParams,
+        mode: TraceMode=['trace'],
+        block_identifier: BlockIdentifier=None
+    ) -> Sequence[BlockTrace]:
```
pipermerriam: Same here with the return type.

PR #1507, opened 2019-11-15
Type hint web3.pm

njgheorghita:
### What was wrong?
Type hints were needed in `web3.pm`

#### Cute Animal Picture
image

### DIFF
--- a/web3/pm.py
+++ b/web3/pm.py
@@ -2,6 +2,9 @@ from abc import (
     ABC,
     abstractmethod,
 )
+from collections import (
+    namedtuple,
+)
 import json
 from pathlib import (
     Path,
@@ -76,6 +79,7 @@ from web3.types import (
 # >>> w3.pm
 # <web3.pm.PM at 0x....>
 
+ReleaseData = namedtuple("ReleaseData", ["package_name", "version", "manifest_uri"])
 T = TypeVar("T")
 
 
@@ -169,7 +173,7 @@ class ERC1319Registry(ABC):
         pass
 
     @abstractmethod
-    def _get_release_data(self, release_id: bytes) -> Iterable[str]:
+    def _get_release_data(self, release_id: bytes) -> ReleaseData:
         """
         Returns a tuple containing (package_name, version, manifest_uri) for the given release id,
         if the release exists on the connected registry.
@@ -282,11 +286,9 @@ class SimpleRegistry(ERC1319Registry):
             yield from reversed(new_ids)
             pointer = new_pointer
 
-    @to_tuple
-    def _get_release_data(self, release_id: bytes) -> Iterable[str]:
-        release_data = self.registry.functions.getReleaseData(release_id).call()
-        for data in release_data:
-            yield data
+    def _get_release_data(self, release_id: bytes) -> ReleaseData:
+        name, version, uri = self.registry.functions.getReleaseData(release_id).call()
+        return ReleaseData(name, version, uri)
 
     def _generate_release_id(self, package_name: str, version: str) -> bytes:
         return self.registry.functions.generateReleaseId(package_name, version).call()
@@ -487,10 +489,10 @@ class PM(Module):
         self._validate_set_registry()
         release_ids = self.registry._get_all_release_ids(package_name)
         for release_id in release_ids:
-            _, version, manifest_uri = self.registry._get_release_data(release_id)
-            yield (version, manifest_uri)
+            release_data = self.registry._get_release_data(release_id)
+            yield (release_data.version, release_data.manifest_uri)
 
-    def get_release_id_data(self, release_id: bytes) -> Tuple[str, ...]:
+    def get_release_id_data(self, release_id: bytes) -> ReleaseData:
         """
         Returns ``(package_name, version, manifest_uri)`` associated with the given
         release id, *if* it is available on the current registry.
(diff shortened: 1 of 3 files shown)


web3/pm.py
```diff
@@ -151,7 +154,7 @@ def _get_release_id(self, package_name: str, version: str) -> bytes:
         pass
 
     @abstractmethod
-    def _get_all_release_ids(self, package_name: str) -> Tuple[bytes]:
+    def _get_all_release_ids(self, package_name: str) -> Iterable[Tuple[bytes]]:
```
pipermerriam: This type  hint means `((b'arst',), (b'tsra'))`, i.e. an iterable of length-1 tuples with a bytes value.  Is that what you intended?

web3/pm.py
```diff
@@ -204,7 +207,7 @@ def _num_release_ids(self, package_name: str) -> int:
         pass
 
     @abstractmethod
-    def deploy_new_instance(cls, w3):
+    def deploy_new_instance(cls, w3: Web3) -> 'ERC1319Registry':
```
pipermerriam: This is *ok* but there are two models that I think improve this:

In the case where you want to stick to a well defined API:

```python
class ThingAPI(ABC):
    ...

class Thing(ThingAPI):
    @classmethod
    def make_thing(cls) -> ThingAPI:
        ...
```

This pattern helps enforce the defined API and is only the right solution in some cases.

In all of the other cases:

```python
T = TypeVar("T")

class Thing:
    @classmethod
    def make_thing(cls: Type[T]) -> T:
        ...
```

This ensures that in the case of subclassing, the return type isn't locked into the base class.

web3/pm.py
```diff
@@ -132,7 +135,7 @@ def _get_package_name(self, package_id: bytes) -> str:
         pass
 
     @abstractmethod
-    def _get_all_package_ids(self) -> Tuple[bytes]:
+    def _get_all_package_ids(self) -> Iterable[bytes]:
```
pipermerriam: I'm confused by these being `Iterable[...]` rather than `Tuple[...]`.  I'm pretty sure they should be `Tuple` types.  Did `mypy` yell at you?

web3/pm.py
```diff
@@ -479,7 +485,7 @@ def get_all_package_releases(self, package_name: str) -> Iterable[Tuple[str, str
             _, version, manifest_uri = self.registry._get_release_data(release_id)
             yield (version, manifest_uri)
 
-    def get_release_id_data(self, release_id: bytes) -> Tuple[str, str, str]:
+    def get_release_id_data(self, release_id: bytes) -> Tuple[str, ...]:
```
pipermerriam: This return type doesn't match the docstring (or my understanding of what this API returns).  Why isn't it `Tuple[str, str, str]` anymore?

web3/pm.py
njgheorghita: >  Did mypy yell at you?

Yup, b/c in the [implementation](https://github.com/ethereum/web3.py/pull/1507/files#diff-b76a8371a5c2d30f739ddcd21e7a6b95R249) we use the `@to_tuple` decorator, which `mypy` recognizes as an `Iterable` b/c of the `yield` pattern rather than a `Tuple`

web3/pm.py
njgheorghita: This is related to [`_get_release_data() -> Iterable[str]`](https://github.com/ethereum/web3.py/pull/1507/files#diff-b76a8371a5c2d30f739ddcd21e7a6b95R168) b/c of using the `@to_tuple` decorator.

cburgdorf: If this is a fixed size Tuple then it would be better to not use `@to_tuple` but rather return the tuple manually (`return (x, y, z)`). The `to_tuple` decorator will always just convert a `Iterable[Something]` into a `Tuple[Something, ...]` but that doesn't convey well the intention of the API if the value is always of the same size. 

In fact, you *could* even consider creating a `NamedTuple` such as `ReleaseData` to get a stronger API. That would also eliminate the need for placeholder variables when you are only interested in one of the three values.

E.g. instead of.

```python
_, __, what_i_need = _get_release_data()
print(what_i_need)
```

You could just write

```python

release_data = _get_release_data()
print(release_data.what_i_need)
```

njgheorghita: Ahh, that's a good point. Made those updates, danke @cburgdorf !!

web3/pm.py
```diff
@@ -71,6 +79,9 @@
 # >>> w3.pm
 # <web3.pm.PM at 0x....>
 
+ReleaseData = namedtuple("ReleaseData", ["package_name", "version", "manifest_uri"])
```
cburgdorf: Unless you have to support Python 3.5 there's a cleaner syntax for `NamedTuple`.

```python
class ReleaseData(NamedTuple):
    package_name: str
    version: str
    manifest_uri: str
```

web3/pm.py
cburgdorf: Gerne :)

web3/pm.py
njgheorghita: Nice, thanks for the tip. I vaguely remember trying to use this syntax before in `3.6` and it resulted in some obscure error that wasn't fixed until `3.7` - maybe had something to do with [setting defaults](https://github.com/python/mypy/issues/4788) :shrug: but everything seems to be working fine here.

PR #1518, opened 2019-11-20
Add type hints to web3.providers.ethtester

njgheorghita:
### What was wrong?
Type hints missing from `web3.providers.ethtester`

#### Cute Animal Picture
image

### DIFF
--- a/ens/utils.py
+++ b/ens/utils.py
@@ -17,6 +17,7 @@ from eth_typing import (
     Address,
     ChecksumAddress,
     HexAddress,
+    HexStr,
 )
 from eth_utils import (
     is_same_address,
@@ -190,7 +191,7 @@ def address_in(address: ChecksumAddress, addresses: Collection[ChecksumAddress])
 
 
 def address_to_reverse_domain(address: ChecksumAddress) -> str:
-    lower_unprefixed_address = remove_0x_prefix(to_normalized_address(address))
+    lower_unprefixed_address = remove_0x_prefix(HexStr(to_normalized_address(address)))
     return lower_unprefixed_address + '.' + REVERSE_REGISTRAR_DOMAIN
 
 
--- a/web3/providers/eth_tester/defaults.py
+++ b/web3/providers/eth_tester/defaults.py
@@ -9,6 +9,7 @@ from typing import (
     NoReturn,
     Tuple,
     Type,
+    TypeVar,
 )
 
 from eth_tester.exceptions import (
@@ -47,6 +48,10 @@ if TYPE_CHECKING:
         EthereumTester,
     )
 
+TReturn = TypeVar("TReturn")
+TParams = TypeVar("TParams")
+TValue = TypeVar("TValue")
+
 
 def not_implemented(*args: Any, **kwargs: Any) -> NoReturn:
     raise NotImplementedError("RPC method not implemented")
@@ -61,18 +66,22 @@ def call_eth_tester(
     return getattr(eth_tester, fn_name)(*fn_args, **fn_kwargs)
 
 
-def without_eth_tester(fn: Callable[..., Any]) -> Callable[..., RPCResponse]:
+def without_eth_tester(
+    fn: Callable[[TParams], TReturn]
+) -> Callable[["EthereumTester", TParams], TReturn]:
     # workaround for: https://github.com/pytoolz/cytoolz/issues/103
     # @functools.wraps(fn)
-    def inner(eth_tester: "EthereumTester", params: Any) -> Callable[..., RPCResponse]:
+    def inner(eth_tester: "EthereumTester", params: TParams) -> TReturn:
         return fn(params)
     return inner
 
 
-def without_params(fn: Callable[..., Any]) -> Callable[..., RPCResponse]:
+def without_params(
+    fn: Callable[[TParams], TReturn]
+) -> Callable[["EthereumTester", TParams], TReturn]:
     # workaround for: https://github.com/pytoolz/cytoolz/issues/103
     # @functools.wraps(fn)
-    def inner(eth_tester: "EthereumTester", params: Any) -> Callable[..., RPCResponse]:
+    def inner(eth_tester: "EthereumTester", params: Any) -> TReturn:
         return fn(eth_tester)
     return inner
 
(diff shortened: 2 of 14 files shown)


ens/utils.py
```diff
@@ -190,7 +191,7 @@ def address_in(address: ChecksumAddress, addresses: Collection[ChecksumAddress])
 
 
 def address_to_reverse_domain(address: ChecksumAddress) -> str:
-    lower_unprefixed_address = remove_0x_prefix(to_normalized_address(address))
+    lower_unprefixed_address = remove_0x_prefix(HexStr(to_normalized_address(address)))
```
pipermerriam: Not for this PR but it would be ideal if we got our typing right so that we didn't need to wrap this in `HexStr`.

web3/providers/eth_tester/defaults.py
```diff
@@ -23,46 +36,61 @@
...
+def call_eth_tester(
+    fn_name: str, eth_tester: "EthereumTester", fn_args: Any, fn_kwargs: Any=None
+) -> RPCResponse:
     if fn_kwargs is None:
         fn_kwargs = {}
     return getattr(eth_tester, fn_name)(*fn_args, **fn_kwargs)
 
 
-def without_eth_tester(fn):
+def without_eth_tester(fn: Callable[..., Any]) -> Callable[..., RPCResponse]:
```
pipermerriam: I think this would be more accurately typed as:

```python
TReturn = TypeVar("TReturn")
TParams = TypeVar("TParams")

def without_eth_tester(fn: Callable[[TParams], TReturn]) -> Callable[[EThereumTester, TParams], TReturn]:
    ...
```

web3/providers/eth_tester/defaults.py
```diff
@@ -23,46 +36,61 @@
...
     # workaround for: https://github.com/pytoolz/cytoolz/issues/103
     # @functools.wraps(fn)
-    def inner(eth_tester, params):
+    def inner(eth_tester: "EthereumTester", params: Any) -> Callable[..., RPCResponse]:
         return fn(params)
     return inner
 
 
-def without_params(fn):
+def without_params(fn: Callable[..., Any]) -> Callable[..., RPCResponse]:
```
pipermerriam: Same here.  Typing can be more precicesly defined using `TypeVar` types.

web3/providers/eth_tester/defaults.py
```diff
@@ -23,46 +36,61 @@
...
-def preprocess_params(eth_tester, params, preprocessor_fn):
+def preprocess_params(
+    eth_tester: "EthereumTester", params: Any, preprocessor_fn: Callable[..., Any]
+) -> Tuple["EthereumTester", Callable[..., Any]]:
     return eth_tester, preprocessor_fn(params)
 
 
-def static_return(value):
-    def inner(*args, **kwargs):
+def static_return(value: Any) -> Callable[..., Any]:
```
pipermerriam: This one should be doable as `def static_return(value: TValue) -> Callable[..., TValue]:`

web3/providers/eth_tester/defaults.py
```diff
@@ -73,7 +101,7 @@ def client_version(eth_tester, params):
 
 
 @curry
-def null_if_excepts(exc_type, fn):
+def null_if_excepts(exc_type: Type[BaseException], fn: Callable[..., Any]) -> Callable[..., Any]:
```
pipermerriam: Same here, types can be tightened using `TypeVar` for callable return type.

web3/providers/eth_tester/defaults.py
```diff
@@ -98,26 +128,28 @@ def get_transaction_by_block_hash_and_index(eth_tester, params):
...
 
-def create_log_filter(eth_tester, params):
+def create_log_filter(eth_tester: "EthereumTester", params: Any) -> int:
     filter_params = params[0]
     filter_id = eth_tester.create_log_filter(**filter_params)
     return filter_id
 
 
-def get_logs(eth_tester, params):
+def get_logs(eth_tester: "EthereumTester", params: Any) -> List[LogReceipt]:
```
pipermerriam: For each of these that is a defined API we should know what the `params` value looks like and thus be able to use stricture types than `Any`.

web3/providers/eth_tester/defaults.py
```diff
@@ -129,11 +161,11 @@ def _generate_random_private_key():
 
 
 @without_params
-def create_new_account(eth_tester):
+def create_new_account(eth_tester: "EthereumTester") -> HexAddress:
     return eth_tester.add_account(_generate_random_private_key())
 
 
-def personal_send_transaction(eth_tester, params):
+def personal_send_transaction(eth_tester: "EthereumTester", params: Any) -> HexStr:
```
pipermerriam: params here should be someting like `Optional[TxParams]` I think.

web3/providers/eth_tester/main.py
```diff
@@ -1,23 +1,45 @@
...
     for testing the async api execution.
     """
-    def __init__(self):
+    def __init__(self) -> None:
         self.eth_tester = EthereumTesterProvider()
 
-    async def make_request(self, method, params):
+    async def make_request(
+        self, method: RPCEndpoint, params: Any
+    ) -> Coroutine[Any, Any, RPCResponse]:
```
pipermerriam: I'm confused by this returning a `Coroutine` type.  I would have expected just `RPCResponse`...

web3/providers/eth_tester/main.py
```diff
@@ -55,7 +81,7 @@ def __init__(self, ethereum_tester=None, api_endpoints=None):
         else:
             self.api_endpoints = api_endpoints
 
-    def make_request(self, method, params):
+    def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
```
pipermerriam: At minimum do we know that params is a `List[Any]`?  I'm not 100% sure but I think that we have some knowledge that params is a list of values.

njgheorghita: > ``params``  This will be a list or other iterable of the parameters for the JSON-RPC method being called.

[Yup!](https://github.com/ethereum/web3.py/blob/0617444b243d878708997009869fe5fd67a57041/docs/internals.rst#writing-your-own-provider), or at minimum an `Iterable`?

njgheorghita: Making this update touches a lot of files, so to avoid from bloating this pr, i'll follow it up with another that tightens this up

web3/providers/eth_tester/defaults.py
njgheorghita: 👍 Will take care of tightening up the params value in a different pr to avoid bloat.

PR #1520, opened 2019-11-21
Type hints web3.providers

njgheorghita:
### What was wrong?
Finish type hinting `web3.providers`

#### Cute Animal Picture
image

### DIFF
--- a/ens/constants.py
+++ b/ens/constants.py
@@ -1,5 +1,6 @@
 from eth_typing import (
     HexAddress,
+    HexStr,
 )
 from hexbytes import (
     HexBytes,
@@ -11,6 +12,6 @@ AUCTION_START_GAS_CONSTANT = 25000
 AUCTION_START_GAS_MARGINAL = 39000
 
 EMPTY_SHA3_BYTES = HexBytes(b'\0' * 32)
-EMPTY_ADDR_HEX = HexAddress('0x' + '00' * 20)
+EMPTY_ADDR_HEX = HexAddress(HexStr('0x' + '00' * 20))
 
 REVERSE_REGISTRAR_DOMAIN = 'addr.reverse'
--- a/ens/main.py
+++ b/ens/main.py
@@ -12,6 +12,7 @@ from eth_typing import (
     ChecksumAddress,
     Hash32,
     HexAddress,
+    HexStr,
 )
 from eth_utils import (
     is_binary_address,
@@ -56,7 +57,9 @@ if TYPE_CHECKING:
     )
 
 
-ENS_MAINNET_ADDR = ChecksumAddress(HexAddress('0x314159265dD8dbb310642f98f50C066173C1259b'))
+ENS_MAINNET_ADDR = ChecksumAddress(
+    HexAddress(HexStr('0x314159265dD8dbb310642f98f50C066173C1259b'))
+)
 
 
 class ENS:
--- a/web3/providers/auto.py
+++ b/web3/providers/auto.py
@@ -1,6 +1,20 @@
 import os
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 from urllib.parse import (
     urlparse,
+)
+
+from eth_typing import (
+    URI,
 )
 
 from web3.exceptions import (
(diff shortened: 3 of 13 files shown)


ens/constants.py
```diff
@@ -11,6 +12,6 @@
 AUCTION_START_GAS_MARGINAL = 39000
 
 EMPTY_SHA3_BYTES = HexBytes(b'\0' * 32)
-EMPTY_ADDR_HEX = HexAddress('0x' + '00' * 20)
+EMPTY_ADDR_HEX = HexAddress(HexStr('0x' + '00' * 20))
```
pipermerriam: Wouldn't this be `HexAddress`?

ens/main.py
```diff
@@ -56,7 +57,9 @@
     )
 
 
-ENS_MAINNET_ADDR = ChecksumAddress(HexAddress('0x314159265dD8dbb310642f98f50C066173C1259b'))
+ENS_MAINNET_ADDR = ChecksumAddress(
+    HexAddress(HexStr('0x314159265dD8dbb310642f98f50C066173C1259b'))
```
pipermerriam: Oh, that's awkward..... that we have to do the double wrap.

web3/providers/auto.py
```diff
@@ -12,20 +26,26 @@
...
+    RPCResponse,
+)
 
 HTTP_SCHEMES = {'http', 'https'}
 WS_SCHEMES = {'ws', 'wss'}
 
 
-def load_provider_from_environment():
-    uri_string = os.environ.get('WEB3_PROVIDER_URI', '')
+def load_provider_from_environment() -> BaseProvider:
```
pipermerriam: Not necessarily for this PR but it might be good to create a `web3.abc` module with a `ProviderAPI` at some point soon.  This will be especially valuable for other libraries that need to do type hinting.

web3/providers/base.py
```diff
@@ -72,9 +97,9 @@ def encode_rpc_request(self, method, params):
         encoded = FriendlyJsonSerde().json_encode(rpc_dict)
         return to_bytes(text=encoded)
 
-    def isConnected(self):
+    def isConnected(self) -> bool:
         try:
-            response = self.make_request('web3_clientVersion', [])
+            response = self.make_request(RPCEndpoint('web3_clientVersion'), [])
```
pipermerriam: Probably nice to do something like have a module that has all of these pre-wrapped for re-use.

web3/providers/ipc.py
```diff
@@ -18,10 +26,12 @@
 try:
     from json import JSONDecodeError
 except ImportError:
-    JSONDecodeError = ValueError
+    # type ignored b/c mypy unhappy w/ conflicting exception types
+    # difficult to cast / resolve w/ conditional import
+    JSONDecodeError = ValueError  # type: ignore
```
pipermerriam: Worth checking if this is still necessary.  Docs say it was introduced in 3.5 which suggests we may not need this anymore.

web3/providers/ipc.py
```diff
@@ -37,18 +47,18 @@ def get_ipc_socket(ipc_path, timeout=0.1):
...
+    def __enter__(self) -> socket.socket:
         if not self.ipc_path:
             raise FileNotFoundError("cannot connect to IPC socket at path: %r" % self.ipc_path)
 
         if not self.sock:
             self.sock = self._open()
         return self.sock
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, exc_type: Type[BaseException], exc_value: Any, traceback: Any) -> None:
```
pipermerriam: `exc_value` should be `BaseException`.  `traceback` should be `types.TracebackType`

web3/providers/rpc.py
```diff
@@ -17,48 +26,54 @@
...
         self._request_kwargs = request_kwargs or {}
         super().__init__()
 
-    def __str__(self):
+    def __str__(self) -> str:
         return "RPC connection {0}".format(self.endpoint_uri)
 
     @to_dict
-    def get_request_kwargs(self):
+    def get_request_kwargs(self) -> Iterable[Tuple[Any, Any]]:
```
pipermerriam: should probably be `Iterable[Tuple[str, Any]]`

web3/providers/websocket.py
```diff
@@ -14,44 +21,50 @@
...
-    async def __aenter__(self):
+    async def __aenter__(self) -> websockets.WebSocketClientProtocol:
         if self.ws is None:
             self.ws = await websockets.connect(
                 uri=self.endpoint_uri, loop=self.loop, **self.websocket_kwargs
             )
         return self.ws
 
-    async def __aexit__(self, exc_type, exc_val, exc_tb):
+    async def __aexit__(self, exc_type: Type[BaseException], exc_val: Any, exc_tb: Any) -> None:
```
pipermerriam: same here for `exc_value` and `exc_tb`

ens/main.py
njgheorghita: Yeah, I can't say I'm a huge fan. It'll probably be somewhat confusing/frustrating to devs as it breaks their mypys - at least initially, until they get used to the convention. But, I'm not so sure the awkwardness is worth relaxing the `HexAddress` back to `str`?

PR #1521, opened 2019-11-21
Upgrade integration tests to run against geth 1.9

kclowes:
### What was wrong?
Integration tests didn't cover the new-ish Geth 1.9 releases. 

### How was it fixed?
Added a fixture for Geth 1.9 and updated tests accordingly.

### Todo:

#### Cute Animal Picture
image


tests/integration/go_ethereum/common.py
```diff
@@ -10,45 +10,47 @@
     Web3ModuleTest,
 )
 
+OLDER_GETH_VERSIONS = ['v1.7.2', 'v1.8.22']
+NEWER_GETH_VERSIONS = ['v1.8.22', 'v1.9.7']
```
pipermerriam: Having `v1.8.22` in both of these is a little awkward.  I see why once I look through the code below but maybe this would be easier to grok if it was done as:

```python
VERSIONS_SUPPORTING_XXX = ...
VERSIONS_WITHOUT_YYY = ...
```

tests/integration/go_ethereum/conftest.py
```diff
@@ -52,14 +53,31 @@ def absolute_datadir(directory_name):
...
+def geth_zipfile_version(get_geth_version):
+    if get_geth_version.major == 1:
+        if get_geth_version.minor == 7:
             return GETH_17_ZIP
-        elif version.minor == 8:
+        elif get_geth_version.minor == 8:
             return GETH_1822_ZIP
+        elif get_geth_version.minor == 9:
+            return GETH_197_ZIP
     assert False, "Unsupported geth version"
```
pipermerriam: FYI: newer version of `mypy` will complain about this.  Changing to `raise AssertionError("...")` is the current canonical way to do this.

tests/integration/go_ethereum/test_goethereum_http.py
```diff
@@ -30,18 +30,36 @@ def endpoint_uri(rpc_port):
...
+                (
+                    '--rpc',
+                    '--rpcport', rpc_port,
+                    '--rpcapi', 'admin,db,eth,net,web3,personal,shh,miner',
+                    '--ipcdisable',
+                )
+            )
+        else:
+            assert False, "Unsupported geth version"
+    else:
```
pipermerriam: I think you can cleanup this function by using a `yield` based approach.

```python
def get_command_args(...):
    yield from base_geth_command_arguments
    if version.major == 1:
        yield from (
            '--rpc',
            '--rpcport', rpc_port,
            '--rpcapi', 'admin,db,eth,net,web3,personal,shh,miner',
            '--ipcdisable',
        )
        if version.minor == 9:
            yield "--allow-insecure-unlock"
```

The code above doesn't contain the "unsupported geth version" checks but should be demonstrative of the concept.

tests/integration/go_ethereum/test_goethereum_ipc.py
```diff
@@ -21,17 +21,24 @@
...
     geth_port = get_open_port()
-    return (
-        geth_binary,
-        '--datadir', str(datadir),
-        '--ipcpath', geth_ipc_path,
-        '--shh',
-        '--nodiscover',
-        '--fakepow',
+    yield from base_geth_command_arguments
+    yield from (
```
kclowes: @pipermerriam is there a better way to clean this up instead of using 2 `yield from`s in a row?

pipermerriam: No, but the two `yield from` statemeents are what I would do.

Issue #1522, opened 2019-11-22
Update docs to reflect methods that raise errors in v5 instead of returning None


kclowes:
* Version: 5.x

### What was wrong?
The docs weren't updated when we started returning a `TransactionNotFound` or `BlockNotFound` error instead of `None` for RPC methods that don't find a transaction. 

### How can it be fixed?
Docs need to be checked and updated for the following methods: 
https://web3py.readthedocs.io/en/stable/v5_migration.html#json-rpc-updates

ashish10677:
Hi. I want to try to contribute to this issue. Can you please assign it to me?

kclowes:
Thanks @ashish10677! Let me know if you have questions!

ashish10677:
@kclowes I had a very silly doubt. In the documentation, these methods: https://web3py.readthedocs.io/en/stable/v5_migration.html#json-rpc-updates , only one of them has a return type of `None` mentioned. Do you want me to include the return type of `TransactionNotFound` or `BlockNotFound` in all?

kclowes:
Yep, let's go ahead and include the return type for all of them.

ashish10677:
@kclowes I have raised a pull request. I am very new to open source contributions. Please guide me if there are any mistakes that I can fix. I will be happy to do it.

Swixx:
@kclowes Can this be closed because of #1527?

kclowes:
@Swixx Yes! Thanks!

PR #1527, opened 2019-11-25
Document update for v5

ashish10677:
### What was wrong?

The docs weren't updated when we started returning a TransactionNotFound or BlockNotFound error instead of None for RPC methods that don't find a transaction.

Related to Issue #1522

### How was it fixed?
Updated all the functions mentioned in https://web3py.readthedocs.io/en/stable/v5_migration.html#json-rpc-updates

### Todo:

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,10 @@ logs
 .idea/dictionaries
 .idea/vcs.xml
 .idea/jsLibraryMappings.xml
+.idea/inspectionProfiles
+.idea/misc.xml
+.idea/web3.py.iml
+.idea/modules.xml
 
 # Sensitive or high-churn files:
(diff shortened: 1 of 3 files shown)


kclowes:
Thanks @ashish10677! This looks good! I noticed that the migration docs are wrong. `getTransactionCount` doesn't return an error, but `getBlockTransactionCount` will raise a `BlockNotfound` if it doesn't find that block. Do you mind updating?

ashish10677:
@kclowes sorry for the mistake. I've updated the docs. Do I need to create a new merge request?

ashish10677:
@kclowes 1 lint test is failing. Can you tell me how to fix it?

PR #1528, opened 2019-11-26
Type hints utils abi

njgheorghita:
### What was wrong?
Type hints added to abi-relevant utility modules
- `web3._utils.abi`
- `web3._utils.contract`
- `web3._utils.events`

#### Cute Animal Picture
image

### DIFF
--- a/web3/_utils/abi.py
+++ b/web3/_utils/abi.py
@@ -80,7 +80,6 @@ from web3.exceptions import (
 )
 from web3.types import (
     ABI,
-    ABIElement,
     ABIEvent,
     ABIEventParams,
     ABIFunction,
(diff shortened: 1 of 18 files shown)


web3/_utils/abi.py
```diff
@@ -564,40 +618,40 @@ def get_constructor_abi(contract_abi):
...
 
 
-def is_string_type(abi_type):
+def is_string_type(abi_type: TypeStr) -> bool:
     return abi_type == 'string'
 
 
 @curry
-def is_length(target_length, value):
+def is_length(target_length: int, value: Any) -> bool:
```
pipermerriam: [`collections.abc.Sized`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sized) might be more appropriate for the `value: Any`

PR #1529, opened 2019-11-27
Type hints in more _utils.

njgheorghita:
### What was wrong?
Type hints added to:
- `web3._utils.datatypes`
- `web3._utils.decorators`
- `web3._utils.empty`
- `web3._utils.ens`
- `web3._utils.filters`

Removed `combomethod` from `web3._utils.decorators` in favor of `eth_utils.combomethod`

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam I'm noticing some duplication b/w `web3` & `eth-utils`. In this pr I deleted `combomethod` from `web3._utils.decorators` in favor of `eth_utils.combomethod`. I'm also noticing some duplication between `web3._utils.encoding` and `eth-utils` (i.e. a lot of the `to_hex`, `to_int`, etc..). Are those safe to remove as well in favor of the `eth-utils` versions? Or should all of these be deprecated instead and removed in the next major bump?

web3/_utils/datatypes.py
```diff
@@ -16,12 +24,25 @@ def verify_attr(class_name, key, namespace):
 
 
 class PropertyCheckingFactory(type):
-    def __init__(cls, name, bases, namespace, **kargs):
+    def __init__(
+        cls,
+        name: str,
+        bases: Tuple[type],
```
pipermerriam: This means a length-1 tuple with a single type.  I think you probably mean: `Tuple[Type[Any], ...]` which may-or-may-not be equivalent to `Tuple[type, ...]`

web3/_utils/decorators.py
```diff
@@ -1,55 +1,46 @@
...
+def deprecated_for(replace_message: str) -> Callable[..., Any]:
     """
     Decorate a deprecated function, with info about what to use instead, like:
 
     @deprecated_for("toBytes()")
     def toAscii(arg):
         ...
     """
-    def decorator(to_wrap):
+    def decorator(to_wrap: Callable[..., Any]) -> Callable[..., Any]:
```
pipermerriam: Generally best to do:

```python
TFunc = TypeVar("TFunc", bound=Callable[..., Any])

def decorator(to_wrap: TFunc) -> TFunc:
    @functools.wraps(to_wrap)
    def inner(...):
        ...
    return cast(TFunc, inner)
```

This preserves the types for the wrapped function.

PR #1530, opened 2019-11-27
Type hints in more _utils pt. 2

njgheorghita:
### What was wrong?
Moar type hints
- `web3._utils.request`
- `web3._utils.shh`
- `web3._utils.threads`
- `web3._utils.transactions`
- `web3._utils.validation`
- `web3._utils.windows`

#### Cute Animal Picture
image


web3/_utils/threads.py
```diff
@@ -60,16 +76,16 @@ def check(self):
...
+    def cancel(self) -> None:
         self.is_running = False
 
-    def sleep(self, seconds):
+    def sleep(self, seconds: float) -> None:
         time.sleep(seconds)
         self.check()
 
 
 class ThreadWithReturn(threading.Thread):
```
pipermerriam: I think that this should be typed with a `Generic[TReturn]`.  This allows you to then have the type of `target: Callable[..., TReturn`, and thus to be able to provide the type for the `def get(self, ...) -> TReturn`

web3/_utils/threads.py
```diff
@@ -91,23 +107,28 @@ def get(self, timeout=None):
...
             self.terminate_event.wait(self.interval)
 
-    def stop(self):
+    def stop(self) -> None:
         self.terminate_event.set()
 
 
-def spawn(target, *args, thread_class=ThreadWithReturn, **kwargs):
+def spawn(
+    target: Callable[..., Any],
```
pipermerriam: Same here, you could use `Callable[..., TReturn]` and have the function return type be `ThreadWithReturn[TReturn]`

web3/_utils/threads.py
```diff
@@ -79,10 +95,10 @@ def __init__(self, target=None, args=None, kwargs=None):
         self.args = args
         self.kwargs = kwargs
 
-    def run(self):
+    def run(self) -> None:
         self._return = self.target(*self.args, **self.kwargs)
 
-    def get(self, timeout=None):
+    def get(self, timeout: int=None) -> Any:
```
pipermerriam: `timeout` should probably be a `float`, same with all of the other timeouts and things that represent time units.

web3/_utils/transactions.py
```diff
@@ -65,7 +80,9 @@ def fill_transaction_defaults(web3, transaction):
     return merge(defaults, transaction)
 
 
-def wait_for_transaction_receipt(web3, txn_hash, timeout, poll_latency):
+def wait_for_transaction_receipt(
+    web3: "Web3", txn_hash: Hash32, timeout: int, poll_latency: float
```
pipermerriam: `timout` should probably be a `float`

web3/_utils/threads.py
njgheorghita: What about timestamps? Shouldn't those be `int`'s?

pipermerriam: Yeah, block timestamps are ints.

PR #1533, opened 2019-12-02
Fix doc about signed transactions

lionel-martin:
### What was wrong?

Account object did not have a sign_transaction function. Actually it is called signTransaction (probably renamed but not corrected in the doc?).

### How was it fixed?

Fixed the "typo" in the documentation for the next one to come. :-)

### Todo:


kclowes:
@lionel-martin Thanks so much for taking initiative! However, this should actually be `sign_transaction`. `signTransaction` is deprecated in eth-account v0.4.0. You may need to make a new virtual environment and redo your pip install so that you have the latest eth-account and therefore access to the `sign_transaction` method. Let me know if you're still having issues and I can help debug!

PR #1535, opened 2019-12-03
Type hint web3._utils.module_testing

njgheorghita:
### What was wrong?
Type hints in `web3._utils.module_testing`

#### Cute Animal Picture
image


web3/_utils/module_testing/eth_module.py
```diff
@@ -84,92 +108,106 @@ def test_eth_accounts(self, web3):
...
         assert is_integer(block_number)
         assert block_number >= 0
 
-    def test_eth_getBalance(self, web3):
+    def test_eth_getBalance(self, web3: "Web3") -> None:
         coinbase = web3.eth.coinbase
 
         with pytest.raises(InvalidAddress):
-            web3.eth.getBalance(coinbase.lower())
+            web3.eth.getBalance(ChecksumAddress(HexAddress(HexStr(coinbase.lower()))))
```
njgheorghita: It's starting to feel quite awkward having to triple cast to go from `str` -> `ChecksumAddress`

web3/types.py
```diff
@@ -37,6 +37,11 @@
 TValue = TypeVar("TValue")
 
 
+HexBytes32 = NewType("HexBytes32", HexBytes)
+HexStr32 = NewType("HexStr32", HexStr)
+
+_Hash32 = Union[Hash32, HexBytes, HexStr]
```
njgheorghita: @pipermerriam I just realized that I've most likely been using the `eth_typing.Hash32` type incorrectly, which is defined as a `bytes` type of len 32. I assumed it was any string or bytes representing a 32 byte hash. I think it would be nice to have a type that consolidates `Union[Hash32, HexBytes, HexStr]` which are all valid throughout `web3`.

One thought is maybe introducing `Bytes32` to replace `Hash32`, and then `Hash32` can be used to represent `Union[Bytes32, HexBytes32, HexStr32]` - But it's unclear how rigid the `Hash32` definition is used in `py-evm`.

web3/_utils/module_testing/eth_module.py
pipermerriam: Agreed...  thoughts on how to fix this?

web3/types.py
pipermerriam: `Hash32` is used very broadly across `py-evm` and `trinity` and this would be a breaking change to `eth-typing` which sounds like a logistical nightmare.  I'm not categorically opposed but the upgrade path seems riddled with insanity.

web3/_utils/module_testing/eth_module.py
pipermerriam: In this specific case I think that you can just `type: ignore` because this is explicitely using the wrong type.

njgheorghita: Not much off that I can think of, besides
1. relaxing the types (i.e. `ChecksumAddress = NewType("ChecksumAddress", str)`)
2. using `cast()` more often (i.e. `cast(ChecksumAddress, "0x123...")`

Unless @cburgdorf - any thoughts on how to simplify `ChecksumAddress(HexAddress(HexStr("0x123...")))`?

cburgdorf: I personally think we could take this:

```python
ChecksumAddress = NewType('ChecksumAddress', HexAddress)
AnyAddress = TypeVar('AnyAddress', Address, HexAddress, ChecksumAddress)
```

and turn it into this:

```python
ChecksumAddress = NewType('ChecksumAddress', HexAddress, str)
AnyAddress = TypeVar('AnyAddress', Address, HexAddress, ChecksumAddress, str)
```

After all, these `NewType`'s are just to define the intend of the underlying data and trace that it is used correctly across function boundaries.

So you might look at this and say: "Wait, but then we can define `ChecksumAddress("Foo")` which is clearly wrong. But I'd argue that you can do the same today just in a slightly more involved form of `CheckSumAddress(HexAddress(HexStr("Foo")))`. So, there is always a "trust gate" at the point where the `NewType` is created. The `NewType` can not actually check if this is a 20-byte string. It can only trace if the correct `NewType` flows from method to method. At the point where the `NewType` is being created we have to put some trust into the developer who's wrapping whatever type as a `NewType`.

njgheorghita: @cburgdorf I'm not sure it's possible to pass two types into a `NewType()`

> NewType accepts exactly two arguments. The first argument must be a string literal containing the name of the new type and must equal the name of the variable to which the new type is assigned. The second argument must be a properly subclassable class, i.e., not a type construct like Union, etc.

cburgdorf: Oh, I got confused because `AnyAddress` is a `TypeVar` and thought `NewType` has some syntactic sugar to wrap `Union` types. But I guess `ChecksumAddress` could still be `NewType('ChecksumAddress', Union[HexAddress, str])` to get the same thing, no?

njgheorghita: That would be nice, but unfortunately `NewType`'s don't accept `Union`s.

> The second argument must be a properly subclassable class, i.e., not a type construct like Union, etc.

cburgdorf: Meh. I give up then :sweat_smile:

PR #1538, opened 2019-12-04
Type hint web3.method

njgheorghita:
### What was wrong?
Type  hints needed in `web3.method`

#### Cute Animal Picture
image


web3/_utils/rpc_abi.py
```diff
@@ -17,16 +17,134 @@
 
 
 class RPC:
```
njgheorghita: @pipermerriam I was trying to find a more elegant solution to create these class variables which might eliminate some of the repetition, but I struggled to come up with anything simple b/c mypy doesn't actually run any code. Any suggestions?

pipermerriam: I think this is fine for now.

PR #1541, opened 2019-12-06
Add Matomo Tracking

cburgdorf:
### What was wrong?

Currently we do not have a good insight into our target group. In order to improve both web3.py itself and the documentation, it makes sense to aggregate some data.

### How was it fixed?

Add Matomo Tracking to Docs site.

Matomo is an Open Source web analytics platform that allows us
to get better insights and optimize for our audience without
the negative consequences of other compareable platforms.

Read more: https://matomo.org/why-matomo/

### Todo:

#### Cute Animal Picture

put a cute animal picture link inside the parentheses


pipermerriam:
I merged and tested and the Lahja PR worked so this is likely good to go.

PR #1542, opened 2019-12-06
Finish type hints in web3._utils

njgheorghita:
### What was wrong?
Type hints in the rest of `web3._utils`

#### Cute Animal Picture
image


web3/_utils/encoding.py
```diff
@@ -88,34 +104,34 @@ def to_hex_twos_compliment(value, bit_size):
...
     """
     value = remove_0x_prefix(value)
     return add_0x_prefix(value.zfill(int(bit_size / 4)))
 
 
-def trim_hex(hexstr):
+def trim_hex(hexstr: HexStr) -> HexStr:
     if hexstr.startswith('0x0'):
-        hexstr = re.sub('^0x0+', '0x', hexstr)
+        hexstr = HexStr(re.sub('^0x0+', '0x', hexstr))
```
pipermerriam: orthogonal to this PR but I'd be curious of the performance difference between this implementation and just converting to an integer and then back to hex...

web3/_utils/formatters.py
```diff
@@ -41,12 +55,14 @@ def apply_formatters_to_args(*formatters):
 
 @curry
 @to_list
-def apply_formatter_to_array(formatter, value):
+def apply_formatter_to_array(
+    formatter: Callable[[TValue], TReturn], value: Collection[TValue]
```
pipermerriam: I think this should be `Sequence` instead of `Collection` since there is an assumption of ordering which a collection doesn't guarantee.

PR #1543, opened 2019-12-06
Tighten up types via Method

njgheorghita:
### What was wrong?
`Method` class now uses `Generic` type to enforce stricter type checking amongst the module methods.

#### Cute Animal Picture
image


web3/_utils/method_formatters.py
```diff
@@ -472,7 +473,7 @@ def get_request_formatters(
 
 def get_result_formatters(
     method_name: Union[RPCEndpoint, Callable[..., RPCEndpoint]]
-) -> Dict[str, Callable[..., Any]]:
+) -> Dict[str, Callable[..., TReturn]]:
```
pipermerriam: I'm not sure these actually improve the situation any.  The idea behind `TypeVar` is that it lets you setup linkages between the various types for something.  Without any of the calling args using `TReturn` there is no way to anchor what the type should be so I suspect this is roughly equivalent to `Any`

Issue #1540, opened 2019-12-06
Missed dependency when import web3: typing_extensions


eugene-vasilev:
* Version: latest
* Python: 3.7
* OS: linux
* `pip freeze` output

```
attrdict==2.0.1
attrs==19.3.0
base58==1.0.3
certifi==2019.11.28
chardet==3.0.4
cytoolz==0.10.1
eth-abi==2.0.0
eth-account==0.4.0
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.1.2
eth-typing==2.2.1
eth-utils==1.8.4
hexbytes==0.2.0
idna==2.8
importlib-metadata==1.2.0
ipfshttpclient==0.4.12
jsonschema==3.2.0
lru-dict==1.1.6
more-itertools==8.0.1
multiaddr==0.0.8
mypy-extensions==0.4.3
netaddr==0.7.19
parsimonious==0.8.1
pkg-resources==0.0.0
protobuf==3.11.1
pycryptodome==3.9.4
pyrsistent==0.15.6
requests==2.22.0
rlp==1.2.0
six==1.13.0
toolz==0.10.0
urllib3==1.25.7
varint==1.0.2
web3==5.3.1
websockets==8.1
zipp==0.6.0

```

### What was wrong?

When I want to import web3, I catch the following error:

> File "/home/evasilev/Documents/Projects/pytests/common/base_test.py", line 14, in <module>
    from web3 import Web3
  File "/home/evasilev/Documents/Projects/pytests/venv/lib/python3.7/site-packages/web3/__init__.py", line 9, in <module>
    from web3.main import (
  File "/home/evasilev/Documents/Projects/pytests/venv/lib/python3.7/site-packages/web3/main.py", line 27, in <module>
    from web3._utils.abi import (
  File "/home/evasilev/Documents/Projects/pytests/venv/lib/python3.7/site-packages/web3/_utils/abi.py", line 70, in <module>
    from web3._utils.ens import (
  File "/home/evasilev/Documents/Projects/pytests/venv/lib/python3.7/site-packages/web3/_utils/ens.py", line 22, in <module>
    from web3.exceptions import (
  File "/home/evasilev/Documents/Projects/pytests/venv/lib/python3.7/site-packages/web3/exceptions.py", line 4, in <module>
    from web3.types import (
[...]

PR #1545, opened 2019-12-06
Add script to test packaging

kclowes:
### What was wrong?
We had a release (5.3.1) that was missing a dependency. Unfortunately, that wasn't caught until after the release. 

Related to Issue #1540 

### How was it fixed?
I stole a script from trinity that installs the release to a temporary virtualenv for testing.

### Todo:
- [x] Add this to CI (ex. https://github.com/ethereum/trinity/pull/106/files#diff-b91f3d5bd63fcd17221b267e851608e8R61-R64)

#### Cute Animal Picture

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -264,6 +264,12 @@ jobs:
     environment:
       TOXENV: py36-integration-ethtester
       ETHEREUM_TESTER_CHAIN_BACKEND: eth_tester.backends.PyEVMBackend
+  py36-wheel-cli:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-wheel-cli
   #
   # Python 3.7
   #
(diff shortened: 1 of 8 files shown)


kclowes:
@pipermerriam It looks like this script only builds from the wheel that is already in `dist/`. Does the Trinity team usually do a new local build and then run this script, or what is your preferred flow?

pipermerriam:
cc @carver did you setup the wheel tests for trinity?

veox:
@kclowes The `dist` is built as part of the `tox` job; here's the particular lines:

https://github.com/ethereum/trinity/pull/106/files#diff-b91f3d5bd63fcd17221b267e851608e8R61-R64

https://github.com/ethereum/trinity/pull/106/files# is the rest of that PR, of course.

kclowes:
Aha! Thanks for looking into that @veox!

kclowes:
Nope, good catch. Instead of adding to CI, I added to the Makefile and the README similar to what Trinity has going on. There is a smoke test in CI that makes sure that the wheel builds and that we can import web3 though.

PR #1548, opened 2019-12-10
Type hints complete

njgheorghita:
### What was wrong?
Finish type hints in `web3`

The way `TypedDict` was being imported in `web3._utils.compat` was buggy (it seems there is a `typing.TypedDict` pre `3.8`, but it's hardly effective in what it checks). After updating to the correct `TypedDict` implementation, naturally a lot of errors in the type-hints appeared. Those were all fixed. 

#### Cute Animal Picture

image

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,4 +63,4 @@ extras=linter
 commands=
   flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/ethpm {toxinidir}/tests
   isort --recursive --check-only --diff {toxinidir}/web3/ {toxinidir}/ens/ {toxinidir}/ethpm/ {toxinidir}/tests/
-  mypy -p web3._utils -p web3.providers -p web3.main -p web3.contract -p web3.datastructures -p web3.eth -p web3.exceptions -p web3.geth -p web3.iban -p web3.logs -p web3.manager -p web3.module -p web3.net -p web3.parity -p web3.middleware -p web3.method -p web3.pm -p web3.auto -p web3.gas_strategies -p web3.testing -p web3.tools -p web3.version -p ethpm -p ens --config-file {toxinidir}/mypy.ini
+  mypy -p web3 -p ethpm -p ens --config-file {toxinidir}/mypy.ini
--- a/web3/_utils/abi.py
+++ b/web3/_utils/abi.py
@@ -86,9 +86,7 @@ from web3.types import (
 
 
 def filter_by_type(_type: str, contract_abi: ABI) -> List[Union[ABIFunction, ABIEvent]]:
-    # fix for this just landed, not yet released : https://github.com/python/mypy/pull/7917
-    # after it's released, update Union[ABIFunction, ABIEvent] -> ABIElement
-    return [abi for abi in contract_abi if abi['type'] == _type]  # type: ignore
+    return [abi for abi in contract_abi if abi['type'] == _type]
 
 
 def filter_by_name(name: str, contract_abi: ABI) -> List[Union[ABIFunction, ABIEvent]]:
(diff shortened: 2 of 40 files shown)


web3/contract.py
```diff
@@ -1339,7 +1350,7 @@ def __getattr__(self, function_name: str) -> Any:
                 "The ABI for this contract contains no function definitions. ",
                 "Are you sure you provided the correct contract ABI?"
             )
-        elif function_name not in self._functions:
+        elif function_name not in [fn['name'] for fn in self._functions]:
```
njgheorghita: I've tried to not change any source code while type hinting, but this one to me seems like a genuine bug

web3/types.py
```diff
@@ -48,38 +46,37 @@
...
-})
+}, total=False)
 
 
-ABIFunctionComponents = TypedDict("ABIFunctionParams", {
+ABIFunctionComponents = TypedDict("ABIFunctionComponents", {
     "name": str,
     "type": str,
-    "components": Sequence["ABIFunctionComponents"],
+    "components": Sequence[Any],
```
njgheorghita: This is better typed as it was originally, `Sequence["ABIFunctionComponents"]` - but it seems like mypy doesn't allow for recursive definitions. Is there a better solution that I'm missing besides using `Any`?

web3/contract.py
pipermerriam: I'd ask that you extract this to a stand-alone PR and write a test that would have exposed this bug.

web3/types.py
pipermerriam: I think typing as `Any` and linking the the relevant mypy issue about recursive types.

tox.ini
```diff
@@ -63,4 +63,4 @@ extras=linter
 commands=
   flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/ethpm {toxinidir}/tests
   isort --recursive --check-only --diff {toxinidir}/web3/ {toxinidir}/ens/ {toxinidir}/ethpm/ {toxinidir}/tests/
-  mypy -p web3._utils -p web3.providers -p web3.main -p web3.contract -p web3.datastructures -p web3.eth -p web3.exceptions -p web3.geth -p web3.iban -p web3.logs -p web3.manager -p web3.module -p web3.net -p web3.parity -p web3.middleware -p web3.method -p web3.pm -p web3.auto -p web3.gas_strategies -p web3.testing -p web3.tools -p web3.version -p ethpm -p ens --config-file {toxinidir}/mypy.ini
+  mypy -p web3 -p ethpm -p ens --config-file {toxinidir}/mypy.ini
```
pipermerriam: sexy

web3/_utils/abi.py
```diff
@@ -97,25 +95,26 @@ def filter_by_name(name: str, contract_abi: ABI) -> List[Union[ABIFunction, ABIE
...
     ]
 
 
-def get_abi_input_types(abi: Union[ABIFunction, ABIEvent]) -> List[str]:
+def get_abi_input_types(abi: ABIFunction) -> List[str]:
     if 'inputs' not in abi and abi['type'] == 'fallback':
         return []
     else:
-        return [collapse_if_tuple(arg) for arg in abi['inputs']]
+        # https://github.com/python/mypy/issues/4976
```
pipermerriam: is this comment outdated?  maybe a text comment with a brief explanation would be good too?

web3/_utils/module_testing/personal_module.py
```diff
@@ -78,19 +84,19 @@ def test_personal_sendTransaction(
...
+            'value': Wei(1),
             'gasPrice': web3.toWei(1, 'gwei'),
         }
         txn_hash = web3.geth.personal.sendTransaction(txn_params, unlockable_account_pw)
         assert txn_hash
         transaction = web3.eth.getTransaction(txn_hash)
 
-        assert is_same_address(transaction['from'], txn_params['from'])
-        assert is_same_address(transaction['to'], txn_params['to'])
+        assert is_same_address(transaction['from'], cast(ChecksumAddress, txn_params['from']))
```
pipermerriam: I think the better way here is to do `ChecksumAddress(txn_params['from'])` since that will throw an error during runtime.  I assume the `cast` is to get around the need for multiple layers of casting....  I'm inclined to prefer the multiple layers over `cast` in this situation now that I realize that it has runtime implications.

web3/contract.py
pipermerriam: Also, when doing inclusion checks you should default to use a `set()`

web3/manager.py
```diff
@@ -134,7 +134,7 @@ def _make_request(
             self.web3,
             self.middleware_onion)
         self.logger.debug("Making request. Method: %s", method)
-        return await request_func(method, params)
+        return await request_func(method, params)  # type: ignore
```
pipermerriam: comment as to why ignored?

web3/middleware/cache.py
```diff
@@ -1,24 +1,30 @@
 import functools
 import threading
 import time
-from typing import (
+from typing import (  # noqa: F401
```
pipermerriam: I would advocate putting the `Union` import inside the `if TYPE_CHECKING` block so that it is a bit clearer that it's the only one that the `F401` applies to.

web3/_utils/module_testing/personal_module.py
njgheorghita: Gotcha, I went back to the layered casting for most examples, but since `transaction['from']` (and `['to']`) are typed as `Union[Address, ChecksumAddress, str]` (to allow for ens) - I think we still need to `cast()` in some situations.

PR #1554, opened 2019-12-17
Update typing-extensions imports to compatible version

njgheorghita:
### What was wrong?
re @veox in the web3 gitter channel - `typing_extensions` was incorrectly used instead of the imported version from `web3._utils.compat`

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/newsfragments/1554.bugfix.rst
+++ b/newsfragments/1554.bugfix.rst
@@ -1,0 +1,1 @@
+Update modules to use compatible typing-extensions import.
--- a/web3/_utils/transactions.py
+++ b/web3/_utils/transactions.py
@@ -16,10 +16,10 @@ from eth_utils.toolz import (
 from hexbytes import (
     HexBytes,
 )
-from typing_extensions import (
+
+from web3._utils.compat import (
     Literal,
 )
-
 from web3._utils.threads import (
     Timeout,
 )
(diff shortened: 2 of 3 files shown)


veox:
I've detected this by running `ipython` and then `import web3` (or `from web3.auto import w3`) with a clean-room Python 3.8 install (via `pip install -e .`).

It resulted in

```
<long traceback snipped>
~/src/web3.py/web3/_utils/transactions.py in <module>
     17     HexBytes,
     18 )
---> 19 from typing_extensions import (
     20     Literal,
     21 )
```

Is there a test case for this?.. It would probably happen without `ipython`, too, with plain `python -c 'from web3.auto import w3'`.

njgheorghita:
@kclowes Correct me if I'm wrong, but I vaguely recall you were writing up some kind of testing script that could handle ^?

veox:
If that was me, then I don't remember doing it, and don't know where it is. I sure hope it was someone else. :sweat_smile:

kclowes:
Yep, its in #1545. I stole it from Trinity and I'm not 100% sure, it looks to me like it's designed to be run before a release, rather than automatically by the test suite. I'm waiting on confirmation to be sure I'm not missing something though.

Issue #1547, opened 2019-12-10
Receipt formatter doesn't format 'from' and 'to' fields to checksum addresses


ignition42:
### What was wrong?

Since `w3.eth.getTransaction()` returns the `from` and `to` as checksum addresses I was expecting the same behaviour for the receipt methods like `w3.eth.waitForTransactionReceipt()` or `w3.eth.getTransactionReceipt()`.

Basically the [receipt formatter](https://github.com/ethereum/web3.py/blob/v5.4.0/web3/_utils/method_formatters.py#L188) used [pythonic middleware](https://github.com/ethereum/web3.py/blob/v5.4.0/web3/middleware/pythonic.py) does not format `from` and `to` addresses to checksum as it does with [transaction formatter](https://github.com/ethereum/web3.py/blob/v5.4.0/web3/_utils/method_formatters.py#L139). So when I call `w3.eth.waitForTransactionReceipt()` or `w3.eth.getTransactionReceipt()`, `from`
 and `to` fields non checksumed addresses in lowercase. 

### How can it be fixed?

Basically add them here: https://github.com/ethereum/web3.py/blob/v5.4.0/web3/_utils/method_formatters.py#L174

As they are here: https://github.com/ethereum/web3.py/blob/v5.4.0/web3/_utils/method_formatters.py#L119

rajeev29five:
I would like to work on this.

ashish10677:
@rajeev29five are you trying to fix this issue or can I also try?

rajeev29five:
@ashish10677 I have already fixed the issue, just need to create a PR.

PR #1562, opened 2020-01-03
Format 'from' and 'to' fields to checksum addresses in RECEIPT_FORMAT…

rajeev29five:
### What was wrong?

Receipt formatter was not formatting **from** and **to** addresses to checksum as it does with transaction formatter.

Related to Issue #1547 

### How was it fixed?

Followed @ignition42's instructions.

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_ethereum_tester.py
+++ b/tests/integration/test_ethereum_tester.py
@@ -298,6 +298,10 @@ class TestEthereumTesterEthModule(EthModuleTest):
         assert is_integer(chain_id)
         assert chain_id == 61
 
+    @pytest.mark.xfail(raises=KeyError, reason="ethereum tester doesn't return 'to' key")
+    def test_eth_getTransactionReceipt_mined(self, web3, block_with_txn, mined_txn_hash):
+        super().test_eth_getTransactionReceipt_mined(web3, block_with_txn, mined_txn_hash)
+
 
 class TestEthereumTesterVersionModule(VersionModuleTest):
     pass
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -767,6 +767,9 @@ class EthModuleTest:
         assert receipt['blockHash'] == block_with_txn['hash']
         assert receipt['transactionIndex'] == 0
         assert receipt['transactionHash'] == HexBytes(mined_txn_hash)
+        assert is_checksum_address(receipt['to'])
+        assert receipt['from'] is not None
+        assert is_checksum_address(receipt['from'])
 
     def test_eth_getTransactionReceipt_unmined(
         self, web3: "Web3", unlocked_account_dual_type: ChecksumAddress
--- a/newsfragments/1562.bugfix.rst
+++ b/newsfragments/1562.bugfix.rst
@@ -1,0 +1,1 @@
+Make 'from' and 'to' fields checksum addresses in returned transaction receipts
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -183,6 +183,8 @@ RECEIPT_FORMATTERS = {
     'contractAddress': apply_formatter_if(is_not_null, to_checksum_address),
     'logs': apply_list_to_array_formatter(log_entry_formatter),
     'logsBloom': to_hexbytes(256),
+    'from': apply_formatter_if(is_not_null, to_checksum_address),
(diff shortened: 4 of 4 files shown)


rajeev29five:
Ok @kclowes  I'll do that. If I'm stuck I'll ask for help.
I think I'll have to add something like ...receipt['from'] too. Right?

kclowes:
That would be ideal, but you may need to also add a check to make sure it's there since we're doing the `is_not_null` check for the `from` formatter. I didn't check that 'from' comes back with the transaction receipts in those tests. Thank you!

rajeev29five:
@kclowes looks like from and to key do not come back with transaction receipt. Little guidance please.
Thank you!

kclowes:
@rajeev29five let's add the xfail decorator (`@pytest.mark.xfail(raises=KeyError)`) to the test for the ethtester integration tests. So in `tests/integration/test_ethereum_tester.py`, make a new method in the `TestEthereumTesterEthModule` class:
```python
@pytest.mark.xfail(raises=KeyError, reason="ethereum tester doesn't return from or to keys")    
def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_account):
        super().test_eth_getTransactionReceipt_unmined(web3, unlocked_account)
```

rajeev29five:
@kclowes Need some help with these Failures

tests/integration/test_ethereum_tester.py
```diff
@@ -298,6 +298,9 @@ def test_eth_chainId(self, web3):
         assert is_integer(chain_id)
         assert chain_id == 61
 
+    @pytest.mark.xfail(raises=KeyError, reason="ethereum tester doesn't return 'to' key")
+    def test_eth_getTransactionReceipt_unmined(self, eth_tester, web3, unlocked_account):
```
kclowes: I believe this should be `def test_eth_getTransactionReceipt_mined`. Sorry I mistyped earlier.

web3/_utils/module_testing/eth_module.py
```diff
@@ -758,6 +758,7 @@ def test_eth_getTransactionByBlockNumberAndIndex(
         assert is_dict(transaction)
         assert transaction['hash'] == HexBytes(mined_txn_hash)
 
+    @pytest.mark.xfail(raises=KeyError)
```
kclowes: This is the higest level that gets run for each different client (ethtester, geth, parity). Since only ethtester was failing, you'll need to remove this here. It will be overridden for ethtester where you have it above in `tests/integration/test_ethereum_tester.py`. Since we don't expect a failure for geth or parity (I don't think), you can remove the xfail here.

tests/integration/test_ethereum_tester.py
```diff
@@ -298,6 +298,9 @@ def test_eth_chainId(self, web3):
         assert is_integer(chain_id)
         assert chain_id == 61
 
+    @pytest.mark.xfail(raises=KeyError, reason="ethereum tester doesn't return 'to' key")
+    def test_eth_getTransactionReceipt_mined(self, eth_tester, web3, unlocked_account):
+        super().test_eth_getTransactionReceipt_unmined(web3, unlocked_account)
```
kclowes: @rajeev29five this line should also be `test_eth_getTransactionReceipt_mined`. I think then you'll be in good shape as soon as you fix the lint tests.

rajeev29five: @kclowes I was about to ask you about that line. I will make the changes test for any other failures. Thanks!

rajeev29five: @kclowes What should I pass for positional argument: 'mined_txn_hash' ?

rajeev29five: I think it should be `super().test_eth_getTransactionReceipt_mined(web3, unlocked_account, mined_txn_hash)` like this. Right?

kclowes: You'll need to pass in `web3`, `block_with_txn`, and `mined_txn_hash`, just like in the original definition in `module_testing/eth_module.py`. So the function should look like: 
```
@pytest.mark.fail(...)
def test_eth_getTransactionReceipt_mined(self, web3, block_with_txn, and mined_txn_hash):
        super().test_eth_getTransactionReceipt_mined(web3, block_with_txn, and mined_txn_hash)
```
Sorry for all the confusion. Let me know if you want a better explanation of why.

rajeev29five:
All checks passed. Thanks for all the help : ) @kclowes

PR #1565, opened 2020-01-11
add support of Solidity 0.6.0 new receive keyword

jerryji:
### What was wrong?

Solidity 0.6.0 adds a new "receive" keyword ( https://solidity.readthedocs.io/en/v0.6.1/060-breaking-changes.html#explicitness-requirements )

web3.py 5.4.0 fails to call _any_ function when ^0.6.0 contract contains "receive" payable fallback function as it chokes on missing 'name' in abi due to type == 'receive'

### How was it fixed?

Adding "receive" to _utils/abi.py row 101

### Todo:
there should probably be a few more places to add "receive" for it to be supported properly -- this is just a starting point that fixed my immediate problem at hand

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis--> Are You Serious?

### DIFF
--- a/tests/core/utilities/test_abi_filter_by_abi_name.py
+++ b/tests/core/utilities/test_abi_filter_by_abi_name.py
@@ -18,7 +18,6 @@ ABI_CONSTRUCTOR = {
 }
 ABI_FALLBACK = {
     "constant": False,
-    #"stateMutability": "nonpayable", # actually this is how 0.6 abi looks like
     "type": "fallback",
 }
 ABI_RECEIVE = {
--- a/tests/core/utilities/test_abi_filter_by_abi_name.py
+++ b/tests/core/utilities/test_abi_filter_by_abi_name.py
@@ -18,7 +18,12 @@ ABI_CONSTRUCTOR = {
 }
 ABI_FALLBACK = {
     "constant": False,
+    #"stateMutability": "nonpayable", # actually this is how 0.6 abi looks like
     "type": "fallback",
+}
+ABI_RECEIVE = {
+    "stateMutability": "payable",
+    "type": "receive"
 }
 ABI_FUNC_2_SIG_A = {
     "constant": False,
@@ -50,6 +55,7 @@ ABI_FUNC_3 = {
 ABI = [
     ABI_CONSTRUCTOR,
     ABI_FALLBACK,
+    ABI_RECEIVE,
     ABI_FUNC_1,
     ABI_FUNC_2_SIG_A,
     ABI_FUNC_2_SIG_B,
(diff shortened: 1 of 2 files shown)


pipermerriam:
Can you please add a test case for this?

jerryji:
Excuse my ignorance, how do I add a test requiring a new contract that needs to be compiled in Solidity ^0.6.0?

kclowes:
@jerryji there is a test [here](https://github.com/ethereum/web3.py/blob/f5fcaa67b5858a8b9f144023c50b73a0914981e1/tests/core/utilities/test_abi_filter_by_abi_name.py) for the function that you changed. You should be able to add the ABI for the new `receive` keyword like [here](https://github.com/ethereum/web3.py/blob/f5fcaa67b5858a8b9f144023c50b73a0914981e1/tests/core/utilities/test_abi_filter_by_abi_name.py#L19). 

There are also a bunch of places that we check for a fallback function that we should probably add a check for `receive` now as well, just to make sure that everything is working as it should. For example, tests where we call `Contract._find_matching_fn_abi` should check to make sure we handle `receive` correctly. A fairly safe bet to make sure all the cases are covered is to search the codebase for `fallback` and add tests for the `receive` function.

Similarly, it's probably also a good idea to write few high-level tests make sure that the correct function from the contract is called based on these bullet points from the docs you linked to (maybe in `tests/core/contracts/test_contract_call_interface.py`): 

> - If present, the receive ether function is called whenever the call data is empty (whether or not ether is received). This function is implicitly payable.
> - The new fallback function is called when no other function matches (if the receive ether function does not exist then this includes calls with empty call data). 

Let me know if that novella doesn't make sense, or if you have other questions. I'm also happy to pick up the testing, it'll just be a bit. Thank you!

jerryji:
Actually the Solidity 0.6 abi looks like the following (i.e, not "constant": True/False) --
'''
    {   
      "stateMutability": "nonpayable",
      "type": "fallback"
    },  
    {   
      "stateMutability": "payable",
      "type": "receive"
    }   
'''

I don't think I have time to dive into the details of tests/core/contracts/test_contract_call_interface.py any time soon.

kclowes:
Ah, okay. Since the ABI is different, I think we'll want to add a new ABI in the test -  one for solidity v0.6 and and one for v0.5. I'll put it on my list to pick up unless someone else beats me to it. Thanks @jerryji!

kclowes:
@jerryji did you get an error when calling the receive function or what happened? I want to make sure we have the right fix. And can you provide an example of how were you trying to call the function? Thanks!

jerryji:
> did you get an error when calling the receive function or what happened?

The error occurs when calling _any_ function if a receive function is present.

I've put together a simple packaged to demonstrate the error, please download and follow the README.md inside --

https://drive.google.com/open?id=1EpLRnO_gdADEK6jZOf6GK3bcfBIUgFjb

Let me know if you need further clarification.

kclowes:
Yikes, I didn't realize it was every function in the contract. Thanks for putting together the package. It felt weird to be pushing significant work to your fork, so I cherry-picked your commits and opened #1623 in my fork. Closing in favor of #1623.

PR #1568, opened 2020-01-21
[WIP] Eth to method

kclowes:
### What was wrong?

Related to Issue #

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/blocks.py
+++ b/web3/_utils/blocks.py
@@ -10,6 +10,9 @@ from eth_utils import (
     is_text,
     remove_0x_prefix,
 )
+from eth_utils.toolz import (
+    curry,
+)
 
 from web3.types import (
     RPCEndpoint,
@@ -51,6 +54,7 @@ def is_hex_encoded_block_number(value: Any) -> bool:
     return 0 <= value_as_int < 2**256
 
 
+@curry
 def select_method_for_block_identifier(
     value: Any, if_hash: RPCEndpoint, if_number: RPCEndpoint, if_predefined: RPCEndpoint
 ) -> RPCEndpoint:
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -39,6 +39,9 @@ from hexbytes import (
     HexBytes,
 )
 
+from web3.exceptions import (
+    BlockNotFound,
+)
 from web3._utils.abi import (
     is_length,
 )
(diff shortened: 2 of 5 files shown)


ethpm/_utils/deployments.py
```diff
@@ -47,7 +47,7 @@ def get_linked_deployments(deployments: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def validate_linked_references(
-    link_deps: Tuple[Tuple[int, bytes], ...], bytecode: HexBytes
+    link_deps: Tuple[Tuple[int, bytes], ...], bytecode: HexStr
```
kclowes, line 50: @njgheorghita I am not sure why I had to make this change to appease the linter, but do you see any problems with it?

setup.py
```diff
@@ -73,7 +73,7 @@
         "eth-account>=0.5.2,<0.6.0",
         "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
         "eth-typing>=2.0.0,<3.0.0",
-        "eth-utils>=1.9.3,<2.0.0",
+        "eth-utils==1.9.3",
```
pipermerriam, line 76: This will cause problems elsewhere if it remains.  For example Trinity, installs Web3.py and I know that trinity requires a later version of `eth-utils`

web3/_utils/validation.py
```diff
@@ -152,10 +155,23 @@ def validate_abi_value(abi_type: TypeStr, value: Any) -> None:
     )
 
 
+def is_ordinary_string(value: Any) -> bool:
+    return (is_string(value) and not is_bytes(value) and not
```
pipermerriam, line 159: I think that the `is_string` check and the `is_bytes` check are mutually exclusive so you should be able to drop the `is_bytes` check.

web3/_utils/validation.py
```diff
@@ -152,10 +155,23 @@ def validate_abi_value(abi_type: TypeStr, value: Any) -> None:
     )
 
 
+def is_ordinary_string(value: Any) -> bool:
```
pipermerriam, line 158: Maybe this would be more accurately named: `is_non_address_string` or something with the same intent?

setup.py
kclowes: That's a bummer. eth-utils 1.9.4 breaks web3's linting so will probably have to update that before this can be shipped

kclowes:
That is a good idea. I'll work on splitting it into a few PRs!

ethpm/_utils/deployments.py
njgheorghita: Hmm, I can't tell either off the top of my head, but this change doesn't seem problematic to me. 👍

kclowes:
This got split into a bunch of PRs referenced above, but they've all been merged. Closing!

PR #1570, opened 2020-01-28
Increase replacement transaction minimum gas price bump to 12.5%

nfeignon:
Parity requires a gas price 12.5% higher than the previous one to accept
a replacement transaction in its transaction queue.
The previous value of 10% only worked for Geth.

See https://wiki.parity.io/Transactions-Queue and
https://github.com/paritytech/parity-ethereum/blob/e69539357f50d20bff51df28c556af8663552eec/miner/src/pool/scoring.rs#L38

### How was it fixed?

Increase to 12.5%.

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/transactions.py
+++ b/web3/_utils/transactions.py
@@ -179,7 +179,10 @@ def assert_valid_transaction_params(transaction_params: TxParams) -> None:
 
 
 def prepare_replacement_transaction(
-    web3: "Web3", current_transaction: TxData, new_transaction: TxParams
+    web3: "Web3",
+    current_transaction: TxData,
+    new_transaction: TxParams,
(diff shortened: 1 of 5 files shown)


web3/_utils/transactions.py
```diff
@@ -195,7 +195,7 @@ def prepare_replacement_transaction(
             raise ValueError('Supplied gas price must exceed existing transaction gas price')
     else:
         generated_gas_price = web3.eth.generateGasPrice(new_transaction)
-        minimum_gas_price = int(math.ceil(current_transaction['gasPrice'] * 1.1))
```
pipermerriam: I'm thinking that the number here should be part of the function signature with `1.125` as the default value.

PR #1571, opened 2020-01-29
[ethPM] Update registry uri

njgheorghita:
### What was wrong?
Updated the Registry URI utils to support URIs with `ethpm` as the scheme and package ids conforming to `pkg_name@version` - containing escaped characters

re: #1486 

### Todo:

#### Cute Animal Picture
image


ethpm/backends/registry.py
```diff
@@ -97,8 +100,17 @@ def parse_registry_uri(uri: str) -> RegistryURI:
...
     return RegistryURI(address, chain_id, pkg_name, pkg_version, ens)
+
+
+def process_pkg_path(pkg_path: str) -> Tuple[str, str]:
+    pkg_id = pkg_path.strip("/")
+    if not pkg_id:
+        return None, None
+
+    if "@" not in pkg_id:
+        return pkg_id, None
```
pipermerriam: This return and the one above this violate the type signagure of this function....  looks like it should be `Tuple[Optional[str], Optional[str]]`

ethpm/backends/registry.py
```diff
@@ -97,8 +100,17 @@ def parse_registry_uri(uri: str) -> RegistryURI:
...
+    pkg_name, safe_pkg_version = pkg_id.split("@")
```
pipermerriam: This line would still be executed if `pkg_id = `a@b@c`.

Since `version` is technically allowed to be any string (i think) then I think you need to do this with `pkg_id.partition('@')` which *might* result in a `safe_pkg_version` with multiple `@` symbols in it.

njgheorghita:
@pipermerriam I made one minor update: making the `chain_id` optional - if it's not included, the default is the mainnet. This goes one step further to making registry uris easy to use - and since we want to encourage registries on the mainnet, it seems like a safe call. 

> Any documentation updates that should go with this?

Yup doc updates are in #1486 and I've already made the updates to docs.ethpm.com which should cover the bases

ethpm/backends/registry.py
njgheorghita: This case is covered in `validate_registry_uri` - which will raise an exception if there is more than 1 `@` symbol - since at this stage all `@` symbols should be properly escaped (i.e. `"safe"`). There's test cases covering both ...
- invalid paths: `pkg@1@0` (version= `1@0`)
- valid paths: `pkg@1%400` (version= `1@0`)

PR #1576, opened 2020-02-07
[ethpm] Allow registry uri to namespace package assets

njgheorghita:
### What was wrong?
Updated the ethpm uri parsing functions to support namespaced assets.
`ethpm://compound.ethpm.eth:1/compound@1.0.0/deployments/cDai`

This will be useful in the ethpm widget - so that all the data needed to fetch the contract address of a deployment is in a single url.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/backends/registry.py
+++ b/ethpm/backends/registry.py
@@ -33,7 +33,10 @@ from ethpm.validation.uri import (
 
 # TODO: Update registry ABI once ERC is finalized.
 REGISTRY_ABI = fetch_standard_registry_abi()
-RegistryURI = namedtuple("RegistryURI", ["address", "chain_id", "name", "version", "ens"])
+RegistryURI = namedtuple(
+    "RegistryURI",
+    ["address", "chain_id", "name", "version", "namespaced_asset", "ens"]
+)
 
 
 class RegistryURIBackend(BaseURIBackend):
@@ -57,7 +60,7 @@ class RegistryURIBackend(BaseURIBackend):
         """
         Return content-addressed URI stored at registry URI.
         """
-        address, chain_id, pkg_name, pkg_version, _ = parse_registry_uri(uri)
+        address, chain_id, pkg_name, pkg_version, _, _ = parse_registry_uri(uri)
         if chain_id != '1':
             # todo: support all testnets
             raise CannotHandleURI(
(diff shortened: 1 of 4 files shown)


ethpm/backends/registry.py
```diff
@@ -33,7 +33,10 @@
 
 # TODO: Update registry ABI once ERC is finalized.
 REGISTRY_ABI = fetch_standard_registry_abi()
-RegistryURI = namedtuple("RegistryURI", ["address", "chain_id", "name", "version", "ens"])
+RegistryURI = namedtuple(
+    "RegistryURI",
+    ["address", "chain_id", "name", "version", "namespaced_asset", "ens"]
```
pipermerriam: FYI, if you aren't aware, what you are calling `namespace_asset` is a [JSON Pointer](https://tools.ietf.org/html/rfc6901)

ethpm/backends/registry.py
```diff
@@ -104,15 +107,35 @@ def parse_registry_uri(uri: str) -> RegistryURI:
...
+            "Invalid registry URI, missing package version."
+            "Version is required if namespaced assets are defined."
+        )
+    return pkg_name, pkg_version, namespaced_asset
+
+
+def _parse_pkg_path(pkg_path: str) -> Tuple[str, Optional[str]]:
+    if "/" in pkg_path:
+        pkg_id = pkg_path.split("/")[0]
+        namespaced_asset = "/".join(pkg_path.split("/")[1:])
```
pipermerriam: These two lines can be replaced with:

```python
pkg_id, _, namespaced_asset = pkg_path.partition('/')
```

ethpm/backends/registry.py
```diff
@@ -104,15 +107,35 @@ def parse_registry_uri(uri: str) -> RegistryURI:
...
 
-def _process_pkg_path(pkg_path: str) -> Tuple[Optional[str], Optional[str]]:
-    pkg_id = pkg_path.strip("/")
-    if not pkg_id:
-        return None, None
 
+def _parse_pkg_id(pkg_id: str) -> Tuple[str, Optional[str]]:
     if "@" not in pkg_id:
         return pkg_id, None
     pkg_name, safe_pkg_version = pkg_id.split("@")
```
pipermerriam: This call still jumps out to me as a minor problem.  It relies on validation that happens outside of this function.  If you change this to `pkg_name, _, safe_pkg_version = pkg_id.partition('@')` then this code would be slightly more robust.  Totally a *nitpick* on my part.

ethpm/backends/registry.py
njgheorghita: Oh, good to know! I'm not sure if it's strictly a JSON pointer? Essentially, when trying to specify a certain deployment - I'm trying to avoid requiring the entire blockchain URI. Do you think that's a bad idea? 

Basically...
`ethpm://maker.ethpm.eth:1/dai@1.0.0/deployments/DeploymentName` 
Is what I'm aiming for. . . . It seems like there might be some edge cases where this isn't ideal - in terms of forked chains that share a chain ID? Since this is for the widget - I was thinking of just specifying that a registry URI with a pointer/namespaced asset only supports the 5 main chains.

PR #1579, opened 2020-02-18
Add GethMiner snake case methods

billbsing:
### What was wrong?

Related to Issue #1429

### How was it fixed?
add new methods, and mark the current non snake case method names as depreciated.


tests/core/mining-module/test_miner_setGasPrice.py
```diff
@@ -10,15 +11,35 @@
...
     initial_gas_price = web3.eth.gasPrice
 
     # sanity check
     assert web3.eth.gasPrice > 1000
 
-    web3.geth.miner.setGasPrice(initial_gas_price // 2)
+    web3.geth.miner.set_gas_price(initial_gas_price // 2)
+
+    with Timeout(60) as timeout:
+        while web3.eth.gasPrice == initial_gas_price:
```
kclowes: even tough I'm pretty sure these are also getting skipped, let's change `gasPrice` to `gas_price`
```suggestion
        while web3.eth.gas_price == initial_gas_price:
```

docs/web3.miner.rst
```diff
@@ -71,23 +88,33 @@ The following methods are available on the ``web3.geth.miner`` namespace.
...
     * Delegates to ``miner_startAutoDag`` RPC Method
 
     Enable automatic DAG generation.
 
     .. code-block:: python
 
-        >>> web3.geth.miner.startAutoDAG()
+        >>> web3.geth.miner.start_auto_dag()
 
+.. py:method:: GethMiner.startAutoDAG()
```
kclowes: This is existing, but we only have a `startAutoDag` method, not `startAutoDAG`. Do you mind changing?
```suggestion
.. py:method:: GethMiner.startAutoDag()
```

docs/web3.miner.rst
```diff
@@ -71,23 +88,33 @@ The following methods are available on the ``web3.geth.miner`` namespace.
...
     * Delegates to ``miner_stopAutoDag`` RPC Method
 
     Disable automatic DAG generation.
 
     .. code-block:: python
 
-        >>> web3.geth.miner.stopAutoDAG()
+        >>> web3.geth.miner.stop_auto_dag()
+
+.. py:method:: GethMiner.stopAutoDAG()
```
kclowes: Same here
```suggestion
.. py:method:: GethMiner.stopAutoDag()
```

tests/core/mining-module/test_miner_setGasPrice.py
billbsing: Not sure I should be doing this in this PR, the property `gasPrice` is from the 'eth' module. I'm happy to do the eth module later..

tests/core/mining-module/test_setEtherBase.py
```diff
@@ -1,7 +1,18 @@
+import pytest
+
+
+def test_miner_set_etherbase(web3_empty):
+    web3 = web3_empty
+    assert web3.eth.coinbase == web3.eth.accounts[0]
+    new_account = web3.personal.newAccount('this-is-a-password')
+    web3.geth.miner.set_etherBase(new_account)
```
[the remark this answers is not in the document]
billbsing: thanks missed that one.. in theory we should also rename the test file, to `test_setEtherbase.py', but I'll leave it later to rename file names correctly to the snake case version 'test_set_etherbase.py'

tests/core/mining-module/test_miner_setGasPrice.py
kclowes: Ha, you're right. My bad!

PR #1581, opened 2020-02-21
add new Net module snake-case properties

billbsing:
### What was wrong?

Related to Issue #1429

### How was it fixed?

Changed the properties of the net module to snake-case

### DIFF
--- a/web3/net.py
+++ b/web3/net.py
@@ -1,6 +1,7 @@
 from typing import (
     NoReturn,
 )
+import warnings
 
 from web3._utils.rpc_abi import (
     RPC,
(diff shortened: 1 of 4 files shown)


web3/net.py
```diff
@@ -1,6 +1,8 @@
+
```
kclowes: _nitpick_: let's get rid of this line

web3/providers/eth_tester/defaults.py
```diff
@@ -198,8 +198,11 @@ def personal_send_transaction(eth_tester: "EthereumTester", params: Any) -> HexS
     },
     'net': {
         'version': static_return('1'),
-        'peerCount': static_return(0),
+        'peer_count': static_return(0),
```
kclowes: This should still be `peerCount`, since this is meant to mirror the JSON_RPC response that would be issued from a client, and we also wouldn't need the deprecated comment on line 203 either.

billbsing: got a bit too carried away..

Issue #941, opened 2018-06-29
Return values for eth_call are truncated when using revert


Uxio0:
* Version: 4.4.0
* Python: 3.6
* OS: linux

### What was wrong?

When calling a contract with `eth_call` to a function that has a `revert(string(abi.encodePacked("Hello")));` I would expect a return value like this:

```
0x08c379a0                                                         // Function selector for Error(string)
0000000000000000000000000000000000000000000000000000000000000020 // Data offset
000000000000000000000000000000000000000000000000000000000000001a // String length
4e6f7420656e6f7567682045746865722070726f76696465642e000000000000 // String data
```
As stated in the solidity documentation: http://solidity.readthedocs.io/en/v0.4.24/control-structures.html

Instead, I get the return value truncated:

`0x8c379a000000000000000000000000000000000000000000000000000000000`

If I see the RPC Call that web3 does in Wireshark I see this as the result: 

`0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000001df7`

### How can it be fixed?

Don't truncate the return value

### UPDATE

I can just reproduce the issue calling `mycontract.functions.myfunction().call()`. If I use `eth.call` I get the non truncated output (as it should be)

dylanjw:
The contract function call() method has no handling for error reasons strings.  It fails to detect the errored return value, and attempts to parse the return value according the the function abi.  We need to add a way to detect the error return string, and then parse it according the reason string format.

carver:
>  We need to add a way to detect the error return string, and then parse it according the reason string format.

Also, probably add a new exception type to raise. I'm not sure how standard this is (or might become) across languages, so maybe a `SolidityError`.

dylanjw:
It looks like py-evm/eth-tester will also need to add support for the error return string spec.

karlb:
I'm interested in writing a PR for this. Is there already a test case including a revert that can be used?

kclowes:
@karlb That would be great! I don't think we have any test cases already. Feel free to ping me if you need anything!

Issue #1580, opened 2020-02-20
Flaky integration tests on a ssh.post in the ssh module


billbsing:
* Version: 5.5.1
* Python: py36-integration-goethereum-http-1.7.2
* OS: linux

### What was wrong?
tox tests fail on py36-integration-goethereum-http-1.7.2 for ssh_module
at https://github.com/ethereum/web3.py/blob/dd28939b351ef72ef038670368eb940669c2c65f/web3/_utils/module_testing/shh_module.py#L374

see the full [circelc logi](https://circleci.com/gh/ethereum/web3.py/97134)

### How can it be fixed?

I think a `time.sleep(1)` after each post in `test_shh_post` and `test_shh_post_deprecated` , to allow the transaction too have enough time to complete?

PR #1585, opened 2020-02-24
Parse revert reason

karlb:
# What was wrong?

Solidity allows messages in `require` statements which can help a lot
when debugging. This commit parses the revert reasons from failing
`call`s and raises them as `SolidityError` exceptions.

Closes https://github.com/ethereum/web3.py/issues/941.

### Todo:
- [x] Add unit test
- [x] Add integration test @kclowes 
- [x] Fix failures from integration test @marcgarreau 
- [x] Bump `eth-utils` (to account for the `replace_exceptions` change)

#### Cute Animal Picture

IMG_20170706_123758


web3/manager.py
```diff
@@ -150,6 +179,9 @@ def request_blocking(
 
         if "error" in response:
             apply_error_formatters(error_formatters, response)
```
karlb: Does this line have any effect? (This comment is unrelated to the changes in this PR)

karlb:
Is [this CI failure](https://app.circleci.com/jobs/github/ethereum/web3.py/98072) really caused by this PR? I don't see how the change could cause that.

kclowes:
@karlb no, that failure is flaky. Issue is [here](https://github.com/ethereum/web3.py/issues/1580). Feel free to ignore!

karlb:
> I think it would be good to add an integration test if you feel up to it, but I'm happy to add one if you don't.

@kclowes I'll gladly take you up on the offer of adding an integration test. I don't see how to fit it in nicely, right now.

I'll take care of the other change tomorrow.

kclowes:
Sounds good, thank you!

karlb:
I've moved the code to `method_formatters.py` and added a basic test for `get_error_formatters`. Did I get the change done the way you intended?

I'll probably have to fix something when there's a proper integration test, since I didn't test across different eth clients, yet.

kclowes:
@karlb I will keep digging into this tomorrow. I realized I read the original issue wrong and misdirected you towards putting the formatting in `method_formatters`. Based on the original issue, the formatting is correct when it goes through`eth_call`, but not when a function is called via `contract.functions.<function_name>.call()`. So I _think_ we'll need to do something around [here](https://github.com/ethereum/web3.py/blob/331f321835de5b842e0929dc6b416720634e46ee/web3/contract.py#L1433-L1437) to catch the `Revert` and re-raise a `SolidityError`, but I'm not 100% sure that is the right path. [This comment](https://github.com/ethereum/web3.py/issues/941#issuecomment-401422928) provides some direction as well. I will keep thinking about it and if you have any thoughts in the meantime, let me know.

I also don't think we need an integration test since `eth_call` is working as expected. I was able to write a quick contract with a revert function that you can use in the core tests to test the contract call interface. The commit is [here](https://github.com/ethereum/web3.py/commit/050bd27c36573b962d69d903a1723294034d265f), but eth-tester doesn't return the error message (it only returns the error type), so that will need to be addressed as well I think.

karlb:
> Based on the original issue, the formatting is correct when it goes through `eth_call`

It's not accidentally truncated. But I think web3py should detect reverts, parse the revert reason and raise that information as a proper Exception in that case, too. Getting
```python
SolidityError('not allowed to monitor'')
```
is much more useful than
```python
'Reverted 0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000166e6f7420616c6c6f77656420746f206d6f6e69746f7200000000000000000000'
```

> The commit is here, but eth-tester doesn't return the error message (it only returns the error type), so that will need to be addressed as well I think.

Yes, eth-tester is special. IIRC, it does not return errors as the other eth clients do, but directly raises python exceptions on transaction failures. This is usually very helpful, but for this specific PR it is causing additional work, since it is just an additional case to handle.

> I was able to write a quick contract with a revert function that you can use in the core tests to test the contract call interface.

Thanks! What is the easiest way (on a linux machine) to run that test trough geth and parity in addition to eth-tester (which seems to be what's used when I just use default pytest)?

kclowes:
> But I think web3py should detect reverts, parse the revert reason and raise that information as a proper Exception in that case, too.

Yes, I agree!

> it does not return errors as the other eth clients do, but directly raises python exceptions on transaction failures

In an effort to be backend-agnostic it changes errors that are specific to one EVM (like Revert) to a TransactionFailed error. 

> What is the easiest way (on a linux machine) to run that test trough geth and parity?

I think we'd have to do something like [this](https://github.com/ethereum/web3.py/blob/master/tests/generate_go_ethereum_fixture.py#L347) using the revert contract, then add it to the geth fixture at the end of the `setup_chain_state` method. After that, it could be added as a fixture in the respective conftest file (for example: https://github.com/ethereum/web3.py/blob/6ce485f554d13b93f43d9d7b415fbf158c65a77d/tests/integration/parity/conftest.py#L127), and then it could get used in the `module_testing` files. I will probably be able to get to that this week, or if you get to it first, that's great. Let me know if you need any more clarification!

karlb:
One sensible addition to the RevertContract could be a revert without message, so that we can test that case, too.

> In an effort to be backend-agnostic it changes errors that are specific to one EVM (like Revert) to a TransactionFailed error.

I'm now catching the eth-tester exception to create the result I intended. We'll see whether that lines up with the results from other backends.

I didn't find any time to look at setting up geth and parity. If you take care of that, I'll gladly have a look at any new failures.

kclowes:
> One sensible addition to the RevertContract could be a revert without message, so that we can test that case, too.

Good idea. I'll add that too. 

> I didn't find any time to look at setting up geth and parity. If you take care of that, I'll gladly have a look at any new failures.

Sounds good, thank you!

Exef:
Is it going to be merged soon?

If not @karlb, can I continue your work on this PR?

kclowes:
@Exef - @karlb was waiting on me to write some tests and then I lost track of this. Sorry about that! If you want to write some integration tests, please feel free!

karlb:
@Exef @kclowes Is one of you working on integration tests? I want to make sure that we don't accidentally have a deadlock where everyone waits for someone else to act next.

kclowes:
@karlb it's on my list, but I've been working on other things. I should be able to dig back in either later today or tomorrow but I'll let you know if that changes! And if @Exef gets to it first, that's great!

kclowes:
@karlb I just finished up the other PR that I was working on, so I should be able to get to this on Monday. Thanks for your patience!

kclowes:
@karlb I was able to make some progress on this today. I got the geth integration fixture up with the revert contract, and a simple test written, but I haven't had time to figure out why it is failing. I don't yet have the parity tests running either, but I should be able to keep working on it tomorrow. 

I also can't remember how I got code to you last time, so feel free to take or leave the commits I pushed to this branch!

karlb:
> @karlb I was able to make some progress on this today. I got the geth integration fixture up with the revert contract, and a simple test written, but I haven't had time to figure out why it is failing.

Unfortunately, I don't get the same error as in the CI locally:
```
pytest tests/integration/go_ethereum/test_goethereum_http.py::TestGoEthereumEthModuleTest::test_eth_call_revert_with_msg -x --tb=short
...
____________________________________________________________________________________________ TestGoEthereumEthModuleTest.test_eth_call_revert_with_msg[<lambda>] _____________________________________________________________________________________________
web3/_utils/module_testing/eth_module.py:664: in test_eth_call_revert_with_msg
    result = web3.codec.decode_single('bool', call_result)
venv/lib/python3.7/site-packages/eth_abi/codec.py:155: in decode_single
    return decoder(stream)
venv/lib/python3.7/site-packages/eth_abi/decoding.py:127: in __call__
    return self.decode(stream)
venv/lib/python3.7/site-packages/eth_abi/decoding.py:198: in decode
    raw_data = self.read_data_from_stream(stream)
venv/lib/python3.7/site-packages/eth_abi/decoding.py:308: in read_data_from_stream
    len(data),
E   eth_abi.exceptions.InsufficientDataBytes: Tried to read 32 bytes.  Only got 0 bytes
```
Full output: [error.txt](https://github.com/ethereum/web3.py/files/4735115/error.txt)

Apparently I get back `call_result = HexBytes('0x')`. Is my local setup broken, do I run the tests the wrong way, or is something else happening?

> I also can't remember how I got code to you last time, so feel free to take or leave the commits I pushed to this branch!

Please just push everything related into this branch!

web3/manager.py
```diff
@@ -146,10 +150,13 @@ def request_blocking(
         """
         Make a synchronous request using the provider
         """
-        response = self._make_request(method, params)
+        try:
+            response = self._make_request(method, params)
+        except eth_tester.exceptions.TransactionFailed as e:
```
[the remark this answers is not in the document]
pipermerriam: I think in the best place to do this would be inside of the `EthereumTesterProvider` somewhere in here:

https://github.com/ethereum/web3.py/blob/master/web3/providers/eth_tester/main.py#L85-L103

That keeps us from having to have `eth-tester` specific code mixed into the main library code.

marcgarreau:
So I've been stepping through the eth_tester case first. Got stumped for a while before realizing the reason that the revert data never made it through to web3 was this `replace_exceptions` decorator; it swaps the revert out for an (empty) `TransactionFailed` exception:

https://github.com/ethereum/eth-tester/blob/51a208e02205639c37108873ccfddc4d845a0ddd/eth_tester/backends/pyevm/main.py#L488

Before I go yanking it out, is there anything I should know about the context around this exception handling? (It's not your code, but maybe you might know, @carver?)

carver:
> So I've been stepping through the eth_tester case first. Got stumped for a while before realizing the reason that the revert data never made it through to web3 was this `replace_exceptions` decorator; it swaps the revert out for an (empty) `TransactionFailed` exception:
> 
> https://github.com/ethereum/eth-tester/blob/51a208e02205639c37108873ccfddc4d845a0ddd/eth_tester/backends/pyevm/main.py#L488
> 
> Before I go yanking it out, is there anything I should know about the context around this exception handling? (It's not your code, but maybe you might know, @carver?)

Yeah, this all sounds right. The wrapping is because eth-tester wraps both pyethereum and py-evm, so we wanted to unify the exceptions in a way that was not tied to either back end.

I'm not totally sure we're ready to give that abstraction up. Maybe we can insert the wrapped exception's message in the abstract `TransactionFailed` exception?

marcgarreau:
Thanks for the tip, @carver. I think the right place to solve that is in eth-utils. Tiny PR up here: https://github.com/ethereum/eth-utils/pull/198.

marcgarreau:
Quick update: have still been digging around in the weeds here. Have learned that as of June, Geth returns revert messages in the JSON-RPC response ([v1.9.15](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.15)). This has lead to some renewed conversation about which versions we should support in our test suite (https://github.com/ethereum/web3.py/issues/1703). If we only support the ~latest version(s), we may be able to forego some of the low-level decoding going on here. That decision/work may be considered a blocker to this PR.

karlb:
> Have learned that as of June, Geth returns revert messages in the JSON-RPC response ([v1.9.15](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.15)). This has lead to some renewed conversation about which versions we should support in our test suite (#1703). If we only support the ~latest version(s), we may be able to forego some of the low-level decoding going on here.

Won't we need the decoding anyway to support revert messages for the OpenEthereum client?

marcgarreau:
> Won't we need the decoding anyway...?

Yeah, that stray idea was over-optimistic.

As for next steps, `call` (and the other `eth` namespaced methods) aren't getting the error formatters passed through, so the decoding logic isn't reached. Getting this PR over the line may be the next blocker: https://github.com/ethereum/web3.py/pull/1568/.

a0969818431:
Close #1585

marcgarreau:
Got some cleanup to do, but at long last, I think the base functionality is there 🎉 

One stylistic choice to highlight: I applied Geth's convention to format all revert messages as `execution reverted: <reason>`.  I'm not attached to this idea; only priority is a standardizing across Geth, Parity/OE, and eth-tester. The alternatives would be to strip off the prefix from Geth or invent our own.

kclowes:
Awesome!!! I think Geth's convention is good, especially since we're raising the same `SolidityError` across the board. And I like that the revert word is in the message to give an extra clue.

marcgarreau:
Ready for review @kclowes (and anyone else interested). 

- More LOC than is ideal, but most are related to fixture generation with the revert contract and deduping two geth generation files. 
- Feedback especially welcome on the revert reason parsing code. I left most of Karl's original implementation to parse the raw data from Parity/OE. An alternative to that would be to use eth-abi's `decode_single` function, but IMO the readability is similar.

Thanks! 🙏

tests/integration/test_ethereum_tester.py
```diff
@@ -135,6 +135,29 @@ def txn_hash_with_log(block_with_txn_with_log):
...
+def revert_contract(web3, revert_contract_factory, revert_contract_deploy_txn_hash):
+    deploy_receipt = web3.eth.waitForTransactionReceipt(revert_contract_deploy_txn_hash)
+    assert is_dict(deploy_receipt)
+    contract_address = deploy_receipt['contractAddress']
+    assert is_checksum_address(contract_address)
+    return revert_contract_factory(contract_address)
+
+
+@pytest.fixture(scope="module")
+def revert_contract_address(revert_contract, address_conversion_func):
```
kclowes, line 157: I don't see this getting used anywhere?

web3/_utils/method_formatters.py
```diff
@@ -486,6 +487,60 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
 ABI_REQUEST_FORMATTERS = abi_request_formatters(STANDARD_NORMALIZERS, RPC_ABIS)
 
 
+def get_revert_reason(response: RPCResponse) -> str:
```
kclowes, line 490: It feels like this `method_formatters` file is becoming a catch-all, so may be worth thinking about how we can split this out. But not for this PR :)

kclowes (approved):
Hooray! This will be so great to get in! I had mostly nitpicky comments and one about using `assert` in the code.

Also, because I can't comment on the file, I don't think we want to commit `tests/integration/tester.zip` :)

tests/integration/generate_fixtures/common.py
```diff
@@ -34,22 +34,28 @@
     "config": {
         "chainId": 131277322940537,  # the string 'web3py' as an integer
         "homesteadBlock": 0,
+        "byzantiumBlock": 0,
+        "constantinopleBlock": 0,
+        #  "petersburgBlock": 0,
```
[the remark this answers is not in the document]
marcgarreau: I left these in because docs for syncing a parity + geth private net is woefully sparse. So, this is part documentation + part todo list. Does that context change anything or would you prefer that it live in an issue? No biggie either way.

tests/integration/generate_fixtures/common.py
```diff
@@ -34,22 +34,28 @@
...
+        COINBASE: {"balance": "1000000000000000000000000000"},
+        UNLOCKABLE_ACCOUNT: {"balance": "1000000000000000000000000000"},
+        RAW_TXN_ACCOUNT: {"balance": "1000000000000000000000000000"},
+        "0000000000000000000000000000000000000001": {"balance": "1"},
+        "0000000000000000000000000000000000000002": {"balance": "1"},
+        "0000000000000000000000000000000000000003": {"balance": "1"},
+        "0000000000000000000000000000000000000004": {"balance": "1"},
+        "0000000000000000000000000000000000000005": {"balance": "1"},
+        "0000000000000000000000000000000000000006": {"balance": "1"},
+        #  "0000000000000000000000000000000000000007": {"balance": "1"},
```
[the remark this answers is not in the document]
marcgarreau: (will address in the same way as the note above)

tests/integration/generate_fixtures/common.py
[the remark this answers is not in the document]
kclowes: Fair enough. I think it's fine here!

tests/integration/generate_fixtures/go_ethereum.py
```diff
@@ -295,7 +176,7 @@ def verify_chain_state(web3, chain_data):
 def mine_transaction_hash(web3, txn_hash):
     web3.geth.miner.start(1)
     try:
-        return web3.eth.waitForTransactionReceipt(txn_hash, timeout=60)
+        return web3.eth.waitForTransactionReceipt(txn_hash, timeout=120)
```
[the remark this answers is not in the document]
marcgarreau: Leaving it for good, since I regularly require it.

tests/integration/generate_fixtures/common.py
```diff
@@ -213,7 +219,7 @@ def mine_block(web3):
 def mine_transaction_hash(web3, txn_hash):
     start_time = time.time()
     web3.geth.miner.start(1)
-    while time.time() < start_time + 120:
+    while time.time() < start_time + 240:
```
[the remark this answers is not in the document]
marcgarreau: this is prob overkill, will revert.

web3/_utils/method_formatters.py
```diff
@@ -486,6 +487,60 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
...
+        "Reverted "
+        Function selector for Error(string): 08c379a (4 bytes)
+        Data offset: 32 (32 bytes)
+        String length (32 bytes)
+        Reason strong (padded, use string length from above to get meaningful part)
+
+    See also https://solidity.readthedocs.io/en/v0.6.3/control-structures.html#revert
+    """
+    assert 'error' in response
+    if not isinstance(response['error'], dict):
```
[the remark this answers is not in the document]
marcgarreau: Good question; not certain. Local linter is grumpy without it: `Item "str" of "Union[RPCError, str]" has no attribute "get"`. Will poke more.

tests/integration/test_ethereum_tester.py
```diff
@@ -309,6 +332,12 @@ def test_eth_chainId(self, web3):
     def test_eth_getTransactionReceipt_mined(self, web3, block_with_txn, mined_txn_hash):
         super().test_eth_getTransactionReceipt_mined(web3, block_with_txn, mined_txn_hash)
 
+    def test_eth_call_revert_with_msg(self, web3, revert_contract, unlocked_account) -> None:
```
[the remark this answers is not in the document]
marcgarreau: Unsure, don't think that was my doing. Tests pass without this code, so I'll remove.

web3/_utils/method_formatters.py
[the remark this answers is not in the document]
marcgarreau: So, I think web3 raises or prevent exceptions before reaching this point, but the RPCResponse type does include string values. I think leaving this guard clause won't do any harm, at least.

sobolev-igor:
Hi everyone! 
Is there anything blocking this PR?

I'm having an issue connected to this PR (#1781) 
I really hope that this PR will fix the issue and I look forward to this PR being merged!

marcgarreau:
@sobolev-igor just leaving space for any final reviews. Expecting to merge later today or tomorrow, but next release is TBD.

web3/_utils/method_formatters.py
```diff
@@ -486,6 +487,58 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
...
+    # Geth case:
+    if 'message' in response['error'] and response['error'].get('code', '') == 3:
+        return response['error']['message']
+
+    return None
+
+
+def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
+    revert_reason = get_revert_reason(response)
+    if revert_reason is None:
```
njgheorghita, line 529: It seems to me that there is some unavoidable complexity here in dealing with the various clients, but in general return value checking is a bit of a code smell - one that I'm very familiar with cause I violate it constantly. But the [style guide](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/antipatterns.md#return-value-checking) goes into a bit more detail. One idea is to raise an exception in `get_revert_reason` if no reason is found and then handle the cases accordingly here in `raise_solidity_error_on_revert`.

web3/_utils/module_testing/eth_module.py
```diff
@@ -741,6 +742,38 @@ def test_eth_call_with_0_result(
         result = web3.codec.decode_single('uint256', call_result)
         assert result == 0
 
+    def test_eth_call_revert_with_msg(
+        self,
+        web3: "Web3",
+        revert_contract: "Contract",
+        unlocked_account: ChecksumAddress,
+    ) -> None:
+        with pytest.raises(SolidityError, match='execution reverted'):
```
njgheorghita, line 751: Can we match a more specific part of the message here and the next test (like `execution reverted: not allowed to monitor` in `test_method_formatters.py`)?

tests/core/utilities/test_method_formatters.py
```diff
@@ -0,0 +1,65 @@
...
+    "jsonrpc": "2.0",
+    "error": {
+        "code": -32601,
+        "message": "Method not found",
+    },
+    "id": 1,
+})
+
+
+def test_get_revert_reason() -> None:
```
njgheorghita, line 53: This test seems like it's worth [parameterizing](https://docs.pytest.org/en/stable/parametrize.html). Right now - for each successful test run we're testing 3 different case. Each test run should assert/disprove 1 case. 

Also the next test seems like it should be parametrized as well for the two cases that test if a `SolidityError` is raised, and then the `OTHER_ERROR` case split into its own test.

njgheorghita (approved):
Looks good to me! I just left a couple of thoughts on the style here & there, but nothing critical.

a0969818431 (commented):
`defi`

tests/integration/generate_fixtures/go_ethereum.py
```diff
@@ -363,6 +232,36 @@ def setup_chain_state(web3):
...
+    block_hash_revert_with_msg = web3.eth.getBlock(txn_receipt_revert_with_msg['blockHash'])
+    print('BLOCK_HASH_REVERT_WITH_MSG:', block_hash_revert_with_msg['hash'])
+
+    txn_hash_revert_with_no_msg = revert_contract.functions.revertWithoutMessage().transact(
+        {'gas': 320000, 'from': web3.eth.coinbase}
+    )
+    print('TXN_HASH_REVERT_WITH_NO_MSG:', txn_hash_revert_with_no_msg)
+    txn_receipt_revert_with_no_msg = common.mine_transaction_hash(web3, txn_hash_revert_with_no_msg)
+    block_hash_revert_no_msg = web3.eth.getBlock(txn_receipt_revert_with_no_msg['blockHash'])
+    print('BLOCK_HASH_REVERT_NO_MSG:', block_hash_revert_no_msg['hash'])
```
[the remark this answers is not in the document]
marcgarreau: While generated fixtures, it has been nice to see these get printed in real time, because it fails/times out a nontrivial % of the time.

web3/_utils/method_formatters.py
```diff
@@ -486,6 +487,58 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
...
+    # Parity/OpenEthereum case:
+    data = response['error'].get('data', '')
+    if data.startswith('Reverted '):
+        # "Reverted", function selector and offset are always the same for revert errors
+        prefix = 'Reverted 0x08c379a00000000000000000000000000000000000000000000000000000000000000020'  # noqa: 501
+        if not data.startswith(prefix):
+            return ''
+
+        reason_length = int(data[len(prefix):len(prefix) + 64], 16)
+        reason = data[len(prefix) + 64:len(prefix) + 64 + reason_length * 2]
```
pipermerriam, line 517: This all feels fragile.  That might be "ok" but it still worries me.  Any thoughts on just doing the simple check that the data contains `"Reverted "` and then skipping this parsing logic and embedding the whole response message in the revert exception?

web3/_utils/method_formatters.py
```diff
@@ -486,6 +487,58 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
...
+        return f'execution reverted: {bytes.fromhex(reason).decode("utf8")}'
+
+    # Geth case:
+    if 'message' in response['error'] and response['error'].get('code', '') == 3:
+        return response['error']['message']
+
+    return None
```
pipermerriam, line 524: gentle nudge to [avoid patterns that require return value checking](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/antipatterns.md#return-value-checking)

pipermerriam: Also, `mypy` should be yelling at you about this since this function is designated to return a `str` so it is probably worth looking into why those checks aren't being enforced.

web3/_utils/method_formatters.py
marcgarreau: I might be misunderstanding, but I'd definitely like to include the parsed revert message and not put that on the end user's shoulders. Are you suggesting returning both the reason and the full error dict, just the error dict, or something else?

marcgarreau: Current thinking in support of :shipit: : 
- in related [issues filed](https://github.com/ethereum/web3.py/search?q=revert&type=issues), users are generally looking for the plaintext reason.
- handling reverts is broken today, so even brittle parsing logic is an improvement. 
- I don't expect clients to change this formatting on a whim. 
- we can iterate on it.

tests/core/utilities/test_method_formatters.py
```diff
@@ -0,0 +1,78 @@
...
+    (
+        (REVERT_WITH_MSG, 'execution reverted: not allowed to monitor'),
+        (REVERT_WITHOUT_MSG, ''),
+    ),
+    ids=[
+        'test-get-revert-reason-with-msg',
+        'test-get-revert-reason-without-msg',
+    ])
+def test_get_revert_reason(response, expected) -> None:
+    with pytest.raises(SolidityError):
```
kclowes, line 64: This is pretty nitpicky, but that assert will never get hit. I think this would be better like:
```
    with pytest.raises(SolidityError, match=expected):
        raise_if_revert_reason(response)
```

web3/_utils/method_formatters.py
pipermerriam: I'm proposing that we return the string which we are trying to parse in full.  basically this whole context block would become:

```python
if data.startswith("Reverted "):
    reason = data
```

I don't feel strongly about this.  It just seems like it would be roughly the same usefulness to the user and would avoid the part that looks brittle.

marcgarreau: Makes good sense, but it feels like signing up for an endless trickle of "what is this? how do I parse this?" questions. I'd prefer to sign up for patching the logic on the hopefully very rare occasions required and deliver the superior UX. Will flag as this something "to harden" though.

web3/exceptions.py
```diff
@@ -192,3 +192,11 @@ class InvalidEventABI(ValueError):
...
     """
     pass
+
+
+class SolidityError(ValueError):
+    # Inherits from ValueError for backwards compatibility
+    """
+    Raised on a solidity require/revert
+    """
+    pass
```
fubuloubu, line 202: This seems like a breaking change to the API of expected failures for web3. What was the reasoning for intercepting `TransactionFailed` failures and converting to this error (which also seems poorly named, perhaps `AssertionError(TransactionFailed)` would be better considering Solidity isn't the only smart contract language out there that has a assertion feature)

PR #1586, opened 2020-02-24
Upgrade eth-tester requirement

kclowes:
### What was wrong?
~Solidity has a sort-of new error called `revert` that eth-tester currently turns into a `TransactionFailed` error. I have a PR up in eth-tester to stop changing the error, but wanted to do an incremental update first to minimize variables~
Just a plain old eth-tester upgrade for now

Related to Issue #1585

### How was it fixed?
Updated!

### Todo:

#### Cute Animal Picture
image


kclowes:
Fixed in #1636.

PR #1587, opened 2020-02-25
fix flaky ssh integration tests

billbsing:
### What was wrong?

Related to Issue #1580 

### How was it fixed?
Checked that each `ssh.post` is followed by a `time.sleep(1)`

### DIFF
--- a/web3/_utils/module_testing/shh_module.py
+++ b/web3/_utils/module_testing/shh_module.py
@@ -367,6 +367,7 @@ class GoEthereumShhModuleTest():
             "payload": web3.toHex(text="testing shh on web3.py"),
             "pubKey": receiver_pub,
         })
+        time.sleep(1)
 
     def test_shh_post_deprecated(self, web3: "Web3") -> None:
         with pytest.warns(DeprecationWarning):
@@ -378,6 +379,7 @@ class GoEthereumShhModuleTest():
                 "payload": web3.toHex(text="testing shh on web3.py"),
                 "pubKey": receiver_pub,
             })
+        time.sleep(1)
 
     #
     # shh_properties
@@ -598,6 +600,7 @@ class ParityShhModuleTest():
             "priority": 40,
             "ttl": 400,
         })
+        time.sleep(1)
 
     def test_shh_info(self, web3: "Web3") -> None:
         info = web3.parity.shh.info()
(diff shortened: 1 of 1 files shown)


kclowes:
@billbsing after running the integration tests a few times, I think they're still flaky. I ran into this in the Parity integration tests a while back, see https://github.com/ethereum/web3.py/pull/1447. I think the best option is just to mark it as flaky like I did in #1447. Let me know if you disagree or if you don't want/have time to implement and I can pick it up. Thanks!

billbsing:
> @billbsing after running the integration tests a few times, I think they're still flaky. I ran into this in the Parity integration tests a while back, see #1447. I think the best option is just to mark it as flaky like I did in #1447. Let me know if you disagree or if you don't want/have time to implement and I can pick it up. Thanks!

I have added xfail on a ValueError around each test method using a post.

I'm ok to try and find a better solution, if this does not work.

web3/_utils/module_testing/shh_module.py
```diff
@@ -246,6 +267,10 @@ def test_shh_remove_filter_deprecated(self, web3: "Web3") -> None:
             except BaseException:
                 assert True
 
+    # Sometimes the post fails because PoW is too low.
+    # We don't care if an error or a True response comes back,
+    # we only care that we're interfacing correctly with Parity
```
kclowes:
```suggestion
    # we only care that we're interfacing correctly with Geth
```

web3/_utils/module_testing/shh_module.py
```diff
@@ -358,6 +383,10 @@ def test_shh_symmetric_key_pair_from_password(self, web3: "Web3") -> None:
     #
     # shh_post
     #
+    # Sometimes the post fails because PoW is too low.
+    # We don't care if an error or a True response comes back,
+    # we only care that we're interfacing correctly with Parity
```
kclowes:
```suggestion
    # we only care that we're interfacing correctly with Geth
```

web3/_utils/module_testing/shh_module.py
```diff
@@ -367,7 +396,12 @@ def test_shh_post(self, web3: "Web3") -> None:
             "payload": web3.toHex(text="testing shh on web3.py"),
             "pubKey": receiver_pub,
         })
+        time.sleep(1)
 
+    # Sometimes the post fails because PoW is too low.
+    # We don't care if an error or a True response comes back,
+    # we only care that we're interfacing correctly with Parity
```
kclowes:
```suggestion
    # we only care that we're interfacing correctly with Geth
```

billbsing:
This PR will not resolve the flaky shh module errors. 
This is now replaced by PR #1590

PR #1592, opened 2020-03-03
Net module v2

billbsing:
### What was wrong?

Related to Issue #1429 

This is a cleanup of the Net module, after changing to snake case methods/properties

### How was it fixed?
+ Documentation has been changed to look the same as the other modules.
+ Documentation shows the properties instead of method calls
+ Better documentation examples
+ The net module as been moved to `ModuleV2' 
+ The methods are now in _utils folder in the net.py module

### DIFF
--- a/web3/net.py
+++ b/web3/net.py
@@ -1,36 +1,39 @@
 from typing import (
     NoReturn,
 )
-import warnings
 
-from web3._utils.rpc_abi import (
-    RPC,
+from web3._utils.net import (
+    listening,
+    peer_count,
+    peerCount,
+    version,
 )
 from web3.module import (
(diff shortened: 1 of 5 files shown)


web3/net.py
```diff
@@ -1,41 +1,44 @@
...
 )
 
 
-class Net(Module):
+class Net(ModuleV2):
     """
         https://github.com/ethereum/wiki/wiki/JSON-RPC
     """
 
+    __listening = listening
```
kclowes: We typically use the pattern of just one underscore for these.

billbsing: sure no problem

Issue #1560, opened 2019-12-26
Contract event overloading breaks `hasattr`


Arachnid:
* Version: 5.0.0a6
* Python: 3.7
* OS: linux

### What was wrong?
Calling `hasattr` on a contract events object should return `True` or `False` as appropriate. Instead, it raises the following exception:

```
Traceback (most recent call last):
  File "tools/get_names.py", line 101, in <module>
    main(parser.parse_args())
  File "tools/get_names.py", line 89, in main
    if hasattr(registrar.events, 'BidRevealed'):
  File "/home/user/.local/lib/python3.7/site-packages/web3/contract.py", line 200, in __getattr__
    "Are you sure you provided the correct contract abi?"
web3.exceptions.MismatchedABI: ("The event 'BidRevealed' was not found in this contract's abi. ", 'Are you sure you provided the correct contract abi?')
```

### How can it be fixed?

Fix the 'helpful' overloading of `__getattr__` to raise the exception expected by Python (preferable), or implement `__hasattr__`.

pipermerriam:
:+1: to making `hasattr` work as expected.  Should be a pretty thing to fix.  It is possible we have a similar problem in `contract.functions`

Swixx:
@pipermerriam Are we looking to change `__getattr__` to raise an `AttributeError` [here](https://github.com/ethereum/web3.py/blob/cdf35dd44b64de5281900a341726c40e8f86377d/web3/contract.py#L237) and [here](https://github.com/ethereum/web3.py/blob/cdf35dd44b64de5281900a341726c40e8f86377d/web3/contract.py#L242), or do we want to implement `__hasattr__` by checking if some `event_name` exists in `self.__dict__` or `self.__dict__['_events']`?

pipermerriam:
@Swixx sorry I missed your comment when it came through.

I believe we want to change `__getattr_` to raise an `AttributeError` (with the same exception messages).  It would be ideal if `hasattr` worked as expected too so if we need a custom implementation of that it could go alongside this fix.

pipermerriam:
cc @kclowes I there is an argument to be made that this qualifies as a breaking change but also a bugfix.  Anyone with existing code catching these exception would end up with broken code.

We could:

1. tell ourselves it is ok because this is a bugfix (is it?)
2. make a new interum exception class that inherits from both `AttributeError` and the existing exception classes that are raised that would then be removed in `v6`.

[...]

PR #1594, opened 2020-03-03
Issue1560

realjohnward:
### What was wrong?
The hasattr method wasn't working for the ContractEvent object (wasn't returning True or False)
AttributeError wasn't being raised in the ContractEvent's getattr method.

Related to Issue #1560 

### How was it fixed?
Added hasattr overloader to ContractEvent class which returns False if an AttributeError is raised during the getattr call.
Added exception class called ABIEventFunctionNotFound, which inherits from AttributeError and MismatchedABI. 

### Todo:

#### Cute Animal Picture


realjohnward:
No problem @kclowes . Thank you for reviewing this and making adjustments!

PR #1601, opened 2020-03-13
Fix return type of web3.eth.getCode

palango:
### What was wrong?

I just ran into the problem of `eth.getCode` announciong the return type to be `HexStr`, but it really returning `HexBytes`.

```
(Pdb++) blockchain_bytecode = web3.eth.getCode(human_standard_token.address)
(Pdb++) type(blockchain_bytecode)
<class 'hexbytes.main.HexBytes'>
```

### How was it fixed?

Adjusted the return type annotation. A different option would be to fix the type.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/ethpm/_utils/deployments.py
+++ b/ethpm/_utils/deployments.py
@@ -6,9 +6,6 @@ from typing import (
     Tuple,
 )
 
-from eth_typing import (
-    HexStr,
-)
 from eth_utils import (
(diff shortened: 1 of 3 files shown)


palango:
> Thanks @palango! I'll get this in today and shoot for a release on Monday.

Thank you!

PR #1602, opened 2020-03-13
Parity to method class

kclowes:
### What was wrong?
The Parity class needed to move over to use the `Method` class and `ModuleV2` class. 

### How was it fixed?
Started using `Method` and `ModuleV2`

### Todo:
- [x] Think about how to get rid of self.web3 = web3 in Module
- [x] Check that all parity tests have all modules running
- [x] Typing
- [x] Cleanup commits

#### Cute Animal Picture

image

### DIFF
--- a/tests/integration/parity/common.py
+++ b/tests/integration/parity/common.py
@@ -54,38 +54,9 @@ class ParityEthModuleTest(EthModuleTest):
         pytest.xfail('eth_uninstallFilter calls to parity always return true')
         super().test_eth_uninstallFilter(web3)
 
-    def test_eth_replaceTransaction(self, web3, unlocked_account):
-        super().test_eth_replaceTransaction(web3, unlocked_account)
-
     @pytest.mark.xfail(reason='Parity is not setup to auto mine')
     def test_eth_replaceTransaction_already_mined(self, web3, unlocked_account):
         super().test_eth_replaceTransaction_already_mined(web3, unlocked_account)
-
-    def test_eth_replaceTransaction_incorrect_nonce(self, web3, unlocked_account):
-        super().test_eth_replaceTransaction_incorrect_nonce(web3, unlocked_account)
-
-    def test_eth_replaceTransaction_gas_price_too_low(self, web3, unlocked_account):
-        super().test_eth_replaceTransaction_gas_price_too_low(web3, unlocked_account)
-
-    def test_eth_replaceTransaction_gas_price_defaulting_minimum(self, web3, unlocked_account):
-        super().test_eth_replaceTransaction_gas_price_defaulting_minimum(web3, unlocked_account)
-
-    def test_eth_replaceTransaction_gas_price_defaulting_strategy_higher(self,
-                                                                         web3,
-                                                                         unlocked_account):
-        super().test_eth_replaceTransaction_gas_price_defaulting_strategy_higher(
-            web3, unlocked_account
-        )
-
-    def test_eth_replaceTransaction_gas_price_defaulting_strategy_lower(self,
-                                                                        web3,
-                                                                        unlocked_account):
-        super().test_eth_replaceTransaction_gas_price_defaulting_strategy_lower(
-            web3, unlocked_account
-        )
-
-    def test_eth_modifyTransaction(self, web3, unlocked_account):
-        super().test_eth_modifyTransaction(web3, unlocked_account)
 
     @flaky(max_runs=MAX_FLAKY_RUNS)
     def test_eth_getTransactionReceipt_unmined(self, web3, unlocked_account):
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -82,6 +82,7 @@ from web3.datastructures import (
 from web3.exceptions import (
     BlockNotFound,
     ContractLogicError,
+    InvalidParityMode,
     TransactionNotFound,
 )
 from web3.types import (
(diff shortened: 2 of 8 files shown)


web3/_utils/method_formatters.py
```diff
@@ -452,6 +457,16 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
 ABI_REQUEST_FORMATTERS = abi_request_formatters(STANDARD_NORMALIZERS, RPC_ABIS)
 
 
+def raise_invalid_parity_mode(response: RPCResponse) -> NoReturn:
+    error_message = response['error']['message']
```
kclowes, line 461: @njgheorghita I'm getting an error from mypy on this line: `error: Invalid index type "str" for "Union[RPCError, str]"; expected type "Union[int, slice]"` and I can't figure out what's wrong with the line. Do you have any ideas?

njgheorghita: Hmmm, this is a weird one. I'm not sure off the top of my head what's going on.... The types `RPCResponse` and `RPCError` look like they should work just fine with this. Can you try overriding it maybe? with something like `error_message: str = response['error']['message']`

kclowes: hmm, that didn't work either :-/ FWIW, it's green until I add on the `['message']` part

kclowes: And when I do: 
```
def raise_invalid_parity_mode(response: RPCResponse) -> NoReturn:
    error: RPCError = response['error']
    error_message: str = error['message']
    raise InvalidParityMode(error_message)
```
I get: 
`error: Incompatible types in assignment (expression has type "Union[RPCError, str]", variable has type "RPCError")`

njgheorghita: hmmm, ok i think if I understand [this](https://stackoverflow.com/questions/43982085/mypy-error-incompatible-type-despite-using-union) correctly - with `RPCResponse['error']: Union[RPCError, str]` mypy is throwing a bug b/c the case when `RPCResponse['error']` is a string is not supported (i.e. you can't get the `['message']` key of a string). In this code we're assuming that it will always be an `RPCError` type  -  which is not always the case if I'm remembering correctly (since mypy throws errors in  `web3/providers/eth_tester/main.py` with `error: RPCError`.... but i could be completely wrong on that). 

So, here.. if it's safe to assume that parity will always respond with an `RPCError` type for `response['error']` then i'd just go ahead and 
```python
from typing import cast
def raise_invalid_parity_mode(response: RPCResponse) -> NoReturn:
    error = cast(RPCError, response['error'])
    error_message = error['message']
    raise InvalidParityMode(error_message)
```

cburgdorf: >mypy is throwing a bug b/c the case when `RPCResponse['error']` is a string is not supported (i.e. you can't get the `['message']` key of a string)

That is correct. `RPCResponse` explicitly defines `error` as being `Union[RPCError, str]` hence it is now rightfully rejects to access an `message` property as that may lead to a runtime error.

My I suggest a different solution than the cast though. Keep in mind a cast always means "step away, `mypy`, I know what I'm doing, please accept `x` to be of type `SomeType`." Mypy will accept anything you feed it even though this may lead to runtime errors.

So in this case, I believe the safer solution would be one of these:

**1. Assuming `string` might be possible**

if isinstance(response.error, RPCError):
   error_message = response.error.message
elif isinstance(response.error, str):
  error_message = response.error
else:
   raise Exception("Invariant") 

**2. Assuming `string` might not be possible** (to be clear the `RPCResponse` type allows for it to be `str`

if isinstance(response.error, RPCError):
   error_message = response.error.message
else:
   raise Exception("Invariant")

The `isinstance` checks allow mypy to ensure type safety which is generally better than giving up type safety by introducing a `cast`.

kclowes: Aha! That makes sense. Thanks @njgheorghita and @cburgdorf!

tests/integration/parity/common.py
```diff
@@ -42,39 +42,10 @@ def test_eth_uninstallFilter(self, web3):
         pytest.xfail('eth_uninstallFilter calls to parity always return true')
         super().test_eth_uninstallFilter(web3)
 
-    def test_eth_replaceTransaction(self, web3, unlocked_account):
```
marcgarreau, line 45: I'm seeing that the inclusion of these tests produces the `FileNotFoundError: [Errno 2] No such file or directory: 'parity'`. Curious to get some more context around this though.

web3/module.py
```diff
@@ -99,3 +99,5 @@ def __init__(self, web3: "Web3") -> None:
             self.retrieve_caller_fn = retrieve_async_method_call_fn(web3, self)
         else:
             self.retrieve_caller_fn = retrieve_blocking_method_call_fn(web3, self)
+        # TODO - remove once web3 calls are moved to middleware
+        self.web3 = web3
```
marcgarreau, line 103: Context would be appreciated here too :pray:

tests/integration/parity/common.py
kclowes: Hmm, that shouldn't be the case. Are you seeing that locally? If so, it may be because you don't have parity installed?

web3/module.py
kclowes: For some reason, moving the child modules to use ModuleV2 doesn't allow them to attach them correctly. So modules like the `ParityPersonal` and `ParityShh` module  won't work without this here. I'll timebox digging in one more time to see if I can figure it out and let you know what I find!

kclowes:
@marcgarreau I think this is (finally) ready to go!

marcgarreau (approved):

PR #1608, opened 2020-03-20
Some fixes to type definitions

palango:
### What was wrong?

Some type definitions were not a strict as they could be.

### How was it fixed?

Tighten type definitions.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/types.py
+++ b/web3/types.py
@@ -161,7 +161,7 @@ TxData = TypedDict("TxData", {
     "gasPrice": Wei,
     "hash": HexBytes,
     "input": HexStr,
-    "nonce": int,
+    "nonce": Nonce,
     "r": HexBytes,
     "s": HexBytes,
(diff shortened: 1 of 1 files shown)


web3/types.py
```diff
@@ -191,9 +191,9 @@ class LogReceipt(TypedDict):
 # syntax b/c "from" keyword not allowed w/ class construction
 TxReceipt = TypedDict("TxReceipt", {
     "blockHash": HexBytes,
-    "blockNumber": int,
+    "blockNumber": BlockNumber,
     "contractAddress": Optional[ChecksumAddress],
-    "cumulativeGasUsed": int,
+    "cumulativeGasUsed": Wei,
```
pipermerriam: This should still be `int`.  `gasPrice` is a `Wei` value but the actual `gas` values are all just integers.

Issue #1572, opened 2020-01-29
getBlockByHash and getBlockByNumber will return a block number if block is pending


kclowes:
### What was wrong?
As of geth 1.9.11, `eth_getBlockByHash` and `eth_getBlockByNumber` will be returning a block number if the block is pending, instead of `null`. Older versions of geth were returning a number, but as of 1.9.7, geth returns `null` if the block is pending. 

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbyhash

Context:
https://github.com/ethereum/go-ethereum/issues/20587

### How can it be fixed?
Check to make sure that web3 can handle either a null response or a block number. This probably means adding some testing. Will also need to check parity's responses.

tmckenzie51:
Hi. I would like to write tests for this fix (#20616). However, I have a few questions on how to go about this. 

1.  I went through the web3 codebase to see where getBlockByHash/getBlockByNumber/getBlock are called and tested. I came across the test below for pending blocks. Given the different return values in the different geth versions, how is it that the test below is currently passing?  

```
def test_eth_getBlockByNumber_pending(
    self, web3: "Web3", empty_block: BlockData
) -> None:
    current_block_number = web3.eth.blockNumber
    block = web3.eth.getBlock('pending')
    assert block['number'] == current_block_number + 1
```

2. I'm not understanding how merge #20616 fixes the issue. From what I understand based on the changed files in the merge it seems to be returning null for pending blocks. How does this ensure that web3 can handle both null and number return values? 

3. Regarding adding appropriate tests in web3 for this issue and its fix, I've attached the small block of code below. Is there a way to specify the version of geth in the code below? How may I go about testing against different versions of geth? 

```
def test_eth_getBlockByNumber_pending(
    self, web3: "Web3", empty_block: BlockData
) -> None:
    #TODO: SPECIFY VERSION
    current_block_number = web3.eth.blockNumber
    block = web3.eth.getBlock('pending')
    assert block['number'] == None
```

Any help would be appreciated. Thank you!

kclowes:
> Given the different return values in the different geth versions, how is it that the test below is currently passing?

[...]

PR #1613, opened 2020-03-25
tests return values against different geth versions

tmckenzie51:
### What was wrong?

Related to Issue #1572

### How was it fixed?
In order to be compatible with geth versions v1.7.0 to v1.9.12+, tests were added to check that web3 can handle both null and number return values from getBlockByHash and getBlockByNumber. The version return value details are based on the following from [this comment](https://github.com/ethereum/web3.py/issues/1572#issuecomment-603694142) on Issue #1572:

```
>= 1.9.11 : returns number
>= 1.9.7 and < 1.9.11: returns null
< 1.9.7 : returns number
```

### Todo:

#### Cute Animal Picture

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ extras_require = {
         "towncrier>=19.2.0,<20",
         "urllib3",
         "web3>=2.1.0",
-        "wheel"
+        "wheel",
     ],
     'dev': [
         "bumpversion",
@@ -46,7 +46,7 @@ extras_require = {
         "tox>=1.8.0",
         "tqdm>4.32,<5",
         "twine>=1.13,<2",
-        "when-changed>=0.3.0,<0.4"
+        "when-changed>=0.3.0,<0.4",
     ]
 }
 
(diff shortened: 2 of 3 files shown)


tmckenzie51:
@kclowes 

> As of geth 1.9.11, eth_getBlockByHash and eth_getBlockByNumber will be returning a block number if the block is pending, instead of null.

From the issue description, I interpreted this as getBlockByHash is also returning the block number. However, based on this CI test error (https://circleci.com/gh/ethereum/web3.py/101945?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), it seems to be returning hexadecimal when for non-null returning versions. How may I test that the pending block's hash is correct in a similar way to ` assert block['number'] == current_block_number + 1` for getBlockByNumber? 

I'm thinking that I could do something like this:
```
    def test_eth_getBlockByHash_pending(
            self, web3: "Web3", empty_block: BlockData
    ) -> None:
        null_versions = ('1.9.8', '1.9.9', '1.9.10', '1.9.11')
        current_block_number = web3.eth.blockNumber
        block = web3.eth.getBlock('pending')
        for v in null_versions:
            if v in web3.clientVersion:
                assert block['hash'] is None
                return
        assert web3.eth.getBlock(block['hash']) == block['hash']
```
This approach seems a bit circular to me though. Is there a better way to do this, such as by checking that the new pending block hash did not previously exist, or something like this? 
Thank you!

tmckenzie51:
@kclowes It also seems as though the getBlockByNumber_pending tests are passing, so it seems as though the version return value details for getBlockByNumber mentioned above may be correct. 

However, for getBlockByHash,  based on the CI tests such as [this one](https://circleci.com/gh/ethereum/web3.py/101948?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), it seems as though getBlockByhash in versions 1.7.2, 1.8.22, and 1.9.7  returns null rather than a number/hash/non-null value, unlike getBlockByNumber. Is this interpretation of the CI test results correct or is there another reason for this? 

Thank you!

Issue #1463, opened 2019-09-27
Support for Weighted Gas Price Strategy


Jmjin:
### What was wrong?

`construct_time_based_gas_price_strategy` is unable to handle cases when network suddenly requires higher gas prices due to many variables such as network congestion. This has resulted in more transactions to revert.

### How can it be fixed?

Currently `construct_time_based_gas_price_strategy` does not factor in weighting when looking over a sample of X number of blocks. If the calculation is weighted towards more recent blocks, `construct_time_based_gas_price_strategy` should be able to handle similars cases like the one mentioned in `What was wrong`.

tmckenzie51:
Hi. I'm interested in working on this issue. I notice that in the construct_time_based_gas_price strategy function that it takes as a parameter 'sample size'. I understand based on your description, that we would want to weight the price towards more recent blocks. What exactly influences this weight? Is it the price associated with the more recent blocks, or the amount of time associated with the more recent blocks? I'm a bit confused about this and would appreciate any help and/or clarification. Thank you.

Jmjin:
> Hi. I'm interested in working on this issue. I notice that in the construct_time_based_gas_price strategy function that it takes as a parameter 'sample size'. I understand based on your description, that we would want to weight the price towards more recent blocks. What exactly influences this weight? Is it the price associated with the more recent blocks, or the amount of time associated with the more recent blocks? I'm a bit confused about this and would appreciate any help and/or clarification. Thank you.

Hey @tmckenzie51 I believe it's the latter. Weighting should be favored more towards the time associated with the more recent blocks.

ulope:
Is there a reason this has been closed? 
This is still a problem (esp. at the moment with very high and fluctuating gas prices on mainnet).

pipermerriam:
@Jmjin historically the weighted gas price strategy was written at a time when this problem was just starting to show up and certainly was not well understood.  Since then it has become clear (to me at least) that gas pricing is a very subjective area.  Because of this, I'm inclined to focus on ensuring that the gas pricing strategy API is well suited for 3rd party extension and to encourage the community to publish their own gas price strategies for different use cases.  If any of these proves broadly useful we can look into including it in the core set of tools that the web3 library provides.

PR #1614, opened 2020-03-25
Tiffany/gas pricing

tmckenzie51:
### What was wrong?

> construct_time_based_gas_price_strategy is unable to handle cases when network suddenly requires higher gas prices due to many variables such as network congestion. This has resulted in more transactions to revert.

Related to Issue #1463

### How was it fixed?
A recency based weighted average of the block time is calculated. This weighted_avg_block_time is then used to calculate the probabilities required to determine the best gas price. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/gas_strategies/time_based.py
+++ b/web3/gas_strategies/time_based.py
@@ -23,15 +23,19 @@ from hexbytes import (
 )
 
 from web3 import Web3
+from web3._utils.encoding import (
+    to_hex_with_size,
+)
+
 from web3._utils.math import (
     percentile,
 )
+
 from web3.exceptions import (
     InsufficientData,
     ValidationError,
 )
 from web3.types import (
-    BlockNumber,
     GasPriceStrategy,
     TxParams,
     Wei,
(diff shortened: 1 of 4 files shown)


tmckenzie51:
Hi there! @jmjin @kclowes 

I just created this weighted average function that uses the weighted average by recency of the block in the sample, in order to calculate the weighted average block time for the time_based_gas_price_strategy function.  I used n/N (where N is the sample size and n is the number of the block in the sample 1 through to N, with the Nth block being equivalent to the 'latest' block) as a simple weighting function that could capture the idea of weighting more recent blocks more heavily. Do you think that such a weighting function would achieve the goal of more accurately deciding a gas price during a time of sudden network congestion?

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
kclowes: This doesn't need to be added to the codebase :) may want to add it to our gitignore though

tmckenzie51:
> This looks like the right approach to me @tmckenzie51! Thank you! I think this deserves to be it's own strategy, meaning that I think we should leave the simple time based strategy, and have a separate weighted strategy. I'd like to see some more testing around this strategy to make sure it's working as expected, and there is also some white space that needs to be cleaned up. I'll plan on addressing this feedback since I'm sure you are busy with other things now :)

@kclowes That makes sense to have it as a separate strategy. Thanks for pointing that out, and thank you for the feedback and for handling this for me. :)

web3/gas_strategies/time_based.py
```diff
@@ -55,6 +54,27 @@ def _get_avg_block_time(w3: Web3, sample_size: int) -> float:
     return (latest['timestamp'] - oldest['timestamp']) / constrained_sample_size
 
 
+def _get_weighted_avg_block_time(w3: Web3, sample_size: int) -> float:
```
kclowes, line 57: @pipermerriam do you have ideas on how we can test this? I did manually test it and confirm it works, but the eth-tester block times are all the same AFAICT.

kclowes:
@marcgarreau I think this is good to go if you want to take a look and give final approval!

Issue #1469, opened 2019-10-08
how can I set the enlarge the pool_connections and pool_maxsize for urllib3 in web3?


ayeteng:
* Version: 5.2.0
* Python: 3.6
* OS: osx

### What was wrong?
If I submit the transaction in multi-threading environment, I got the error message as below:

WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: localhost

I think it's because of the connection not released until I get the transaction receipt. So, is it possible to set larger the pool_connections and pool_maxsize for urllib3 in web3 to avoid such error?

```
def send_tx(unit):
    while True:
            try:
                tx_hash = self.contract_instance[recorder.sol'].functions.add_value(
                    str(unit['sys_id']).encode('utf-8'),
                    ).transact({'from': self.w3.toChecksumAddress(unit['iot_working_update_account'])})
                return tx_hash
            except Exception as e:
                print("Cannot connect to geth and tx fail. Reason: %s, Retry: %d" % (e, count))

def get_receipt(tx_hash):
    while True:
            try:
                ret_value = self.w3.eth.waitForTransactionReceipt(tx_hash, timeout=600)
                return ret_value
            except Exception as e:
                print("Cannot get the transaction receipt. Reason: %s, Retry: %d" % (e, count))

tx_hash = send_tx(unit)
receipt = get_receipt(tx_hash)

```

kclowes:
Sorry @ayeteng this was raised right at the beginning of Devcon and I missed it. Are you using the HTTPProvider? I'm not aware of a way to increase `pool_connections` or `pool_maxsize` in web3. I believe the urllib3 error is actually being bubbled up from the requests library and it looks like they don't do any requests that are non-blocking. @pipermerriam do you have any ideas here?

ayeteng:
Thanks @kclowes , 
About this issue, because the DEFAULT_POOLSIZE is set to 10 in requests package, I'd like to check if possible to set the pool connection to 100 with following trick in web3 request connection.

import requests
sess = requests.Session()
adapter = requests.adapters.HTTPAdapter(pool_connections=100, pool_maxsize=100)
sess.mount('http://', adapter)
resp = sess.get("/mypage")

EdNoepel:
Web3 seems to assume that kwargs passed to the `request` method (likely just _timeout_) are the ones which should be hashed in the `cache_key`.  I'd argue the following points:
- `HttpProvider` should accept arguments passed to the `HttpAdapter` constructor.
[...]

PR #1620, opened 2020-04-03
Pass kwargs to HTTPAdapter, removing pool size constraints

EdNoepel:
### What was wrong?

Related to Issue #1469 
`HTTPAdapter` connection pool size was not tunable.  As a result, applications which submit transactions (or make on-chain method calls) from multiple threads can easily reach the queue limit, restricting flow of data to an Ethereum node which could otherwise handle it.

### How was it fixed?

Since the `requests` library's `Session` object doesn't allow these parameters to be passed either, `request.py` was modified to explicitly reassign a new `HTTPAdapter` with the desired parameters.  To prevent changing the interface, `kwargs` passed to `HTTPProvider` are categorized based upon whether they are for the _request_ (`timeout` for example), or for the _adapter_ (`pool_maxsize` for example).

One downside with this implementation is that the categorization of arguments must be explicitly maintained with new releases of `requests`.  Discussion of alternate implementations is welcomed.

#### Cute Animal Picture

three baby ducklings

### DIFF
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -145,12 +145,11 @@ def test_web3_auto_infura_with_secret(monkeypatch, caplog, environ_name):
     monkeypatch.setenv(environ_name, 'test')
     monkeypatch.setenv('WEB3_INFURA_API_SECRET', 'secret')
 
-    with pytest.raises(DeprecationWarning):
-        importlib.reload(infura)
+    importlib.reload(infura)
 
-        w3 = infura.w3
-        assert isinstance(w3.provider, HTTPProvider)
-        expected_url = 'https://%s/v3/test' % (infura.INFURA_MAINNET_DOMAIN)
-        expected_auth_value = ('', 'secret')
-        assert getattr(w3.provider, 'endpoint_uri') == expected_url
-        assert w3.provider.get_request_kwargs()['auth'] == expected_auth_value
+    w3 = infura.w3
+    assert isinstance(w3.provider, HTTPProvider)
+    expected_url = 'https://%s/v3/test' % (infura.INFURA_MAINNET_DOMAIN)
+    expected_auth_value = ('', 'secret')
+    assert getattr(w3.provider, 'endpoint_uri') == expected_url
+    assert w3.provider.get_request_kwargs()['auth'] == expected_auth_value
--- a/tests/core/providers/test_auto_provider.py
+++ b/tests/core/providers/test_auto_provider.py
@@ -145,11 +145,12 @@ def test_web3_auto_infura_with_secret(monkeypatch, caplog, environ_name):
     monkeypatch.setenv(environ_name, 'test')
     monkeypatch.setenv('WEB3_INFURA_API_SECRET', 'secret')
 
-    importlib.reload(infura)
+    with pytest.raises(DeprecationWarning):
+        importlib.reload(infura)
 
-    w3 = infura.w3
-    assert isinstance(w3.provider, HTTPProvider)
-    expected_url = 'https://%s/v3/test' % (infura.INFURA_MAINNET_DOMAIN)
-    expected_auth_value = ('', 'secret')
-    assert getattr(w3.provider, 'endpoint_uri') == expected_url
-    assert w3.provider.get_request_kwargs()['auth'] == expected_auth_value
+        w3 = infura.w3
+        assert isinstance(w3.provider, HTTPProvider)
+        expected_url = 'https://%s/v3/test' % (infura.INFURA_MAINNET_DOMAIN)
+        expected_auth_value = ('', 'secret')
+        assert getattr(w3.provider, 'endpoint_uri') == expected_url
+        assert w3.provider.get_request_kwargs()['auth'] == expected_auth_value
--- a/tests/core/utilities/test_request.py
+++ b/tests/core/utilities/test_request.py
@@ -1,0 +1,75 @@
+from web3._utils import request
+from requests import Session, Response
+from requests.adapters import (
+    DEFAULT_POOLSIZE,
+    HTTPAdapter
+)
+
+
+class MockedResponse:
+    def __init__(self, text="", status_code=200):
+        assert (isinstance(text, str))
+        assert (isinstance(status_code, int))
+        self.status_code = status_code
+        self.ok = 200 <= status_code < 400
+        self.text = text
+        self.reason = None
+        self.content = "content"
+
+    def raise_for_status(self):
+        pass
+
+
+uri = "http://mynode.local:8545"
+
+
+def check_adapters_mounted(session: Session):
+    assert isinstance(session, Session)
+    print(session.adapters)
+    assert len(session.adapters) == 2
+    assert isinstance(session.adapters['http://'], HTTPAdapter)
+    assert isinstance(session.adapters['https://'], HTTPAdapter)
+    assert session.adapters['http://'] == session.adapters['https://']
+
+
+def test_make_post_request_no_args(mocker):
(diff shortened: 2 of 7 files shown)


web3/_utils/request.py
```diff
@@ -23,15 +24,34 @@ def _remove_session(key: str, session: requests.Session) -> None:
...
     return response.content
+
+
+def extract_adapter_kwargs(**kwargs: Any) -> Dict[str, Any]:
+    adapter_args = ['pool_connections', 'pool_maxsize', 'max_retries', 'pool_block']
+    return {key: value for key, value in kwargs.items() if key in adapter_args}
+
+
+def extract_request_kwargs(**kwargs: Any) -> Dict[str, Any]:
+    request_args = ['params', 'params', 'headers', 'cookies', 'files',
```
njgheorghita, line 54: duplicate `'params'`

tests/core/utilities/test_request.py
```diff
@@ -0,0 +1,80 @@
...
+    def raise_for_status(self):
+        pass
+
+
+uri = "http://mynode.local:8545"
+
+
+def check_adapters_mounted(session: Session):
+    assert isinstance(session, Session)
+    print(session.adapters)
```
njgheorghita, line 33: it seems like this `print` can be removed

web3/_utils/request.py
```diff
@@ -23,15 +24,34 @@ def _remove_session(key: str, session: requests.Session) -> None:
...
+        adapter = requests.adapters.HTTPAdapter(**extract_adapter_kwargs(**kwargs))
+
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+
+        _session_cache[cache_key] = session
     return _session_cache[cache_key]
 
 
 def make_post_request(endpoint_uri: URI, data: bytes, *args: Any, **kwargs: Any) -> bytes:
```
njgheorghita, line 37: If you find this feature useful, it's probably worth adding a small blurb about it in the docs.

web3/_utils/request.py
```diff
@@ -23,15 +24,34 @@ def _remove_session(key: str, session: requests.Session) -> None:
...
-    response = session.post(endpoint_uri, data=data, *args, **kwargs)  # type: ignore
+    response = session.post(endpoint_uri, data=data, *args,
+                            **extract_request_kwargs(**kwargs))  # type: ignore
     response.raise_for_status()
 
     return response.content
+
+
+def extract_adapter_kwargs(**kwargs: Any) -> Dict[str, Any]:
+    adapter_args = ['pool_connections', 'pool_maxsize', 'max_retries', 'pool_block']
```
njgheorghita, line 49: I'd agree that this isn't the most maintainable design, i'd try playing around with these helper functions to automatically detect the available parameters - similar to [how it's done here](https://stackoverflow.com/questions/55590419/filter-keyword-arguments-when-calling-function).

tests/core/utilities/test_request.py
```diff
@@ -0,0 +1,80 @@
...
+        self.ok = 200 <= status_code < 400
+        self.text = text
+        self.reason = None
+        self.content = "content"
+
+    def raise_for_status(self):
+        pass
+
+
+uri = "http://mynode.local:8545"
```
njgheorghita, line 28: total nitpick: but across our libraries we usually uppercase constants

EdNoepel:
> the _real_ solution to solve the underlying problem of the user having more control over the HTTP connection is to give the user full control

Thank you for your comment.  I'd like to help make this happen.

Currently the session is lazily constructed and cached upon the first `post` for the URI.  Since we don't want to pass a session each time we `post`, I could extend `request.py` to allow  `HttpProvider.init` to cache the session (assuming a session for that URI isn't already cached).

First I'd like to sanity check the motivation for `request.py`.  It seems this exists to share a session across multiple `HttpProviders` pointing to the same endpoint within the same application.  Is this useful in applications which create short-lived instances of `Web3` on-the-fly?  Are there other use cases I should understand?

EdNoepel:
I made the changes as directed, however I'm having second thoughts about deprecating the `kwargs`.  Currently, those args are passed to the `post` request.  Without them, I don't see any way to set timeout (a default cannot be passed to `Session` or `HttpAdapter`).  Perhaps we should retain `kwargs` as arguments passed to the `post`, yet still allow consumers to pass their own session to `HttpProvider`?

pipermerriam:
> Perhaps we should retain kwargs as arguments passed to the post, yet still allow consumers to pass their own session to HttpProvider?

This seems to make sense.  I did a quick scan through the requests docs and it doesn't appear there is a way to provide a timeout outside of passing it in as a kwarg to the request method.

EdNoepel:
Great; changes are applied and documentation has been updated.

web3/providers/rpc.py
```diff
@@ -49,12 +50,19 @@ class HTTPProvider(JSONBaseProvider):
...
             self.endpoint_uri = get_default_endpoint()
         else:
             self.endpoint_uri = endpoint_uri
+
+        if request_kwargs:
+            raise DeprecationWarning("request_kwargs will not be supported in future versions.")
         self._request_kwargs = request_kwargs or {}
+
+        if session:
+            cache_session(self.endpoint_uri, session)
```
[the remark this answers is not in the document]
pipermerriam: Not proposing we should make this change in this PR, just making a note here for my own purposes.

tests/core/providers/test_auto_provider.py
```diff
@@ -145,11 +145,12 @@ def test_web3_auto_infura_with_secret(monkeypatch, caplog, environ_name):
     monkeypatch.setenv(environ_name, 'test')
     monkeypatch.setenv('WEB3_INFURA_API_SECRET', 'secret')
 
-    importlib.reload(infura)
+    with pytest.raises(DeprecationWarning):
```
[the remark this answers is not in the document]
EdNoepel: Per previous discussion, I'm no longer deprecating the `kwargs` because they are still needed for params passed to each `post` (`timeout`, specifically).

marcgarreau:
@EdNoepel can you rebase onto master? that should address the failing docs test

EdNoepel:
> It looks like the signature for `HTTPProvider` needs to be updated to include the `session` argument.

Not sure I understand; the `__init__` method includes `session: Any=None` as the last argument.  Do you also want `session` stored in a class variable?  I didn't see reason to do so, since it's merely passed to `cache_session()`.

EdNoepel:
> @EdNoepel can you rebase onto master? that should address the failing docs test

Yes, and it did.  Thank you.

pipermerriam:
@EdNoepel sorry, my previous comment was lacking any context.

I was referring to the documentation.  The documentation does not currently include the `session` argument.

EdNoepel:
> I was referring to the documentation. The documentation does not currently include the `session` argument.

I documented the `session` argument in `providers.rst` line 217, and provided an example on line 241.  Is there another place I should update the documentation?  If so, please provide guidance.

pipermerriam:
@EdNoepel something is broken with your git history in this branch.  If you look at the diff you'll see a large number of changes that aren't related to this PR.  Can you rebase this onto the current `master` to clean that up?

pipermerriam:
Regarding the documentation.  Look around like 207 in `docs/providers.rst` where it documents the `HTTPProvider` API.  It appears to be missing the `session` argument there.

EdNoepel:
That makes sense now.  I updated the signature in the documentation.  Also did another rebase and force-pushed to my fork, and that seems to have cleaned up the git history.

EdNoepel:
I just rebased to resolve the conflict from https://github.com/ethereum/web3.py/commit/aa30b3bf97d67b8434cb399c659d8b6425a9bec1 .  What's the status on this PR?

PR #1622, opened 2020-04-08
Upgrade eth account to >= 0.5.2

kclowes:
### What was wrong?
Want to upgrade eth-account to >=0.5.0 for HD Account features. 

### How was it fixed?
Upgraded `eth-account`, but ran into a dependency resolution problem when running `pip install -e .[dev]` because `eth-keys` >v0.3 isn't compatible with the current released versions of `eth-tester` and `py-evm`. Both `py-evm` and `eth-tester` have the `eth-key` >v0.3 support in their master branches, but they aren't released. We'll need to release `py-evm` first and then upgrade the `py-evm` dependency in `eth-tester` (and make sure everything works okay) before the `eth-keys` pinning can be removed. 

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,9 @@ from setuptools import (
 
 extras_require = {
     'tester': [
+        # TODO - remove this eth-keys pinning once eth-tester >0.4.0-beta.1 is released
+        # and py-evm v0.3.0-alpha.15 is released
+        "eth-keys>=0.2.1,<0.3",
         "eth-tester[py-evm]==v0.2.0-beta.2",
         "py-geth>=2.2.0,<3",
     ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -7,6 +7,9 @@
 
 extras_require = {
     'tester': [
+        # TODO - remove this eth-keys pinning once eth-tester >0.4.0-beta.1 is released
+        # and py-evm v0.3.0-alpha.15 is released
+        "eth-keys>=0.2.1,<0.3",
```
carver, line 12: Eth-tester v0.4.0-beta.2 looks like it has what we need here (and it's released).
So with a change to the line below, it looks like we can drop this eth-keys version constraint ^.

Issue #1566, opened 2020-01-13
Support Solidity v0.6.x


kclowes:
* Version: 5.x

### What was wrong?

Placeholder PR to check web3 support for Solidity v0.6. I'll fill in what needs to be done when I dig in, but if anyone knows something that needs to be changed, feel free to comment on this issue. 

https://solidity.readthedocs.io/en/v0.6.1/060-breaking-changes.html

PR #1565 has already begun to address the new `receive` keyword.

mswilkison:
Do you all have an estimate as to when v0.6 support will land in web3.py? Thanks!

kclowes:
@mswilkison it's not on the immediate docket, but knowing there is interest helps. It looks like supporting `receive()` is the feature you need? Anything else that would be handy?

vzotova:
@kclowes thank you, only `receive()`, other features that we use are already working with latest version of web3.py

kclowes:
@mswilkison (or anyone else) On a quick glance, I think we can close this? Are there other v0.6.x features we still need to support?

vzotova:
I think it can be closed, thank you. We are using 0.7.x for now and everything works good

PR #1623, opened 2020-04-09
Solidity receive

kclowes:
### What was wrong?
web3 didn't support Solidity's new `receive` keyword. 

Broken docs build has been fixed in #1622 

Related to Issue #1565, #1566 

### How was it fixed?
Added a new contract API: `contract.receive`, that basically mirrors the fallback function. 

### Todo:

#### Cute Animal Picture

image


web3/_utils/contracts.py
```diff
@@ -265,6 +272,16 @@ def get_fallback_function_info(
     return fn_abi, fn_selector, fn_arguments
 
 
+def get_receive_function_info(
+    contract_abi: ABI=None, fn_abi: ABIFunction=None
+) -> Tuple[ABIFunction, HexStr, Sequence[Any]]:
```
pipermerriam, line 277: We have a general convention of *strict* typing for return values and *generic* typing for inputs.

- *generic* types for inputs
    - `Sequence`, `Collection`, `Iterable`
- *concrete* types for return values
    - `List`, `Tuple`, etc

Unless I'm missing something, I believe that the `Sequence` here should be `Tuple[Any, ...]` since it is a return value.

web3/contract.py
```diff
@@ -808,14 +826,23 @@ def __getattr__(self, attr: Any) -> Callable[[], None]:
         return NonExistentFallbackFunction._raise_exception
 
 
+class NonExistentReceiveFunction:
+    @staticmethod
+    def _raise_exception() -> NoReturn:
+        raise FallbackNotFound("No receive function was found in the contract ABI.")
+
+    def __getattr__(self, attr: Any) -> Callable[[], None]:
+        return NonExistentReceiveFunction._raise_exception
```
pipermerriam, line 835: Does `return self._raise_exception` not work?

web3/contract.py
```diff
@@ -808,14 +826,23 @@ def __getattr__(self, attr: Any) -> Callable[[], None]:
...
+
 class ContractFunction:
     """Base class for contract functions
 
     A function accessed via the api contract.functions.myMethod(*args, **kwargs)
     is a subclass of this class.
     """
     address: ChecksumAddress = None
-    function_identifier: Union[str, Type[FallbackFn]] = None
+    function_identifier: Union[str, Type[FallbackFn], Type[ReceiveFn]] = None
```
pipermerriam, line 845: You might consider aliasing this:

```python
# web3/typing.py

FunctionIdentifier = Union[str, Type[FallbackFn], Type[ReceiveFn]]
```

Then you can reuse it across the codebase.

kclowes: Good idea!

web3/contract.py
kclowes: It does. I was just following the pattern from `NonExistentFallbackFunction` above. Will change 👍

PR #1624, opened 2020-04-13
add transaction options to docs

### LINKED ISSUE #1610: Better documentation for signing and sending transactions
### What was wrong?

There are three different ways to sign and send transactions: 
1. a user can use an unlocked account and then `w3.eth.sendTransaction`, 
2. Building a transaction, manually sign, then send with w3.eth.sendRawTransaction.
3. The `sign_and_send_middleware`. 

It would be nice to detail the steps to use each one and list reasons for using it over the others

marcgarreau:
### What was wrong?
There's a few ways to send transactions with web3.py, which may be confusing for newcomers in particular.

Closes #1610 

### How was it fixed?
Added some notes about when to use each method/middleware in the Examples page:

Screen Shot 2020-04-13 at 1 25 18 PM

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -705,6 +705,9 @@ Taking the following contract code as an example:
     ValidationError:
     Could not identify the intended function with name `setBytes2Value`
 
+
+.. _contract-functions:
+
 Contract Functions
 ------------------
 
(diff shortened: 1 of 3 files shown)


kclowes:
I just merged the doc fix and then you should be able to re-run the tests to fix that integration test. I haven't seen that particular flake before, so probably not worth fixing right at the moment, but we can keep an eye on it and fix when it gets annoying.

PR #1630, opened 2020-04-24
Fix key remapping for eth-tester client

pipermerriam:
### What was wrong?

- The key for the receipts root is wrong when using the `EthereumTesterProvider`
- The key for the bloom filter was also wrong
- The standard formatter for `logsBloom` wasn't setup to correctly handle the integer format used by eth-tester.

### How was it fixed?

- Fixed the typo on the key remappings for `logsBloom` and `receiptRoot`
- Adjusted formatter for `logsBloom` to correctly handle integer values that aren't the full bit-length

#### Cute Animal Picture

cats-dogs-and-human-furniture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -453,7 +453,7 @@ The following methods are available on the ``web3.eth`` namespace.
             'nonce': '0x3b05c6d5524209f1',
             'number': 2000000,
             'parentHash': '0x57ebf07eb9ed1137d41447020a25e51d30a0c272b5896571499c82c33ecb7288',
-            'receiptRoot': '0x84aea4a7aad5c5899bd5cfc7f309cc379009d30179316a2a7baa4a2ea4a438ac',
+            'receiptsRoot': '0x84aea4a7aad5c5899bd5cfc7f309cc379009d30179316a2a7baa4a2ea4a438ac',
             'sha3Uncles': '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347',
             'size': 650,
             'stateRoot': '0x96dbad955b166f5119793815c36f11ffa909859bbfeb64b735cca37cbf10bef1',
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -2889,6 +2889,8 @@ class EthModuleTest:
     def test_eth_getBlockByHash(self, w3: "Web3", empty_block: BlockData) -> None:
         block = w3.eth.get_block(empty_block["hash"])
         assert block["hash"] == empty_block["hash"]
+        assert block["receiptsRoot"] == empty_block["receiptsRoot"]
+        assert block["logsBloom"] == empty_block["logsBloom"]
 
     def test_eth_getBlockByHash_not_found(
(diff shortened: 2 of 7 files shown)


pipermerriam:
I'll deal with the failures here and ping you when this is actually ready for review @marcgarreau

marcgarreau:
Beside the point, but I'm interested in more context around `receiptsRoot` if you have it, e.g., are you using it for anything in particular? With the exception of this value, the getBlock payload is identical between web3py and [web3js](https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#id42).

pipermerriam:
I'm not actually using it.  I just ran into these small key/naming discrepancies while working on an app that consumes all the block data and loads it into a relational database.  I didn't go as far as actually using the `receiptsRoot` to verify the receipts but if you're question is related to how the value is constructed I can point you at the right place in the py-evm code.

web3/_utils/module_testing/eth_module.py
```diff
@@ -765,6 +767,7 @@ def test_eth_getTransactionReceipt_mined(
         assert is_dict(receipt)
         assert receipt['blockNumber'] == block_with_txn['number']
         assert receipt['blockHash'] == block_with_txn['hash']
+        assert receipt['receiptsRoot'] == block_with_txn['receiptsRoot']
```
marcgarreau, line 770: looks like [`eth_getTransactionReceipt`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt) returns either `root` or `status`, but not `receiptsRoot`. I expected `receipt['root'] == block_with_txn['receiptsRoot'].hex()`, but that's apparently not the case ¯\_(ツ)_/¯

docs/web3.eth.rst
```diff
@@ -348,7 +348,7 @@ The following methods are available on the ``web3.eth`` namespace.
             'nonce': '0x3b05c6d5524209f1',
             'number': 2000000,
             'parentHash': '0x57ebf07eb9ed1137d41447020a25e51d30a0c272b5896571499c82c33ecb7288',
-            'receiptRoot': '0x84aea4a7aad5c5899bd5cfc7f309cc379009d30179316a2a7baa4a2ea4a438ac',
+            'receiptsRoot': '0x84aea4a7aad5c5899bd5cfc7f309cc379009d30179316a2a7baa4a2ea4a438ac',
```
pipermerriam, line 351: I believe this is technically a breaking change, but it is also a bugfix...  I'm inclined to just highlight this well in the release notes.  Alternative would be to duplicate this value into both versions of the key (plural and singular) and then remove the singular one in v6...

kclowes: I'd probably lean towards keeping both, but I don't have a strong preference either way. Maybe we could note that in the newsfragment?

kclowes:
It looks like this is ready to be merged after a newsfragment gets added. Want me to pick that up @pipermerriam?

fselmo (approved):
This seems like a low-hanging fruit to scoop into `v6`. I can rebase, add a newsfragment, and merge it if this looks good to you @kclowes 👀.

kclowes:
Yep, sounds good to me! Thanks @fselmo!

kclowes (approved):
LGTM! :shipit:

PR #1636, opened 2020-05-01
Upgrade eth-tester to v0.4.0-beta.2

kclowes:
### What was wrong?
With pip's lack of dependency resolution, we had an incompatible eth-keys version that was raising errors in a fresh `pip install -e .[dev]` install, so it had to be pinned at `"eth-keys>=0.2.1,<0.3"`. This is fixed in the v0.4.0-beta2 release of eth-tester. 

Related to issue #1622 

### How was it fixed?
Upgraded eth-tester

### Todo:
- [x] Get rid of the first commit. 

#### Cute Animal Picture

image

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -694,8 +694,8 @@ Taking the following contract code as an example:
 
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
-    >>> array_contract.functions.setBytes2Value([b'a']).transact()
-    HexBytes('0x39bc9a0bf5b8ec8e8115ccb20bf02f5570351a20a8fd774da91353f38535bec1')
+    >>> tx_hash = array_contract.functions.setBytes2Value([b'a']).transact({'gas': 420000, 'gasPrice': 21000})
+    >>> assert tx_hash == b'\x89\xf9\xb3\xa0\x06Q\xe4\x06\xc5h\xe8\\\x1d#6\xc6kN\xc4\x0b\xa8,^rro\xbd\x07"0\xa4\x1c', tx_hash
     >>> array_contract.functions.getBytes2Value().call()
     [b'a\x00']
     >>> w3.enable_strict_bytes_type_checking()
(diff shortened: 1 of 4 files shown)


docs/contracts.rst
```diff
@@ -695,7 +695,7 @@ Taking the following contract code as an example:
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
     >>> array_contract.functions.setBytes2Value([b'a']).transact()
-    HexBytes('0x39bc9a0bf5b8ec8e8115ccb20bf02f5570351a20a8fd774da91353f38535bec1')
+    HexBytes('0x977bd66e0ef59433a02c2f71f61ec3ca48bf7a5a19d10aece84570061fe9d4f3')
```
carver, line 698: Hm, I don't follow, why did this change?

kclowes: Good question. I didn't think too hard about it changing. Would the different VM rules (istanbul vs. constantinople) do it?

carver: Hm, the only thing I can think of is that maybe the estimated gas usage changed? I would definitely want to make sure that's the only thing that changed before merging, though. Maybe by getting the transaction in the old version and new version. Could even add that to the doc test, like
```py
>>> transaction_hash = array_contract.functions.setBytes2Value([b'a']).transact()
>>> w3.eth.getTransaction(transaction_hash)
{
  value: 0,
  gas_price: 1000000000,
  // etc
}
```
Then it should be really obvious if the gas price is the only thing that changes. We could potentially skip the gas cost in the test with `...` to be a bit more flexible without losing much.

carver: So one option is to add that ^ doctest change as a separate PR without the eth-tester upgrade. Then rebase this PR on top, to see how it changes cleanly.

(or if you're feeling a little more adventurous with local rebasing, you could push it as a separate commit in this PR, but need to push it alone to trigger CI and let it finish, before adding the eth-tester upgrade)

kclowes: Yep, looks like it was gas/gasPrice changing in the tx_params. I printed out the result from `w3.eth.getTransaction` and took the higher gas and gasPrice and rounded up then added it to the transaction params. Let me know if you'd like me to do something different/see something else!

docs/contracts.rst
```diff
@@ -694,8 +694,9 @@ Taking the following contract code as an example:
 
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
-    >>> array_contract.functions.setBytes2Value([b'a']).transact()
-    HexBytes('0x977bd66e0ef59433a02c2f71f61ec3ca48bf7a5a19d10aece84570061fe9d4f3')
+    >>> tx_hash = array_contract.functions.setBytes2Value([b'a']).transact({'gasPrice': 200000})
+    >>> tx_dict = w3.eth.getTransaction(tx_hash)
+    >>> assert tx_hash == tx_dict.hash
```
carver, line 699: This assertion looks a little superfluous, it just checks if `getTransaction()` isn't broken I guess?

docs/contracts.rst
```diff
@@ -694,8 +694,9 @@ Taking the following contract code as an example:
 
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
-    >>> array_contract.functions.setBytes2Value([b'a']).transact()
-    HexBytes('0x977bd66e0ef59433a02c2f71f61ec3ca48bf7a5a19d10aece84570061fe9d4f3')
+    >>> tx_hash = array_contract.functions.setBytes2Value([b'a']).transact({'gasPrice': 200000})
```
carver, line 697: Setting gas price shouldn't be necessary here, right? I'm guessing we'd just need to set the gas, and then check the hash is a specific value, like you did with `my_token_contract.constructor()` below.

docs/contracts.rst
kclowes: Oops, you're right. This was a remnant from debugging. Thanks!

PR #1652, opened 2020-05-15
Ethpm v3

njgheorghita:
### What was wrong?
The `ethpm` module and `web3.pm` needed updating to support the [latest ethpm version](http://ethpm.github.io/ethpm-spec/v3-package-spec.html). The most significant changes to an ethpm schema include....
- snake_case -> camelCase
- `package_name` -> `name`
- `manifest_version` -> `manifest`
- `compilers` was added as a new top-level field
- `name` and `version` are no longer strictly required fields

### How was it fixed?
We decided to make a hard update from v2 -> v3 (as opposed to deprecating and gradually phasing in v3 support) for reasons mentioned in a comment below.

- Updated v2 schema to v3 and all examples
- Updated code & tests to support new manifests/schema
- Submoduled [ethpm-spec](https://github.com/ethpm/ethpm-spec) which allowed us to remove a ton of redundant assets

### Todo:
- [x] Add flag / check for v2 packages to announce deprecated support
- [x] Update web3 docs
- [x] Update `docs.ethpm.com`
- [x] Update `ethpm/tools`
- [x] Update `web3.pm`
- [x] Submodule `ethpm-spec`
- [x] Remove outdated assets

#### Cute Animal Picture

image


njgheorghita:
@pipermerriam Yup, we landed on this a while back b/c it's behind the `enable_unstable_pm_api` flag, there's low V2 adoption, and it helps simplify the update vs. supporting both v2 & v3 simultaneously.

njgheorghita:
@pipermerriam @marcgarreau Okey dokey, this turned into a doozy of a pr. But most of the changes are fairly cosmetic, and the majority of files changed are simply removing redundant assets. I'm happy to hop on a call and walk-through the changes. If anybody is uncomfortable with merging this branch directly into master, I'm definitely open to suggestions / alternative approaches.

web3/tools/pytest_ethereum/linker.py
```diff
@@ -100,10 +100,11 @@ def link(contract: ContractName, linked_type: str, package: Package) -> Package:
             "so it is not a valid contract type for link()"
         )
     linked_factory = unlinked_factory.link_bytecode({linked_type: deployment_address})
+    # ???
```
marcgarreau, line 103: ???

tests/ethpm/validation/test_manifest.py
```diff
@@ -109,13 +91,11 @@ def test_extract_contract_types_from_deployments(data, expected):
...
-@pytest.mark.parametrize("version", ("2"))
-def test_validate_manifest_version_validates_version_two_string(version):
-    validate = validate_manifest_version(version)
-    assert validate is None
+def test_validate_manifest_version_validates_version_three_string():
+    assert validate_manifest_version("ethpm/3") is None
 
 
-@pytest.mark.parametrize("version", (1, 2, "1" "3", b"3"))
+@pytest.mark.parametrize("version", (2, 3, "2" "3", b"3"))
```
marcgarreau, line 98: Is `ethpm/3` the only acceptable answer here?

marcgarreau: And should `"2"` and `"3"` be separated by a comma?

ethpm/backends/ipfs.py
```diff
@@ -154,7 +154,7 @@ def base_uri(self) -> str:
 MANIFEST_URIS = {
     "ipfs://QmVu9zuza5mkJwwcFdh2SXBugm1oSgZVuEKkph9XLsbUwg": "standard-token",
```
pipermerriam, line 155: I'm curious why these didn't change?

tests/ethpm/validation/test_manifest.py
njgheorghita: > Is ethpm/3 the only acceptable answer here?
Yup, that's the only valid version string for an ethpm v3 manifest.

ethpm/backends/ipfs.py
njgheorghita: Ahh good catch, well it's because this is just a small cache for the `DummyIPFSBackend` and `DummyIPFSBackend.fetch_uri_contents()` simply fetches the local asset rather than fetching it from IPFS, so the hash isn't really used in any meaningful way. But, they're all updated now.

PR #1653, opened 2020-05-18
permit pathlib.Path type as IPCProvider input

### LINKED ISSUE #1647: IPCProvider argument type bug
### What was wrong?

IPCProvider correctly handles a `pathlib.Path` as its `ipc_path` argument, but the type annotation says it only accepts `str`.

### How can it be fixed?

Change it to `Union[str, Path]`

marcgarreau:
### What was wrong?
"IPCProvider correctly handles a pathlib.Path as its ipc_path argument, but the type annotation says it only accepts str." - @carver 

### How was it fixed?
`str` -> `Union[str, Path]`
Closes #1647

### Todo:

### DIFF
--- a/newsfragments/1647.bugfix.rst
+++ b/newsfragments/1647.bugfix.rst
@@ -1,0 +1,1 @@
+IPCProvider correctly handled Path input, but warned against its type. Fixed to permit Path objects in addition to strings.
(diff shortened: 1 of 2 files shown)


newsfragments/1647.bugfix.rst
```diff
@@ -0,0 +1 @@
+IPCProvider correctly handled Path input, but warned against its type. Fixed to permit Path objects in addition to strings.
```
pipermerriam, line 1:
```suggestion
IPCProvider correctly handled ``pathlib.Path`` input, but warned against its type. Fixed to permit Path objects in addition to strings.
```

PR #1660, opened 2020-05-26
fixed error when accessing blocks by blocknumber using eth_tester

MatthiasLohr:
### What was wrong?

When accessing a block by its block number using eth_tester, e.g. 3, the following error occured:
```
Block number must be a positive integer or one of the strings 'latest', 'earliest', or 'pending'. Got: 0x3
```

### How was it fixed?

With this pull request, the ethereum_tester_middleware is extended by typing instructions for `eth_getBalance`.

#### Cute Animal Picture

Too tired for searching for a cat picture. Meow.

### DIFF
--- a/newsfragments/1660.bugfix.rst
+++ b/newsfragments/1660.bugfix.rst
@@ -1,1 +1,1 @@
-- Added formatter rules for eth_tester middleware to allow getBalance by using integer block numbers
+Added formatter rules for eth_tester middleware to allow :meth:`~web3.eth.Eth.getBalance` by using integer block numbers
--- a/newsfragments/1660.bugfix.rst
+++ b/newsfragments/1660.bugfix.rst
@@ -1,0 +1,1 @@
+- Added formatter rules for eth_tester middleware to allow getBalance by using integer block numbers
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -126,6 +126,17 @@ class EthModuleTest:
         assert is_integer(balance)
         assert balance >= 0
 
+    def test_eth_getBalance_with_block_identifier(self, web3: "Web3") -> None:
+        coinbase = web3.eth.coinbase
+
+        with pytest.raises(InvalidAddress):
+            web3.eth.getBalance(ChecksumAddress(HexAddress(HexStr(coinbase.lower()))))
+
+        balance = web3.eth.getBalance(coinbase, 0)
+
+        assert is_integer(balance)
+        assert balance >= 0
+
     def test_eth_getStorageAt(
         self, web3: "Web3", emitter_contract_address: ChecksumAddress
     ) -> None:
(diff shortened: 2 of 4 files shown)


carver:
For reference:
https://github.com/ethereum/web3.py/blob/6258d2e1c807aee892481296c7f115399ce6a6f8/web3/_utils/module_testing/eth_module.py#L205-L210

Something like that, but with balance. I'd try to go one step further and say that you can verify that the balances are different on different blocks.

Notice that coinbase has a balance at the latest block:
https://github.com/ethereum/web3.py/blob/6258d2e1c807aee892481296c7f115399ce6a6f8/web3/_utils/module_testing/eth_module.py#L124-L127

It seems a good bet that that balance is lower at the genesis block. See if you can verify that in a test. I'd highly recommend reverting the implementation commit, then writing the test and confirming that it fails. (In `tox -e py36-integration-ethtester`)

marcgarreau:
@carver the genesis balance vs. later balance idea is a no-go in at least one case:
```
(Pdb++) web3.eth.getBalance(coinbase)
1000000000000000000000000
(Pdb++) web3.eth.blockNumber
0
```
however, testing of `Literal(0)` is enough to demonstrate that without the fix, an error is thrown: 
```
ValidationError: Block number must be a positive integer or one of the strings 'latest', 'earliest', or 'pending'.  Got: 0x0
```

carver:
> genesis balance vs. later balance idea is a no-go in at least one case:

Sure, you would need to mine a block before that idea would work.

newsfragments/1660.bugfix.rst
```diff
@@ -0,0 +1 @@
+- Added formatter rules for eth_tester middleware to allow getBalance by using integer block numbers
```
carver, line 1: Oh the `-` isn't necessary here -- towncrier adds that on its own. You only need to add one if you have *sub*-bullets for a particular issue. It might be nice to link to :meth:`~web3.Eth.getBalance` here. Do the docs mention being able to use block number? If not, adding to the docs would be good to :+1:

marcgarreau: I believe its clear enough that a block identifier [can be used](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.getBalance) here. There's a lot of contextual hints about block identifiers in the docs, but I'm realizing that some more formal definition could be helpful, with the options for specifying them. Noted in my notebook; I'll look for an opportunity to add it if one doesn't already exist.

newsfragments/1660.bugfix.rst
```diff
@@ -1 +1 @@
-- Added formatter rules for eth_tester middleware to allow getBalance by using integer block numbers
\ No newline at end of file
+Added formatter rules for eth_tester middleware to allow :meth:`~web3.eth.Eth.getBalance` by using integer block numbers
```
carver, line 1: FWIW I often make mistakes on the exact ReST syntax. Definitely something worth watching for during release time. I like to build the docs locally to see if the links work (after running `make notes`, but before `make release`).

PR #1666, opened 2020-06-03
add ExtraDataLengthError

iamdefinitelyahuman:
### What was wrong?
When a POA network returns >32 bytes of `extraData` a `ValidationError` is raised.  Because this exception is also used for other errors, it is difficult to specifically target and handle the issue.

Originally discussed on [Gitter](https://gitter.im/ethereum/web3.py?at=5ed7723d225dc25f54c2b38c).

### How was it fixed?
 To more effectively handle this specific situation, I had added a new exception `ExtraDataLengthError` which is only raised in this case.  The new exception is a subclass of `ValidationError` so this shouldn't break any existing code.

#### Cute Animal Picture
image

### DIFF
--- a/newsfragments/1666.feature.rst
+++ b/newsfragments/1666.feature.rst
@@ -1,0 +1,1 @@
+Introduce a more specific validation error, ``ExtraDataLengthError``. This enables tools to detect when someone may be connected to a POA network, for example, and provide a smoother developer experience.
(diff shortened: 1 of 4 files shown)


marcgarreau:
Thanks for the PR @iamdefinitelyahuman. Just needs a newsfragment, which I'll add then merge in.

PR #1670, opened 2020-06-12
Use Optional for typing on contract.py

Uxio0:
### What was wrong?

Mypy complaining about using `None` in functions where typing was set to non `None` values.

### How was it fixed?
- `Optional[type]` is used for nullable arguments
- Fixed formatting of default arguments when using type `argument: type = default`, with spaces around `=`

### Todo:
- Fix another files

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1668.bugfix.rst
+++ b/newsfragments/1668.bugfix.rst
@@ -1,1 +1,1 @@
-Fix type annotations within the ``eth.py`` module. Several arguments that defaulted to ``None`` were not declared ``Optional``.
+Fix type annotations within the ``eth.py`` and ``contract.py`` modules. Several arguments that defaulted to ``None`` were not declared ``Optional``.
--- a/newsfragments/1668.bugfix.rst
+++ b/newsfragments/1668.bugfix.rst
@@ -1,1 +1,1 @@
-Fix type annotations within the ``eth.py`` and ``contract.py`` modules. Several arguments that defaulted to ``None`` were not declared ``Optional``.
+Fix type annotations within the ``eth.py`` module. Several arguments that defaulted to ``None`` were not declared ``Optional``.
--- a/ens/main.py
+++ b/ens/main.py
@@ -292,7 +292,8 @@ class ENS:
             self._claim_ownership(new_owner, unowned, owned, super_owner, transact=transact)
             return new_owner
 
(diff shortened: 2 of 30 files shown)


Uxio0:
> This looks like it is taking the type hints for the library in the direction of using [`--no-implicit-optional`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional) which I support. I think you need to update the `mypy.ini` to include this option and to remove the disabling of strict optional types here:

Hi @pipermerriam , I have no strong opinion here, but in Python explicit better than implicit, so I lean towards your approach. This is just a follow up on my other PR (https://github.com/ethereum/web3.py/pull/1668) as it was accepted and merged. I think some modifications still need to be done before updating `mypy.ini`

newsfragments/1668.bugfix.rst
```diff
@@ -1 +1 @@
-Fix type annotations within the ``eth.py`` module. Several arguments that defaulted to ``None`` were not declared ``Optional``.
+Fix type annotations within the ``eth.py`` and ``contract.py`` modules. Several arguments that defaulted to ``None`` were not declared ``Optional``.
```
marcgarreau, line 1: Let's get a separate newsfragment going, just because it'll automagically link back to this PR.

pipermerriam:
@Uxio0 my suggestion about adding `--no-implicit-optional` was that it *should* enforce the changes you are already making in our CI environment which will tell us whether *all* of the places have been properly updated.

newsfragments/1668.bugfix.rst
Uxio0: Ok!

Uxio0:
> @Uxio0 my suggestion about adding `--no-implicit-optional` was that it _should_ enforce the changes you are already making in our CI environment which will tell us whether _all_ of the places have been properly updated.

I understand, but there's still a lot of changes to do and I think this is something to do step by step on multiple PRs

Uxio0:
> Thanks for the mass update. Another freshly merged PR created a couple small conflicts; this looks good once those are resolved and a separate newsfragment is written. I can take it from here if you like, though.

I fixed the newsfragment and rebased the new commit, everything should be alright now

PR #1671, opened 2020-06-12
Add context to index.rst and docs categories

marcgarreau:
### What was wrong?

The index page of the docs is (naturally) the most visited page in the documentation, but it does little for our cause as an unstructured table of contents dump.

### How was it fixed?

Added some introductory text, next step links, and broke the pages into three categories.

This content is a starting point. Opinions desired. 

### Todo:

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,32 +1,63 @@
-Web3.py
-=======
+Introduction
+============
 
-Web3.py is a python library for interacting with Ethereum.  Its API is derived
-from the `Web3.js`_ Javascript API and should be familiar to anyone who has
-used ``web3.js``.
+Web3.py is a Python library for interacting with Ethereum.
 
+It's commonly found in `decentralized apps (dapps)`_ to help with
+sending transactions, interacting with smart contracts, reading
+block data, and a variety of other use cases.
 
-Contents
---------
+The original API was derived from the `Web3.js`_ Javascript API,
+but has since evolved toward the needs and creature comforts of
+Python developers.
+
+Getting Started
+---------------
+
+Your next steps depend on where you're standing:
+
+- Unfamiliar with Ethereum? `ethereum.org`_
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -1,43 +1,70 @@
-Web3.py
-=======
+Introduction
+============
 
-Web3.py is a python library for interacting with Ethereum.  Its API is derived
-from the `Web3.js`_ Javascript API and should be familiar to anyone who has
-used ``web3.js``.
+Web3.py is a python library for interacting with Ethereum.
```
cburgdorf, line 4: nit: I think it should be Python, with capital P

docs/index.rst
```diff
@@ -1,43 +1,70 @@
...
-Contents
---------
+The original API was derived from the `Web3.js`_ Javascript API,
+but has since evolved toward the needs and creature comforts of
+Python developers.
+
+Getting Started
+---------------
+
+Your next steps depend on where you're standing:
```
cburgdorf, line 17: Skimming over the web3.py docs I noticed that the docs are using the *second person "you" form* a lot so I'm not suggesting to change it here. I just wanted to point out that the majority of our other projects use the *first-person "we"* form. There's no right or wrong, it's just a matter of picking one and staying consistent. I just wanted to point it out and draw your attention to our little guide line on docs: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/documentation.md#narrative-perspective

But then again, I'm not suggestion to change the narrative perspective for web3.py, I just wanted to raise awareness that this is slightly different then e.g. what Trinity uses.

marcgarreau: Thanks for highlighting this. I like to keep a casual tone, but I'm probably guilty of some inconsistency, so I'll keep an eye on it. Indeed the rest of the docs (or at least the few pages I just scanned) utilize the "you" form, so seems reasonable to continue the trend. If we decide to change course at some point, I won't fight it too hard ;)

PR #1676, opened 2020-06-22
docs: getBalance example and denomination context

marcgarreau:
### What was wrong?
While thinking more about new user onboarding, I often circle back to a tutorial. The examples page is roughly that already, so what I have in mind is just to connect more dots between each example. Will tackle that in small chunks for ease of review.

### How was it fixed?
- Introduced `getBalance` example
- Added more context to `fromWei` and `toWei` examples

### Todo:

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -41,6 +41,7 @@ representation.  Block numbers
         'transactionsRoot': '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421',
         'uncles': [],
     }
+
     # get a block by it's hash
     >>> web3.eth.getBlock('0x767c2bfb3bdee3f78676c1285cd757bcd5d8c272cef2eb30d9733800a78c0b6d')
     {...}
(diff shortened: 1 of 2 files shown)


docs/examples.rst
```diff
@@ -122,15 +139,36 @@ Web3 can help you convert between denominations.  The following denominations ar
...
+use case, so it's necessary to cast the value to a
+`Decimal <https://docs.python.org/3/library/decimal.html>`_ if it isn't already.
+
+.. code-block:: python
+
+    >>> from decimal import Decimal
+    >>> web3.toWei(Decimal('3841357.360894980500000001'), 'ether')
+    3841357360894980500000001
+
+Best practice: If you need to work with multiple currency denominations, default
```
cburgdorf, line 162: I would probably put that into a `.. note:` but no strong feelings.

marcgarreau: I like the idea, but I'll push back, because I'd want to include the code sample that follows. No other `Note`s in the docs include a code snippet and I think it'd make for an overly large aside.

PR #1682, opened 2020-07-09
Fix buggy support of ethpm-spec submodule

njgheorghita:
### What was wrong?
For developer setup / to run the `ethpm` tests successfully, developers will have to initialize the `ethpm-spec/` submodule.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/ethpm/__init__.py
+++ b/ethpm/__init__.py
@@ -3,7 +3,18 @@ from pathlib import Path
 
 ETHPM_DIR = Path(__file__).parent
 ASSETS_DIR = ETHPM_DIR / "assets"
-ETHPM_SPEC_DIR: Path = ETHPM_DIR / "ethpm-spec"
+
+
+def get_ethpm_spec_dir() -> Path:
+    ethpm_spec_dir = ETHPM_DIR / "ethpm-spec"
+    v3_spec = ethpm_spec_dir / "spec" / "v3.spec.json"
+    if not v3_spec.is_file():
+        raise FileNotFoundError(
+            "The ethpm-spec submodule is not available. "
+            "Please import the submodule with `git submodule update --init`"
+        )
+    return ethpm_spec_dir
+
 
 from .package import Package  # noqa: F401
 from .backends.registry import RegistryURI  # noqa: F401
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,3 +7,5 @@ recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
 
 recursive-include ethpm/assets/ *
+recursive-include ethpm/ethpm-spec/examples *.json
+recursive-include ethpm/ethpm-spec/spec *.json
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Read more in the [documentation on ReadTheDocs](http://web3py.readthedocs.io/). 
 ## Developer Setup
 
 ```sh
-git clone git@github.com:ethereum/web3.py.git
+git clone --recursive git@github.com:ethereum/web3.py.git
 cd web3.py
 ```
 
--- a/docs/ethpm.rst
+++ b/docs/ethpm.rst
@@ -35,11 +35,12 @@ new ``Package`` class for a given package.
 
 .. doctest::
 
-   >>> from ethpm import Package, ETHPM_SPEC_DIR
+   >>> from ethpm import Package, get_ethpm_spec_dir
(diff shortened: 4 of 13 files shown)


marcgarreau:
Just checked out Trinity for a reference:
Screen Shot 2020-07-09 at 4 12 47 PM

Actually, pretty sure the Py-EVM references are a copy pasta bug? But if the recursive flag works to the same effect, might be nice to keep that consistency across libs.

njgheorghita:
@marcgarreau Good call on the `--recursive` flag, seems to work great on my end and removes an extra command. I also think it would be a good idea to put a check in the respective tests like @pipermerriam suggested, since it's possible users are just regular cloning / not directly copying the command. I'll write up a check in the morning for the ethpm tests that will tell users what to do if they don't have the submodule available.

njgheorghita:
@marcgarreau Ok, I feel much better about this now. 
- I updated the clone command in the README to have the `--recursive` flag.
- If users clone this repo without the `--recursive` flag, and use a function that requires an `ethpm-spec` asset, there's now a check that will tell them how to import the files if the files are unavailable. I also moved this down from the module-level, so if users don't have the submodule but don't need it for whatever they're doing, then it won't bug them to update their submodule.
- I added two lines to `MANIFEST.in` that ensure the necessary `ethpm-spec` json assets are included in the builds. (I built the dist locally, and tested importing it into another project and the assets are there, and everything's working nicely)

ethpm/__init__.py
```diff
@@ -3,7 +3,18 @@
...
-ETHPM_SPEC_DIR: Path = ETHPM_DIR / "ethpm-spec"
+
+
+def get_ethpm_spec_dir() -> Path:
+    ethpm_spec_dir = ETHPM_DIR / "ethpm-spec"
+    v3_spec = ethpm_spec_dir / "spec" / "v3.spec.json"
+    if not v3_spec.is_file():
+        raise FileNotFoundError(
+            "The ethpm-spec submodule is not available. "
+            "Please import the submodule with `git submodule update --init`"
```
pipermerriam, line 14: I think this error message should be expanded since there are two main paths that it seems like a user would find themselves in this situation.

1. `pip install web3` with a broken package that doesn't include the assets
2. `get clone ...` that doesn't pull the submodules.

Maybe lets expand this error message to include sections for both of those.  In the case where the problem is `get clone` based, the user has agency to fix it.  In the `pip install ...` case, they have no agency to fix the problem because it is likely indicative of a broken release....  

The current exception messaging would be confusing in the `pip install `case since it would not actually fix the user's problem.

Alternatively, we could allow use of an environment variable to point at a different spec directory than the default one... but I'm skeptical that is the right approach because it would only serve as a way to make broken web3.py releases work, when we really should be focusing on not issuing broken releases....

njgheorghita: @pipermerriam My assumption was that adding the [spec json assets in `MANIFEST.in`](https://github.com/ethereum/web3.py/pull/1682/files#diff-97c91a104c431d0c365565d3ac03ac13R10) (as far as I understand) guarantees that the spec files will always be included in the distribution, so I wasn't really considering case #1 to be a probable scenario. That being said, things can always go wrong I guess, but it just seemed unlikely enough that it doesn't warrant its own message. But, if you feel like this is a case we should guard against, I can definitely update the warning strategy.

marcgarreau: _queues up `beta.2`_

pipermerriam: What about someone who does `pip install GIT-HUB_URL`?  I suspect that would put us in that position as well?

njgheorghita: I just tested a fresh install with `pip install git+https://github.com/ethereum/web3.py/` and the `ethpm-spec` submodule assets are all included in that install. This should mean all the install stories are covered (at least the ones I'm familiar with), but if we want to go with a better safe than sorry approach here, I can update the warning.

pipermerriam: Interesting.  I would not have expected the submodules assets to make their way in.  In fact, I'm surprised enough by that to be suspicious (but not enough to dig in myself).

If you are confident that there are not any obvious cases where a user would find themselves in this position, I don't think the warning/error-messaging change is necessary.

njgheorghita: I was also curious, [these docs] seemed the most helpul to me, but I was unable to glean any relevant/meaningful insight from those docs.

PR #1683, opened 2020-07-13
Update example docs with latest ethpm v3 packages

njgheorghita:
### What was wrong?
`examples.rst` docs needed an update to reflect the latest ethpm v3 upgrade.

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/docs/ethpm.rst
+++ b/docs/ethpm.rst
@@ -46,6 +46,9 @@ new ``Package`` class for a given package.
    >>> OwnedPackage = Package.from_file(owned_manifest_path, w3)
    >>> assert isinstance(OwnedPackage, Package)
 
+For a closer look at how to interact with EthPM packages using web3, check out the
+:ref:`examples page <ethpm_example>`.
+
 Properties
 ~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


docs/ethpm.rst
```diff
@@ -46,6 +46,9 @@ new ``Package`` class for a given package.
    >>> OwnedPackage = Package.from_file(owned_manifest_path, w3)
    >>> assert isinstance(OwnedPackage, Package)
 
+For a closer look at how to interact with EthPM packages using web3, check out the
+`examples page <https://web3py.readthedocs.io/en/stable/examples.html#working-with-contracts-via-ethpm>`__.
```
marcgarreau, line 50: nit: I've been following an existing pattern linking to references instead of full URLs. For example, create an reference (terminology?) above the example with, e.g., `.. _ethpm_example:`, then here you can link to it like 
```
...check out the :ref:`examples page <ethpm_example>`
```
This way the link won't break if the name changes or it moves around.

njgheorghita: Nice, yeah that's a good call. I use that style for same-page references, but wasn't sure about cross-page refs. But, I tested it locally and it works just as you typed out, thanks!

PR #1684, opened 2020-07-14
Update deploy example in docs

### LINKED ISSUE #1679: deploy() is used anywhere in the stable docs, and it was removed
As a sample, I link this documentation pages:
https://web3py.readthedocs.io/en/stable/examples.html#working-with-contracts

marcgarreau:
### What was wrong?

A docs example contained a stale (deprecated) `deploy()` reference. I ran through the example locally and found I was unable to get off the ground without some other cleanup too. 

Closes #1679 

### How was it fixed?

- replaced `deploy()` with `constructor().transact()`
- added PyEVMBackend. Without it, you get `ValueError: RPC Endpoint has not been implemented: eth_estimateGas`.
- latest solidity requires `emit` keyword to trigger events
- added `py-solc-x` for latest solidity compilation
- other cleanup: formatting, etc. (some done by `black`)

### Todo:

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -278,7 +278,7 @@ Given the following solidity source file stored at ``contract.sol``.
 
         function setVar(uint8 _var) public {
             _myVar = _var;
-            MyEvent(_var);
+            emit MyEvent(_var);
         }
 
         function getVar() public view returns (uint8) {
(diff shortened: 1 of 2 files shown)


pipermerriam:
This would be something I would expect to be caught by doctest.  I would recommend doing some investigating to 1) enable doctest on these examples if it isn't enabled or 2) figure out why doctest isn't catching these if it is already enabled.

marcgarreau:
Will check into bolstering these examples via doctests next. Shipping this update to get the next beta release out :shipit:

PR #1686, opened 2020-07-14
Include ethpm example solidity files in distribution

njgheorghita:
### What was wrong?
Ethpm example solidity contracts were not included in distribution. 

### How was it fixed?
Added `ethpm/examples *.sol` to `MANIFEST.in`

### Todo:

#### Cute Animal Picture
image

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,5 +7,5 @@ recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
 
 recursive-include ethpm/assets/ *
-recursive-include ethpm/ethpm-spec/examples *.json
-recursive-include ethpm/ethpm-spec/spec *.json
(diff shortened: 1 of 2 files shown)


njgheorghita:
@marcgarreau So there are no "mission-critical" bugs from what I have tested with the latest release (aka all the ethpm / web3.pm tooling works lovely), since it turns out that none of that tooling depends on the source solidity contracts included in the `ethpm-spec` examples. In another project (`ethpm-cli`) I do import those contracts directly from `ethpm-spec`, and they were missing from the latest release (aside, it's not imperative that I have the raw source contracts in `ethpm-cli` but it's easier than parsing the ethpm manifest for the source). I've updated `MANIFEST.in` to include the solidity contracts in the distribution. IMO this probably does not warrant another beta release to validate, but that's just an opinion and I'll happily run another beta release through the ringer.

PR #1687, opened 2020-07-15
Unbundle release notes generation from release command

marcgarreau:
### What was wrong?
The `make release` command compiles and publishes release notes as part of the process. Being able to manually update the auto-generated notes before releasing is useful in the case of moving from `beta` to `stable`, for example.

### How was it fixed?
Borrows the scripts from Trinity's [Makefile](https://github.com/ethereum/trinity/blob/master/Makefile#L63-L84).

### Todo:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,17 +50,24 @@ validate-docs: build-docs
 linux-docs: build-docs
 	readlink -f docs/_build/html/index.html
 
-release: clean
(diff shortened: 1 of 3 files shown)


marcgarreau:
👍 on the todo list is bringing contribution/maintenance docs into RTD, like Trinity. But, for now, README tweak incoming.

PR #1690, opened 2020-07-16
adopt trinity's newsfragment validation

marcgarreau:
### What was wrong?
Updated the `make docs` and `make release` commands to line up with Trinity's, but didn't realize the `/newsfragments/validate_files.py` file also diverged. 

`./newsfragments/validate_files.py is-empty` is called in the `make release` command, but the `is-empty` arg goes unhandled in web3 at the moment.

### How was it fixed?
Borrowed Trinity's `validate_files.py`. 

### Todo:

### DIFF
--- a/newsfragments/validate_files.py
+++ b/newsfragments/validate_files.py
@@ -12,6 +12,7 @@ ALLOWED_EXTENSIONS = {
     '.doc.rst',
     '.feature.rst',
     '.misc.rst',
+    '.removal.rst',
 }
 
 ALLOWED_FILES = {
--- a/newsfragments/validate_files.py
+++ b/newsfragments/validate_files.py
@@ -5,12 +5,12 @@
 
 import os
 import pathlib
+import sys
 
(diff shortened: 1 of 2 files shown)


newsfragments/validate_files.py
```diff
@@ -5,12 +5,12 @@
 
 import os
 import pathlib
+import sys
 
 ALLOWED_EXTENSIONS = {
-    '.feature.rst',
     '.bugfix.rst',
     '.doc.rst',
-    '.removal.rst',
```
carver, line 13: Maybe this is meant to be for backwards incompatible changes? It probably does make sense to have its own category for that, but we don't need to figure it out right now. :+1: for leaving it out.

marcgarreau: Ah, didn't even realize `removal.rst` was in the existing extentions - thought that came over from Trinity. Wasn't trying to make a statement there, so I'll add it back in.

Issue #1302, opened 2019-03-26
Future of cloud-based auto-providers


pipermerriam:
### What was wrong?

- Recently Infura changed it's API to require an API key.
- There are [new cloud based JSON-RPC providers](https://github.com/ethereum/web3.py/pull/1300) showing up.

The maintenance overhead of this API is not something I'd like the core development team to end up spending their time on, **but** there is value in having a way for users to *quickly* connect to a reliable cloud based provider.

### How can it be fixed?

I see two main options.

1. Use something like setuptools entry_points to allow 3rd party providers to hook into the `web3` namespace so that a user would install `web3-infura` and in doing so, be able to import the web3 providers from `web3.auto.infura.*`
2. Just relegate all of the cloud ones to 3rd party packages.

Downsides I see with the entry points based approach is that it makes the imports *look* like they come from the main web3 library when actually you're importing 3rd party code.  This has the potential to mask bad actors, or worse, have malicious code insert itself masquerading as "infura" but actually doing something malicious.  This is a **major** downside.

Because of this I'm inclined to go with option two and to deprecate our existing `web3.auto.infura` imports and move them into our own `web3-infura` package that we maintain.  We could even go as far as providing a template for these repos to make it easier for people like nodesmith to manage their own packages.

Last, we could have a curated list of these packages listed in the docs to make them more discoverable.

marcgarreau:
Found my way here after a PR was opened to add documentation for another provider here: https://github.com/ethereum/web3.py/pull/1692.

I like the idea of deprecating native Infura-specific APIs in favor of a more even playing field. Would be lovely to handle all providers via generic env vars, e.g., `WEB3_REMOTE_NODE_URL`, `...SCHEMA`, `...KEY`, but may be unrealistic.

pipermerriam:
I think it is reasonable to define a URI scheme for providers that would allow something like `Web3.from_provider_uri(...)` however I think this is still different from the `auto` style imports which are supposed to *just work*.

I'm still in favor of splitting off and releasing a `web3-infura` library that can serve as a template for other hosted providers who wish to provide similar ease-of-onboarding along with listing any other libraries like `web3-quicknode` if they choose to make it in the documentation.

marcgarreau:
Closing since the decision has been made to move to 3rd party packages. Newer issue here: https://github.com/ethereum/web3.py/issues/1736.

PR #1692, opened 2020-07-22
Add quik node as web3 provider in docs

Sahilsen:
### What was wrong?
QuikNode provides ETH nodes, which allows users to connect to blockchain in a hassle free manner(eliminating a need to host a node locally).

### How was it fixed?
Added QuikNode Support in the docs

### Todo:

#### Cute Animal Picture

slothsarefar

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -161,6 +161,10 @@ To setup QuikNode, then set the following environment variable::
 
     $ export WEB3_PROVIDER_URI=https://young-early-dawn.quiknode.pro/<your-token-here>/
     
+Now, let's check if this actaully worked, first we'll import the web3 instance(we'll add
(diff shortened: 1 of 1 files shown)


marcgarreau:
I think this is a nod to a conversation that was not had: https://github.com/ethereum/web3.py/issues/1302.

My understanding is that Infura has special status in the library, only because it was the first of its kind. I like @pipermerriam's suggestion to deprecate third-party specific code in web3.py in favor of auxiliary packages. If/when that happens, we can include QuikNode in the list of supported packages and put all remote nodes on equal footing.

Thanks for the effort, but I'm declining this change set, because it doesn't scale well. It's unwieldy to add similar language for every competitor in this space. Infura's "advantage" here should be a short-term one.

Issue #1704, opened 2020-08-11
Events cannot retreive their ABI if multiple events have the same name


Pet3ris:
* Version: 5.12.0
* Python: 3.7
* OS: osx

### What was wrong?

`_get_event_abi()` doesn't work with overloaded events (same name, but different argument counts).

From [here](https://github.com/ethereum/web3.py/blob/aa30b3bf97d67b8434cb399c659d8b6425a9bec1/web3/contract.py#L1125), we can see that `_get_event_abi()` calls `find_matching_event_abi()`, which has the problem of not working if two events of the same name are found.

In particular, consider the following set of overloaded events (found in the Compound `Comptroller` contract for example):

```
/// @notice Emitted when an action is paused globally
event ActionPaused(string action, bool pauseState);

/// @notice Emitted when an action is paused on a market
event ActionPaused(CToken cToken, string action, bool pauseState);
```

### How can it be fixed?

Expand `find_matching_event_abi` to find the appropriate event through its signature, not through the event name.

PR #1705, opened 2020-08-11
Support overloaded events when finding event ABI

Pet3ris:
### What was wrong?

Related to Issue https://github.com/ethereum/web3.py/issues/1704.

### How was it fixed?

Make sure event arguments are used to disambiguate events (in addition to names) when overloaded events are present.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1122,13 +1122,18 @@ class ContractEvent:
         else:
             self.argument_names = argument_names
 
-        self.abi = self._get_event_abi()
+        self.abi = self._get_event_abi(argument_names)
 
     @classmethod
-    def _get_event_abi(cls) -> ABIEvent:
+    def _get_event_abi(cls, argument_names: Optional[Tuple[str]]=None) -> ABIEvent:
+        if argument_names is None:
+            return find_matching_event_abi(
+                cls.contract_abi,
+                event_name=cls.event_name)
         return find_matching_event_abi(
             cls.contract_abi,
-            event_name=cls.event_name)
+            event_name=cls.event_name,
+            argument_names=argument_names)
 
     @combomethod
     def processReceipt(


marcgarreau:
Thanks for the PR! If we don't have a test that includes overloaded events, we ought to include one. If you're up for taking that on, great, otherwise I can take over.

Pet3ris:
@marcgarreau please do - this was a very incremental change without broader context!

Pet3ris:
I don't think I fully solved the underlying root cause. One observation:

- For an event that's generated from factory `self.abi` can sometimes be `None` and calling `self.get_event_abi()` directly cannot be used as a work-around as there is now way to pass in the arguments

marcgarreau:
Yeah, there's a bit more to the story here. In doing some quick tests, looks like the effects can be seen through more APIs. For example, `contract_instance.events.OverloadedEvent.getLogs()` returns a `ValueError: Multiple events found`. Will dig in deeper tomorrow.

pipermerriam:
A brief scan of our code that handles contract events suggests we have no handling in place for multiple events with the same name.

one example here: https://github.com/ethereum/web3.py/blob/b9d7677e2488b59dd8733ce194f09ab912e0c3e7/web3/contract.py#L231-L244

It will take a bit of clever problem solving to improve these APIs to support this, but it should be largely doable.

1. If we have two similarly named events we should probably expose a mechanism that would allow direct access by `topic` if the topic is known.
2. If the user is creating a filter, and the filter contains any filters against the arguments (indexed or not) we typically be capable of determining which of the duplicate named events the filter is intended for by matching against argument types.
3. In the case that there are duplicate named events and we **cannot** determine which event is intended which is definitely going to be an unavoidable case, we need to be clever and construct the filter such that it matches all of the event topics for the various overloaded events.

We already do similar things for functions, but there we have the luxury of always having the arguments and thus only rarely can we not differentiate between functions with duplicate names.  But I think with events we have the advantage of not actually ending up in a situation where we cannot do what the user wants, it just might end up containing extra logs that they don't want since we'll over-match on the filter parameters since we'll have to provide multiple topics.

Pet3ris:
FWIW, my specific need (that this PR doesn't solve yet) was to get access to an overloaded event's ABI. Seems like neither `<event>.abi` nor `<event>._get_event_abi()` works at this stage but I'm not 100% this is an intended part of the `web3.py` API. As you both indicate, there's lots of other positive implications related to fixing overloaded events.

ptrcarta:
Is there a suggested workaround? I'm incurring in this when trying to query the ActionPaused events of [Compound Comptroller](https://etherscan.io/address/0x3d9819210a31b4961b30ef54be2aed79b9c9cd3b), which is overloaded

dthinkr:
Here is a further fix of the issue. Change `find_matching_event_abi` to avoid it from throwing an error. Instead, let it return the first element. 

`    if len(event_abi_candidates) >= 1:
        return event_abi_candidates[0] 
`
After the fix, I was able to run this with no problem. 

`            events = contract.events.File(*argument_names).getLogs(fromBlock=start_block, toBlock=end_block)
`

BZAghalarov:
I think in the condition of ` if len(event_abi_candidates) >= 1`  returning the first element is not good idea.  @pacrob , @reedsa I am ready to implement final decision (waiting from you) and cover with tests. Could I take a look to this issue?

reedsa:
@BZAghalarov thank you for your work on this. A solution is now implemented for all events, functions and the caller API in #3491. Closing this PR and we'll get those updates released very soon!

Issue #1703, opened 2020-08-10
Update Geth Version Support


kclowes:
This can probably be split up into two related issues, but since the Geth team doesn't support any old versions, we can at least drop support for Geth versions prior to the most recent hard fork. Dropping support includes: 
- removing logic around the old versions in integration tests 
- updating the integration test fixture to the most recent Geth and then fixing any broken tests
   - We will probably also need to add the most recent Geth version(s) to py-geth
- Making sure we have all breaking changes accounted for since 1.9.7

marcgarreau:
Geth version was bumped to 1.9.19 in this PR: https://github.com/ethereum/web3.py/pull/1708. A couple patch releases have come out since, but the purpose of this issue has been achieved. (Bumping to the latest version will cause all Whisper-related tests to fail. This has been noted in the v6 breaking changes list.)

PR #1709, opened 2020-08-17
Remove old geth versions

kclowes:
### What was wrong?
We decided we didn't need to support Geth versions past the most recent hard fork, since the Geth team doesn't support anything beyond the latest version. 

Related to Issue #1703 

### How was it fixed?
Removed all old Geth tests and fixtures.

### Todo:

#### Cute Animal Picture

image


.gitignore
```diff
@@ -121,3 +121,6 @@ fabric.properties
 
 # pip wheel metadata
 pip-wheel-metadata/
+
+# jupyter notebook files
+*.ipynb
```
marcgarreau, line 126: :pray:

setup.py
```diff
@@ -73,7 +73,7 @@
         "eth-account>=0.5.2,<0.6.0",
         "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
         "eth-typing>=2.0.0,<3.0.0",
-        "eth-utils>=1.8.4,<2.0.0",
+        "eth-utils>=1.8.4,<1.9.1",
```
marcgarreau, line 76: FYI: the revert message work will require a later eth-utils version. Without it, EthereumTesterProvider won't receive revert messages. Can be figured out later.

web3/_utils/module_testing/shh_module.py
```diff
@@ -393,10 +393,7 @@ def test_shh_post_deprecated(self, web3: "Web3") -> None:
     #
     def test_shh_version(self, web3: "Web3") -> None:
         version = web3.geth.shh.version()
-        if '1.7' in web3.clientVersion:
-            assert version == '5.0'
-        else:
-            assert version == '6.0'
+        assert version == '6.0'
```
marcgarreau, line 396: FYI - latest version(s) of Geth removed Whisper from the distro, so when we bump to the latest version, maybe we want to look at dropping support for it in Web3.py.

Issue #1719, opened 2020-08-26
Flaky Geth websocket tests


kclowes:
### What was wrong?
[This PR](https://github.com/ethereum/web3.py/pull/1708) introduced flaky tests in the Geth Websocket module. 
Failures look like: https://app.circleci.com/pipelines/github/ethereum/web3.py/331/workflows/b033c86b-528a-47eb-a699-44027cb1e46b/jobs/122605

carver:
Oops, didn't realize that linking it from the PR would auto-close it on the merge (without the "close" tag)... weird.

kclowes:
Yeah, super weird. At first I thought you just figured it out really quickly 😆

PR #1727, opened 2020-09-02
Handle Flaky Geth WS Tests

kclowes:
### What was wrong?
Since we updated the version of Geth that integration tests run against, the start/stop websocket and start/stop_rpc tests periodically cause all tests after them to fail, but only on CI. I suspect it has something to with the port, or setup/teardown between tests but couldn't figure the exact problem out, and feel that there is a diminishing return in spending more time on it. It does work as expected when checked manually. 

Related to Issue #1719

### How was it fixed?
Just added an xfail for now :(

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/1727.misc.rst
+++ b/newsfragments/1727.misc.rst
@@ -1,0 +1,1 @@
+"Fixed" flaky geth integration tests by adding an xfail.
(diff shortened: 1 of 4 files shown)


marcgarreau (approved):
Manual testing of both start and stop methods worked as expected. Agree that these tests can be revisited and aren't highest priority. 

My naive theory is just to stuff both the rpc and ws start and stop methods in one test to ensure a given state. Can test that and other uninformed theories separately, though.

PR #1733, opened 2020-09-09
Change eth_getBalance and eth_getCode to use Method

kclowes:
### What was wrong?
The Eth module needs to be moved to use Method. Part 2 includes changes to eth_getBalance and eth_getCode.

Related to Issue #1568

### How was it fixed?

Updated eth.py to use Method for selected methods.

### Todo:
- [x] Manually test
- [x] squash

#### Cute Animal Picture

### DIFF
--- a/ens/utils.py
+++ b/ens/utils.py
@@ -212,13 +212,3 @@ def assert_signer_in_modifier_kwargs(modifier_kwargs: Any) -> ChecksumAddress:
 
 def is_none_or_zero_address(addr: Union[Address, ChecksumAddress, HexAddress]) -> bool:
     return not addr or addr == EMPTY_ADDR_HEX
-
-
-def is_valid_domain(domain: str) -> bool:
-    split_domain = domain.split('.')
-    if len(split_domain) == 1:
-        return False
-    for name in split_domain:
-        if not is_valid_name(name):
-            return False
-    return True
--- a/ens/utils.py
+++ b/ens/utils.py
@@ -212,3 +212,13 @@ def assert_signer_in_modifier_kwargs(modifier_kwargs: Any) -> ChecksumAddress:
 
 def is_none_or_zero_address(addr: Union[Address, ChecksumAddress, HexAddress]) -> bool:
     return not addr or addr == EMPTY_ADDR_HEX
+
+
+def is_valid_domain(domain: str) -> bool:
+    split_domain = domain.split('.')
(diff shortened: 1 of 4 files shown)


ens/utils.py
```diff
@@ -212,3 +212,13 @@ def assert_signer_in_modifier_kwargs(modifier_kwargs: Any) -> ChecksumAddress:
 
 def is_none_or_zero_address(addr: Union[Address, ChecksumAddress, HexAddress]) -> bool:
     return not addr or addr == EMPTY_ADDR_HEX
+
+
+def is_valid_domain(domain: str) -> bool:
```
marcgarreau, line 217: Didn't this util have a test before being broken out?

web3/eth.py
```diff
@@ -222,6 +214,17 @@ def block_identifier_munger(
             block_identifier = self.defaultBlock
         return [*args, block_identifier]
 
+    def address_resolver_munger(self, account, block_identifier=None):
+        resolved_addr = abi_ens_resolver(self.web3, 'address', account)[1]
```
kclowes, line 218: I'm fairly sure that having this `web3` here is going to cause problems when we actually implement async, but I think it's the best solution for the moment. I believe the issue is that the middleware used to resolve an ENS address before it hit `validate_address`, but now `validate_address` is called before an ENS address is resolved. This results in an `InvalidAddress` error in the case where an ENS name is passed because `validate_address` expects a hexstring. For posterity's sake, the approach I took earlier was to add a conditional in `validate_address` to check that the domain was valid according to the idna spec, without actually resolving it, and then allowing the middleware to resolve it after the validation. 

@marcgarreau let me know if you have a preference on approach!

marcgarreau: I'm not too concerned about the added "cost" to the async work at this point, but this context may be long forgotten by the time its needed. Maybe a note on [this issue](https://github.com/ethereum/web3.py/issues/1413) is the way to address that? 

Conceptually this is easy enough to reason about, though: if you pass an ENS name into one of these methods, the first that that happens is a resolution check. 👍

pipermerriam: I'm fine kicking the can down the road till async hits.

pipermerriam: This will also fail as soon as the method is used on a `ModuleV2` since the current architecture there removes `self.web3` on the module class.

PR #1735, opened 2020-09-11
Eth -> Method getBlock

kclowes:
### What was wrong?
The Eth module needs to be moved to use Method. Part 3 includes changes to eth_getBlock

Related to Issue #1568

### How was it fixed?

Updated eth.py to use Method for getBlock.

### Todo:
- [x] Squash
- [x] Manually Test

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -19,8 +19,7 @@ def _get_block_by_something(method, params):
     block_identifier = params[0]
     if (
         block_identifier == 'latest' or
-        block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000005' or
-        block_identifier == 5
+        block_identifier == '0x5'
     ):
         return {
             'hash': '0x0000000000000000000000000000000000000000000000000000000000000005',
@@ -35,12 +34,12 @@ def _get_block_by_something(method, params):
                 {'gasPrice': 5},
                 {'gasPrice': 50},
             ],
-            'miner': '0xA',
+            'miner': '0x' + 'AA' * 20,
             'timestamp': 120,
         }
     elif (
         block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000004' or
-        block_identifier == 4
+        block_identifier == '0x4'
     ):
         return {
             'hash': '0x0000000000000000000000000000000000000000000000000000000000000004',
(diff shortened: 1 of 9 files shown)


kclowes:
I think the ethpm tests are failing because IPFS is having trouble today 🤷 I'll make sure they're passing before merging.

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
```diff
@@ -35,12 +34,12 @@ def _get_block_by_something(method, params):
                 {'gasPrice': 5},
                 {'gasPrice': 50},
             ],
-            'miner': '0xA',
+            'miner': '0x' + 'AA' * 20,
```
marcgarreau, line 37: context for miner hash change?

tests/core/middleware/test_filter_middleware.py
```diff
@@ -38,8 +49,8 @@ def iterator():
 @pytest.fixture(scope='function')
 def result_generator_middleware(iter_block_number):
     return construct_result_generator_middleware({
-        'eth_getLogs': lambda *_: ["middleware"],
-        'eth_getBlockByNumber': lambda *_: type('block', (object,), {'hash': 'middleware'}),
+        'eth_getLogs': lambda *_: FILTER_LOG,
```
marcgarreau, line 52: TIL `*_`

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
```diff
@@ -19,8 +19,7 @@ def _get_block_by_something(method, params):
     block_identifier = params[0]
     if (
         block_identifier == 'latest' or
-        block_identifier == '0x0000000000000000000000000000000000000000000000000000000000000005' or
```
marcgarreau, line 22: curious about the difference/requirement of prefixed zeros here 🤔

marcgarreau (approved):
thanks for the walkthrough!

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
kclowes: This test provides its own middleware, which did not include any normalization, so previously it didn't matter what got passed back. Now that the normalization is unavoidable, '0xA' throws an error when it tries to convert the address to a checksum address.

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
kclowes: This particular block hash just wasn't used in the tests. I could leave it in if you want, I don't have a strong opinion either way. The other block identifier hashes are used with the `getBlockByHash` method below. But I did have to change `5` to `0x5` to account for data normalization that wasn't happening before, but now is.

marcgarreau: 👍 If unused, def leave it out.

tests/core/gas-strategies/test_time_based_gas_price_strategy.py
marcgarreau: 👍 thanks for the explanation!

PR #1741, opened 2020-09-15
Eth to method 4

kclowes:
### What was wrong?
The Eth module needs to be moved to use Method. Part 4 includes changes to:
- RPC.eth_getBlockTransactionCountByHash
- RPC.eth_getBlockTransactionCountByNumber
- RPC.eth_getUncleCountByBlockHash
- RPC.eth_getUncleCountByBlockNumber
- RPC.eth_getUncleByBlockHashAndIndex
- RPC.eth_getUncleByBlockNumberAndIndex
- RPC.eth_getTransactionByHash

Related to Issue #1568

### How was it fixed?

Updated eth.py to use Method for the above methods

### Todo:
- [x] Manually test

#### Cute Animal Picture

image

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -92,6 +92,7 @@ def bytes_to_ascii(value: bytes) -> str:
 to_ascii_if_bytes = apply_formatter_if(is_bytes, bytes_to_ascii)
 to_integer_if_hex = apply_formatter_if(is_string, hex_to_integer)
 block_number_formatter = apply_formatter_if(is_integer, integer_to_hex)
+to_hex_if_integer = apply_formatter_if(is_integer, integer_to_hex)
 
 
 is_false = partial(operator.is_, False)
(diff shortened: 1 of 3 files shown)


web3/_utils/method_formatters.py
```diff
@@ -360,7 +362,7 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
     RPC.eth_getUncleCountByBlockNumber: apply_formatter_at_index(block_number_formatter, 0),
     RPC.eth_getUncleByBlockNumberAndIndex: compose(
         apply_formatter_at_index(block_number_formatter, 0),
-        apply_formatter_at_index(integer_to_hex, 1),
+        apply_formatter_at_index(block_number_formatter, 1),
```
marcgarreau, line 368: This has a smell. The formatter is being applied to the index, right?

web3/_utils/method_formatters.py
```diff
@@ -505,9 +510,31 @@ def raise_block_not_found(params: Tuple[BlockIdentifier, bool]) -> NoReturn:
     raise BlockNotFound(f"Block with id: {block_identifier} not found.")
 
 
+def raise_block_not_found_for_uncle_at_index(
+    params: Tuple[BlockIdentifier, Union[HexStr, int]]
+) -> NoReturn:
+    block_identifier = params[0]
+    uncle_index = to_integer_if_hex(params[1])
```
marcgarreau, line 517: (Carrying on from the last comment) I do think this explicit pattern is easier to reason about than the `block_number_formatter`. What do you think about renaming `block_number_formatter` to `to_hex_if_integer`, since this formatter has bled outside of just block formatting? (Okay if not in this PR.)

marcgarreau (approved):
Comment above is nonblocking 🚢

web3/_utils/method_formatters.py
kclowes: Yep, agreed. I had a duplicate method in the original PR, so I'll add that here, and then add an issue to remove the duplicated method in a separate PR.

a0969818431:
`defi`

Issue #1721, opened 2020-08-28
pypiwin32 vs pywin32


step21:
* Version: since 4.2.0
* Python: 2.7/3.4/3.5/*
* OS: osx/linux/win

### What was wrong?

pypiwin32 is an outdated package, that basically only requires pywin32. To make things less complex and tidier and easier on environments without that packages (as it basically does not do anything) it would be nice to switch to pywin32. It should just work, as abi wise they are the same and even from the same guy, just this one is oudated/no longer updated.

### How can it be fixed?

pywin32 in requirements.txt etc.
I could propose a PR.

kclowes:
@step21 it would be awesome if you wanted to propose a PR! I think this is the line that needs to be changed:
https://github.com/ethereum/web3.py/blob/master/setup.py#L82. Thanks!

step21:
I omitted release notes for now, because the actual imports didn't change and users will not import this directly anyway.

PR #1742, opened 2020-09-16
Changed pypiwin32 to pywin32

step21:
### What was wrong?
pypiwin32 is just an empty package requiring pywin32, so this can just be required directly. As far as I know, pypiwin32 is only there for historical reasons.

Related to Issue #1721

### How was it fixed?
change it in setup.py

### Todo:

#### Cute Animal Picture


kclowes:
Thank you @step21! It looks like 228 is the most recent version of pywin32, so I think we might as well use that unless things break? And I'll double check in the next few days, but will you confirm that web3 installs correctly on your windows machine in a fresh environment with this change?

step21:
I didn't test if things break, but if you did not require a higher version until now, leaving the older one keeps compatibility so I would just leave it. Web3 def installs and imports correctly. However I cannot run the whole test suite atm (but not because of this change)

PR #1749, opened 2020-09-21
Allow for ENS name resolution

kclowes:
### What was wrong?
ENS names were throwing an `InvalidAddress` in methods that were using `Method`.

Related to Issue #1585 

### How was it fixed?
Added a new method that does a high-level validation of an ENS address, before it gets resolved in the middleware.

### Todo:
- [x] Manually test
- [x] Address PR feedback
- [x] Make sure all changed methods are programmatically tested
- [x] Squash/cleanup

### DIFF
--- a/ens/utils.py
+++ b/ens/utils.py
@@ -212,3 +212,13 @@ def assert_signer_in_modifier_kwargs(modifier_kwargs: Any) -> ChecksumAddress:
 
 def is_none_or_zero_address(addr: Union[Address, ChecksumAddress, HexAddress]) -> bool:
     return not addr or addr == EMPTY_ADDR_HEX
+
+
+def is_valid_ens_name(ens_name: str) -> bool:
+    split_domain = ens_name.split('.')
+    if len(split_domain) == 1:
+        return False
+    for name in split_domain:
+        if not is_valid_name(name):
+            return False
+    return True
--- a/tests/core/eth-module/test_transactions.py
+++ b/tests/core/eth-module/test_transactions.py
@@ -1,6 +1,10 @@
 import pytest
 
+from web3._utils.ens import (
+    ens_addresses,
+)
 from web3.exceptions import (
+    NameNotFound,
     TimeExhausted,
     TransactionNotFound,
     ValidationError,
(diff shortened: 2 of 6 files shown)


ens/utils.py
```diff
@@ -212,3 +212,13 @@ def assert_signer_in_modifier_kwargs(modifier_kwargs: Any) -> ChecksumAddress:
 
 def is_none_or_zero_address(addr: Union[Address, ChecksumAddress, HexAddress]) -> bool:
     return not addr or addr == EMPTY_ADDR_HEX
+
+
+def is_valid_domain(domain: str) -> bool:
```
pipermerriam, line 217: maybe `is_valid_ens_name`?

tests/core/eth-module/test_transactions.py
```diff
@@ -1,6 +1,10 @@
 import pytest
 
+from web3._utils.ens import (
```
pipermerriam, line 3: Random plumbing work that should be done at some point.  Anything under the `web3` namespace like this that is purely a testing tool should get moved to live under `web3.tools`

tests/core/eth-module/test_transactions.py
```diff
@@ -41,6 +45,46 @@ def test_send_transaction_with_valid_chain_id(web3, make_chain_id, expect_succes
         assert 'chain ID' in str(exc_info.value)
 
 
+@pytest.mark.parametrize(
+    'to, _from, expect_success',
+    (
+        (
+            'thedao.eth',
+            'notanaddress.eth',
```
pipermerriam, line 53: maybe slightly less confusing if you used something like `not_a_registered_ens_name.eth`.

tests/core/eth-module/test_transactions.py
```diff
@@ -41,6 +45,46 @@ def test_send_transaction_with_valid_chain_id(web3, make_chain_id, expect_succes
...
+        ('thedao.eth', web3.eth.accounts[1]),
+        ('rainbow.eth', web3.eth.accounts[0])
+    ]):
+        transaction = {
+            'to': to,
+            'chainId': web3.eth.chainId,
+            'from': _from,
+        }
+
+        if expect_success:
```
pipermerriam, line 79: We've been moving away from this pattern.  It's not a big deal but separating the success and failure cases reduces the `if/else` clause context switches in the test body which can make for easier to read tests.

marcgarreau (approved):
Looks good via manual testing!

Issue #1745, opened 2020-09-17
remove block_number_formatter in favor of to_hex_if_integer


kclowes:
* Version: 5.13 (to be released)

### What was wrong?
In `web3/_utils/method_formatters.py` there is a method called `block_number_formatter` that is used for all of the block numbers. However, I also want to use it for a transaction index. I added a method called `to_hex_if_integer` that is the exact same functionality with a different name. 

### How can it be fixed?

This should be a straightforward replacement of `block_number_formatter` for `to_hex_if_integer`.

tmckenzie51:
I can start working on this

PR #1757, opened 2020-10-01
Update method_formatters.py

tmckenzie51:
removes duplicate block_number_formatter method in favor of to_hex_if_integer method.

### What was wrong?
The methods block_number_formatter and to_hex_if_integer were duplicate functionality.

Related to Issue #1745

### How was it fixed?
block_number_formatter  method was replaced with the to_hex_if_integer method.

### Todo:

#### Cute Animal Picture

Red-Tibetan-Mastiff

### DIFF
--- a/newsfragments/1757.misc.rst
+++ b/newsfragments/1757.misc.rst
@@ -1,0 +1,2 @@
+Duplicate block_number_formatter method removed and replaced with the to_hex_if_integer method.
+
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -91,7 +91,6 @@ def bytes_to_ascii(value: bytes) -> str:
 
 to_ascii_if_bytes = apply_formatter_if(is_bytes, bytes_to_ascii)
 to_integer_if_hex = apply_formatter_if(is_string, hex_to_integer)
-block_number_formatter = apply_formatter_if(is_integer, integer_to_hex)
 to_hex_if_integer = apply_formatter_if(is_integer, integer_to_hex)
 
 
@@ -320,7 +319,7 @@ estimate_gas_with_block_id: Callable[
 ]
 estimate_gas_with_block_id = apply_formatters_to_sequence([
     transaction_param_formatter,
-    block_number_formatter,
+    to_hex_if_integer,
 ])
 
 SIGNED_TX_FORMATTER = {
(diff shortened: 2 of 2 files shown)


njgheorghita:
So, it looks to me like Infura recently updated the version of their IPFS endpoint to `7.0` (running `curl -X POST "https://ipfs.infura.io:5001/api/v0/version"` from your terminal shows the currently supported version). And `py-ipfs-http-client` only supports `0.4.22 ≤ … < 0.7.0`. There are a couple open issues that appear to reference the exact same problem, I've left a comment on one of them. It seems like web3 is blocked until we resolve this, which is definitely not ideal.

Lmk what you think the best way to proceed is, as I see it we have a couple of options.
- Wait until a new version of `py-ipfs-http-client` is released (it's hard to say how soon we can expect this to happen)
- Temporarily skip all tests that use `InfuraIPFSBackend` until `py-ipfs-http-client` is updated
- Reconfigure the tests to use the `LocalIPFSBackend`
- Find a different public gateway we can use. (I've spent a little time trying to connect to a different [public gateway](https://ipfs.github.io/public-gateway-checker/), but have had trouble finding something that works with the [multiaddr](https://github.com/multiformats/multiaddr#string-format) format required by `py-ipfs-http-client`) - imo this is probably the best temporary solution to get us until `py-ipfs-http-client` is updated

tmckenzie51:
> here are a couple open issues that appear to reference the exact same problem, I've left a comment on one of them.

tmckenzie51:
> So, it looks to me like Infura recently updated the version of their IPFS endpoint to `7.0` (running `curl -X POST "https://ipfs.infura.io:5001/api/v0/version"` from your terminal shows the currently supported version). And `py-ipfs-http-client` only supports `0.4.22 ≤ … < 0.7.0`. There are a couple open issues that appear to reference the exact same problem, I've left a comment on one of them. It seems like web3 is blocked until we resolve this, which is definitely not ideal.
> 
> Lmk what you think the best way to proceed is, as I see it we have a couple of options.
> 
> * Wait until a new version of `py-ipfs-http-client` is released (it's hard to say how soon we can expect this to happen)
> * Temporarily skip all tests that use `InfuraIPFSBackend` until `py-ipfs-http-client` is updated
> * Reconfigure the tests to use the `LocalIPFSBackend` (might require a significant chunk of refactoring)
> * Find a different public gateway we can use. (I've spent a little time trying to connect to a different [public gateway](https://ipfs.github.io/public-gateway-checker/), but have had trouble finding something that works with the [multiaddr](https://github.com/multiformats/multiaddr#string-format) format required by `py-ipfs-http-client`) - imo this is probably the best temporary solution to get us until `py-ipfs-http-client` is updated

Okay, I see. Thank you for explaining that. Those sound like decent options that we can consider. @pipermerriam @kclowes  Of these options, which do you think would be the best for us to proceed with?

marcgarreau:
@njgheorghita pointed out a [fix PR](https://github.com/ipfs-shipyard/py-ipfs-http-client/pull/241) waiting for review, but it's not a very active repo. I'm not opposed to just xfail these tests for now, but is that doable for doctests?

kclowes (approved):
This looks good to me as long as the tests pass after you merge in master!

kclowes:
@tmckenzie51 You can try and push an empty commit to see if that readthedocs failure will work itself out: `git commit --allow-empty -m "Trigger Build"` Let me know if you have questions or want to pair!

Issue #1720, opened 2020-08-27
Update Flake8


kclowes:
* Version: 5.12.0
* Python: 3.8

### What was wrong?
flake8 3.4.1 doesn't run with python 3.8, but upgrading flake8 will fail our linting.

### How can it be fixed?
Upgrade flake8 to the most recent version and fix all the errors. 

...Or maybe this is a good time to pull in black

pratik-vii:
@kclowes I will pick this up. please assign to me.

kclowes:
Closed via #1759.

PR #1759, opened 2020-10-02
flake8 upgrade to 3.8.3

pratik-vii:
### What was wrong?

Related to Issue #1720

### How was it fixed?
* upgraded flake8 to 3.8.3
* W503 ignored where required.

#### Cute Animal Picture

Koala!!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ extras_require = {
         "py-geth>=2.4.0,<3",
     ],
     'linter': [
-        "flake8==3.4.1",
+        "flake8==3.8.3",
         "isort>=4.2.15,<4.3.5",
         "mypy==0.730",
     ],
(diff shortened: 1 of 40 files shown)


setup.py
```diff
@@ -51,10 +51,10 @@
 }
 
 extras_require['dev'] = (
-    extras_require['tester'] +
-    extras_require['linter'] +
-    extras_require['docs'] +
-    extras_require['dev']
+    extras_require['tester']
+    + extras_require['linter']  # noqa: W503
```
pipermerriam, line 55: I would be ok with adding `W503` to the ignored warnings in our `tox.ini` file.

pratik-vii: yeah. right! will do

pipermerriam (approved):
Thanks for doing this cleanup for us!  Would you be up for adding W503 to the ignored linting codes so we can drop a few of the `# noqa: W503` comments?

pipermerriam:
(also, you should work on getting CI passing)

pratik-vii:
> (also, you should work on getting CI passing)

some of the CI will fail because of [Known issue](https://github.com/ipfs-shipyard/py-ipfs-http-client/issues/239) with Infura IPFS endpoint

https://github.com/ethereum/web3.py/pull/1757#issuecomment-702628285

kclowes:
@pratik-vii thanks for taking this on! I just pulled in the most recent master that should have the tests fixed, so as soon as those pass, I'll pull this in. :tada:

PR #1763, opened 2020-10-06
Generate parity v2.5.13 fixture

### LINKED ISSUE #1753: Upgrade Parity version in test infrastructure
### What was wrong?

Our Parity tests run on version 2.3.5. According to a recent OpenEthereum [blog post](https://medium.com/openethereum/vision-for-openethereum-ex-parity-client-eb7b11f6eef8), latest stable recommendation is 2.5.13.

marcgarreau:
### What was wrong?
- A stale Parity version (2.3.5) is used for running Parity tests. Updating is another stepping stone on the way to handling revert messages. Closes #1753. 
- We have two files for generating geth fixtures: one to generate geth fixtures and one to generate the Parity fixtures off of a (separate) Geth fixture. I'm unaware of the context to get to this point, but its a bit of a maintenance headache and is confusing to figure out what to use where.

### How was it fixed?
- Removed the old Geth fixture generation file in favor of the newer one, with minor tweaks. (This accounts for most of the diff.)
- Added Parity config params to account for later forks (and enforce at block zero)
- Generated the new Parity fixture
- Handle tests
   - Unsure of why tracing isn't working in tests; exceeded my timebox for investigating
   - Whisper functionality is deprecated, but I didn't think this was the right PR to start ripping things out

### Todo:

#### Cute Animal Picture

image


tests/integration/parity/common.py
```diff
@@ -181,7 +176,17 @@ def test_invalid_eth_signTypedData(self,
 
 
 class ParityTraceModuleTest(ParityTraceModuleTest):
-    pass
+    def test_trace_block(self, web3):
+        pytest.xfail("TODO: tracing not working on v2.5.13")
```
kclowes, line 180: In general, we prefer using the `@pytest.mark.xfail` decorator instead of using `pytest.xfail`, since pytest.xfail doesn't run anything else in the tests unless there is a reason for not running the rest of the test. Would it be possible to use the decorator here?

Here's some [documentation](https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail)

kclowes (approved):
Nice! Just a couple comments:
- one about `@pytest.mark.xfail` vs. `pytest.xfail` in the code
- Did you check that you can generate a geth fixture (and tests pass) without the `tests/generate_go_ethereum_fixture.py` file?

marcgarreau:
@kclowes 1) Confirmed geth fixture generation still works 👍 2) Got inconsistent/flaky results from letting those tests run under an xfail decorator. Gave up with one of the flaky whisper tests and outright failed it. The rest of the suite is happy for the moment, but I'm not 100% convinced the flakiness is resolved. If it acts up, we just go more aggressive with the failures, since Whisper will be ripped out for v6.

kclowes:
@marcgarreau sweet, thank you! 

> If it acts up, we can just go more aggressive with the failures, since Whisper will be ripped out for v6.

Sounds good! 

🚀

Issue #1729, opened 2020-09-04
Get Rid of Hypothesis Deprecation Warnings


kclowes:
* Version: 5.12.1

### What was wrong?

At the end of our core tests, we have a Deprecation Warning that looks like: 
`HypothesisDeprecationWarning: tests/core/utilities/test_event_filter_builder.py::test_match_any_bytes_type_properties uses the 'web3' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.`

Actual output can be found in this [test run](https://app.circleci.com/pipelines/github/ethereum/web3.py/368/workflows/a61f1619-cbb4-4c43-b0f5-a3c49cee62a5/jobs/123892).

### How can it be fixed?

Add the appropriate scope to each fixture. This will need to be done in multiple PRs.

tmckenzie51:
I'm interested in working on this issue.

kclowes:
Yes it does, and we'd love help! You can see the full list of warnings on any of the latest core test runs on CircleCI. The py36-core tests can be found [here](https://app.circleci.com/pipelines/github/ethereum/web3.py/1440/workflows/5b42450d-b9d8-4d5a-bd6e-e82a1641764b/jobs/159645), and you can see the warnings in the `run tox` dropdown. I'd recommend tackling a few at a time so that we can review smaller PRs. Let us know if you need more direction!

reyaw:
Couple of questions: 
1) Is this the correct way to define the scope of a fixture? (top line of the screenshot)

![fixture_scope](https://user-images.githubusercontent.com/82685635/122440542-a8c68e80-cf6a-11eb-84d2-7e85e03190c1.png)

kclowes:
> 1. Is this the correct way to define the scope of a fixture? (top line of the screenshot)

yep!

> 2. How can I determine the correct scope of each fixture?

Since we're not having problems with the scope now, I'd start with the largest scope you can (which I believe is session, but I'm not sure on that) and then make the scope narrower as you run into issues. We've done a few of these before - [this PR](https://github.com/ethereum/web3.py/pull/1780/files) is a good example. We ended up moving all of the functions related to the Emitter fixture to a module scope, and moved it into the file where it was actually used. You'll likely need to do something similar. Let me know if you have more questions!

[...]

PR #1767, opened 2020-10-08
specifies 'module' scope for relevant pytest fixtures

tmckenzie51:
specifies 'module' scope for pytest fixtures 'emitter_event_ids', 'w3_strict_abi', and 'create_filter'.

### What was wrong?
At the end of our tests we have hypothesis deprecation warnings that look like - `HypothesisDeprecationWarning: tests/core/utilities/test_event_filter_builder.py::test_match_any_bytes_type_properties uses the 'web3' fixture, which is reset between function calls but not between test cases generated by @given(...). You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function. See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope`.

Related to Issue #1729 and PR #1766 

### How was it fixed?
'module' scope was specified for pytest fixtures 'Emitter', 'emitter_event_ids', and 'w3_strict_abi' pytest fixtures.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -101,7 +101,7 @@ def web3():
     return Web3(provider)
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def w3_strict_abi():
     w3 = Web3(EthereumTesterProvider())
     w3.enable_strict_bytes_type_checking()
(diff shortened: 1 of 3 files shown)


kclowes (approved):
LGTM! I would probably rebase this down to one commit, and then once you add the release notes and the tests pass, you can merge! 🚢

kclowes (approved):
Nice! As soon as this has a release note, I'd say it's ready to be merged!

PR #1768, opened 2020-10-08
Mark failing tests as such until ipfshttpclient is updated

kclowes:
### What was wrong?
Infura's IPFS gateway updated to go-ipfs v0.7.0 which `ipfshttpclient` doesn't support, so all of our IPFS tests that use the Infura backend are broken. 

Related to Issue #1757

### How was it fixed?
See comment [here](https://github.com/ethereum/web3.py/pull/1757#issuecomment-702628285) for a better description of the problem. If `ipfshttpclient` doesn't update soon (PR up [here](https://github.com/ipfs-shipyard/py-ipfs-http-client/pull/241)), it may be worth pursuing one of the more long-term solutions, but we can decide that in a few weeks. For now, this unblocks a bunch of PRs that need to be merged.

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #1775, opened 2020-10-14
Update ipfshttpclient

kclowes:
### What was wrong?
The `ipfshttpclient` package updated, and now our tests can pass again!

### How was it fixed?
Updated, but not sure if pinning to the alpha version is any better. If it's not better, I can close this PR until it's out of pre-release. I pinned instead of providing a range because it sounds like there will be breaking changes still. 

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/tests/ethpm/_utils/test_backend_utils.py
+++ b/tests/ethpm/_utils/test_backend_utils.py
@@ -31,8 +31,7 @@ from ethpm.uri import (
     ),
 )
 @pytest.mark.skipif('WEB3_INFURA_PROJECT_ID' not in os.environ, reason='Infura API key unavailable')
-@pytest.mark.xfail(reason="py-ipfs-http-client library doesn't support go-ipfs v0.7.0")
-def test_get_resolvable_backends_for_supported_uris(uri, backends):
+def test_get_resolvable_backends_for_supported_uris(dummy_ipfs_backend, uri, backends):
     good_backends = get_resolvable_backends_for_uri(uri)
     assert good_backends == backends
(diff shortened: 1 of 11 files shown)


tests/ethpm/_utils/test_backend_utils.py
```diff
@@ -45,7 +44,6 @@ def test_get_resolvable_backends_for_supported_uris(uri, backends):
     ),
 )
 @pytest.mark.skipif('WEB3_INFURA_PROJECT_ID' not in os.environ, reason='Infura API key unavailable')
-@pytest.mark.xfail(reason="py-ipfs-http-client library doesn't support go-ipfs v0.7.0")
 def test_get_translatable_backends_for_supported_uris(uri, backends):
```
njgheorghita, line 47: I guess it doesn't make a difference to the test passing, but maybe we want to add the `dummy_ipfs_backend` back in the args here like it [was before](https://github.com/ethereum/web3.py/pull/1768/files#diff-14b4b74dd69848bdda95565da963b1b113afbf93430b9036b57a9a36bb2b574bL48)

njgheorghita (approved):
Looks good to me! just one missing thing is to rollback the docstring from `python` back to `doctest` [here](https://github.com/ethereum/web3.py/pull/1768/files#diff-9ad3cecf35af643e0c370e53929562f6366c0b51c3b75dec4fe5dc22e00a789fR608)

kclowes:
Aha, good memory! thank you!

PR #1778, opened 2020-10-21
Move filter method to use Method class

kclowes:
### What was wrong?
Filter is the last method in Eth that wasn't using Method. This PR fixes that. 

### Todo:
- [x] Make sure tests are using the correct middleware
- [x] Better solution than passing around a string in place of a method
- [x] Lint
- [x] Get rid of TODOs
- [x] Manually test

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/filtering/test_existing_filter_instance.py
+++ b/tests/core/filtering/test_existing_filter_instance.py
@@ -17,7 +17,7 @@ def filter_id(web3):
     return block_filter.filter_id
 
 
-def test_instatiate_existing_filter(web3, sleep_interval, wait_for_block, filter_id):
+def test_instantiate_existing_filter(web3, sleep_interval, wait_for_block, filter_id):
     with pytest.raises(TypeError):
         web3.eth.filter('latest', filter_id)
     with pytest.raises(TypeError):
--- a/web3/_utils/filters.py
+++ b/web3/_utils/filters.py
@@ -9,6 +9,7 @@ from typing import (
     Optional,
     Sequence,
     Tuple,
+    Union,
 )
 
 from eth_abi.codec import (
@@ -23,6 +24,7 @@ from eth_typing import (
     TypeStr,
 )
 from eth_utils import (
+    is_hex,
     is_list_like,
     is_string,
     is_text,
@@ -46,15 +48,24 @@ from web3._utils.events import (
 from web3._utils.validation import (
     validate_address,
 )
+from web3.exceptions import (
+    ValidationError,
+)
 from web3.types import (
     ABIEvent,
     BlockIdentifier,
     FilterParams,
     LogReceipt,
+    RPCEndpoint,
 )
 
 if TYPE_CHECKING:
     from web3 import Web3  # noqa: F401
+    from web3.module import (  # noqa: F401
+        Module,
+        ModuleV2,
+    )
+    from web3.eth import Eth  # noqa: F401
 
 
 def construct_event_filter_params(
@@ -122,14 +133,16 @@ class Filter:
     poll_interval = None
     filter_id = None
 
-    def __init__(self, web3: "Web3", filter_id: HexStr) -> None:
-        self.web3 = web3
+    def __init__(self,
+                 filter_id: HexStr,
+                 eth_module: "Eth") -> None:
+        self.eth_module = eth_module
         self.filter_id = filter_id
         self.callbacks = []
         super().__init__()
 
     def __str__(self) -> str:
-        return "Filter for {0}".format(self.filter_id)
+        return f"Filter for {self.filter_id}"
 
     def format_entry(self, entry: LogReceipt) -> LogReceipt:
         """
@@ -148,11 +161,11 @@ class Filter:
         return filter(self.is_valid_entry, entries)
 
     def get_new_entries(self) -> List[LogReceipt]:
-        log_entries = self._filter_valid_entries(self.web3.eth.getFilterChanges(self.filter_id))
+        log_entries = self._filter_valid_entries(self.eth_module.getFilterChanges(self.filter_id))
         return self._format_log_entries(log_entries)
 
     def get_all_entries(self) -> List[LogReceipt]:
-        log_entries = self._filter_valid_entries(self.web3.eth.getFilterLogs(self.filter_id))
+        log_entries = self._filter_valid_entries(self.eth_module.getFilterLogs(self.filter_id))
         return self._format_log_entries(log_entries)
 
     def _format_log_entries(self,
@@ -203,7 +216,7 @@ class LogFilter(Filter):
         """
         self.data_filter_set = data_filter_set
         if any(data_filter_set):
-            self.data_filter_set_function = match_fn(self.web3, data_filter_set)
+            self.data_filter_set_function = match_fn(self.eth_module.codec, data_filter_set)
 
     def is_valid_entry(self, entry: LogReceipt) -> bool:
         if not self.data_filter_set:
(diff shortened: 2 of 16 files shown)


web3/eth.py
```diff
@@ -462,44 +459,81 @@ def estimate_gas_munger(
...
             raise TypeError("Must provide either filter_params as a string or "
                             "a valid filter object, or a filter_id as a string "
                             "or hex.")
 
+    filter: Method[Callable[..., Any]] = Method(
+        method_choice_depends_on_args=select_filter_method(
+            is_new_block_filter=RPC.eth_newBlockFilter,
+            is_new_pending_transaction_filter=RPC.eth_newPendingTransactionFilter,
+            is_new_filter=RPC.eth_newFilter,
+            is_existing_filter='existing_filter',
```
kclowes, line 494: @pipermerriam @marcgarreau would love some feedback on this janky handling of the existing filter case. If a user passes in a filter_id (ex. '0x0'), then we don't want to make the request to the server, we just want to return a `LogFilter` object which needs the module and the filter_id.  Currently, I'm handling it [here](https://github.com/ethereum/web3.py/pull/1778/files#diff-c473cbb4ff9e3ac6c462a6585df86227779156085f03fc5a37aad66f1e6e5a6eR47-R48), but it doesn't seem like it belongs there. Any ideas?

marcgarreau: 🤔 Is there a not awful way to check for an existing filter prior to `select_filter_method` and bail out early before going through the Method workflow?

nitpicky: looks like the other `depends_on` methods use `if_condition` instead of `is_condition`, like you've got here.

tests/core/filtering/test_existing_filter_instance.py
```diff
@@ -17,7 +17,7 @@ def filter_id(web3):
     return block_filter.filter_id
 
 
-def test_instatiate_existing_filter(web3, sleep_interval, wait_for_block, filter_id):
```
pipermerriam, line 20: insatiable filters!

web3/_utils/method_formatters.py
```diff
@@ -580,14 +593,45 @@ def raise_transaction_not_found_with_index(params: Tuple[BlockIdentifier, int])
...
+    module: Union["Module", "ModuleV2"],
+    method: RPCEndpoint,
+    filter_id: HexStr,
+):
+    if method == RPC.eth_newBlockFilter:
+        return BlockFilter(filter_id, module=module)
+    elif method == RPC.eth_newPendingTransactionFilter:
+        return TransactionFilter(filter_id, module=module)
+    elif method == RPC.eth_newFilter:
+        return LogFilter(filter_id, module=module)
```
pipermerriam, line 606: I would add a final `else` clause here that raises an exception assuming you expect that logic branch to never be exercised.

web3/module.py
```diff
@@ -40,6 +43,8 @@ def retrieve_blocking_method_call_fn(
 ) -> Callable[..., RPCResponse]:
     def caller(*args: Any, **kwargs: Any) -> RPCResponse:
         (method_str, params), response_formatters = method.process_params(module, *args, **kwargs)
+        if method_str == 'existing_filter':
```
pipermerriam, line 46: First thought is that we can at least get rid of the polymorphic return type for the `method_str` by converting this to use an exception instead of a return value.

```python
try:
    (method_str, params), response_formatters = method.process_params(...)
except _UseExistingFilterID as err:
    return LogFilter(module=module, filter_id=err.filter_id)
...
```

pipermerriam: Sorta ugly but at least the mechanism doesn't mix-up the return types.....

kclowes: Ooh, I like that better 👍

web3/module.py
```diff
@@ -34,12 +44,18 @@ def apply_result_formatters(
...
-) -> Callable[..., RPCResponse]:
-    def caller(*args: Any, **kwargs: Any) -> RPCResponse:
-        (method_str, params), response_formatters = method.process_params(module, *args, **kwargs)
+    w3: "Web3", module: Union["Module", "ModuleV2"], method: Method[Callable[..., TReturn]]
+) -> Callable[..., Union[TReturn, LogFilter]]:
+    def caller(*args: Any, **kwargs: Any) -> Union[TReturn, LogFilter]:
+        try:
+            (method_str, params), response_formatters = method.process_params(module, *args, **kwargs)  # noqa: E501
+        except _UseExistingFilter as err:
+            return LogFilter(eth_module=module, filter_id=err.filter_id)
```
kclowes, line 58: @pipermerriam I can't figure out how to pull this out without going back to passing around a string method name, since all of the formatters are chosen based on the method name. Let me know if you see another way!

kclowes:
I think this is ready for a final look @marcgarreau and @pipermerriam!

web3/_utils/filters.py
```diff
@@ -46,15 +48,25 @@
 from web3._utils.validation import (
     validate_address,
 )
+from web3.exceptions import (
+    ValidationError,
+    _UseExistingFilter,
```
pipermerriam, line 53: most of the time when I define an "internal" exception like this that should never leak into userspace I'll simply define it in the module where it is used.  This helps prevent it from leaking into the public API namespace.

web3/_utils/filters.py
```diff
@@ -260,3 +274,29 @@ def match_fn(w3: "Web3", match_values_and_abi: Collection[Tuple[str, Any]], data
...
+            raise _UseExistingFilter(value)
+        else:
+            raise ValidationError("Filter argument needs to be either 'latest',"
+                                  " 'pending', or a hex-encoded filter_id")
+    elif isinstance(value, dict):
+        return if_new_filter
+    else:
+        raise ValidationError("Filter argument needs to be either the string "
+                              "'pending' or 'latest', a filter_id, "
+                              "or a filter params dictionary")
```
pipermerriam, line 302: When someone encounters either of these `ValidationError` cases, it tends to be useful to include both this information about what was expected, but also to include information about what we actually got.  It can be helpful to include both the `repr(value)` and `type(value)` in the error message.

web3/main.py
```diff
@@ -258,6 +259,18 @@ def isConnected(self) -> bool:
     def is_encodable(self, _type: TypeStr, value: Any) -> bool:
         return self.codec.is_encodable(_type, value)
 
+    @property
+    def codec(self) -> ABICodec:
+        return self._codec
+
+    @codec.getter
+    def codec(self) -> ABICodec:
```
pipermerriam, line 267: I don't follow the need for these property methods.  Why not just assign `self.codec = ABICodec(...)`?

web3/middleware/filter.py
```diff
@@ -235,14 +244,19 @@ def __init__(
 
     @property
     def from_block(self) -> BlockNumber:
-        return self._from_block
+        if is_string(self._from_block) and is_hex(self._from_block):
```
pipermerriam, line 247: we should probably be normalizing the `self._from_block` value before assigning it which should make this un-necessary.

web3/module.py
pipermerriam: If we can't get this out, lets consider making it generic.  I'd suggest doing that as a follow up to this PR.

Roughly speaking, it looks like we need an official way for a `Method` to short-circut the JSON-RPC request and to return a value that is generated locally.  That could be using something like an exception like we did here, but I think that's probably not the best choice here because it feels like a heavy way to implement what is effectively an `if/else`.

We might need to return something more *structured* from `process_params` that we can then switch on.

```python
request = method.preprocess_request(module, *args, **kwargs)
if request.is_json_rpc:
    ...  # do the actual request
elif request.has_static_response:
    ...  # directly return the response value.
else:
    ...  # dunno, maybe just raise Exception("Invariant")
```

I don't know if this is going to work, or be clean.  Just my stab at what the API could look like.  I think you'll be doing yourself and everyone on the team a favor to spend some time working on these APIs, especially in making them less cryptic in how they work.

pipermerriam (approved):

web3/main.py
kclowes: oops, forgot to take this out, thanks!

web3/module.py
kclowes: Yeah, agreed. I'll do some mucking around with it in another PR.

PR #1780, opened 2020-10-23
Fixes emitter fixture scoping issues

tmckenzie51:
## What was wrong?
At the end of our tests we have hypothesis deprecation warnings that look like - HypothesisDeprecationWarning: tests/core/utilities/test_event_filter_builder.py::test_match_any_bytes_type_properties uses the 'web3' fixture, which is reset between function calls but not between test cases generated by @given(...). You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function. See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.

Related to Issue #1729

### How was it fixed?
emitter fixtures ('emitter', 'Emitter', 'EMITTER', EMITTER_ABI, 'EMITTER_RUNTIME', 'EMITTER_CODE')  were changed to have  'module' scope for test_contract_data_filters and function scope otherwise.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -1,17 +1,41 @@
-# -*- coding: utf-8 -*-
-
 import pytest
 
 from eth_utils.toolz import (
     dissoc,
 )
 
+from web3 import Web3
 from web3.exceptions import (
     ValidationError,
 )
+from web3.middleware import (
+    local_filter_middleware,
+)
+from web3.providers.eth_tester import (
+    EthereumTesterProvider,
+)
+
+# -*- coding: utf-8 -*-
+
 
 # Ignore warning in pyethereum 1.6 - will go away with the upgrade
 pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
+
+
+# Ignore warning in pyethereum 1.6 - will go away with the upgrade
+pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
+
+
+@pytest.fixture(
+    params=[True, False],
+    ids=["local_filter_middleware", "node_based_filter"])
+def web3(request):
+    use_filter_middleware = request.param
+    provider = EthereumTesterProvider()
+    w3 = Web3(provider)
+    if use_filter_middleware:
+        w3.middleware_onion.add(local_filter_middleware)
+    return w3
 
 
 @pytest.fixture()
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -4,38 +4,11 @@ from eth_utils.toolz import (
     dissoc,
 )
 
-from web3 import Web3
 from web3.exceptions import (
     ValidationError,
 )
-from web3.middleware import (
-    local_filter_middleware,
-)
-from web3.providers.eth_tester import (
-    EthereumTesterProvider,
-)
 
 # -*- coding: utf-8 -*-
-
-
-# Ignore warning in pyethereum 1.6 - will go away with the upgrade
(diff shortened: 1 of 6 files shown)


a0969818431:
`defi`

tests/core/contracts/test_contract_buildTransaction.py
```diff
@@ -1,19 +1,43 @@
...
+    local_filter_middleware,
+)
+from web3.providers.eth_tester import (
+    EthereumTesterProvider,
+)
+
+# -*- coding: utf-8 -*-
+
 
 # Ignore warning in pyethereum 1.6 - will go away with the upgrade
```
kclowes, line 21: Looks like this is a duplicate, and we don't need it anyway, so let's delete lines 21-26

tests/core/filtering/test_contract_data_filters.py
```diff
@@ -6,6 +6,83 @@
...
+    CONTRACT_EMITTER_RUNTIME,
+)
+from web3.middleware import (
+    local_filter_middleware,
+)
+from web3.providers.eth_tester import (
+    EthereumTesterProvider,
+)
+
+# Ignore warning in pyethereum 1.6 - will go away with the upgrade
```
kclowes, line 22: These two lines can be removed too!

tests/core/filtering/test_contract_on_event_filtering.py
```diff
@@ -4,10 +4,30 @@
...
 
+from web3 import Web3
+from web3.middleware import (
+    local_filter_middleware,
+)
+from web3.providers.eth_tester import (
+    EthereumTesterProvider,
+)
+
 # Ignore warning in pyethereum 1.6 - will go away with the upgrade
```
kclowes, line 15: And these can be removed :)

tests/core/filtering/conftest.py
```diff
@@ -27,7 +27,6 @@ def tester_snapshot(web3):
 
 
 @pytest.fixture(
-    scope='function',
```
kclowes, line 30: Although it doesn't really matter, I think you can put this line back just to cut down on lines changed here.

tests/core/filtering/test_contract_on_event_filtering.py
```diff
@@ -4,10 +4,30 @@
...
 pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
 
 
+@pytest.fixture(
```
kclowes, line 19: This looks like the same fixture from `tests/core/filtering/conftest.py`. If we remove it, do the tests fail?

tests/core/contracts/test_contract_buildTransaction.py
```diff
@@ -1,19 +1,43 @@
...
 pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
 
 
+# Ignore warning in pyethereum 1.6 - will go away with the upgrade
+pytestmark = pytest.mark.filterwarnings("ignore:implicit cast from 'char *'")
+
+
+@pytest.fixture(
```
kclowes, line 29: I think this is a different web3 than what we want in the contract tests. Do tests fail if you remove this? Or do any warnings come back?

kclowes (commented):
Looking good :boom: !

tests/core/filtering/test_contract_on_event_filtering.py
tmckenzie51: No failure when they're removed!

tests/core/contracts/test_contract_buildTransaction.py
tmckenzie51: No failure, and no new warnings! :)

kclowes (approved):
Nice!

Issue #1655, opened 2020-05-19
abi['name'] == name KeyError: 'name'


reinisb123:
* Version: 5.9.0
* Python: 3.7
* OS: win
* `pip freeze` output

```
aiodns==1.1.1
aiohttp==3.0.1
altgraph==0.16.1
appdirs==1.4.3
asn1crypto==0.24.0
async-generator==1.10
async-timeout==2.0.1
asyncio-run-in-process==0.1.0a4
attrdict==2.0.1
attrs==18.2.0
autobahn==19.1.1
Automat==0.7.0
base58==0.2.5
beautifulsoup4==4.8.2
bitmex==0.2.2
bitmex-ws==0.4.0
bitshares==0.3.2
bloom-filter==1.3
bravado==10.4.3
bravado-core==5.16.0
broker-client==1.0.0
bs4==0.0.1
ccxt==1.17.434
certifi==2018.11.29
cffi==1.11.5
chardet==3.0.4
Click==7.0
click-datetime==0.2
cloudpickle==1.3.0
constantly==15.1.0
cryptography==2.3
cycler==0.10.0
cytoolz==0.10.1
dateparser==0.7.0
dateutils==0.6.6
docopt==0.6.2
ecdsa==0.13.2
eth-abi==2.1.1
eth-account==0.4.0
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.1.2
eth-typing==2.2.1
eth-utils==1.8.4
Events==0.3
future==0.17.1
graphenelib==1.1.18
hexbytes==0.2.0
hyperlink==18.0.0
idna==2.7
importlib-metadata==1.3.0
incremental==17.5.0
ipfshttpclient==0.4.12
jsonpointer==2.0
jsonref==0.2
jsonschema==3.2.0
jwt==0.6.1
kiwisolver==1.0.1
liquidtap==1.0.1
lru-dict==1.1.6
macholib==1.11
matplotlib==3.0.2
monotonic==1.5
more-itertools==8.0.2
msgpack-python==0.5.6
multiaddr==0.0.9
multidict==4.5.2
mypy-extensions==0.4.3
netaddr==0.7.19
numpy==1.16.0
oauthlib==3.0.0
pandas==0.25.3
parsimonious==0.8.1
pefile==2018.8.8
protobuf==3.11.3
pyasn1==0.4.5
pyasn1-modules==0.2.3
pyblake2==1.1.2
[...]

PR #1786, opened 2020-10-30
Gracefully handle when the name key does not exist in the contract abi

RoyLee:
### What was wrong?
Some abi contracts do not have a name key. This will gracefully handle those cases by preventing an errors when that occurs.

Related to Issue [#1655](https://github.com/ethereum/web3.py/issues/1655)

### How was it fixed?
Use `get` method to return a default value of '' when there is no `name` key.

### Todo:

#### Cute Animal Picture
🐱

### DIFF
--- a/newsfragments/1786.feature.rst
+++ b/newsfragments/1786.feature.rst
@@ -1,0 +1,1 @@
+When finding a contract function gracefully handle abis that do not have a name key.
--- a/tests/core/utilities/test_abi_filter_by_abi_name.py
+++ b/tests/core/utilities/test_abi_filter_by_abi_name.py
@@ -50,6 +50,13 @@ ABI_FUNC_3 = {
     "outputs": [],
     "type": "function",
 }
+ABI_FUNC_4_NO_NAME = {
+    "constant": False,
+    "inputs": [],
+    "outputs": [],
+    "type": "function",
+}
+
 
 ABI = [
     ABI_CONSTRUCTOR,
@@ -59,6 +66,7 @@ ABI = [
     ABI_FUNC_2_SIG_A,
     ABI_FUNC_2_SIG_B,
     ABI_FUNC_3,
+    ABI_FUNC_4_NO_NAME,
 ]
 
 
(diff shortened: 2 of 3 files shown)


kclowes:
@RoyLee Thanks for opening this PR! Do you have a real ABI example with a function that has no name? I may be wrong, but it looks to me from the [Solidity ABI spec](https://solidity.readthedocs.io/en/v0.7.0/abi-spec.html#json) that the only functions that don't have a name are fallback and constructor functions, which means that if this `NameError` is raised, the ABI is malformed. If that's case, I think the better change would be to raise a different error with a better message, rather than ignoring it with a `get`. And probably the best solution is to have an ABI validator as discussed [here](https://github.com/ethereum/web3.py/issues/1655#issuecomment-630908773).

@pipermerriam I'm curious to hear your thoughts here too!

RoyLee:
@kclowes In my specific case the name I was trying to call did exist in the application binary interface but there just happened to be an element in the ABI that did not have a name key in it. Viewing the solidity documentation I do not see any of the fields including name are required and do not believe an exception should be raised because the `filter_by_name`'s role is not to validate/enforce the ABI.

marcgarreau:
@RoyLee - if you've used a tool that generated an ABI function without a name, please let us know which one/how to reproduce it.

While still unclear to me when we'd run into one of these nameless functions outside of fallback/constructor, I do agree that this `filter_by_name` function is probably the wrong place to enforce the validation. As Piper points out in the linked issue, a separate feature for validation/normalization might be the 'real' answer here, but I'm okay with this change set alone.

RoyLee:
@marcgarreau An example of an ABI that I assume is highly used via the web3 library is the Uniswap v2 router smart contract: https://etherscan.io/address/0x7a250d5630b4cf539739df2c5dacb4c659f2488d#code
There are multiple elements without a name.

kclowes:
@RoyLee I grabbed the ABI at that address and ran a quick script on it, but the only functions that didn't have names are the `fallback` and `constructor` functions. We have separate logic for these functions that account for the functions not having a name. I am hesitant to merge this without an example of an ABI where functions don't have names. Let us know if you're using a different tool or something to get the ABI, and I'd be happy to take another look.

fselmo:
Closing due to inactivity and missing clarifying information. Feel free to re-open if there is more information here and if this is still relevant.

PR #1790, opened 2020-11-09
tests return values against different geth versions

tmckenzie51:
### What was wrong?

Related to Issue #1572

### How was it fixed?
In order to be compatible with the most recent geth versions v 1.9.19+, test_eth_getBlockByNumber was modified to ensure that a block number is being returned, rather than null. Also, a test for getBlockByHash for pending blocks was added. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_ethereum_tester.py
+++ b/tests/integration/test_ethereum_tester.py
@@ -248,7 +248,7 @@ class TestEthereumTesterEthModule(EthModuleTest):
     test_eth_submitWork = not_implemented(EthModuleTest.test_eth_submitWork, ValueError)
 
     def test_eth_getBlockByHash_pending(
-        self, web3: "Web3", empty_block: BlockData
+        self, web3: "Web3"
     ) -> None:
         block = web3.eth.getBlock('pending')
         assert block['hash'] is not None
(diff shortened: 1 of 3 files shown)


web3/_utils/module_testing/eth_module.py
```diff
@@ -808,6 +808,12 @@ def test_eth_getBlockByHash_not_found(
         with pytest.raises(BlockNotFound):
             web3.eth.getBlock(UNKNOWN_HASH)
 
+    def test_eth_getBlockByHash_pending(
+        self, web3: "Web3", empty_block: BlockData
```
kclowes, line 812: same here re. `empty_block`

tests/integration/test_ethereum_tester.py
```diff
@@ -244,6 +247,12 @@ class TestEthereumTesterEthModule(EthModuleTest):
     test_eth_submitHashrate = not_implemented(EthModuleTest.test_eth_submitHashrate, ValueError)
     test_eth_submitWork = not_implemented(EthModuleTest.test_eth_submitWork, ValueError)
 
+    def test_eth_getBlockByHash_pending(
+        self, web3: "Web3", empty_block: BlockData
```
kclowes, line 251: I don't think this `empty_block` arg is getting used. Let's remove it!

kclowes (approved):
Looks good to me! I just left the same comment twice :)

PR #1791, opened 2020-11-11
Remove Whisper API

marcgarreau:
### What was wrong?
- Geth removed Whisper in [v1.9.21](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.21)
- Parity/OE removed Whipser in [v3.1.0](https://github.com/openethereum/openethereum/releases/tag/v3.1.0) (and earlier versions, but there's a funky transition in the inherited codebase).

### How was it fixed?
Burn it down :fire:

Since this functionality is outright unavailable from the clients, I don't think a deprecation path makes sense / is worth the effort. Anyone relying on Whisper is probably forced into using an old version of web3.py to begin with.

### Todo:


kclowes (approved):
-2,182 lines, nice! This actually helps me out with filtering too, thanks! 🚢

Issue #1793, opened 2020-11-17
Solidity error parsing doesn't work with Ganache


dmitry-tk:
* Version: 5.13.0
* Python: 3.7.6
* OS: osx
* `pip freeze` output

```
astroid==2.4.2
asyncio==3.4.3
attrdict==2.0.1
attrs==19.3.0
base58==2.0.0
bitarray==1.2.2
bleach==3.1.5
bump2version==1.0.0
bumpversion==0.6.0
certifi==2020.4.5.1
chardet==3.0.4
click==7.1.2
coverage==5.1
cytoolz==0.10.1
docutils==0.16
ECPy==1.2.3
entrypoints==0.3
eth-abi==2.1.1
eth-account==0.5.4
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.1.2
eth-typing==2.2.1
eth-utils==1.9.5
flake8==3.7.9
future==0.18.2
hexbytes==0.2.0
hidapi==0.9.0.post2
idna==2.9
importlib-metadata==1.6.0
ipfshttpclient==0.7.0a1
isort==4.3.21
jsonschema==3.2.0
keyring==21.2.1
lazy-object-proxy==1.4.3
ledgerblue==0.1.31
lru-dict==1.1.6
mccabe==0.6.1
mock==4.0.2
more-itertools==8.3.0
multiaddr==0.0.9
netaddr==0.7.19
nose==1.3.7
packaging==20.4
parsimonious==0.8.1
pathtools==0.1.2
Pillow==7.1.2
pkginfo==1.5.0.1
pluggy==0.13.1
protobuf==3.12.2
py==1.8.1
pycodestyle==2.5.0
pycryptodome==3.9.7
pycryptodomex==3.9.7
pyflakes==2.1.1
Pygments==2.6.1
pylint==2.5.3
pyparsing==2.4.7
pyrsistent==0.16.0
pytest==5.4.2
pytest-cov==2.8.1
python-u2flib-host==3.0.3
PyYAML==5.3.1
Random-Word==1.0.4
readme-renderer==26.0
requests==2.23.0
requests-toolbelt==0.9.1
rlp==1.2.0
-e git+git@github.com:skalenetwork/sgx.py.git@d2a82b04eb788cf8894da0ac4b387a6055c24b37#egg=sgx.py
six==1.15.0
[...]

PR #1794, opened 2020-11-19
Add handler for Ganache revert message format

dmitry-tk:
### What was wrong?

Related to Issue #1793 

### How was it fixed?

The case when `data` field in the RPC response is `dict` was handled, so Ganache revert reasons could be parsed correctly. 

### Todo:

#### Cute Animal Picture

Cute possum here

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -485,8 +485,13 @@ def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
     if not isinstance(response['error'], dict):
         raise ValueError('Error expected to be a dict')
 
-    # Parity/OpenEthereum case:
     data = response['error'].get('data', '')
+
+    # Ganache case:
+    if isinstance(data, dict) and response['error'].get('message'):
+        raise SolidityError(response['error']['message'])
+
+    # Parity/OpenEthereum case:
     if data.startswith('Reverted '):
         # "Reverted", function selector and offset are always the same for revert errors
         prefix = 'Reverted 0x08c379a00000000000000000000000000000000000000000000000000000000000000020'  # noqa: 501
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -489,7 +489,7 @@ def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
 
     # Ganache case:
     if isinstance(data, dict) and response['error'].get('message'):
-        raise SolidityError(response['error']['message'])
+        raise SolidityError(f'execution reverted: {response["error"]["message"]}')
 
     # Parity/OpenEthereum case:
     if data.startswith('Reverted '):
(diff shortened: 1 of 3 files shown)


web3/_utils/method_formatters.py
```diff
@@ -485,8 +485,13 @@ def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
     if not isinstance(response['error'], dict):
         raise ValueError('Error expected to be a dict')
 
-    # Parity/OpenEthereum case:
     data = response['error'].get('data', '')
+
+    # Ganache case:
+    if isinstance(data, dict) and response['error'].get('message'):
+        raise SolidityError(response['error']['message'])
```
kclowes, line 492: I think it would be a good idea to standardize this response with the Geth response. Do you mind adding an `execution reverted` to the beginning of this error message? So:

`raise SolidityError(f"execution reverted: {response['error']['message']}")`

kclowes (commented):
Thanks for the contribution @dmitry-tk! I just left one comment on standardizing the error message with the geth response.

web3/_utils/method_formatters.py
dmitry-tk: thanks for the review, fixed 👌

dmitry-tk:
@kclowes could you please take a look at this failing `ethpm` test?

tmckenzie51:
I was able to rerun it for you @dmitry-tk. Seems to be working now!

dmitry-tk:
@tmckenzie51 @kclowes could this PR be merged now or some other modifications needed?

web3/_utils/method_formatters.py
```diff
@@ -485,8 +485,13 @@ def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
     if not isinstance(response['error'], dict):
         raise ValueError('Error expected to be a dict')
 
-    # Parity/OpenEthereum case:
     data = response['error'].get('data', '')
+
+    # Ganache case:
+    if isinstance(data, dict) and response['error'].get('message'):
```
kclowes, line 491: I believe this will swallow the Geth case, so I think we need to move it below line 507.

kclowes (commented):
@marcgarreau I'm going to leave this for your final approval!

web3/_utils/method_formatters.py
dmitry-tk: In Geth case type of the `data` is `str` if I'm not mistaken.  
And we should perform this check before `if data.startswith('Reverted '):` since it's exactly what causes errors on Ganache (see #1793).

kclowes: Ah, you're right 👍

marcgarreau: Test case(s) should be added to cover the different response types. @dmitry-tk if you have time for this and are comfortable doing it, great - if not, I can take over after I finish another todo.

marcgarreau: Your Ganache test looks good. Looks like we just need to add one for a Geth-specific payload.

marcgarreau (approved):

Issue #1755, opened 2020-09-28
Document `Method` class to allow for extensibility


kclowes:
With the move toward `Method`, we can now allow users to extend the library to include RPC methods other than what's included in web3 by default. We should document how it works so people can use it!

pacrob:
Done!

PR #1797, opened 2020-11-21
[WIP] Document Method

kclowes:
### What was wrong?

Method needed some documentation to be used by outside parties. 

Related to Issue #1755 

### How was it fixed?
Added some documentation!

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/extending_method.rst
+++ b/docs/extending_method.rst
@@ -47,9 +47,11 @@ and make sure it inherits from ``ModuleV2``:
 
 .. doctest::
 
+    >>> from web3 import Web3
     >>> from web3.method import Method
     >>> from web3.module import ModuleV2
-    >>> from web3.auto.infura import w3
+
+    >>> w3 = Web3(Web3.EthereumTesterProvider())
 
     >>> class Eth(ModuleV2):
     ...     def block_id_munger(self, account, block_identifier = None):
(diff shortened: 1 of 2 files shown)


pipermerriam (commented):
I think you should do some experiments on how these APIs could be simplified/streamlined before documenting the current API.  My guess is that we can find a cleaner more intuitive way to do this.

pipermerriam:
Experiments that come to mind:

1. Is there a class-based way to do this.  A base class that requires you override certain methods...
2. Is there a decorator based way that really *simple* methods could be written.
3. Are there utilities that we can write that could simplify method creation.  Can we make a "Recipe" or "template" that could be copy/pasta to simplify method creation.

kclowes:
Yeah, good idea. I was thinking about that as I was documenting it. I'll play around with some of those options! It seems like the base class would be the most straightforward so I'll probably start there first, but a decorator might be nice too 🤔

pipermerriam:
Worth noting that these options don't necessarily require reworking the existing API.  Maybe it makes sense to build them as a layer on top of it?  Or to build it as a layer on top of it and then re-factor the middle part once you know the new layer works...  Otherwise you'll be up against refactoring all the work you just did while also trying to refactor the API itself....

fselmo:
@marcgarreau I think we could use some related material from the web3 customization blog post and revamp the documentation around `Method` and attaching methods here to modernize this issue a bit more. I realized we didn't really update the docs after adding that feature in. Thoughts?

kclowes:
Closing - no longer relevant. There is a nice blog post here: https://snakecharmers.ethereum.org/web3-py-patterns-customizations/

PR #1805, opened 2020-12-09
Remove module

kclowes:
### What was wrong?
Out with the old, in with the new! Got rid of the last `Module` references, in favor of `ModuleV2` which supports async. The next PR, I'll change the ModuleV2 to Module, since it no longer makes sense to have it versioned any more. 

I debated putting the attach_modules function on the web3 class, and ultimately decided to keep it in _utils/module. If anyone has a strong opinion one way or the other, I will happily move it!

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/main.py
+++ b/web3/main.py
@@ -285,7 +285,7 @@ class Web3:
     def enable_unstable_package_management_api(self) -> None:
         from web3.pm import PM  # noqa: F811
         if not hasattr(self, '_pm'):
-            PM.attach(self, '_pm')
+            attach_modules(self, {'_pm': (PM,)})
 
     def enable_strict_bytes_type_checking(self) -> None:
         self.codec = ABICodec(build_strict_registry())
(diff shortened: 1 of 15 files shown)


web3/main.py
```diff
@@ -285,7 +296,34 @@ def pm(self) -> "PM":
...
-            PM.attach(self, '_pm')
+            self.attach_modules({'_pm': (PM,)})
 
     def enable_strict_bytes_type_checking(self) -> None:
         self.codec = ABICodec(build_strict_registry())
+
+    def attach_modules(
+        self,
+        module_definitions: Dict[str, Sequence[Any]],
+        parent: Optional[T] = None
```
pipermerriam, line 307: This typing looks suspect.  Typically `TypeVar` is used to say something like "this function takes a value of type `T` and returns a `Sequence[T]`".  Roughly speaking, it's for establishing a linkage between different arguments and return values.  On it's own, I don't know exactly how it behaves, but I think this is effectively an `Any` type.

kclowes: Oh thanks! That makes sense. It should actually either be a Web3 instance or a Module, so I'll update that!

marcgarreau (approved):
thanks for the tour! newsfragment, then :ship:

PR #1808, opened 2020-12-12
Parity modules to snake_case

ticekev:
First open source contribution. Learning the lay of the land

Parity class modules updated from camelCase to snake_case

Related to Issue #1429

### Todo:

(
image
)


kclowes:
Thanks for the contribution @ticekev! Unfortunately, we need to deprecate these methods before moving them over to snake case, so we'll need to follow the `DeprecatedMethod` pattern that is laid out in #1326. This PR should also build on #1602, so if you want, once that's merged in (should be in the next couple days), then we can switch these to snake case.

kclowes:
Thanks for the work @ticekev! @tmckenzie51 has taken over this task

PR #1809, opened 2020-12-14
[WIP] Handle Empty Revert Message

kclowes:
### What was wrong?
EDIT: I'm going to split this into 2 PRs - one for the bugfix, one for making the error more generic.

This is a backwards incompatible bugfix that returns eth-tester functionality to what it was before 5.13.0 

@fubuloubu pointed out that vyper tests were failing since web3 converts eth-tester's `TransactionFailed` error to a `SolidityError`, and we weren't handling the empty response message quite right. eth-tester couldn't decode the empty error message on a `revert`. I added logic to try and `decode_single` on the error message, and if that fails, just return the plain, undecoded error message. I also discovered that we didn't have any tests that check what happens if the revert error message is empty, so I added some. 

When estimateGas gets reverted and doesn't have an error message, Parity returns a generic error message, so I opted to leave it as a ValueError, instead of catching it and returning a SolidityError, since I'm not sure we can reliably determine if it comes from a `revert` or not.

### How was it fixed?
I did a couple things in this PR:
- Added integration tests for the case where there is no error message that comes back from a revert
- Added a new case for an empty error message in the raise_solidity_error_on_revert function
- Eth-tester returns a TransactionFailed error again
- Made a new `TransactionError` class that inherits from `SolidityError` to make it more generic, without breaking existing functionality.

~Something that I considered doing was changing `SolidityError` to inherit from eth-tester's `TransactionFailed`, but ultimately decided not to put that in so that we didn't have to add the eth-tester library to our dependency list. I could be persuaded that it really should be a `TransactionFailed` however, since it is a breaking change.~

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -9,6 +9,9 @@ import json
 import pytest
 
 import eth_abi
+from eth_tester.exceptions import (
+    TransactionFailed,
+)
 from eth_utils import (
     is_text,
 )
@@ -30,7 +33,6 @@ from web3.exceptions import (
     MismatchedABI,
     NoABIFound,
     NoABIFunctionsFound,
-    SolidityError,
     ValidationError,
 )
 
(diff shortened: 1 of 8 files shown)


kclowes:
I'm going to split this PR into 2. One is #1813, and the other is #1814.

Issue #1758, opened 2020-10-01
Eth 2 Endpoints


kclowes:
### What was wrong?
We'll need to add the Eth2 endpoints so that people can access that data. We'll also need good documentation and it would probably be nice to have a tutorial or two.

### How can it be fixed?
For the initial pass, @pipermerriam and I talked about making a new provider to access the endpoints and then putting the provider behind a flag much like the `enable_unstable_package_management_api` flag. The standardized API is found [here](https://github.com/ethereum/eth2.0-apis)

marcgarreau:
MVP merged in. Work remains for input/output validation, documentation, etc., but we're early to the party, so we'll keep iterating. (Only Teku and Lighthouse support the API so far.)

PR #1810, opened 2020-12-16
Read-only eth2 beacon API

marcgarreau:
### What does it do?

- Introduces a naive eth2 beacon API interface
- Adds smoke tests (outside of the CI pipeline) for testing clients locally
- Related to #1758

### Background

- Looks like there's still some churn with these APIs and they're in various phases of development within each client.
- Most clients seem to have a few additional endpoints unique to their client.
- I used Teku for most of my local testing; a Teku beacon node produces passing tests in these smoke tests.
- The web3.js team has a TypeScript [schema](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth2-beacon/src/schema.ts) up for some of the endpoints, so I used that to standardize on some of the method names.
- Interface:
``` python
from web3.beacon import Beacon
beacon = Beacon("http://localhost:5051")
beacon.get_genesis()
```
- ~I'm still unclear if there's agreement on where all the APIs will be served from. I made an assumption in this PR that the validator endpoints would need to be served by the validator client process, but I'm not seeing this spelled out clearly anywhere. The validator endpoints are not implemented by every client yet.~ Edit: Eth R&D Discord convo revealed that for security reasons, validator clients typically don't expose any http api, so amending for a single beacon url.
- TODO: some input and output validation/optimization, error handling. Some or all could be handled in follow-up PRs if we want to get v1 out quicker.

### Todo:

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -60,6 +60,7 @@ Table of Contents
 
     web3.main
     web3.eth
+    web3.beacon
     web3.pm
     web3.net
     web3.miner
--- a/newsfragments/1758.feature.rst
+++ b/newsfragments/1758.feature.rst
@@ -1,0 +1,1 @@
+Introduce experimental Ethereum 2.0 beacon node API
--- a/tests/beacon/test_beacon.py
+++ b/tests/beacon/test_beacon.py
@@ -1,0 +1,195 @@
+import json
+import pytest
+
+import requests
+
+import requests_mock
+from web3.beacon import (
+    Beacon,
+)
+
+GENESIS_RESPONSE = {
+    "data": {
+        "genesis_time": "1590832934",
+        "genesis_validators_root": (
+            "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
+        ),
+        "genesis_fork_version": "0x00000000",
+    }
+}
+
+
+@pytest.fixture
+def session():
+    session = requests.Session()
+    adapter = requests_mock.Adapter()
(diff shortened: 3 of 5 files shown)


kclowes (commented):
:tada: eth 2!! :tada: I like this as a first pass! 

I had a few thoughts that you can take or leave:
- I think that we'll want to attach this to a w3 instance eventually to have access to things like middleware, the codec, etc
- Docs - Do you think we should add docs in this PR or would you rather add those later? 
- I think it's probable that everyone who is using these APIs knows this, but I wonder if it's worth adding a warning to users that this API is inherently unstable, maybe in Eth2 init?

marcgarreau:
- After chatting in Eth R&D discord, discovered there's no need for the majority of users/use cases to interact with the validator client and most don't expose a public HTTP API for security reasons anyway. Removed those endpoints.
- Renamed `Eth2` to `Beacon` for accuracy and longevity.

@kclowes:
- Added a warning to the import.
- The docs framework required that I add the beacon page to the table of contents, but I'd like to put some real energy into docs in a follow-up PR. 
- Good point, re: middleware and utils, but I'd like to talk through some usage patterns. For example, it seems an awkward hierarchy to instantiate a w3 instance with an eth1 provider and only then having the ability to configure and use a Beacon object.

marcgarreau:
I thought we were playing catch-up, but we're definitely early here. Support in each client is still patchy. Given the "experimental" warning, I feel comfortable merging this in as is and evolving it in tandem with clients.

kclowes (approved):
LGTM! :shipit:

PR #1813, opened 2020-12-17
Remove SolidityError from eth-tester

kclowes:
### What was wrong?

This is a backwards incompatible bugfix that returns eth-tester functionality to what it was before 5.13.0 

@fubuloubu pointed out that vyper tests were failing since web3 converts eth-tester's `TransactionFailed` error to a `SolidityError`, and we weren't handling the empty response message quite right. eth-tester couldn't decode the empty error message on a `revert`. I added logic to try and `decode_single` on the error message, and if that fails, just return the plain, undecoded error message. I also discovered that we didn't have any tests that check what happens if the revert error message is empty, so I added some. 

When estimateGas gets reverted and doesn't have an error message, Parity returns a generic error message, so I opted to leave it as a ValueError, instead of catching it and returning a SolidityError, since I'm not sure we can reliably determine if it comes from a `revert` or not.

### How was it fixed?
I did a couple things in this PR:
- Added integration tests for the case where there is no error message that comes back from a revert
- Added a new case for an empty error message in the raise_solidity_error_on_revert function
- Eth-tester returns a TransactionFailed error again

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/providers/eth_tester/main.py
+++ b/web3/providers/eth_tester/main.py
@@ -9,12 +9,12 @@ from typing import (
 from eth_abi import (
     decode_single,
 )
+from eth_abi.exceptions import (
+    InsufficientDataBytes,
+)
 
 from web3._utils.compat import (
     Literal,
-)
-from web3.exceptions import (
-    SolidityError,
 )
 from web3.providers import (
     BaseProvider,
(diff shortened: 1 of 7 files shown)


marcgarreau (approved):
Looks good; thanks for the repair!

web3/providers/eth_tester/main.py
```diff
@@ -106,11 +106,11 @@ def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
         except TransactionFailed as e:
-            if type(e.args[0]) == str:
-                reason = e.args[0]
-            else:
+            try:
                 reason = decode_single('(string)', e.args[0].args[0][4:])[0]
-            raise SolidityError(f'execution reverted: {reason}')
+            except (InsufficientDataBytes, AttributeError):
+                reason = e.args[0]
+            raise TransactionFailed(f'execution reverted: {reason}')
```
fubuloubu, line 113: 2 of our test cases pick this up, but this can be fixed

I wish there were an ERC for revert reasons and how to handle them in web3 libraries

fubuloubu (approved):
This fixes the issue introduced in v5.13.1 for Vyper (see https://github.com/vyperlang/vyper/issues/2259)

Issue #1812, opened 2020-12-17
Support for Solidity 0.8.x


kclowes:
### What was wrong?
Solidity 0.8 just came out, and we should check to see what we need to update. The Release blog post can be found [here](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/).

### How can it be fixed?
This will likely be an issue that needs to be broken up into a few issues, but for starters, go through the changelog and write some tests for changes. On first glance, here are a few items that we'll probably need to handle:
- updates to `revert` - new error codes and a new `Panic`. EDIT: See comment [here](https://github.com/ethereum/web3.py/pull/1814#discussion_r546875490)
- Make sure we support ABI encoder v2

PR #1814, opened 2020-12-18
Swap SolidityError for ContractLogicError

kclowes:
This is branched off of #1813, so should be merged after.

### What was wrong?
The SolidityError introduced in #1585 should be more generic to account for other languages.

Related to Issue #1813, #1585

### How was it fixed?
Swapped out `SolidityError` for a `ContractLogicError`. `ContractLogicError` inherits from the SolidityError for backwards compatibility, and we'll remove `SolidityError` in v6. Although the history doesn't reflect this any more, I made the change in the code without making any changes to the tests to make sure we had backwards compatibility, then made the change in the tests. 

### Todo:
- [x] Cleanup commit history

#### Cute Animal Picture
image

### DIFF
--- a/web3/exceptions.py
+++ b/web3/exceptions.py
@@ -197,6 +197,15 @@ class InvalidEventABI(ValueError):
 class SolidityError(ValueError):
     # Inherits from ValueError for backwards compatibility
     """
-    Raised on a solidity require/revert
+    Raised on a contract revert error
+    """
+    pass
+
+
+class ContractLogicError(SolidityError, ValueError):
+    # Inherits from ValueError for backwards compatibility
+    # TODO: Remove SolidityError inheritance in v6
+    """
+    Raised on a contract revert error
     """
     pass
--- a/newsfragments/1814.bugfix.rst
+++ b/newsfragments/1814.bugfix.rst
@@ -1,0 +1,2 @@
+Added a new ``ContractLogicError`` for when a contract reverts a transaction.
+``ContractLogicError`` will replace ``SolidityError``, in v6.
--- a/tests/core/utilities/test_method_formatters.py
+++ b/tests/core/utilities/test_method_formatters.py
@@ -8,7 +8,7 @@ from web3._utils.rpc_abi import (
     RPC,
 )
 from web3.exceptions import (
-    SolidityError,
+    ContractLogicError,
 )
 from web3.types import (
     RPCResponse,
@@ -94,7 +94,7 @@ GANACHE_RESPONSE = RPCResponse({
         'test_get-ganache-revert-reason',
     ])
 def test_get_revert_reason(response, expected) -> None:
-    with pytest.raises(SolidityError, match=expected):
+    with pytest.raises(ContractLogicError, match=expected):
         raise_solidity_error_on_revert(response)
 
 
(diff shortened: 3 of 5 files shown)


web3/providers/eth_tester/main.py
```diff
@@ -106,11 +106,11 @@ def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
         except TransactionFailed as e:
-            if type(e.args[0]) == str:
-                reason = e.args[0]
-            else:
+            try:
                 reason = decode_single('(string)', e.args[0].args[0][4:])[0]
-            raise SolidityError(f'execution reverted: {reason}')
+            except (InsufficientDataBytes, AttributeError):
+                reason = e.args[0]
+            raise TransactionFailed(f'execution reverted: {reason}')
```
marcgarreau, line 113: The `TransactionError` vs. `TransactionFailed` distinction is a little fuzzy. If it happens in eth-tester, its the latter?

kclowes: Yeah, agreed. `TransactionError` comes from web3, whereas `TransactionFailed` is from eth-tester. I'm not opposed to making them both `TransactionFailed`, but thought it would be confusing to have them both be the same thing.

marcgarreau: My first instinct is to pick one generic error for both libs if they're used in the same context. (Might be wise to check what else can trigger this exception in eth-tester first.) Adopting eth-tester's `TransactionFailed` in web3 seems reasonable to me on the surface, though.

kclowes: Yeah, I'm good with that. It looks to me like eth-tester is using it in the same context. It looks like it's used for `revert` and `invalid` opcodes on `call` and `estimateGas` AFAICT.

fubuloubu: Why not `AssertionError` since it stems from a `REVERT` opcode, which usually in context is from an `assert`/`require` statement triggering (for any EVM language)?

Should add a test using a v0.8.0 Solidity contract or Vyper contract with checked arithmetic, since that's not a user revert but the language implementing a logical revert (that doesn't stem from an EVM exception)

kclowes: After polling the team, we landed on a `RevertTransaction` since that's the opcode. I added an explicit comment to #1812 to check for the logical revert. Thanks @fubuloubu!

web3/exceptions.py
```diff
@@ -197,6 +197,15 @@ class InvalidEventABI(ValueError):
 class SolidityError(ValueError):
     # Inherits from ValueError for backwards compatibility
     """
-    Raised on a solidity require/revert
+    Raised on a contract revert error
+    """
+    pass
+
+
+class RevertTransaction(SolidityError, ValueError):
```
fubuloubu, line 205: I wonder if something more like `ContractError` or `ContractLogicException` would be more descriptive

marcgarreau: We've done an admirable job of bike shedding on this one, but I do find it appealing to lump all asserts, reverts, requires, and related future keywords into one exception type.

fubuloubu: Yeah, it makes sense to me because it's sort of like "this is an intended application exception" vs. "this is an unexpected VM exception"

kclowes: I'm good with that. Will change in a bit!

marcgarreau (approved):
the sweet satisfaction of a well-named thing :pray:

(hopefully this ages well 😅 )

PR #1823, opened 2020-12-30
Fix missing argument in Eth.signTransaction documentation

DrankoLQ:
### What was wrong?
The documentation for the `Eth.signTransaction` method had a typo: an argument was missing

### How was it fixed?
Update the documentation to correctly show how to call `signTransaction` method

### Todo:


kclowes (approved):
Thank you!

PR #1827, opened 2021-01-08
Add requests-mock dependency to dev requirements

kclowes:
### What was wrong?
`requests-mock` is used in the beacon tests and I found I couldn't run my tests locally from the top-level `tests/` dir without it installed. 

### How was it fixed?
Added requests-mock dependency to setup.py. 

@marcgarreau let me know if this is the wrong solution!

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/tests/beacon/test_beacon.py
+++ b/tests/beacon/test_beacon.py
@@ -1,45 +1,15 @@
-import json
 import pytest
 
 import requests
 
-import requests_mock
 from web3.beacon import (
     Beacon,
 )
 
-GENESIS_RESPONSE = {
-    "data": {
-        "genesis_time": "1590832934",
-        "genesis_validators_root": (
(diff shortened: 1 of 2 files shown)


marcgarreau:
Ah, since these tests are not included in CI and are only intended as local smoke tests, perhaps the more appropriate route is to comment out the import or remove the mock template code altogether. No strong feels, but I think I favor the latter.

tests/beacon/test_beacon.py
```diff
@@ -1,45 +1,19 @@
...
-    )
-    return session
-
 
 @pytest.fixture
 def beacon(session):
     #  return Beacon(
     #  beacon_url="mock://example.com/api",
     #  session=session,
     #  )
```
marcgarreau, line 15: thanks for cleaning up. this comment and the `session` arg in this func definition can get removed too 👍

marcgarreau (approved):

PR #1833, opened 2021-01-11
Mark flaky ethpm test as flaky

kclowes:
### What was wrong?
We have a test in the ethpm module that consistently runs into a timeout error during testing, especially when it gets run over and over again, like for small PR changes.

### How was it fixed?
The PR pretty much speaks for itself, but added an xfail that specifies what error it's all right to fail with. If it fails with a different error, it will show up in the test output and fail the build.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/1833.misc.rst
+++ b/newsfragments/1833.misc.rst
@@ -1,0 +1,1 @@
+Mark ethpm's local IPFS backend test as flaky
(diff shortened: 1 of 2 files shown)


marcgarreau (approved):

njgheorghita (approved):
🚢

PR #1836, opened 2021-01-14
Introduce beacon API docs

marcgarreau:
### What was wrong?
Beacon docs were a TODO. 

### How was it fixed?
Here's a start that includes quick setup instructions and methods + response examples. Will continue to iterate on it.

### Todo:


kclowes (approved):
:shipit: Looks good!

Issue #1842, opened 2021-01-20
ipc socket timeout unrecoverable doing unlock_account on a local node


joseluu:
* Version: 5.15.0
* Python: 3.6
* OS: linux
* `pip freeze` output

```
attrdict==2.0.1
attrs==20.3.0
base58==2.1.0
bitarray==1.2.2
certifi==2020.11.8
chardet==3.0.4
click==7.1.2
cytoolz==0.11.0
eth-abi==2.1.1
eth-account==0.5.4
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.2.1
eth-typing==2.2.2
eth-utils==1.9.5
Flask==1.1.2
hexbytes==0.2.1
idna==2.10
importlib-metadata==3.4.0
importlib-resources==5.1.0
ipfshttpclient==0.7.0a1
itsdangerous==1.1.0
Jinja2==2.11.2
jsonschema==3.2.0
lru-dict==1.1.6
MarkupSafe==1.1.1
multiaddr==0.0.9
netaddr==0.8.0
numpy==1.19.5
pandas==1.1.5
parsimonious==0.8.1
pkg-resources==0.0.0
protobuf==3.14.0
pycryptodome==3.9.9
pyrsistent==0.17.3
python-dateutil==2.8.1
pytz==2020.5
requests==2.24.0
rlp==2.0.0
six==1.15.0
toolz==0.11.1
typing-extensions==3.7.4.3
urllib3==1.25.11
varint==1.0.2
web3==5.15.0
websockets==8.1
Werkzeug==1.0.1
zipp==3.4.0
```

### What was wrong?

Please include any of the following that are applicable:

* The code which produced the error
*w3.geth.personal.unlockAccount(QAXH_ADDRESS, QAXH_PASSWORD)
* The full output of the error
Traceback (most recent call last):
  File "/var/www/test_id/plateform/venv/lib/python3.6/site-packages/web3/providers/ipc.py", line 246, in make_request
    raw_response += sock.recv(4096)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/www/test_id/plateform/venv/lib/python3.6/site-packages/flask/app.py", line 2464, in _call_
    return self.wsgi_app(environ, start_response)
  File "/var/www/test_id/plateform/venv/lib/python3.6/site-packages/flask/app.py", line 2450, in wsgi_app
[...]

PR #1843, opened 2021-01-20
fix ipc socket timeout doing unlock_account

joseluu:
### What was wrong?

Related to Issue #
https://github.com/ethereum/web3.py/issues/1842

### How was it fixed?
Race condition between 2 timeouts set at same value
* the socket timeout
* the guard timeout at line 244 "with Timeout..."
then the socket timeout expires before the guard timeout the socket.timeout exception is triggered which in turn attempts to make the guard timeout expire which raises an exception within the exception handling

The solution I propose is to start a sleep in the exception handler  which will be (very) soon interrupted by the guard timeout.
### Todo:

#### Cute Animal Picture

https://images.app.goo.gl/W8VHpCHRePnu45WVA

### DIFF
--- a/web3/contract/async_contract.py
+++ b/web3/contract/async_contract.py
@@ -58,6 +58,7 @@ from web3.contract.utils import (
     async_estimate_gas_for_function,
     async_transact_with_contract_function,
     find_functions_by_identifier,
+    get_function_by_identifier,
 )
 from web3.types import (  # noqa: F401
     ABI,
@@ -190,6 +191,12 @@ class AsyncContract(BaseContract):
         return find_functions_by_identifier(  # type: ignore
             contract_abi, w3, address, callable_check, AsyncContractFunction
         )
+
+    @combomethod
+    def get_function_by_identifier(
+        cls, fns: Sequence[BaseContractFunction], identifier: str
+    ) -> BaseContractFunction:
+        return get_function_by_identifier(fns, identifier)
 
 
 class AsyncContractConstructor(BaseContractConstructor):
--- a/web3/contract/base_contract.py
+++ b/web3/contract/base_contract.py
@@ -99,7 +99,7 @@ from web3.exceptions import (
     NoABIEventsFound,
     NoABIFound,
     NoABIFunctionsFound,
-    ValidationError,
+    Web3ValidationError,
 )
 from web3.logs import (
     DISCARD,
@@ -218,7 +218,7 @@ class BaseContract:
         fns = self.find_functions_by_identifier(
             self.abi, self.w3, self.address, callable_check
         )
-        return get_function_by_identifier(fns, "signature")
+        return self.get_function_by_identifier(fns, "signature")
 
     @combomethod
     def find_functions_by_name(self, fn_name: str) -> "BaseContractFunction":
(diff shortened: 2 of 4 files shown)


kclowes:
Thanks for the contribution @joseluu! Based on the problem and solution you describe, this looks like a good change. Before I merge though, I'll need to either have an automated test for this change or at the very least I'll need to manually test. Is there a way I can see it fail even intermittently?

joseluu:
I am not sure as to why no one noticed it as code itself  is rather plain:
-unlock
-send tx
-lock
Without the change, it happens here most of the time (>50% occurence)

Specifics:
The code runs within a flask server
The configuration is:
google VM n1-standard-4 (4 vCPUs, 15 GB memory)
local geth in fast mode
There is another python thread running a block filter w3.eth.filter('latest'), handling the events and sleeping

joseluu:
To reproduce, try a much shorter socket timeout like 0.001 at  line 34: def get_ipc_socket

perusing the code, I am realizing that it is not quite a race condition between two equal timeouts, it is a condition where the default value for the socket timeout is 0.1s and the value for the guard timeout is 10s, therefore socket timeout *will* happen before when the geth server is busy.

joseluu:
Because it is not a race condition, I submitted a better and simpler fix which is simply to continue the loop doing socket.recv until either something good is received of the guard timeout expires.

Testing it here shows a performance improvement from the user point of vue.

web3/providers/ipc.py
```diff
@@ -245,7 +245,6 @@ def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
                     try:
                         raw_response += sock.recv(4096)
                     except socket.timeout:
-                        timeout.sleep(0)
```
pipermerriam, line 248: Without this sleep, the the `with Timeout(...) as timeout` block won't get a chance to raise the actual timeout.  This could result in an infinite loop if the socket continually times out on the call to `socket.recv(...)` as the context manager would never get a chance to enforce the timeout.

I don't think removing this timeout is appropriate.

joseluu:
timeout.sleep(0) raises an exception within the exception handler which is
the initial problem , what about time.sleep(0)  ?

pipermerriam:
The only way that the timeout context raises an exception is via a call to either `timeout.sleep(...)` or `timeout.check()` so we have to preserve one of those calls.  I don't actually understand the core problem you're trying to solve here?  What is wrong with an exception being raised from inside the `try/except` handler?

joseluu:
The main point I am trying to solve is the socket timeout being 0.1s while
the "with Timeout" is 10s, so when the the geth server is slow to answer
the IPC requests, the socket times out and by doing timeout.sleep(0) we are
ejected out of the context without getting the response, in addition we
have a bad looking stack trace (exception within exception handler) and in
our case the eth transaction fails and the application is stuck.
What I would like is for this function to get the chance to complete the
IPC call.

This can be reproduced by making the socket timeout 0.001s at line 38 or
forcing geth to be slow at answering.

pipermerriam:
I must be missing something.  Is there a reason why you cannot do `provider = IPCProvider(ipc_path, timeout=30)` to just extend the total timeout time?

joseluu:
The probleme comes from the shorter timeout of the socket itself see line
43:  sock.settimeout(timeout) timeout is not the same as in the "with
Timeout" construct but 0.1s
the sock.recv exits after 0.1s when the geth node is slow/busy but we need
to restart other sock.recv (meaning we should remain in the loop) until
something good is received within the timeout (the 10s timeout of the "with
Timeout")

Jay54520:
Is the problem fixed? I entered this problem when my geth is slower because of receiving more requests.

PR #1860, opened 2021-01-28
document contract function struct args

marcgarreau:
### What was wrong?
I don't believe the docs capture function input types anywhere - specifically how to pass in a Solidity struct was a question asked recently in Gitter.

### How was it fixed?
Add a nested struct example within a new Contract FAQ section.

### Todo:


docs/contracts.rst
```diff
@@ -1211,3 +1212,57 @@ provides methods to interact with contract functions.
...
+You can interact with Web3.py contract API as follows:
+
+.. code-block:: python
+
+   # deploy or lookup the deployed contract, then:
+
+   >>> deployed_contract.functions.retrieve().call()
+   '0x0000000000000000000000000000000000000000'
+
+   >>> deployed_contract.functions.update({'s1': ['0x0000000000000000000000000000000000000001', '0x0000000000000000000000000000000000000002'], 'S2': [b'', b''], 'users': []}).transact()
```
kclowes, line 1265: Do you have `s1` and `S2` capitalized differently here on purpose? 

And then for the S2 `bytes32` args - in v6 we're planning on making it an error to pass in less than the specified number of bytes, so may be worth passing in 32 bytes here 🤷

kclowes (approved):
Just a few nits! LGTM!

docs/contracts.rst
marcgarreau: good catch :)

Issue #728, opened 2018-03-29
Add a "Common Problems" page to the documentation


pipermerriam:
### What was wrong?

There are certain issues that are encountered regularly by users.  

### How can it be fixed?

We should add a "Common Problems" page to the documentation which tries to capture these succinctly and provides answers/solutions.

- using `web3.eth.sendTransaction` when connected to infura
- trying to access `web3.personal` API when using parity or when the `personal` API is not enabled with geth.
- trying to call methods off of the `Web3` class rather than an instance of the class.
- improper instantiation of contract instances like trying to call functions off of the contract class.

marcgarreau:
Finally closing this ancient issue after making a couple doc tweaks. I like the pattern of adding an FAQ to the bottom of specific guides, e.g., contracts, otherwise questions that don't fit nicely into one bucket wind up in `Troubleshooting`.

PR #1865, opened 2021-02-01
troubleshooting docs for instance configuration

marcgarreau:
### What does it do?
Adds troubleshooting docs for the case where someone is using the Web3 class vs. configuring an instance.

Related to Issue #728

### Todo:


kclowes (approved):
Sweet! :shipit:

PR #1884, opened 2021-02-23
fixed import eth_account

99Kies:
### What was wrong?
```bash
ImportError: cannot import name 'Account' from 'eth_account' 
```
Related to Issue #

### How was it fixed?
Account should be import like this:
from eth_account.account import Account

### Todo:

#### Cute Animal Picture


kclowes:
Thanks for taking the initiative! However, you should be able to import `Account` directly from eth_account. If you can't, please raise an issue and we can help troubleshoot!

Issue #1892, opened 2021-03-04
add package to conda-forge


spinoch:
pip environments are easily broken, (conda)[https://docs.conda.io/] solves most of these issues by tracking package versions and their dependencies to prevent broken environments

All of `web3`'s dependencies are already on `conda` except for `pypiwin32`. Adding `web3` to conda-forge would be trivial if it were not for this dependency. I am not entirely aware what the value is of having `pypiwin32` as a hard requirement for web3, could it be an optional requirement or if it's no longer used, removed altogether?

kclowes:
It's for windows users to allow them to use the package, so we can't delete it. However, it's pretty outdated AFAIK, and there is a PR up to swap it out for `pywin32` that keeps getting deprioritized. I don't use conda, but it looks like pywin32 is on conda-forge, is that right? If so, I can work on spinning up a Windows virtualenv to test/review the PR.

spinoch:
Thanks for clarifying.

> t looks like pywin32 is on conda-forge, is that right?

That's right: https://anaconda.org/conda-forge/pywin32

>  If so, I can work on spinning up a Windows virtualenv to test/review the PR.

That'd be great

spinoch:
I'll close this as the conda-forge recipe already exists

PR #1907, opened 2021-03-11
Add windows CI run, and update circleci version

kclowes:
### What was wrong?
We didn't have any windows testing in CI. 

Related to Issue #1892, #166, #1742

### How was it fixed?
Added windows to our CI. I couldn't figure out how to test against different versions of Python in windows, so we're currently only testing against Python 3.7.

Also did some caching cleanup, and had to update the CircleCI version.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2.0
+version: 2.1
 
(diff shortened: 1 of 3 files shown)


marcgarreau (approved):
Thanks for diving down the rabbit hole 🐰 

Checking my understanding, the wheel tests are primarily a smoke test to verify that the lib will build for a given environment and can be imported? If accurate, looks like a good checkpoint.

kclowes:
@marcgarreau Yep, exactly. Making sure we don't have any errors when installing in a fresh env and that we can import our web3 instance!

PR #1909, opened 2021-03-14
Update contributing.rst

miohtama:
Doc update: In some shells (zsh), `[]` are reserved characters, so we need to escape them. The old example only works in Bash.

#### Cute Animal Picture

```
\|/          (__)    
     `\------(oo)
       ||    (__)
       ||w--||     \|/
   \|/
```


marcgarreau (approved):
Good catch, thanks!

PR #1933, opened 2021-04-05
Add existing contract interaction to docs example

marcgarreau:
### What was wrong?

Most folks are looking to interact with an existing contract, so the example with a full deployment is overkill.

### Todo:


kclowes (approved):
Sweet! LGTM!

PR #1937, opened 2021-04-07
Move _block_number to get_block_number

### LINKED ISSUE #1875: Getting the block_number asynchronously is hacky
* Version: 5.16.0
* Python: 3.8.5

### What was wrong?

PR #1851 `Moving blockNumber -> block_number` replaced a function with a `@property`.

Since that peoperty does RPC under the hood, which blocks on the network, using a property here does not seem intuitive, but more importantly it makes it more difficult to get that information in a non-blocking way.

This is an issue when running this in a Thread, Process or Asyncio eventloop. In an asyncio process for example, one would have to do the following replacement:
```diff
- return await loop.run_in_executor(None, w3.eth.block_number)
+ return await loop.run_in_executor(None, getattr, (w3.eth, 'block_number'))
```

or alternatively to call the underlying method, which the underscore marks as private and best practices recommend against using:
```python
- return await loop.run_in_executor(None, w3.eth.block_number)
- return await loop.run_in_executor(None, w3.eth._block_number)
```

### How can it be fixed?

The underlying methods could be made public instead of private:
```diff
-    _block_number: Method[Callable[[], BlockNumber]] = Method(
+    get_block_number: Method[Callable[[], BlockNumber]] = Method(
        RPC.eth_blockNumber,
        mungers=None,
    )

    @property
    def block_number(self) -> BlockNumber:
-        return self._block_number()
+        return self.get_block_number()
```

Alternatively, an async API on web3 would be great, since parts of it already such as WebsocketProvider use asyncio under the hood.

kclowes:
### What was wrong?
`block_number` was a property, then made a public method, then made a private method, which introduced some confusion. This PR adds the `get_block_number` public Method. 

Closes #1875 

### Todo:

#### Cute Animal Picture

image


marcgarreau (approved):
🚢

PR #1953, opened 2021-04-23
Remove an if statement

Harmouch101:
Signed-off-by: Harmouch101 <mahmoudddharmouchhh@gmail.com>

### What was wrong?
unnecessary if statement in a manifest file.
Related to Issue #
no issues.
### How was it fixed?
by understanding the logic of the if statement.
```
                                if condition    return value of the func
manifest = {'deployments': None}  --->  True  ---> False
manifest = {'deployments': 'something'}  --->  False  ---> True
manifest = {'not_deployments': 'something'}  --->  True  ---> False

final statement

"deployments" in manifest and bool(manifest.get("deployments", None))
```
### Todo:

#### Cute Animal Picture

cute little cat

### DIFF
--- a/ethpm/validation/manifest.py
+++ b/ethpm/validation/manifest.py
@@ -91,9 +91,7 @@ def validate_manifest_against_schema(manifest: Dict[str, Any]) -> None:
 
 
 def check_for_deployments(manifest: Dict[str, Any]) -> bool:
-    if "deployments" not in manifest or not manifest["deployments"]:
-        return False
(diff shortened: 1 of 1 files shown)


Harmouch101:
Wait, What! `get()` is slower than []. I've learned something today. Noice!

pipermerriam:
Unfortunately, this reduces readability and we won't be merging this patch.  The logic is much easier to reason about in the expanded form.

PR #1960, opened 2021-04-26
Add trace_filter, deprecate traceFilter

tmckenzie51:
### What was wrong?
add trace_Filter, deprecate traceFilter

Related to Issue #1429

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1960.feature.rst
+++ b/newsfragments/1960.feature.rst
@@ -1,0 +1,1 @@
+Add trace_filter deprecate traceFilter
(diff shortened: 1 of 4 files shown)


kclowes (commented):
In `tests/integration/parity/common.py`, the trace_filter tests are marked as `xfail`, so even though we're not running our parity tests right now, we'll want to add the deprecated test to the common file and mark it as xfailing too. Let me know if that doesn't make sense. 

Looks good other than that!

kclowes (approved):
Looks good after tests pass! 🚀

Issue #1968, opened 2021-04-28
Broken docs styles


marcgarreau:
### What was wrong?

Some of the styling in the docs broke recently.

For example, landing page should look like: 

<img width="1101" alt="Screen Shot 2021-04-28 at 12 45 41 PM" src="https://user-images.githubusercontent.com/3621728/116456747-d2432380-a81f-11eb-9712-c8f59bc0e320.png">

but instead looks like:

<img width="1090" alt="Screen Shot 2021-04-28 at 12 45 52 PM" src="https://user-images.githubusercontent.com/3621728/116456778-dcfdb880-a81f-11eb-9648-2aaf4ffba1c1.png">

### How can it be fixed?

Unsure yet, but building the docs locally works fine (`make docs`).

tserg:
I tried running `make doctest` locally and got the following errors:

> WARNING: Failed to get a method signature for ethpm.Package.build_dependencies: <ethpm._utils.cache.cached_property object at 0x7f50b9674ac0> is not a callable object
WARNING: Failed to get a method signature for ethpm.Package.deployments: <ethpm._utils.cache.cached_property object at 0x7f50b9674b20> is not a callable object

Making the following changes to `ethpm.rst` resolves the errors:

Original
```
.. autoclass:: ethpm.Package
   :members: name, version, manifest_version, uri, __repr__, contract_types, build_dependencies, deployments
```
Amended

```
.. autoclass:: ethpm.Package
   :members: name, version, manifest_version, uri, __repr__, contract_types

   .. automethod:: build_dependencies(self)
   .. automethod:: deployments(self)
```

I am not sure if this is causing the errors though.

kclowes:
This has been fixed in #1977!

Issue #1700, opened 2020-08-06
Docs fail to build 5.12.0


nolash:
* Version: 5.12.0
* Python: 3.8.4
* Sphinx: 3.1.2
* OS: linux 5.7.0

* `pip freeze` output

```
https://gateway.ethswarm.org/files/4f052600b57618886473a6665a1f8d971a379be7fb0af4ed8807bc5bba6f1884
```

```
reading sources... [100%] web3.pm                                                                                                                                                                                                                                              

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/usr/lib/python3.8/site-packages/sphinx/application.py", line 348, in build
    self.builder.build_update()
  File "/usr/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 297, in build_update
    self.build(to_build,
  File "/usr/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 311, in build
    updated_docnames = set(self.read())
  File "/usr/lib/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
  File "/usr/lib/python3.8/site-packages/sphinx/util/logging.py", line 213, in pending_warnings
    memhandler.flushTo(logger)
  File "/usr/lib/python3.8/site-packages/sphinx/util/logging.py", line 178, in flushTo
    logger.handle(record)
  File "/usr/lib/python3.8/logging/__init__.py", line 1587, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.8/logging/__init__.py", line 1649, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.8/logging/__init__.py", line 946, in handle
    rv = self.filter(record)
  File "/usr/lib/python3.8/logging/__init__.py", line 807, in filter
    result = f.filter(record)
  File "/usr/lib/python3.8/site-packages/sphinx/util/logging.py", line 421, in filter
    raise exc
sphinx.errors.SphinxWarning: /home/lash/src/ext/ethereum/web3.py/docs/contracts.rst:243:duplicate object description of web3.contract.Contract.constructor, other instance in contracts, use :noindex: for one of them

Warning, treated as error:
/home/lash/src/ext/ethereum/web3.py/docs/contracts.rst:243:duplicate object description of web3.contract.Contract.constructor, other instance in contracts, use :noindex: for one of them
[...]

Issue #1644, opened 2020-05-08
Upgrade Sphinx to v3.0+


kclowes:
* Version: 5.9+

### What was wrong?
The Sphinx dependency wasn't pinned and when Sphinx 3.0 came out some of our doctests broke. It's currently pinned to  >=2.4.4,<3. 

### How can it be fixed?
Pin Sphinx between 3.0 and 4, and then fix any failing tests.

rajeev29five:
I would like to work on this.

dbfreem:
This issue appears to already be fixed in this [PR](https://github.com/ethereum/web3.py/pull/1977/files). Sphinx was upgraded in the [setup.py](https://github.com/ethereum/web3.py/blob/master/setup.py#L27). This issue can probably be closed.

kclowes:
Yep, you're right @dbfreem. Closing!

PR #1969, opened 2021-04-29
Sphinx Upgrade

Harmouch101:
Signed-off-by: Harmouch101 <mahmoudddharmouchhh@gmail.com>

### What was wrong?

Related to Issues: #1644 & #1968 & #1700

### How was it fixed?

By solving each error generated when upgrading to sphinx >=3.0. It is not the perfect fix. However, the docs files would be generated successfully. You can merge this branch as a quick fix for the docs. In the meantime, I am working on figuring out a better solution.

### Todo:

#### Cute Animal Picture

🐱


marcgarreau:
Thanks for the PR! The affected methods in the `contracts.rst` file get omitted from the docs altogether; I'll see if I can sort that out. FYI - [this link](https://web3py--1969.org.readthedocs.build/en/1969/contracts.html#methods) is handy for seeing the build preview.

Harmouch101:
Hey @marcgarreau,
I have updated the PR. Now, the docs look clean!

ethpm/_utils/cache.py
```diff
@@ -17,8 +17,9 @@ class cached_property:
     def __init__(self, func: Callable[..., Any], name: str = None) -> None:
         self.func = func
         self.name = name or func.__name__
+        self.__doc__ = getattr(func, '__doc__')
         # type ignored b/c self is 'cached_property' but expects 'Callable[..., Any]'
-        functools.update_wrapper(self, func)  # type: ignore
+        functools.update_wrapper(func, self)  # type: ignore
```
marcgarreau, line 22: why were these args swapped?

Harmouch101: I've updated the PR. Since building the docs generates the following error message:
```sh
sphinx.errors.SphinxWarning: Failed to get a method signature for ethpm.Package.build_dependencies: <ethpm._utils.cache.cached_property object at 0x7f710d9f1b20> is not a callable object
```
I have fixed it to make the object callable by adding the __call__ func.
Because cached_property is a wrapper object, you are right. we need to keep the order of the parameters(wrapper, wrapped).

Harmouch101: Actually, this class doesn't work at all. It is supposed to convert the returned value of a given function to an instance whose attribute holds that value. I have figured out a solution that works on python >= 3.6 as follows:

```python
from typing import (
    Any,
    Callable,
)

class cached_property:
    """
    Decorator that converts a method with a single self argument into a
    property cached on the instance.

    Optional ``name`` argument allows you to make cached properties of other
    methods. (e.g.  url = cached_property(get_absolute_url, name='url') )
    """

    def __init__(self, func: Callable[..., Any], name: str = None) -> None:
        self.wrapped_func = func
        self.name = name
        self.__doc__ = getattr(func, '__doc__')

    def __get__(self, instance: Any, cls: Any = None) -> Any:
        """
        Call the function and put the return value in instance.__dict__ so that
        subsequent attribute access on the instance returns the cached value
        instead of calling cached_property.__get__().
        """
        if instance is None:
            return self
        res = instance.__dict__[self.name] = self.func(instance)
        return res

    def __set_name__(self, cls: Any = None, name: str = None) -> None:
        """
        The function is called at the time the cls class is created.
        The descriptor would be assigned to name.
        """
        if self.name is None:
            self.name = name
            self.func = self.wrapped_func
        if name != self.name:
            raise TypeError(
                "Unable to assign cached_property for two different names "
                f"(%{self.name} and {name})."
            )
    @staticmethod
    def func():
        raise TypeError(
            'Unable to use an instance of this class without invoking __set_name__!'
        )

import random
class Bar:
    foo = cached_property(lambda _: random.random())

if __name__ == "__main__":
    #print(cached_property.func())
    bar = Bar()
    print(bar.foo)  # 0.6638420043967768
    print(bar.foo)  # 0.6638420043967768
    bar = Bar()
    print(bar.foo)  # 0.5112941702023637
    print(bar.foo)  # 0.5112941702023637
```

Harmouch101: i will commit these changes

Harmouch101: Metaprogramming is lit!

PR #1971, opened 2021-05-01
Docs: Explicitly note solidityKeccak performs a packed encoding.

gnattishness:
### What was wrong?

The docs diddn't explicitly state what type of encoding `web3.solidityKeccak` performs.
As of Solidity v0.5.x, `keccak256` doesn't do any automatic encoding.
People less familiar with Solidity v0.4.x could find the description ambiguous.

### How was it fixed?

Added to relevant docs section.

I also thought it'd be nice to add examples of the equivalent solidity (which should be valid for v0.5.x - v0.8.x at least)

----

Thanks! Let me know if there's any quibbles or things I can adjust.

### DIFF
--- a/docs/web3.main.rst
+++ b/docs/web3.main.rst
@@ -33,7 +33,7 @@ Attributes
     Returns the current Web3 version.
 
     .. code-block:: python
-     
+
        >>> web3.api
        "4.7.0"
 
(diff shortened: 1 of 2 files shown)


marcgarreau (approved):
Thanks for the additional context! I'll add a newsfragment and any other wordsmithing, then merge it in.

PR #1980, opened 2021-05-06
Update mypy to 0.812

kclowes:
### What was wrong?
mypy v0.730 was giving me syntax errors with python 3.9. 

### How was it fixed?
Updated to the most recent version and fixed any problems.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/middleware/stalecheck.py
+++ b/web3/middleware/stalecheck.py
@@ -26,7 +26,10 @@ SKIP_STALECHECK_FOR_METHODS = set([
 
 
 def _isfresh(block: BlockData, allowable_delay: int) -> bool:
-    return block and time.time() - block['timestamp'] <= allowable_delay
+    if block and (time.time() - block['timestamp'] <= allowable_delay):
+        return True
+    else:
+        return False
 
 
 def make_stalecheck_middleware(
--- a/web3/pm.py
+++ b/web3/pm.py
@@ -389,13 +389,13 @@ class PM(Module):
             addr_lookup = self.web3.ens.address(str(address))
             if not addr_lookup:
                 raise NameNotFound(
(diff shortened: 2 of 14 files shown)


web3/middleware/stalecheck.py
```diff
@@ -26,7 +26,10 @@
 
 
 def _isfresh(block: BlockData, allowable_delay: int) -> bool:
-    return block and time.time() - block['timestamp'] <= allowable_delay
```
marcgarreau, line 29: mypy didnt like this with parens?

web3/pm.py
```diff
@@ -389,13 +389,13 @@ def set_registry(self, address: Union[Address, ChecksumAddress, ENS]) -> None:
             addr_lookup = self.web3.ens.address(str(address))
             if not addr_lookup:
                 raise NameNotFound(
-                    "No address found after ENS lookup for name: {0}.".format(address)
+                    f"No address found after ENS lookup for name: {address!r}."
```
marcgarreau, line 392: insisted on `repr` over `str`?

marcgarreau (approved):
picky mypy 📏

web3/pm.py
kclowes: Yeah, we use bytestrings and strings interchangeably, and mypy wants to make sure that's intentional.

web3/middleware/stalecheck.py
kclowes: I suspect the problem was the `return block` because it was complaining that the return value could either be `BlockData` or a `bool`, but we only specify bool. I don't love the explicit `return True/False` but it is explicit and it makes mypy happy 🤷

PR #1993, opened 2021-05-12
Allow both 0.7.0 and 0.7.0a1 for ipfshttpclient

brunneis:
### What was wrong?
The `ipfshttpclient` version was fixed to `0.7.0a1`. However, `0.7.0` was released already with "No changes compared to 0.7.0a1" (https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/master/CHANGELOG.md).

> py-ipfs-http-client 0.7.0 (15.03.2021)
> * No changes compared to 0.7.0a1 – breaking changes delayed to the unknown future

Since the version is fixed, web3.py cannot be installed when `ipfshttpclient==0.7.0` is installed.

### How was it fixed?
Under `setup.py`, `install_requires` was updated so both versions are supported.

### DIFF
--- a/newsfragments/1993.misc.rst
+++ b/newsfragments/1993.misc.rst
@@ -1,0 +1,1 @@
+Update ipfshttpclient to use v0.7.0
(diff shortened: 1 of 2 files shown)


kclowes:
Thanks for the contribution @brunneis! Since there aren't any changes between 0.7.0a1 and 0.7.0, let's just specify 0.7.0. I'll update and merge!

Issue #1988, opened 2021-05-10
Session Caching in Async HTTP Provider


kclowes:
### What was wrong?
The sync HTTP Provider has some nice session caching that isn't yet present in the async HTTP Provider. 

### How can it be fixed?
Add it! Looking at the caching from the sync HTTP Provider will be a good place to start. 

Will build on work done in #1986

kclowes:
Closed with #1999

Issue #1987, opened 2021-05-10
Handle middleware in Async HTTP Provider


kclowes:
### What was wrong?
Web3py relies on middleware for pre- and post-processing of requests. We'll need to combine middleware in the async HTTP provider and send requests and responses through the middleware onion. 

### How can it be fixed?

This will require at least one of the middlewares to be async, and probably another async method or two. Looking at how the sync HTTP provider handles middleware combining is probably a good place to start. 

Will build on #1986.

kclowes:
Closed with #1999

PR #1999, opened 2021-05-19
AsyncHTTPProvider accepts middleware

kclowes:
### What was wrong?
The Async RPC provider didn't accept any middleware. 

Related to Issue #1987, #1988, #2000

### How was it fixed?
Added the ability for the AsyncHTTPProvider to accept middleware and added an async `gas_price_strategy` middleware to test it out. Also adds session caching, and now the `EthereumTesterProvider` inherits from the `AsyncBaseProvider`

### Todo:

### DIFF
--- a/web3/middleware/__init__.py
+++ b/web3/middleware/__init__.py
@@ -44,6 +44,7 @@ from .formatting import (  # noqa: F401
 )
 from .gas_price_strategy import (  # noqa: F401
     gas_price_strategy_middleware,
+    async_gas_price_strategy_middleware,
 )
 from .geth_poa import (  # noqa: F401
     geth_poa_middleware,
--- a/web3/middleware/__init__.py
+++ b/web3/middleware/__init__.py
@@ -89,3 +89,26 @@ def combine_middlewares(
         reversed(middlewares),
         provider_request_fn,
     )
+
+
+async def async_combine_middlewares(
+    middlewares: Sequence[Middleware],
+    web3: 'Web3',
+    provider_request_fn: Callable[[RPCEndpoint, Any], Any]
+) -> Callable[..., RPCResponse]:
+    """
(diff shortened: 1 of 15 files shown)


web3/middleware/__init__.py
```diff
@@ -88,3 +89,26 @@ def combine_middlewares(
...
+    function wrapped with all of the middlewares.
+    """
+    accumulator_fn = provider_request_fn
+    for middleware in reversed(middlewares):
+        accumulator_fn = await apply_middleware(middleware, accumulator_fn, web3)
+    return accumulator_fn
+
+
+async def apply_middleware(
+    middleware: Middleware,
```
pipermerriam, line 110: This name through me.  I assumed this was the actual middleware function, but it's the "middleware constructor"?  No direct suggestion on naming but it might be worth establishing conventions for these two so that they can be logically separated.

pipermerriam (approved):
I wasn't able to glean "why" this middleware is being applied at the provider level rather than the web3 level.  Otherwise, this looks structurally good.  Still probably worth getting a quick review from @marcgarreau as I didn't use my fine-toothed-comb

web3/middleware/__init__.py
kclowes: whoops, you're right. Good suggestion!

marcgarreau (approved):
LGTM! A good sanity check could be to plug the provider w/ middleware into the benchmark suite. I'll get that wrapped up and merged now.

marcgarreau (approved):
thanks for benchmarking! huge difference in `sendTransaction`; we may need to scale down the number of calls there.

appreciate the import warning too. still looks sane after the updates, and plenty of time to keep feeling out if we've got the API and code location right before we call it 'stable'. :shipit:

PR #2002, opened 2021-05-21
benchmark async vs. sync http providers

marcgarreau:
### What was wrong?
Async http provider is under way; time to benchmark vs. the sync provider.

Update: turns out not much in the case of our CI resources and a locally running Geth node. Local tests with a remote node provider produced more dramatic results.

I think this is a reasonable starting point; the scope of benchmarks can grow from here. 

To run the benchmarks locally: `make benchmark`

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ envlist=
     py{36,37,38,39}-integration-{goethereum,ethtester,parity}
     lint
     docs
+    benchmark
     py{36,37,38,39}-wheel-cli
 
 [isort]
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -63,9 +64,16 @@ basepython =
...
-  mypy -p web3 -p ethpm -p ens --config-file {toxinidir}/mypy.ini
+    flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/ethpm {toxinidir}/tests --exclude {toxinidir}/ethpm/ethpm-spec
+    isort --recursive --check-only --diff {toxinidir}/web3/ {toxinidir}/ens/ {toxinidir}/ethpm/ {toxinidir}/tests/
+    mypy -p web3 -p ethpm -p ens --config-file {toxinidir}/mypy.ini
+
+[testenv:benchmark]
+basepython=python
+commands=
+    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 5
+    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 50
```
kclowes, line 75: Just curious - what's the reasoning behind testing 5, 50, and 100 calls here and not just checking one?

kclowes (approved):
LGTM! 🚀

tox.ini
marcgarreau: Once the suite is more mature, we should have vectors for benchmarking (the average of) 1 call vs. 1 call; this would be useful for comparing the HTTPProvider vs. IPCProvider, for example. I think making many calls is the best way to  showcase the difference between AsyncHTTPProvider vs. HTTPProvider performance.

web3/tools/benchmark/main.py
```diff
@@ -0,0 +1,140 @@
...
+
+def main(logger: logging.Logger, num_calls: int) -> None:
+    fixture = GethBenchmarkFixture()
+    for built_fixture in fixture.build():
+        for process in built_fixture:
+            w3_http = build_web3_http(fixture.endpoint_uri)
+            loop = asyncio.get_event_loop()
+            async_w3_http = loop.run_until_complete(build_async_w3_http(fixture.endpoint_uri))
+
+            methods = [
```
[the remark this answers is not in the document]
marcgarreau: There's probably a good way to clean this up a bit, but the methods rely on the providers built within this context. Will revisit next iteration.

PR #2005, opened 2021-05-26
Support state_override in contract function call

zzh1996:
### What was wrong?

`state_override` is not supported in contract function call.

### How was it fixed?

`state_override` is accepted in contract function call and passed to `web3.eth.call`

### Todo:

### DIFF
--- a/newsfragments/2005.feature.rst
+++ b/newsfragments/2005.feature.rst
@@ -1,0 +1,1 @@
+Support state_override in contract function call.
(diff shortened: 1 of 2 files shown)


kclowes:
Thanks for the contribution! This looks good to me on first glance, but I want to test it manually before merging since eth-tester doesn't support `state_override`.

kclowes (approved):
Was able to test and everything looks good here! Thanks @zzh1996!

Issue #1938, opened 2021-04-08
Update Geth test fixture to 1.10.3


marcgarreau:
### What was wrong?
Geth 1.10.3 was recently released. Time for a fixture bump.

### How can it be fixed?
The [docs](https://web3py.readthedocs.io/en/stable/contributing.html#generating-new-fixtures) got a little overhaul after the last update, so time to see if they're sufficient or still need improvement.

PR #2019, opened 2021-06-08
[#1938] - Generate geth fixture for geth version 1.10.3

fselmo:
### What was wrong?
- Geth fixture needed updating to 1.10.3

Related to Issue # 1938

### How was it fixed?
- Updated versioning across relevant files
- Minor updates to documentation (contributing.rst) for more clarification and geth version bumps where needed

### Todo:

#### Cute Animal Picture

2661_784981887893_5232436_50505893_3356746_n


marcgarreau (commented):
looks good at a glance; once py-geth is bumped, we can rerun the test suite and that's where the moneys at. thanks for PR#1 😎

fselmo:
> looks good at a glance; once py-geth is bumped, we can rerun the test suite and that's where the moneys at. thanks for PR#1 😎

Sounds good. Thanks for lookin' !

I just pushed up a minor commit. I meant to add "installing test dependencies" before attempting to generate new fixtures. When I first tried it gave me some missing library error and this was the missing step. I added it above near the "testing" section in the previous commit but I think it would be clarifying, as a second check, here as well.

PyCharm also caught some spelling mistakes in some other docs I was reading through so I pushed those here as well.

tests/integration/go_ethereum/common.py
```diff
@@ -20,12 +20,10 @@ def _check_web3_clientVersion(self, client_version):
 
 
 class GoEthereumEthModuleTest(EthModuleTest):
-    @pytest.mark.xfail(reason='eth_submitHashrate deprecated in 1.8.22 for ethash_submitHashRate')
     def test_eth_submit_hashrate(self, web3):
         # https://github.com/ethereum/go-ethereum/commit/51db5975cc5fb88db6a0dba1826b534fd4df29d7
         super().test_eth_submit_hashrate(web3)
 
-    @pytest.mark.xfail(reason='eth_submitHashrate deprecated in 1.8.22 for ethash_submitHashRate')
```
marcgarreau, line 28: If these couple of tests just work as expected now, we can delete them from this file. They're only listed here if we want to hijack the default behavior in some way, e.g., mark them as expected failures.

fselmo: Heard. Just pushed this up.

marcgarreau (approved):
LGTM! 🚢

Issue #1990, opened 2021-05-10
Async name_to_address middleware


kclowes:
### What was wrong?

The name to address middleware relies on calls to an RPC server. In order to have a full async pathway, web3py needs an async version of the middleware that will use an async w3 instance.

### How can it be fixed?

Add an async_name_to_address_middleware that can be used in the async pathway. This issue builds on #1986, and will depend on #1987.

dbfreem:
So I started working on this the other day since I felt like it was one of the shortcomings of the Async Provider currently. There were a couple of module level methods that needed to be asynced in `contracts` to call the eth.call async method. Also, I was thinking of creating an `AsyncContractFuntion`. This wouldn't be a full implementation of `ContractFunction` in Async but just enough to get bye to make the `ENS`  contract call. Also, was thinking of doing it in pieces. One PR for `AsyncContractFunction`, one for `ENS` methods needed for the middleware, and one PR for the middleware itself. This might make it more manageable and easier to review the PRs. 

@kclowes  or anyone else have any thoughts on this approach.

kclowes:
@dbfreem that approach sounds like what I would do as well, thanks for taking it on! Smaller PRs are preferred since they tend to be so much easier to review, and breaking it up like you suggest makes sense to me. 

/cc @pacrob since you've been working on an async `Contract`

PR #2044, opened 2021-06-21
Async net module

kclowes:
### What was wrong?
Name to address middleware uses `w3.net.version` to check that the user is attached to mainnet before resolving an ENS name, but there is no async net module yet. This PR adds it. 

Related to Issue #1990 

### How was it fixed?
Added a `BaseNet`, and an `AsyncNet` class, along with some tests.

### Todo:
- [x] Manual testing

#### Cute Animal Picture


web3/net.py
```diff
@@ -31,14 +42,34 @@ def listening(self) -> bool:
...
+    #
+    peerCount = DeprecatedMethod(peer_count, 'peerCount', 'peer_count')  # type: ignore
+
+
+class AsyncNet(BaseNet):
+    is_async = True
+
+    @property
+    async def listening(self) -> bool:
+        # types ignored b/c mypy conflict with BlockingNet properties
```
kclowes, line 64: @cburgdorf I have a mypy question for you when you get a chance - We are having to ignore all of the types for methods in our Async modules, because the methods in the Blocking modules have the same name and signature. Other than ignoring the type or changing the names of the methods, do you know of a way around this? The idea was to be able to keep the same API for both async and sync, but if the cost of that is not having types for half the codebase, I'm not sure that's a good tradeoff.

cburgdorf: As far as I can tell you should be able to define `async def listening -> bool` here because the sync version isn't defined here or in the parent but in `Net` which isn't in our inheritance chain.

What it is complaining about is that you use `await self._listening()`. If you drop the `await` all of these should be fine and don't need the ignore. The reason is that you are calling the synchronous version so there isn't actually anything to `await` here.

```
    @property
    async def listening(self) -> bool:
        # types ignored b/c mypy conflict with BlockingNet properties
        return self._listening()

    @property
    async def peer_count(self) -> int:
        # types ignored b/c mypy conflict with BlockingNet properties
        return self._peer_count()

    @property
    async def version(self) -> str:
        # types ignored b/c mypy conflict with BlockingNet properties
        return self._version()
```

kclowes:
This is ready for a look when you get a chance @fselmo!

fselmo (approved):
LGTM 👍

PR #2048, opened 2021-06-23
Clarify that using `block_identifier` with .call() comes with its caveats

miohtama:
To manage the expectations of the client library users.

### What was wrong?

Programmers were using `call` with a block number, but did not know it might not work as it.

### How was it fixed?

Document `block_identifier` behavior with various Ethereum nodes.

#### Cute Animal Picture

```
-=[ bull ]=-  12/96

   ,/         \,
  ((__,-"""-,__))
   `--)~   ~(--`
  .-'(       )`-,
  `~~`d\   /b`~~`
      |     |
 jgs  (6___6)
       `---`
```

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -826,6 +826,11 @@ Methods
 
         # You can check the state after your pending transactions (if supported by your node):
         >>> token_contract.functions.myBalance().call(block_identifier='pending')
+        
+     Passing `block_identifier` parameter for past block numbers requires that your Ethereum API node 
+     is running an more expensive archive node mode. Normally synced Ethereum nodes will fail with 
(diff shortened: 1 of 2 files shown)


miohtama:
Looks like the docs fails in Circle CI, but when I click "Details" it asks me for a login. Are the CI reports publicly available?

kclowes:
Thanks for the clarification @miohtama! You should definitely be able to see the CI from a branch you wrote but you may need to log in to CircleCI with your Github creds. But to fix the docs build, you just need to unindent the paragraph you wrote. The geth failure is a flaky failure.

kclowes (approved):
Thanks!

Issue #2039, opened 2021-06-18
Add eth_getRawTransactionByHash


kclowes:
### What was wrong?

We don't have support for the `eth_getRawTransactionByHash` RPC method. 

### How can it be fixed?

Add it to `BaseEth`, and add it to both `Eth` and `AsyncEth` following established patterns. Tests will need to be added as well.

Moras-del:
Hello! I would like to work on this. Could you assign this task to me? :)

kclowes:
Done! Let me know if you need any guidance! PR-ing early and often is best! Thanks!

kclowes:
Fixed in #2051

PR #2051, opened 2021-06-24
Add support for eth_getRawTransactionByHash

Moras-del:
### What was wrong?

RPC method `eth_getRawTransactionByHash` was not supported. This PR adds support for this method.

Related to Issue #2039

### How was it fixed?
Implemented similarly to the method `eth_getTransactionByHash`
### Todo:
- Proper exception handling (transaction not found, method not supported)

#### Cute Animal Picture

obraz

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -576,6 +576,20 @@ The following methods are available on the ``web3.eth`` namespace.
       :attr:`~web3.eth.Eth.get_transaction`
 
 
+.. py:method:: Eth.get_raw_transaction(transaction_hash)
+
+    * Delegates to ``eth_getRawTransactionByHash`` RPC Method
+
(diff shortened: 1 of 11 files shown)


kclowes:
Thanks @Moras-del! This is looking good! Since we don't have the `getRawTransactionByHash` RPC method implemented in eth-tester, you'll want to go in to `tests/integration/test_ethereum_tester.py` and add an `xfail` similar to what we do [here](https://github.com/ethereum/web3.py/blob/master/tests/integration/test_ethereum_tester.py#L381-L383). Let me know if that doesn't make sense!

kclowes:
Thanks for getting this rolling @Moras-del! I made some updates and will do a final check before merging, probably on Monday!

kclowes (approved):
Thanks again!

Issue #1835, opened 2021-01-14
Support for EIP-1559


kclowes:
* Version: v5.14+

### What was wrong?
It looks like [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) is coming up soon (maybe in the London hard fork), so we'll need to update web3py to support it. This is a placeholder issue that will likely get broken down into a few issues. 

### How can it be fixed?
From a high level, it looks like we need to change our transaction object:
- New fields in the transaction:
    - fee_cap_per_gas
    - max_miner_bribe_per_gas
- ~Remove gas_price (will probably need to keep it in in v5 and then need to make a note to remove in v6)~

BlockData will need to change as well:
- Add base_fee field
- ~move gas_limit to gas_target (In v5, we'll need to keep both gas_limit and gas_target and remove gas_limit in v6).~ 

Also probably simplifies (possibly removes the need for) our gas price strategies.

@abdelhamidbakhta or @pipermerriam let me know if I'm missing anything!

abdelhamidbakhta:
@kclowes Thank you very much for starting looking at this. 
You don't have to remove `gas_price`. There is no plan/timeline to remove legacy transaction support for the moment.  @timbeiko can you confirm ?
Also IMO the `move gas_limit to gas_target` is not necessary. It is a nice to have though.
Otherwise the high level approach makes a lot of sense.

timbeiko:
No plan on removing legacy transactions in the foreseeable future!

kclowes:
There are a few issues open that still need some work, but initial MVP work on 1559 support has been completed. See #2065 and #2066.

PR #2053, opened 2021-06-25
Fix tests from 1835 changes and default to 1559 fields for testing

fselmo:
### What was wrong?

Related to Issue #1835

1. We skipped some tests in order to create the working london branch. Needed to isolate what was going wrong.
2. We should update default params in our existing transactions in our tests to use the new fields

### How was it fixed?

1. Some older tests seemed to be breaking others because they did not meet updated requirements to be included in a pending block, seemingly creating a txn jam. Increasing `gasPrice` for these tests seemed to move things along. Another issue we have is that some tests have their context dirtied by other tests. I included a way to move some tests to the beginning of the test suite to prevent this dirtied context and allow these tests to pass.

2. Updated default tests to use the new 1559 fields for our transactions.

### Todo:
- [X] Fix broken / skipped tests from london branch creation
- [X] Update existing tests to use the new 1559 fields `maxFeePerGas` and `maxPriorityFeePerGas` over `gasPrice` as defaults
- [X] Added documentation on unit and integration testing and how to contribute to our test suite

#### Cute Animal Picture

IMG_20210615_092412_01

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ develop-eggs
 lib
 lib64
 venv*
+.python-version
 
 # Installer logs
 pip-log.txt
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -126,7 +126,7 @@ with the exception of ``self`` and ``cls`` as seen in the following example.
         self.reset()
 
 
-Running the tests
+Running The Tests
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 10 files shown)


kclowes (approved):
Awesome! Thanks for being so persistent with these tests! I just left one comment about putting back the error expectation that you can take or leave, whatever makes sense

PR #2057, opened 2021-07-01
Bugfix failing tests

fselmo:
### What was wrong?
- Some of our tests have been having some issues due to transactions hanging in the tx pool and inconsistent mining of the pending block for other tests

### How was it fixed?
- I traced most of the issues to low `gasPrice` values for some of the `send_transaction` / `replace_transaction` tests and raised these numbers so the txns didn't hang in the tx pool.
- I added a method to mine a block before a test if that is desired as well. This helped ensure that `get_transaction_count` returns a correct number for a nonce for tests explicitly passing in the `nonce` value.

### Todo:

#### Cute Animal Picture

bckgrnd

### DIFF
--- a/newsfragments/2057.bugfix.rst
+++ b/newsfragments/2057.bugfix.rst
@@ -1,0 +1,1 @@
+Turned some tests back on and reduced cross-test noise that was causing some tests to fail.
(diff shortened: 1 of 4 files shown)


kclowes (approved):
Nice! LGTM! 🚀

PR #2060, opened 2021-07-02
Merge bulk of London support to master

fselmo:
### What was wrong?
- London support was needed, mainly support for the new transaction parameters `maxFeePerGas` and `maxPriorityFeePerGas`

Related to Issue # 1835

### How was it fixed?
- Support `maxFeePerGas` and `maxPriorityFeePerGas` was added, mainly to the `sendTransaction` RPC call and internal `replace_transaction` and `modify_transaction` methods.

### Todo:
~~- [ ] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)~~
^ The separate commits in this branch have their appropriate release notes

#### Cute Animal Picture

Screen Shot 2021-07-02 at 14 48 43


kclowes (commented):
Hooray! We're so close! 

There were a few things I noticed when manually testing: 
- I think `sign_transaction` still needs to be updated to support 1559. If I send a tx with both 1559 and legacy fields, I get a ValueError back from Geth, rather than our web3 `TransactionTypeMismatch` error
- In the same vein, if I `sign_transaction` with an int in the `maxFeePerGas` field: 
```
w3.eth.sign_transaction({
    'from': w3.eth.default_account, 
    'to': acct2, 
    'value': 100, 
    'maxFeePerGas': 3000000000,
    'nonce': w3.eth.get_transaction_count(w3.eth.default_account),
})
```
I get an `invalid argument 0: json: cannot unmarshal non-string into Go struct field TransactionArgs.maxFeePerGas of type *hexutil.Big"` error.
- It's probably also worth writing a few integration tests for `replace_transaction` and `modify_transaction` to prove to ourselves that we can handle both legacy and 1559 transaction params with those two methods. A quick look suggests that we'll at least need to add the new keys to the `VALID_TRANSACTION_PARAMS` in `web3/_utils/transactions.py` but there may be other things that need to be done as well. 
- It may be worth adding a note to the docs for the Gas Price API and/or `w3.eth.generate_gas_price` that those will only work for legacy transactions

Thanks for all of your work on this!

fselmo:
> It's probably also worth writing a few integration tests for replace_transaction and modify_transaction to prove to ourselves that we can handle both legacy and 1559 transaction params with those two methods.

I did change most of the relevant `replace_transaction` tests to use the new params in this PR and kept a `test_eth_replace_transaction_legacy`. I left `gasPrice` only on the tests that are testing gas price strategies for `replace_transaction`. It would be good to add the params to `modify_transaction` though, thanks for catching that.

I'll also add some clarification on the gas price strategy to the docs 👌 

And just going to mark it here that, as we saw earlier, geth's `eth_signTransaction` doesn't appear to be supporting the new params at the moment so we can circle back to that if this PR is still open when they implement it. Otherwise I would suggest we add that support as a separate PR.

kclowes (approved):
Awesome! This looks good to me! 🚀

qiushui777:
Hi, I was trying to sign a tx with the latest web3.py which is 5.21.0.

`
    txn_dict = contract.functions.test(int(amount)).buildTransaction({
        'chainId': 5,
        'gas': 2000000,
        'maxFeePerGas': w3.toWei(7, 'gwei'),
        'nonce': nonce,
    })
    print(txn_dict)
    signed_txn = w3.eth.account.signTransaction(txn_dict, private_key=privatekey)
`

and it gave me this error
```
TypeError: Transaction must not include unrecognized fields: {'maxFeePerGas'}
```

Any idea on how to solve it? I think it should support maxFeePerGas when signing the tx

fselmo:
Hey @qiushui777. These changes are going to be on the next release. Sorry for the inconvenience. For now you can get around this by installing web3py from the latest `master` branch with `pip install https://github.com/ethereum/web3.py/archive/master.zip`.

This next release will come out early this next week if not tomorrow.

fselmo:
@qiushui777 Version `5.22.0` is out now and should include signing support for 1559

Issue #2043, opened 2021-06-21
Bug with `to_hex` in `wait_for_transaction_receipt` function


sobolev-igor:
* Version: 5.20.0
* Python: 3.9
* OS: osx

### What was wrong?

Hey! I've got an issue with `to_hex` and `wait_for_transaction_receipt` functions. 
I use the function as in the example

> web3.eth.wait_for_transaction_receipt('0x5c504ed432cb51138bcf09aa5e8a410dd4a1e204ef84bfed1be16dfba1b22060')

But if the tx doesn't get mined within timeout it raises the exception which uses `to_hex` function with a tx hash passed as a parameter (which is a string in an example). It raises another exception as a string cannot be passed to `to_hex` function.

> Unsupported type: The primitive argument must be one of: bytes,bytearray, int or bool and not str

### How can it be fixed?

Allow `to_hex` accept a string or convert it to bytes in `wait_for_transaction_receipt` function.

kclowes:
Thanks for raising this! Option 2 (convert it to bytes in `wait_for_transaction_receipt`) is going to be the most straightforward I think, so I'd vote for that solution.

KingCreate:
Hi, am I able to give this a crack as my first issue?

kclowes:
@KingCreate yes please! let me know if you need more direction!

ericscagnelli:
@KingCreate @kclowes Is this one still being worked or is any help needed? I'd like to give this one a try as a first issue 

Also - it looks like #2061 is a duplicate of this one

kclowes:
@ericscagnelli I haven't seen a PR yet, so as far as I can tell, it's up for grabs. Thanks!

KingCreate:
@ericscagnelli I think I need to go back and learn some more, so feel free to take this one

abdulqgg:
Could i give this a go as my first issue?

fselmo:
> Could i give this a go as my first issue?

Hey @abdulqgg. Sorry about that, looks like this was picked up and changes were merged as part of #2068 but we forgot to close it. I'm going to close this but I'd still encourage looking for a `Good First Issue` tag or anything else that seems interesting to you if you are still interested in contributing. Thanks for the interest!

PR #2068, opened 2021-07-10
To hex string parameter

ericscagnelli:
### What was wrong?
Hey! I've got an issue with to_hex and wait_for_transaction_receipt functions.
I use the function as in the example

web3.eth.wait_for_transaction_receipt('0x5c504ed432cb51138bcf09aa5e8a410dd4a1e204ef84bfed1be16dfba1b22060')

But if the tx doesn't get mined within timeout it raises the exception which uses to_hex function with a tx hash passed as a parameter (which is a string in an example). It raises another exception as a string cannot be passed to to_hex function.

Unsupported type: The primitive argument must be one of: bytes,bytearray, int or bool and not str

Related to Issue #2043 

### How was it fixed?
In wait_for_transaction_receipt function when TimeExhausted exception is raised the parameter type being passed to to_hex is checked. When it's a string it is now converted to bytes.


ericscagnelli:
@kclowes Hi I'm not very familiar with lint checks and it looks like my solution keeps getting caught up there. I fixed the basic stylistic problems (not enough newlines, trailing whitespaces, etc..) but now it looks like it's trying to do a type check based on type hints and I'm hesitant to alter the parameter types of the wait_for_transaction_receipt function

```
web3/eth.py:539: error: Item "Hash32" of "Union[Hash32, HexBytes, HexStr]" has no attribute "encode"
web3/eth.py:539: error: Item "HexBytes" of "Union[Hash32, HexBytes, HexStr]" has no attribute "encode"
web3/eth.py:539: error: Incompatible types in assignment (expression has type "Union[Any, bytes]", variable has type "Union[Hash32, HexBytes, HexStr]")


536             # to_hex function does not accept string parameters
537             # Convert to bytes if needed
538             if type(transaction_hash) == str:
539                 transaction_hash_bytes = transaction_hash.encode('utf-8')

```

kclowes:
I'll take a look! Thanks for the PR @ericscagnelli!

web3/eth.py
```diff
@@ -531,9 +531,16 @@ def wait_for_transaction_receipt(
...
+
+            # to_hex function does not accept string parameters
+            # Convert to bytes if needed
+            if type(transaction_hash) == str:
+                transaction_hash_bytes = transaction_hash.encode('utf-8')
+
             raise TimeExhausted(
                 "Transaction {} is not in the chain, after {} seconds".format(
-                    to_hex(transaction_hash),
+                    to_hex(transaction_hash_bytes),
```
kclowes, line 543: Instead of trying to do the `encode` on line 539, you can just do something like: 
```suggestion
                    HexBytes(transaction_hash),
```
And then I believe that line 542 will need to become: 
```
"Transaction {!r} is not in the chain, after {} seconds".format(
```

ericscagnelli:
Thank you @kclowes! I should read the documentation more carefully next time... didn't see HexBytes() before so I tried to reinvent the wheel.

There is one final (I hope) issue coming up that I don't see how is related to this change since it's in the go-ethereum test module. Could you take a look? Sorry, I'm not really sure how to move forward with this one. 

Also, I'm new to open source development and was wondering if you have any suggestions/tips for getting to know a large preexisting code base when you're just making a minor localized fix like this one. Thanks again!

kclowes:
@ericscagnelli thanks for making that change! The last failure is just one of our flaky tests. I'll re-run and then we should be good!

> Also, I'm new to open source development and was wondering if you have any suggestions/tips for getting to know a large preexisting code base when you're just making a minor localized fix like this one. 

My suggestion would be to just jump in like you did! Sometimes OSS repos have a Good First Issue tag (or something equivalent) that indicates an issue that should be straightforward and doesn't require deep knowledge of the code base. I recommend starting there to get your feet wet. After that, you can always do a few more of those or you can try your hand at another, more complicated issue. If you do jump into a bigger issue, before you start, I'd run your solution by the maintainers to make sure that's what they were thinking so that you don't waste any work. And then once you're working on it, our team likes to see WIP PRs so that we can provide guidance or support as needed!

kclowes (approved):
Looks good to me! Thanks @ericscagnelli! I'll add a newsfragment and merge!

PR #2085, opened 2021-07-23
Update to using geth fixture 1.10.6

fselmo:
### What was wrong?
- `geth` put out a hot fix in v1.10.6 and we needed to update the fixture to use this version and make sure all was good on our end.

### How was it fixed?
- `geth` fixture updated to use v1.10.6

### Todo:

#### Cute Animal Picture

jax


kclowes (approved):
LGTM! 🚀

PR #2092, opened 2021-07-29
Remove default gasPrice when 1559 params present

fselmo:
### What was wrong?

- `sign_and_send_raw_middleware` was broken for EIP-1559 transactions since it made use of `fill_transaction_defaults()` and this was always adding a `gasPrice` if none was present. This would cause both legacy and 1559 params to be present leading to errors.

### How was it fixed?

- For code making use of `fill_transaction_defaults()`, removed the default value for `gasPrice` when 1559 parameters are present. This keeps things backwards compatible and also provides support for 1559 for the signing middleware.
- `fill_transaction_defaults()` also defaults the transaction `type` to `'0x2'` when 1559 params are present as this implies a 1559-style transaction. This is important because transaction signing (done via `eth-account`) relies on the `type` for validating 1559 transactions. We don't require the `type` for `send_transaction` so this makes the user experience smoother when using this middleware since they may already be used to sending transactions without providing a `type`.

### Todo:

### DIFF
--- a/tests/core/middleware/test_transaction_signing.py
+++ b/tests/core/middleware/test_transaction_signing.py
@@ -15,6 +15,8 @@ from eth_utils import (
     to_hex,
 )
 from eth_utils.toolz import (
+    assoc,
+    dissoc,
     identity,
     merge,
     valfilter,
@@ -136,7 +138,6 @@ def hex_to_bytes(s):
 )
 def test_sign_and_send_raw_middleware(
         w3_dummy,
-        w3,
         method,
         from_,
         expected,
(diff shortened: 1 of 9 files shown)


tests/core/middleware/test_transaction_signing.py
```diff
@@ -277,7 +280,6 @@ def test_signed_transaction(
 
     if isinstance(expected, type) and issubclass(expected, Exception):
         with pytest.raises(expected):
-            start_balance = w3.eth.get_balance(_transaction.get('from', w3.eth.accounts[0]))
```
fselmo, line 280: This is just cleanup for context... this line was not used anywhere

kclowes (approved):
Looks good to me! Thanks for the fix! :shipit:

PR #2093, opened 2021-07-30
fix eth.account documentation to reflect required EIP-1559 params

mike6649:
### What was wrong?

- In the documentation for `send_raw_transaction()`, the example `account.sign_transaction()` call was missing the `type` and `chainId` arguments.

### How was it fixed?

- Added the arguments to the sample method call.


fselmo (approved):
Looks good to me. Thanks for the contribution!

Issue #2069, opened 2021-07-13
Could not decode contract function call {function_identifier} return data {return_data!r} for output_types {output_types}


Uxio0:
* Version: 5.21.0 (detecting this since 5.20)
* Python: 3.9
* OS: linux

### What was wrong?

Previously when a contract reached a Solidity `require` the message was returned in the exception. Currently I just see the placeholders.

Errors can be seen in our library (gnosis-py) tests:
- https://github.com/gnosis/gnosis-py/pull/43
- https://github.com/gnosis/gnosis-py/pull/43/checks?check_run_id=3058214761

When 5.19.0 version is used, no problems:
- https://github.com/gnosis/gnosis-py/runs/3058176262?check_suite_focus=true

Is this a bug or do I need to change anything? I couldn't find anything in the changelog

fselmo:
@Uxio0 Thanks for reporting this! Something did change in the string interpolation here near 5.20... I'm going to check if any other files were affected and I'll submit a quick fix.

fselmo:
@Uxio0  I didn't see any other affected files and I pushed a hotfix for this directly to `master` since it was quite a small change. It should be included in the next release but you can you master for now if this helps.

I'm going to close this. Feel free to re-open if you are still having issues.

Uxio0:
Thanks for the quick response and fix @fselmo

Uxio0:
Hi @fselmo , I've just tried the new release and there's still different behaviour. Revert messages are not decoded. For example, now I get:

`Could not decode contract function call execTransaction return data HexBytes('0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000195369676e617475726573206461746120746f6f2073686f727400000000000000') for output_types ['bool']`

If I convert that `HexBytes` string to `bytes` I can see the revert reason `Signatures data too short`, that was the previous behaviour. As a `fstring` is used to raise the message is not even easy to extract the parameter from the exception unless using a regular expression

Uxio0:
I think the issue is the `!r` in the string interpolation:

```
In [3]: str(HexBytes('0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000195369676e617475726573206461746120746f6f2073686f727400000000000000'))
[...]

PR #2096, opened 2021-08-05
Fix string interpolation issues in exception message.

fselmo:
### What was wrong?
- Exception message for calling a contract function did not return data in a manner where a meaningful exception message could be extracted even though we used to return it in this way (breaking change)

Related to Issue #2069

### How was it fixed?
- Print the result_data as bytes in exception message in order to extract a meaningful exception message from the data. This is how we used to return this message and some users were using it to extract the meaningful exception message and were no longer able to.

### Todo:

#### Cute Animal Picture


kclowes (approved):
🚀

PR #2103, opened 2021-08-11
[WIP] Revert eth.fee_history

kclowes:
### What was wrong?

Geth has the fee history endpoint implemented inconsistently, but they have the PR up to fix. We'll put this endpoint and docs back once 1.10.7 is out.

### How was it fixed?
Removed the squashed commit (d722f3107612a4b713d8ad360a3ca2d637964721), and the doc commits that followed.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -116,40 +116,6 @@ The following properties are available on the ``web3.eth`` namespace.
         >>> web3.eth.hashrate
         906
 
-.. py:attribute:: Eth.fee_history
-
-    * Delegates to ``eth_feeHistory`` RPC Method
-
-    Returns transaction fee data for up to 1024 blocks.
-
-    Parameters:
-
-    * ``block_count``: The number of blocks in the requested range. This value is a number between 1 and 1024.
-      Less than requested may be returned if not all blocks are available.
-    * ``newest_block``: The newest, highest-numbered, block in the requested range.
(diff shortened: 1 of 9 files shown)


banteg:
There are other nodes which implement it correctly though. Web3.py uses an incorrect formatter for `eth_feeHistory` input though. It should be `to_hex_if_integer` for both args if two args are specified, and an array of floats for the third arg if it's not None.

kclowes:
@banteg Do you know if Erigon can take an int for the first param?

The majority of people use Geth (according to Etherscan), so we were thinking it will be less confusing if we just wait for the next Geth release to include this RPC method and then everything will be standard. Geth hasn't done a release in a while, so it should come out soon. But if you need feeHistory now for some reason, I could be convinced to fix the formatter and keep it in for this next release.

kclowes:
Actually, running a quick manual test, it looks to me like the Geth and Erigon implementations are the same. I can't send in a hexstring on either one, but they both take either a base 10 or base 16 int. But, in Geth 1.10.7 they will accept a hexstring like the spec says (see https://github.com/ethereum/go-ethereum/pull/23239, and https://github.com/ethereum/eth1.0-apis/issues/20). I'll raise an issue on Erigon and see if they will standardize too. I'm going to close this for now, and leave the formatter off of the blockCount arg so that users can explicitly pass in what their client wants, and add a note to the docs.

PR #2104, opened 2021-08-11
Update docs for block_count arg

kclowes:
### What was wrong?

There is some standardization lacking between clients in regards to the type of the `block_count` argument for `eth.feeHistory` RPC Method. In the next release Geth will accept a hexstring for the `block_count` argument, but some other clients only allow ints. 

Related to Issue: https://github.com/ethereum/eth1.0-apis/issues/20

### How was it fixed?
Added some docs to clarify that `block_count` may either take a hexstring or an int, depending on the client. @fselmo - let me know if you think this needs more/better/different clarification!

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -1053,9 +1053,10 @@ The following methods are available on the ``web3.eth`` namespace.
 
     Returns transaction fee data for up to 1,024 blocks.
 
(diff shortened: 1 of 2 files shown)


fselmo (approved):
LGTM!

Issue #2095, opened 2021-08-03
Provide constants for e.g. the zero address and max uint256


KaiRo-at:
* Version: 5.22.0
* Python: 3.8
* OS: linux

### What was wrong?

If I need to compare an address returned from a contract or variable to the zero address or want to e.g. use the maximum allowed uint256, I need to define those myself.

### How can it be fixed?

Provide values like this as constants from Web3.py - see also what Ethers has available at https://docs.ethers.io/v5/api/utils/constants/ - for me, the zero address (needed often) and the max uint256 (sometimes needed to e.g. set a maximum allowance on ERC20) are the most important ones, I also sometimes have use cases for a full 32byte hex value of zero, which is "HashZero" in the Ethers list. The wei per ether can be retrieved via `Web3.toWei(1, 'ether')` but may be nice, the others in the Ethers list I think are not very useful in Python.

kclowes:
Sure, I like that idea. On first glance, what I would do is make a new file called `constants.py` in the `web3/` dir and then go search the codebase for the hard-coded constants and replace them. Docs should be added too.

KaiRo-at:
Sounds great!
I was looking at code my colleagues did with Ethers and found there's a constant there for the zero address, and told them to use that, and then thought about how I always define this in my Python code somewhere and wondered if Web3.py has that to refer to as well, so I looked for it and when I didn't find it, I thought I probably should file this as an enhancement request. ;-)

ernestosperanza:
Hello, I'm interested in working on this issue and contribute to the project, let me know if I can help.
Thanks!

smishy05:
Hey there! I am interested in working on this issue. Please let me know if I can contribute.
Many thanks!

kclowes:
It looks like @ernestosperanza got there first, @smishy05. But if there isn't a PR in the next few days, feel free to pick it up. Thanks for the interest!

ernestosperanza:
Yes sorry, it's my first issue. It might take some time but I am working on it.

smishy05:
@kclowes Thanks for the reply! Sounds fair enough!

@ernestosperanza No issues! Good luck!

PR #2109, opened 2021-08-19
Issue #2095 - Provide constants for e.g. the zero address and max uint256

ernestosperanza:
### What was wrong?
If I need to compare an address returned from a contract or variable to the zero address or want to e.g. use the maximum allowed uint256, I need to define those myself. (issue #2095)

### How was it fixed?
I added the constants.py module and created constants for the following commonly used values:
As strings: ADDRESS_ZERO, MAX_INT, HASH_ZERO
As integer: WEI_PER_ETHER

I replaced the hardcoded values in the codebase with the new constants.
And write a little doc for the constants.

### Todo:
If is required I can replace the hardcoded values in docs two, I didn't want to change documentation without permission.

Let me know if there's any improvement that I can apply, thanks :)


kclowes:
Thanks for the PR @ernestosperanza! It looks like the `core` and `ethpm` tests are failures that we have in master, but the `docs` and the `lint` tests are failures related to this PR. Let me know if you want help fixing those!

ernestosperanza:
@kclowes let me do some documentation reading and follow all the instructions, then if I'm still having problems I will get back to you, thanks

ernestosperanza:
@kclowes I'm still getting errors in the docs but I think that those belong to the ethpm.rst, let me know If I have to modify anything, thanks in advance

PR #2136, opened 2021-09-08
AsyncHTTPProvider extra dependencies

ediskandarov-cur:
### What was wrong?

Installation of `web.py` python package by default required `aiohttp` in dependencies.

Async HTTP provider is not required until you specifically write async code(which is not a general case).

Thus it would be rational to move this dependency to extras.

### How was it fixed?

Moved `aiohttp` to extra requirements.

### Todo:

#### Cute Animal Picture

photo-1605206731460-0bbdfadf9808

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -369,6 +369,14 @@ AsyncHTTPProvider
 .. warning:: This provider is unstable and there are still gaps in
     functionality. However, it is being actively developed.
 
+.. NOTE:: To install the needed dependencies to use ``AsyncHTTPProvider``, you can install the
+    pip extras package that has the correct interoperable version of the ``aiohttp``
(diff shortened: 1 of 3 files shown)


emcpow2:
@kclowes @fselmo please review

emcpow2:
if there any way to push this forward?

kclowes:
This seems like a reasonable change to me @emcpow2, especially since async is not in a complete state yet. I'll do some manual testing and merge soon! Thanks!

emcpow2:
Hey @kclowes . Yes, it's solely to keep dependencies as slim as possible.

Many projects will not welcome `aiohttp` in dependencies. (I have no attitude to the `aiohttp` itself, it just seems an overkill to explicitly require it)

PR #2146, opened 2021-09-12
Edit some ValueError() messages for more clarity

Snugface:
### What was wrong?
A few of the error messages were confusing

### How was it fixed?
With additional formatting and clarification

### Todo:

### Background
Errors such as "Cannot set data in estimateGas transaction" make me think that estimateGas() failed to set some data values in transaction provided to it. While instead what it means is that I myself, as an operator, am not supposed to set transaction['data'] field. And should leave it blank. This might be obvious to those experienced with web3.py, but it's extremely confusing to me.
Same with the generic ValueError at line 693. It took me an extra hour to understand what the error was actually saying, because of the format it was in: "cannot set to in estimateGas transaction", for example. Just how the hell is one supposed to understand that line? It is VERY confusing.

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2146.doc.rst
+++ b/newsfragments/2146.doc.rst
@@ -1,0 +1,1 @@
+Clarify some contract ``ValueError`` error messages.
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -689,7 +689,9 @@ class ContractConstructor:
     ) -> None:
         keys_found = set(transaction.keys()) & set(forbidden_keys)
         if keys_found:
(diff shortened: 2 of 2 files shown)


kclowes:
Thanks for the changes @Snugface! I'll add a newsfragment and merge!

Snugface:
> Thanks for the changes @Snugface! I'll add a newsfragment and merge!

Well, it's a crude edit. I could have spent more time working on it, tbh
But I am glad to see that you are willing to merge it.
Thanks!

Issue #2145, opened 2021-09-11
Any request return "Parser Error"


mr-tron:
* Version:5.23.1
* Python: 3.6.8/3.8.10
* OS: linux
* `pip freeze` output

```
aiohttp==3.7.4.post0
async-timeout==3.0.1
attrs==21.2.0
base58==2.1.0
bitarray==1.2.2
certifi==2021.5.30
chardet==4.0.0
charset-normalizer==2.0.4
cytoolz==0.11.0
eth-abi==2.1.1
eth-account==0.5.5
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.3.3
eth-rlp==0.2.1
eth-typing==2.2.2
eth-utils==1.10.0
hexbytes==0.2.2
idna==3.2
idna-ssl==1.1.0
importlib-metadata==4.8.1
importlib-resources==5.2.2
ipfshttpclient==0.8.0a2
jsonschema==3.2.0
lru-dict==1.1.7
more-itertools==8.8.0
multiaddr==0.0.9
multidict==5.1.0
netaddr==0.8.0
packaging==21.0
parsimonious==0.8.1
protobuf==3.17.3
py-bip39-bindings==0.1.7
py-ed25519-bindings==0.1.3
py-sr25519-bindings==0.1.3
pycryptodome==3.10.1
pyparsing==2.4.7
pyrsistent==0.18.0
requests==2.26.0
rlp==2.0.1
scalecodec==0.11.19
six==1.16.0
substrate-interface==0.13.12
toolz==0.11.1
typing-extensions==3.10.0.2
urllib3==1.26.6
varint==1.0.2
web3==5.23.1
websocket-client==0.59.0
websockets==9.1
xxhash==2.0.2
yarl==1.6.3
zipp==3.5.0

```

### What was wrong?

Can't do any request to ethereum node. Server always returns "Parse Error".
```
python3
Python 3.6.8 (default, Mar 19 2021, 05:13:41) 
[GCC 8.4.1 20200928 (Red Hat 8.4.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from web3 import Web3, HTTPProvider
>>> 
>>> w3 = Web3(HTTPProvider('https://cloudflare-eth.com'))
>>> w3.isConnected()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/test/.env/lib/python3.6/site-packages/web3/main.py", line 327, in isConnected
[...]

PR #2151, opened 2021-09-15
JSON RPC request id from 1.

dingguijin:
Since https://cloudflare-eth.com can not parse id from 0.

### What was wrong?

Related to Issue #2145.

Change request_counter start from 1.

### DIFF
--- a/web3/providers/base.py
+++ b/web3/providers/base.py
@@ -82,7 +82,7 @@ class BaseProvider:
 
 class JSONBaseProvider(BaseProvider):
     def __init__(self) -> None:
-        self.request_counter = itertools.count()
+        self.request_counter = itertools.count(start=1)
 
     def decode_rpc_response(self, raw_response: bytes) -> RPCResponse:
         text_response = to_text(raw_response)


dingguijin:
JSON RPC request id from 1.

fselmo:
@dingguijin Thanks for submitting this. We did some discussing and think it would be ideal to instead add an optional kwarg when initializing the provider that would set the start `id` for the JSON-RPC call. This would help not introduce a possible breaking change if anyone is tracking these id's in their code and allows for customization of this value on instantiation.

It would be good to include tests for this as well... let me know if you want to work on this otherwise I can submit a quick PR no problem.

Thanks again for your contribution!

dingguijin:
@fselmo Obviously It is a better idea to add an optional kwarg to set id for JSON-RPC call. I am glad to do that changing code and adding unittest. But I do not know the PR process of this project. It would be good if you could help me do that.

Thanks for your reply.

fselmo:
> @fselmo Obviously It is a better idea to add an optional kwarg to set id for JSON-RPC call. I am glad to do that changing code and adding unittest. But I do not know the PR process of this project. It would be good if you could help me do that.
> 
> Thanks for your reply.

Absolutely. Glad to point you in the right direction.

[Here](https://web3py.readthedocs.io/en/latest/contributing.html#pull-requests) is a good reference for the PR process. We would be adding a `2151.feature.rst` file with a brief description of the task. Let me know if you have any specific questions along the way and I'd be happy to help!

dingguijin:
> @dingguijin Thanks for submitting this. We did some discussing and think it would be ideal to instead add an optional kwarg when initializing the provider that would set the start `id` for the JSON-RPC call. This would help not introduce a possible breaking change if anyone is tracking these id's in their code and allows for customization of this value on instantiation.
> 
> It would be good to include tests for this as well... let me know if you want to work on this otherwise I can submit a quick PR no problem.
> 
> Thanks again for your contribution!

dingguijin:
> > @fselmo Obviously It is a better idea to add an optional kwarg to set id for JSON-RPC call. I am glad to do that changing code and adding unittest. But I do not know the PR process of this project. It would be good if you could help me do that.
> > Thanks for your reply.
> 
> Absolutely. Glad to point you in the right direction.
> 
> [Here](https://web3py.readthedocs.io/en/latest/contributing.html#pull-requests) is a good reference for the PR process. We would be adding a `2151.feature.rst` file to the newsfragments with a brief description of the task. We should also update the docs with the new feature in the appropriate section.
> 
> Let me know if you have any specific questions along the way and I'd be happy to help!

@fselmo Thanks,

After some investigations about id of JSON-RPC spec (https://www.jsonrpc.org/specification), when id is equal to 0, it would be inexplicit to NULL, JSON-RPC server could handle this request as a notification and imagine JSON-RPC client will not care the response.

So, JSON-RPC request id from 1 or adding a request_id as construtor parameter of HTTPProvider?

PR #2159, opened 2021-09-26
Avalance has extra data 80

eino-makitalo:
### What was wrong?

Avalanche network seem to use extra data up to 80 bytes

### Todo:
- Inform me what is right way to fix this library also to support Avalance blockchain - maybe not like this

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -210,7 +210,7 @@ receipt_formatter = apply_formatters_to_dict(RECEIPT_FORMATTERS)
 
 BLOCK_FORMATTERS = {
     'baseFeePerGas': to_integer_if_hex,
-    'extraData': to_hexbytes(32, variable_length=True),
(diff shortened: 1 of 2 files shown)


fselmo:
@eino-makitalo please try injecting the [geth_poa_middleware](https://web3py.readthedocs.io/en/latest/middleware.html?highlight=geth_poa_middleware#geth-style-proof-of-authority) and take a look at this linked section for an example of how to implement it and why this is the recommended approach for now.

Let us know here if that doesn't work out for Avalanche network. I'll keep this open for now.

eino-makitalo:
I tried to inject that but no help. I also found out that extra data need to be even over 80 in avalanche.  I study how this poa middleware is implemented.  Is it a way to change those validation limits in general?

fselmo:
Tested locally with `geth_poa_middleware` and it works connecting to Avalanche node.

Issue #2191, opened 2021-10-29
BaseChainBackend type annotation missing


unparalleled-js:
* Version: master
* Python: 3.9.7
* OS: macOS

### What was wrong?

`BaseChainBackend` missing from `tester` annotation in `EthereumTesterProvider`

### How can it be fixed?

Add `BaseChainBackend` type annotation to `EthereumTesterProvider` `__init__()` `tester` parameter.
---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
please refer to our [documentation](https://web3py.readthedocs.io/en/latest/) and/or join the conversation 
on [discord](https://discord.gg/GHryRvPB84).

PR #2192, opened 2021-10-29
fix: include backend type for tester param in eth test provider ctor

unparalleled-js:
### What was wrong?

fixes: #2191 

The parameter type seems to also permit `BaseChainBackend` by certain static type checkers fussed over it because it was not included in the type annotation.

### How was it fixed?

Add `BaseChainBackend` to the type annotation fort the tester parameter.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2191.misc.rst
+++ b/newsfragments/2191.misc.rst
@@ -1,0 +1,1 @@
+Add BaseChainBackend type annotation to EthereumTesterProvider tester constructor parameter
--- a/web3/providers/eth_tester/main.py
+++ b/web3/providers/eth_tester/main.py
@@ -4,6 +4,7 @@ from typing import (
     Callable,
     Dict,
     Optional,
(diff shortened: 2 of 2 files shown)


unparalleled-js:
> I think this is a good change. Out of curiosity, are you passing in something that inherits from `BaseChainBackend` instead of an instance of `EthereumTester`?

I am inheriting from PyEVMBackend and creating the account_data param that funds the first 10 accounts from a user-set mnemonic and then also providing that mnemonic to super().__init__(). I did this because I noticed you can provide this type to EthereumTesterProvder ethereum_tester init parameter.

unparalleled-js:
However... I really need this change https://github.com/ethereum/eth-tester/pull/209 to make what I am doing actually work.

kclowes:
Got it. Thanks! I'll take a look at that one next.

PR #2197, opened 2021-11-02
Eth tester london updates

fselmo:
### What was wrong?

- Need to update `eth-tester` version to newly released `v0.6.0-beta.2` which includes London support
- Turn on London integration tests and update key mappings for newly introduced params in London (max fees, base fee, effective gas price, etc...)
- Update core tests and integration tests to reflect London changes for eth-tester

### How was it fixed?

- eth-tester version bumped to `v0.6.0-beta.2`
- London core tests updated for eth-tester
- Key mappings updated in `providers/eth-tester/middleware` to match new params for integration tests
- gas price values updated across tests to be >= genesis base fee

### Todo:
- [x] Fill in this template

#### Cute Animal Picture

IMG_20211021_085817

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -699,8 +699,8 @@ Taking the following contract code as an example:
 
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
-    >>> array_contract.functions.setBytes2Value([b'a']).transact({'gas': 420000, 'gasPrice': 21000})
-    HexBytes('0x89f9b3a00651e406c568e85c1d2336c66b4ec40ba82c5e72726fbd072230a41c')
+    >>> array_contract.functions.setBytes2Value([b'a']).transact({'gas': 420000, 'gasPrice': Web3.toWei(1, 'gwei')})
+    HexBytes('0xc5377ba25224bd763ceedc0ee455cc14fc57b23dbc6b6409f40a557a009ff5f4')
     >>> array_contract.functions.getBytes2Value().call()
     [b'a\x00']
     >>> w3.enable_strict_bytes_type_checking()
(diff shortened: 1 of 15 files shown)


unparalleled-js:
I've checked out your branch and am trying to see if it works for me (as you saw, I had a similar branch where I was able to get it to work for me).

But I am running into this error when calling `web3.eth.max_priority_fee`:

```
RPC Endpoint has not been implemented: eth_maxPriorityFeePerGas
```

fselmo:
> ```
> RPC Endpoint has not been implemented: eth_maxPriorityFeePerGas
> ```

This endpoint has not  been implemented in `eth-tester`. To be certain we're talking about the same thing, you are calling an RPC endpoint here that estimates what the `maxPriorityFeePerGas` should be. Whereas sending a transaction with a `maxPriorityFeePerGas` should work, the estimation endpoint is not implemented in eth-tester.

kclowes (approved):
LGTM! Thank you for doing all this work! The only thing I was thinking was that it might be nice to do the conversion to Wei in the examples instead of a big number. For example, instead of `1000000000`, do `Web3.toWei(1, 'gwei')`. But feel free to take it or leave it!

fselmo:
> LGTM! Thank you for doing all this work! The only thing I was thinking was that it might be nice to do the conversion to Wei in the examples instead of a big number. For example, instead of `1000000000`, do `Web3.toWei(1, 'gwei')`. But feel free to take it or leave it!

I do like that, it is more clear. Even over 10 ** 9. I will add this. Thanks!

unparalleled-js:
> > ```
> > RPC Endpoint has not been implemented: eth_maxPriorityFeePerGas
> > ```
> 
> This endpoint has not been implemented in `eth-tester`. To be certain we're talking about the same thing, you are calling an RPC endpoint here that estimates what the `maxPriorityFeePerGas` should be. Whereas sending a transaction with a `maxPriorityFeePerGas` should work, the estimation endpoint is not implemented in eth-tester.
> 
> Edit: I think I see what you mean... you had a static return there. I would be good with the static return @unparalleled-js, is this what you meant?

Weird, for whatever reason it works on my branch still! Does this something in this work explicitly make it raise an exception or something?

fselmo:
> > > ```
> > > RPC Endpoint has not been implemented: eth_maxPriorityFeePerGas
> > > ```
> > 
> > 
> > This endpoint has not been implemented in `eth-tester`. To be certain we're talking about the same thing, you are calling an RPC endpoint here that estimates what the `maxPriorityFeePerGas` should be. Whereas sending a transaction with a `maxPriorityFeePerGas` should work, the estimation endpoint is not implemented in eth-tester.
> > Edit: I think I see what you mean... you had a static return there. I would be good with the static return @unparalleled-js, is this what you meant?
> 
> Weird, for whatever reason it works on my branch still! Does this something in this work explicitly make it raise an exception or something?

You actually had implemented a static return for that endpoint which I am perfectly fine with... it will always return 1 gwei but it should be enough to not be broken for testing.

PR #2208, opened 2021-11-16
Fix typo web3.pm.rst

RH-sdavey:
### What was wrong?

using web3.enable_unstable_package_management_api() had no effect and the AttributeError persisted when trying to access
w3.pm object.

### How was it fixed?

Fixed the typo, so now the enable...() method is called on the import w3 object instead of the w3 library. After this change the pm.PM object is now accessible.

### Todo:

#### Cute Animal Picture


kclowes (approved):
Thank you!

PR #2217, opened 2021-11-20
Update typing extensions to allow v4

### LINKED ISSUE #2216: Using typing-extensions 4 with web3.py and python < 3.8
Hey guys,

This is question on typing-extensions package. In setup.py https://github.com/ethereum/web3.py/blob/74fbcd1761407ee37808173b089d8e44e537fb99/setup.py#L88-L89

you require <4 and python <3.8. From what I understand that is since these extensions you use have made it into stdlib's typing module from 3.8 and on.

But from what I see this requirement creates a probem for projects like ours which are still in python 3.7, use web3.py and want to use typing-extensions 4. Is there any reason for the < 4 requirement?

kclowes:
### What was wrong?

Allow `typing-extensions` dependency to use v4.

Fixes #2216

### How was it fixed?

Loosened the dependency pinning. I did check for the removed modules listed here: https://github.com/python/typing/blob/master/typing_extensions/CHANGELOG, and we are not using any of them. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2217.misc.rst
+++ b/newsfragments/2217.misc.rst
@@ -1,0 +1,1 @@
+Update ``typing-extensions`` dependency to allow for v4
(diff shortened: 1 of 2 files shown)


fselmo (approved):
lgtm !

Issue #2220, opened 2021-11-22
Wrong timeout parameter typing in wait_for_transaction_receipt


maconard:
### What was wrong?

The typing for the `timeout` input to the exposed `wait_for_transaction_receipt` is falsely listed as `int`, seeing as the internal function it calls of the same name from `_utils.transactions` accepts float.

Exposed function falsely showing int: https://github.com/ethereum/web3.py/blob/e01e6ca9e765b54430a152ac18f0890dfa8ae69e/web3/eth.py#L670

Internal function accepting float: https://github.com/ethereum/web3.py/blob/e01e6ca9e765b54430a152ac18f0890dfa8ae69e/web3/_utils/transactions.py#L129

### How can it be fixed?

Simply change line 671 of the `eth` module to use `float` instead of `int` for the `timeout` parameter. 

Thanks!

kclowes:
Thanks! Feel free to submit a PR if you have time :) Otherwise it may be a bit until we can get to it!

broper2:
I can scratch up a quick PR to fix this, will be a good first issue to contribute on web3.py

fselmo:
closed by #2221

PR #2221, opened 2021-11-23
fix timeout typing to float in web3.eth.wait_for_transaction_receipt

broper2:
What was wrong?
The typing for the timeout input to the exposed wait_for_transaction_receipt is falsely listed as int, seeing as the internal function it calls of the same name from _utils.transactions accepts float.

Exposed function falsely showing int:

web3.py/web3/eth.py

Line 670 in e01e6ca

 def wait_for_transaction_receipt( 
Internal function accepting float:

web3.py/web3/_utils/transactions.py

Line 129 in e01e6ca

 def wait_for_transaction_receipt( 
How can it be fixed?
Simply change line 671 of the eth module to use float instead of int for the timeout parameter.

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -668,7 +668,7 @@ class Eth(BaseEth, Module):
         return self.wait_for_transaction_receipt(transaction_hash, timeout, poll_latency)
 
     def wait_for_transaction_receipt(
-        self, transaction_hash: _Hash32, timeout: int = 120, poll_latency: float = 0.1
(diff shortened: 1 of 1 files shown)


broper2:
Related to #2220 
@kclowes any idea with the failing CircleCI checks, I am struggling to reproduce failure locally

kclowes (approved):
Thanks for the PR! Those are just flaky tests. Sorry for the confusion. This change looks good to me!

fselmo (approved):
LGTM 🚀

PR #2224, opened 2021-11-26
Fix access list storage keys and chain id mapping for eth-tester

fselmo:
### What was wrong?

`eth-tester` middleware key mapping needed updates for access list `storageKeys` key

### How was it fixed?

`eth-tester` middleware key mapping was updated for `storage_keys` <-> `storageKeys`

### Todo:

### DIFF
--- a/web3/providers/eth_tester/middleware.py
+++ b/web3/providers/eth_tester/middleware.py
@@ -64,7 +64,58 @@ to_integer_if_hex = apply_formatter_if(is_hexstr, hex_to_integer)
 is_not_named_block = complement(is_named_block)
 
 
-TRANSACTION_KEY_MAPPINGS = {
+# --- Request Mapping --- #
+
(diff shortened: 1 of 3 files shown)


web3/providers/eth_tester/middleware.py
```diff
@@ -64,7 +64,58 @@ def is_hexstr(value: Any) -> bool:
 is_not_named_block = complement(is_named_block)
 
 
-TRANSACTION_KEY_MAPPINGS = {
+# --- Request Mapping --- #
```
fselmo, line 67: I just re-organized / re-named a few of these so that it is more clear which are request formatters and which are result formatters.

kclowes (approved):
🚢 !

web3/providers/eth_tester/middleware.py
kclowes: Yeah, that's much better, thanks!

PR #2228, opened 2021-12-01
Fix for HypothesisDeprecationWarning in tox run test CircleCI #1729

dbfreem:
### What was wrong?
HypothesisDeprecationWarning in tox run test CircleCI
Related to Issue #[ 1729](https://github.com/ethereum/web3.py/issues/1729)

### How was it fixed?

I took some of the suggestions made by @kclows in the #1729  on how to fix this.

Added conftest.py to test/core/utilities and moved web3 fixture to it
Moved fixtures into test_contract_topic_filter.py
Both of these were to fix HypothesisDeprecationWarning while running test

### Todo
I only ran this against tox py38-core since I don't currently have all the interpreters that tox is looking for on my machine

Hope this helps!!

### DIFF
--- a/newsfragments/1729.misc.rst
+++ b/newsfragments/1729.misc.rst
@@ -1,0 +1,1 @@
+Remove HypothesisDeprecationWarnings in tests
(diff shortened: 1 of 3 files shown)


kclowes (approved):
Thank you! This looks much better! I'll add a newsfragment and merge!

PR #2233, opened 2021-12-02
`tools/benchmark/utils`: Use `asyncio.sleep`

mhchia:
### What was wrong?

In async functions, we should use async sleep instead to avoid blocking.

### How was it fixed?

Change `time.sleep` to `asyncio.sleep`. I choose `asyncio.sleep` since it is used in other modules as well. Please correct me if other async frameworks are more suitable.

### DIFF
--- a/newsfragments/2233.misc.rst
+++ b/newsfragments/2233.misc.rst
@@ -1,0 +1,1 @@
+Use non-blocking ``asyncio.sleep`` instead of ``time.sleep``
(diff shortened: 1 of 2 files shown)


kclowes (approved):
Thanks @mhchia! I think those failures are just flaky. I'll add a newsfragment and merge.

mhchia:
NP. Thanks for review:)

PR #2235, opened 2021-12-03
Benchmark Test Fix

### LINKED ISSUE #2234: Benchmarking CI run is silently failing
### What was wrong?

The Benchmarking suite is has these errors on CI: 
```
asyncio.exceptions.TimeoutError
Task exception was never retrieved
```

But is still "passing". An example can be seen [here](https://app.circleci.com/pipelines/github/ethereum/web3.py/1920/workflows/0dae272b-fac5-4713-b1d0-01f1292d9f54/jobs/177486).

### How can it be fixed?

I haven't dug into when or why this is happening, but wanted to leave a placeholder.

dbfreem:
### What was wrong?

Related to Issue [#2234 ](https://github.com/ethereum/web3.py/issues/2234)

### How was it fixed?

Removed the try catch around the sync and async benchmark methods. This was causing the benchmark test to fail silently. 

### Todo:
Still need to figure out what is causing the POST request in request.py to timeout.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -66,8 +66,8 @@ commands=
 basepython=python
 commands=
     python {toxinidir}/web3/tools/benchmark/main.py --num-calls 5
-    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 50
-    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 100
+    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 25
+    python {toxinidir}/web3/tools/benchmark/main.py --num-calls 70
 
 [common-wheel-cli]
 deps=wheel
--- a/web3/tools/benchmark/main.py
+++ b/web3/tools/benchmark/main.py
@@ -81,26 +81,20 @@ async def build_async_w3_http(endpoint_uri: str) -> Web3:
 
 
 def sync_benchmark(func: Callable[..., Any], n: int) -> Union[float, str]:
-    try:
-        starttime = timeit.default_timer()
-        for _ in range(n):
-            func()
-        endtime = timeit.default_timer()
-        execution_time = endtime - starttime
-        return execution_time
-    except Exception:
-        return "N/A"
+    starttime = timeit.default_timer()
+    for _ in range(n):
+        func()
+    endtime = timeit.default_timer()
+    execution_time = endtime - starttime
+    return execution_time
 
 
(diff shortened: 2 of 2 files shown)


dbfreem:
still going to look at why the POST in request.py is getting a timeout

dbfreem:
@kclowes something I noticed while testing this some more. There is a test Geth that gets spun up as part of the benchmark. After a large number of sendTransaction request are submitted to it, it just stops responding. Even if I send a curl from the terminal to it, it will never respond after it takes about 100 to 200 request. It does respond to just a simple get request but the sendTransacation stops working. It seems to me that this is more of an issue with the test Geth node than web3.py benchmarking itself. 

The screenshot below was taken after the TimeOut is received in the benchmark test but before I shut down the debug session, so the node is still running. The request never gets a response.  In the second screenshot I got a 200 on a POST request to the node with no body, so it seems this is specific to the node and sendTransaction

image

image

kclowes:
That is super strange. Thanks for looking into it. Maybe that's why we keep getting timeouts on our integration tests too 🤔

dbfreem:
@kclowes at this point the above PR would be my recommendations. I can't quit figure out why the request to the node start to timeout after a certain number of request. I changed the number of request that are run in the benchmark test, also I removed the throw catch that was causing the benchmark test to fail silently.

One thing that I did note is that a lot of the timeouts were happening in the async request. I noticed that in the `request.py` module that the [async request](https://github.com/ethereum/web3.py/blob/master/web3/_utils/request.py#L54) are not caching the session like they are on the [requests](https://github.com/ethereum/web3.py/blob/master/web3/_utils/request.py#L46) http request. 

geth mentions issues can arise from too many sessions. [issue](https://github.com/ethereum/go-ethereum/issues/17069)

also the aiohttp library recommends [reusing sessions](https://docs.aiohttp.org/en/stable/http_request_lifecycle.html#how-to-use-the-clientsession)

I was thinking of implementing the caching of the session on the aiohttp request in the request.py on a different PR if that makes sense to you.

kclowes:
Thanks for looking into this @dbfreem! I just added a comment to the original issue. I was mistaken, the benchmarking tests are actually running as they should, there are just a bunch of `asyncio.exceptions.TimeoutError`s that need to be cleaned up.

dbfreem:
I added some comments on that issue with some of my findings.

PR #2239, opened 2021-12-07
Include web3 middleware in ENS.fromWeb3()

fselmo:
### What was wrong?

From recent comments on issue #425:
- ``ENS.fromWeb3()`` does not make use of existing middlewares present in the ``Web3`` instance that is passed in to the method. This led to issues using ``geth_poa_middleware``.

Closes #1657 

### How was it fixed?

- Inherit the ``Web3`` instance's middlewares when instantiating the ``ENS`` instance from a ``Web3`` instance.
- Adds a ``middlewares`` property to the ``NamedElementOnion`` data structure / ``web3.middleware_onion`` that returns a list of all current middlewares in appropriate order to be passed in to a new ``Web3`` instance.

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

PR #2241, opened 2021-12-08
Move middlewares to third argument -> no longer breaking change

fselmo:
### What was wrong?

Related to Issue #2239
- Breaking change if ENS was instantiated with `provider` and `addr` as positional arguments as such:

``` python
ens = ENS(web3.manager.provider, address)
```

### How was it fixed?

- Move the new `middlewares` argument to the third positional argument in order to avoid a breaking change.


kclowes (approved):
LGTM! 🚢

PR #2242, opened 2021-12-09
Update mypy to 0.910

kclowes:
### What was wrong?

Somewhere along the way in a downstream library, the `typing-extensions` version requirement got updated.

### How was it fixed?

mypy needs to be updated, along with the py-trie dependency in py-evm. The py-trie dependency has already been merged into py-evm. Once we release py-evm, we'll need to update eth-tester, and pull that in here. Then we can bump `typing-extensions` to v4+.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2242.misc.rst
+++ b/newsfragments/2242.misc.rst
@@ -1,0 +1,1 @@
+Update mypy==0.910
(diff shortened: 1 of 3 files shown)


fselmo (approved):
LGTM 👍

PR #2245, opened 2021-12-10
Increase timout on IPC and WS providers for Geth Integration tests

kclowes:
### What was wrong?
Websocket and IPC integration tests were failing and having to be re-run more often than they weren't. 

### How was it fixed?
Increased the timeout on the IPC and Websocket providers. 

### Todo:

#### Cute Animal Picture


fselmo (approved):
~~Thanks for these changes! We discussed tracking these timeouts in a separate issue as well. Let's make sure we create that and link to this PR there. For now, this is much better than re-running the failed tests manually, as we have been, due to timeouts. LGTM~~ 👍 

The above comment was for marking tests flaky with retries. Increasing the timeout is best, as we discussed via Discord, since we can catch other errors as well. Happy that increasing the timeouts worked here. This will save many headaches 😃  

LGTM! 🎉 👍

Issue #2016, opened 2021-06-07
Allow users to pass a session object to aiohttp


kclowes:
The HTTPProvider allows users to pass a session object for advanced configuration of the underlying `requests` library request. That logic should be emulated for the AsyncHTTPProvider and aiohttp.

fselmo:
closed by #2254

PR #2246, opened 2021-12-11
WIP async session cache

dbfreem:
### What was wrong?
When I was researching the timeouts in the benchmark test I noticed the session is not cached in the request.py for Async aiohttp request. Mainly when I read [this](https://github.com/ethereum/go-ethereum/issues/17069) issue on the geth repo about multiple http sessions I realized this might be an issue.

Also, the aiohttp library recommends reusing [sessions](https://docs.aiohttp.org/en/stable/http_request_lifecycle.html#how-to-use-the-clientsession)

I feel like as a follow up to this PR the request.py should be broken apart into two classes. One for Async and one for standard HTTP. This would better organize the code and make it more testable. 

For the test I started with what was in the test_precached_session and modified it for the async request.
I also started with the code that was used in request.py and HTTPProvider for the caching and modified it for async.

I did find an issue later on related to this in the web3.py issues list.
Related to Issue  #2016

### Todo:


dbfreem:
Looks like py36 and py37 failed. Need to look at why.

Issue #1847, opened 2021-01-22
lru.LRU is not thread-safe; session cache maintenance (web3._utils.request) can crash the interpreter


jamadden:
* Version: 5.15.0, but current master too.
* Python: Any
* OS: Any

### What was wrong?

In https://github.com/ethereum/web3.py/pull/876, code was added to close sessions as they are evicted from the session cache. But in the presence of multiple threads this can lead to interpreter crashes, [because `lru.LRU` is not thread safe](https://github.com/amitdev/lru-dict/issues/22). If, while running the callback function a thread switch happens, and that other thread *also* attempts to add a new session or otherwise manipulate the session cache, the interpreter can crash. (When we re-enter the `lru.LRU` object, the internal state of the object is inconsistent while the callback is running. Because the callback is arbitrary Python code, it can easily drop the GIL.)

This was originally reported in https://github.com/rotki/rotki/pull/2132. There, they're using greenlets, but its easy to demonstrate the crash also using threads (a sample script is in that issue).

### How can it be fixed?

Adding a lock around uses of `_session_cache` ought to fix the problem.

kclowes:
Wow, thanks for the report @jamadden! Will put this near the top of our todo list.

Eenae:
The issue occurred in our setup (> 8 threads, various endpoints) quite stably. Synchronizing the cache access indeed solved the problem.

kclowes:
@Eenae when you say: 

> Synchronizing the cache access indeed solved the problem.

Do you mean PR #2407 fixed your issue?

Eenae:
@kclowes no, it seems https://github.com/ethereum/web3.py/issues/2407 is not a PR.
Did it via a monkey-patch:
```python
from typing import Callable
import threading
import web3._utils.request

lock = threading.Lock()

def locked(fn: Callable) -> Callable:
    def inner(*args, **kwargs):
        with lock:
            return fn(*args, **kwargs)
    return inner

web3._utils.request.cache_session = locked(web3._utils.request.cache_session)
web3._utils.request._get_session = locked(web3._utils.request._get_session)
```

kclowes:
Oops, I meant #2409. I think that's about equivalent to what you have. Thanks for the example!

fselmo:
Resolved in https://github.com/ethereum/web3.py/issues/2049 for web3.py `v6`. Needs a back-port to `v5` PR.

fselmo:
[...]

PR #2254, opened 2021-12-15
Added caching to the async session in request.py and AsyncHTTPProvider

dbfreem:
### What was wrong?
When I was researching the timeouts in the benchmark test I noticed the session is not cached in the request.py for Async aiohttp request. Mainly when I read [this](https://github.com/ethereum/go-ethereum/issues/17069) issue on the geth repo about multiple http sessions I realized this might be an issue.

Also, the aiohttp library recommends reusing [sessions](https://docs.aiohttp.org/en/latest/http_request_lifecycle.html)

I feel like as a follow up to this PR the request.py should be broken apart into two classes. One for Async and one for standard HTTP. This would better organize the code and make it more testable.

### How was it fixed?
For the test I started with what was in the `test_precached_session` and modified it for the async request. I also used the `pytest.mark.asyncio` that I found else where in the project for async testing. 
I also started with the code that was used in request.py and HTTPProvider for the caching and modified it for async.

I upgraded pytest-mock trying to get the Async Mocking to work for the aiohttp session with no success in python 3.6. Since I couldn't get the mock working on the async `session.post` I only tested the session caching and not the `post` call on the session in request.py. Since there didn't seem to be any conflicts with the upgrade as long as it was less than 3.3.0 I left it in this PR.

I did find an issue later on related to session caching in the web3.py issues list.
Related to Issue #2016

### Todo:
- [x] Add entry to the release notes


dbfreem:
@kclowes do you have any feedback on this PR. Am I on the right track?

kclowes:
This looks good @dbfreem! Thanks for looking into this and for all the PRs! 

On first glance, this is about what I would expect in regards to caching the session. I'd probably drop the `pytest-mock` dependency upgrade if you're not really using it in this PR. If you feel strongly about upgrading it in general, it probably should be in its own PR. 

If you pull in the latest master, some of those integration test failures should be less flaky.

dbfreem:
I removed the the pytest-mock upgrade.

kclowes (commented):
Thanks for this PR @dbfreem and for responding to so many of our issues! 

There is an existing bug with the current sync implementation where the LRU cache isn't thread safe (see here: https://github.com/ethereum/web3.py/issues/1847), and I wonder what it would take to make this async version thread safe right off the bat. It may be as easy as pulling in the threading library and using `threading.Lock` as a context manager, but I'm not sure. I am also fine putting that in a different PR (that either you or someone else can tackle, no pressure).
 
Other than that, I just had a few nitpicks! Thanks again!

dbfreem:
I was actually wondering if issue #2253 was a thread safe issue on the standard http side. 

I will take a look at your comments and thread safety.

dbfreem:
So I haven't done a lot of threading before but based on what I say in cache.py plus reading up on python.org about threading lock I think this will work. Unfortunately, this is really hard to test with a unit test. 

If you agree with this threading approach I can make the same change in another PR for #1847

kclowes (commented):
This looks good @dbfreem. Thank you for the work! I've been looking a little bit at how to test the threading, so I'll let you know if I see anything promising.

dbfreem:
This still needs a little work so don't merge this. I will take a look this weekend.

dbfreem:
So here is where I am with this. After I started writing a test to test the thread safety stuff I realized that `ClientSession.close` is an async method and I needed to wrap it in an async/await. When I did this the callback from the cache stopped working. So I decided to write a simple caching class since the requirements here are not that complex.

`SessionCache` is a class backed by an `OrderedDict` that simply removes the FIFO value from the dict when it reaches it's preset capacity. I also moved the closing of the connection to the calling code in `request.py` . IMO that code is more capable of knowing the best way to handle the closing of the connection either async or not instead of the cache handling the closing. 

One other side effect that came of this is that I am no longer able to reproduce the threading issue that was there before I changed to `SessionCache`. Before implementing `SessionCache` if I ran multiple thread inserting into the cache I would get weird errors when the callback was called. After implementing `SessionCache` I am no longer able to reproduce these. I did leave the thread lock in for now just for good measure. But we may be able to try `SessionCache` without the thread lock and get users to test it out if they have seen threading issues before. 

I do see that lru-dict is used in `cache.py` so this doesn't remove the dependency on it but I do feel like the use case for the cache in `request.py` is pretty simple and `SessionCache` provides a lot more control over the cache. 

Thoughts?? I can go another route if you all do not agree with the `SessionCache` class. I would need to figure out how to get lru-dict to call the async callback, which could have been user error on my end but I couldn't get that to work. If you do agree with the new approach I can move the `request.Session` cache over to `SessionCache`.

dbfreem:
Ok, final answer here. Once, I built the unit test for the `AsyncHTTPProvider` I realized that the session wasn't being cached since the `cache_async_session` method is async. I moved this to an async method on the `AsyncHTTPProvider` to fix this.

All these changes started from the fact that the `ClientSession.close()` is an asnyc method itself. When caching an item in a fixed sized cache the eviction of some items need to be handled when new items enter the cache. This means calling `ClientSession.close()` which again is async. 

Also, on the new `test_async_http_provider` module I added. I got the http://mynode.local:8545 from the `test_http_provider` module.

kclowes:
Thanks @dbfreem! One of us will take a look this week!

dbfreem:
@pacrob  or @kclowes  I wanted to check in and see if I could get a review on this one.

kclowes:
@dbfreem haven't forgotten this one. Will take a look! Thank you!

dbfreem:
Thanks @kclowes. I know this turned into a big PR, sorry about that. It kind of spiraled once I realized that the session close for the aiohttp was an async method and it wasn't actually being closed the way I had it at first.

docs/providers.rst
```diff
@@ -379,9 +379,11 @@ AsyncHTTPProvider
       be omitted from the URI.
     * ``request_kwargs`` should be a dictionary of keyword arguments which
       will be passed onto each http/https POST request made to your node.
+    * the ``cache_async_session()`` method allows you to use your own ``aiohttp.ClientSession`` object. This is and async method and not part of the consturctor
```
pacrob, line 382: an async method and not part of the constructor

pacrob, line 382:
```suggestion
    * the ``cache_async_session()`` method allows you to use your own ``aiohttp.ClientSession`` object. This is an async method and not part of the constructor
```

web3/_utils/request.py
```diff
@@ -18,6 +23,37 @@
...
+
+    def __init__(self, size: int):
+        self._size = size
+        self._data: OrderedDict[str, Any] = OrderedDict()
+
+    def cache(self, key: str, value: Any) -> Dict[str, Any]:
+        evicted_items = None
+        # If the key is already in the OrderedDict just update it
+        # and don't evict any values. Ideally, we could still check to see
+        # if there are too many items in the OrderDict but that may rearange
```
pacrob, line 36:
```suggestion
        # if there are too many items in the OrderedDict but that may rearrange
```

dbfreem:
@pacrob I committed your suggestions. Thanks

pacrob:
Thanks for adding this, @dbfreem ! I added the _ myself just to cut cycle time. Otherwise, awesome work as usual 👍

dbfreem:
No worries about the _ @pacrob and you are welcome.

DefiDebauchery:
Just a note that I'm having trouble utilizing this. I typically run web3 within a pycord (fork of discord.py) bot, which until this change has been working without issue.

I applied `request.py` and `async_rpc.py`. The conflict may be coming from the session lock, but hard to really debug.. I turned on the logging module and it seems to fail after initializing 10 RPC instances (needed for different evm networks), only one of them (not even using CustomSession yet). Will dig more, but wanted to mention this in case anyone might have thoughts.

kclowes:
@DefiDebauchery thanks for the report. Any chance you can open up a new issue with a minimum reproducible example?

Issue #2129, opened 2021-09-02
Support get_logs in AsyncHTTPProvider


tsudmi:
Hi,

Fetching the logs (especially if it's a long list) is one of the most time-consuming operations.  It would be great to add support for executing `get_logs` using `AsyncHTTPProvider`.

Thanks!

kclowes:
Thanks for the request! It's near the top of our priority list!

AngeloCa:
Any update on an async-ification of web3.eth.get_transaction_receipt ?

pacrob:
Thanks for the reminder! Working on it now.

pacrob:
fixed in pr #2310

PR #2265, opened 2021-12-16
Asyncify eth.get_transaction_receipt and eth.wait_for_transaction_receipt

pacrob:
### What was wrong?

Added async `eth.get_transaction_receipt` and `eth.wait_for_transaction_receipt`

Related to Issue #1413 
Closes Issue #2129

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -62,7 +62,6 @@ from web3._utils.transactions import (
     extract_valid_transaction_params,
     get_required_transaction,
     replace_transaction,
-    wait_for_transaction_receipt as utils_wait_for_transaction_receipt,
 )
 from web3.contract import (
     ConciseContract,
@@ -71,6 +70,7 @@ from web3.contract import (
 )
 from web3.exceptions import (
     TimeExhausted,
+    TransactionNotFound,
 )
 from web3.iban import (
     Iban,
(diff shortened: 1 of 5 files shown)


web3/eth.py
```diff
@@ -277,6 +277,17 @@ def call_munger(
         mungers=None,
     )
 
+    _get_transaction_receipt: Method[Callable[[_Hash32], TxReceipt]] = Method(
+        RPC.eth_getTransactionReceipt,
+        mungers=[default_root_munger]
+    )
+
+    # _get_transaction_receipt_awaitable: Method[Callable[[_Hash32],
```
kclowes, line 285: this can go :)

kclowes (approved):
This looks good to me! Looks like there is a left over comment that can be removed. Did you manually test it to make sure it works like you think it should?

web3/eth.py
```diff
@@ -403,6 +414,36 @@ async def get_transaction_count(
...
+                while True:
+                    try:
+                        tx_receipt = await self._get_transaction_receipt(  # type: ignore
+                            transaction_hash
+                        )
+                    except TransactionNotFound:
+                        tx_receipt = None
+                    if tx_receipt is not None:
+                        break
+                    _timeout.sleep(poll_latency)
```
kclowes, line 436: Sorry to pop up again after approving this, but I think we want to be able to await this so it's not blocking. I think the most straightforward route is to get rid of the timeout context and use `await asyncio.sleep` here. If you want to see if it makes a difference at all, you could add it to our benchmarking suite and see what happens :)

web3/eth.py
pacrob: Yup, thanks for catching

web3/eth.py
pacrob: Makes sense. _timeout.sleep replaced with await asyncio.sleep

PR #2266, opened 2021-12-17
Rethink `eth_chainId` calls

fselmo:
### What was wrong?

- We can often end up making multiple calls to `eth_chainId` to validate the `chainId` param in a transaction, or to add the default value to each transaction making use of default values (i.e. `contract.buildTransaction()`).

Related to Issue #2207 

### How was it fixed?

- One possible solution is setting an internal chain id value on provider instantiation

- Another solution would be to cache the call to `eth_chainId`
   - Downside to this is default caching happens once per instance so each call for `w3.eth.chain_id`, on its own, would return the cached value and would not be an earnest call.

### DIFF
--- a/web3/__init__.py
+++ b/web3/__init__.py
@@ -45,7 +45,6 @@ __all__ = [
     "HTTPProvider",
     "IPCProvider",
     "WebsocketProvider",
-    "TestRPCProvider",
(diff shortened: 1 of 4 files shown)


fselmo:
closed since caching seems like the best idea here.

PR #2269, opened 2021-12-17
clean up per comments in #2265

pacrob:
### What was wrong?

Changes requested in #2265 didn't stick

Related to Issue #2265

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_ethereum_tester.py
+++ b/tests/integration/test_ethereum_tester.py
@@ -1,4 +1,5 @@
 import functools
+from _pytest.outcomes import Failed
 import pytest
 
 from eth_tester import (
@@ -24,6 +25,7 @@ from web3._utils.module_testing import (
 from web3._utils.module_testing.emitter_contract import (
     EMITTER_ENUM,
 )
+from web3.exceptions import TransactionNotFound
 from web3.providers.eth_tester import (
     EthereumTesterProvider,
 )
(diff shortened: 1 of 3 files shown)


fselmo:
I had a couple nice-to-have nits since we already edited the code in the previous commit too 😅 ...

- We can remove the comma from the `TimeoutError` message across both the async and sync versions of `wait_for_transaction_receipt()` since it isn't necessary. 

- We can add an integration test for the synchronous method if you are feeling up for it and since you already have the meat and potatoes of the test with the async counterpart it should be easy to copy / paste and remove the `await`s. We should've had this at some point and this will be nice to have.

- We can also check the error message in the tests to make sure it is the `TimeExhausted` exception that we expect (as opposed to another one with a different exception message that could be thrown sometime along the way) by loosely checking the messaging around the variables we expect to be in the message.

  Some variation of the following... I couldn't get the `match` kwarg in `pytest.raises()` to pick up the `txn_hash` but that might work too:

``` python
@pytest.mark.asyncio
async def test_async_eth_wait_for_transaction_receipt_unmined(
    self, async_w3: "Web3", unlocked_account_dual_type: ChecksumAddress
) -> None:
    txn_hash = await async_w3.eth.send_transaction({  # type: ignore
        'from': unlocked_account_dual_type,
        'to': unlocked_account_dual_type,
        'value': Wei(1),
        'gas': Wei(21000),
        'maxFeePerGas': async_w3.toWei(3, 'gwei'),
        'maxPriorityFeePerGas': async_w3.toWei(1, 'gwei')
    })

    timeout = 2
    with pytest.raises(TimeExhausted) as exc_info:
        await async_w3.eth.wait_for_transaction_receipt(  # type: ignore
            txn_hash, timeout=timeout
        )
    assert (_ in str(exc_info) for _ in [repr(txn_hash), timeout])
```

fselmo (approved):
LGTM! Thanks again. Let's squash and merge as we discussed since some of the commit history was dirtied a bit. 🎉

PR #2271, opened 2021-12-19
Fixes #2259

### LINKED ISSUE #2259: Move usage of `eth_maxPriorityFeePerGas` to `eth_feeHistory`
Since `eth_maxPriorityFeePerGas` is Geth-specific, we should stop using it to fill in gas values. We should use `eth_feeHistory` instead. I think it still makes sense to keep the RPC endpoint in `web3/eth.py`, but all places we call `eth_maxPriorityFeePerGas` should be changed.

Comment from #2237: 

> Hi, I'd like to add that `eth_maxPriorityFeePerGas` isn't part of [ETH JSON-RPC spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=true&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false). It is an endpoint Geth added on its own. Clearly some ETH providers don't have a plan to add it https://github.com/nomiclabs/hardhat/issues/1664. Maybe using `eth_feeHistory` is a better idea since it's part of the spec?

> Also, it would be great if this can be a built-in middleware so that users could choose between `eth_maxPriorityFeePerGas` (or `eth_feeHistory`) and `eth_gasPrice` based on the chain they want to use.

_Originally posted by @guoyiteng in https://github.com/ethereum/web3.py/issues/2237#issuecomment-993133677_

broper2:
### What was wrong?

Dependency on eth_maxPriorityFeePerGas

Related to Issue #2259

### How was it fixed?

See analysis on issue #2259, eth_maxPriorityFeePerGas == (gas_price - latest block base fee)

Tests expected results were updated to new logic implemented, and verified they make sense against test provider api 
endpoint defaults. When connected to ethereum network, this will be a no-impact change to an api supporting eth_maxPriorityFeePerGas

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -54,8 +54,8 @@ def test_build_transaction_not_paying_to_nonpayable_function(
         'to': payable_tester_contract.address,
         'data': '0xe4cb8f5c',
         'value': 0,
-        'maxFeePerGas': 2750000000,
-        'maxPriorityFeePerGas': 10 ** 9,
+        'maxFeePerGas': 1875000000,
+        'maxPriorityFeePerGas': 125000000,
         'chainId': 61,
     }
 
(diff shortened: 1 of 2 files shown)


broper2:
Checking some of these tests....

PR #2276, opened 2021-12-24
Fixes #2259, remove dependency on eth_maxPriorityFeePerGas

### LINKED ISSUE #2259: Move usage of `eth_maxPriorityFeePerGas` to `eth_feeHistory`

broper2:
### What was wrong?

Related to Issue #2259:
~Removal of eth_maxPriorityFeePerGas in determining transaction param defaults, as its Geth-specific.~

### How was it fixed?

Create native tip estimator based on Geth eth_maxPriorityFeePerGas implementation details (see https://docs.alchemy.com/alchemy/guides/eip-1559/gas-estimator). 

Testing: Logic was tested using unit tests, I added a static return for feeHistory endpoint that evaluated max_priority_fee to double what was current testing default for easy verification. Also ran some functional testing....I looped for about an hour comparing this native implementation to what eth_maxPriorityFeePerGas was returning. On average, the different was about 359394279 Wei. I think this is reasonable....but also know its not my opinion that counts :) Feel free to test further.

---

**@fselmo:**
**I abstracted out the logic from above into a utility method, changed some of the params, added min and max limits to the estimated fee,  added tests. I also kept `eth_maxPriorityFeePerGas` as the default if it is available and only resort to the fee history calculation if it is not (while warning the user that the fallback method was triggered).**

### TODO:

- [x] Add newsfragment
- [x] Add cute animal picture

### DIFF
--- a/newsfragments/2259.feature.rst
+++ b/newsfragments/2259.feature.rst
@@ -1,0 +1,1 @@
+Calculate a default ``maxPriorityFeePerGas`` using ``eth_feeHistory`` when ``eth_maxPriorityFeePerGas`` is not available, since the latter is not a part of the Ethereum JSON-RPC specs and only supported by certain clients.
--- a/tests/core/contracts/test_contract_buildTransaction.py
+++ b/tests/core/contracts/test_contract_buildTransaction.py
@@ -54,8 +54,8 @@ def test_build_transaction_not_paying_to_nonpayable_function(
         'to': payable_tester_contract.address,
         'data': '0xe4cb8f5c',
         'value': 0,
-        'maxFeePerGas': 2750000000,
-        'maxPriorityFeePerGas': 10 ** 9,
+        'maxFeePerGas': 3750000000,
+        'maxPriorityFeePerGas': 2 * (10 ** 9),
         'chainId': 61,
     }
 
@@ -76,8 +76,8 @@ def test_build_transaction_with_contract_no_arguments(web3, math_contract, build
         'to': math_contract.address,
         'data': '0xd09de08a',
         'value': 0,
-        'maxFeePerGas': 2750000000,
-        'maxPriorityFeePerGas': 10 ** 9,
+        'maxFeePerGas': 3750000000,
+        'maxPriorityFeePerGas': 2 * (10 ** 9),
         'chainId': 61,
     }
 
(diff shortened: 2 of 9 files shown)


fselmo:
Hey @broper2 thanks for spending some time on this! I took your suggestion but made some tweaks and abstracted it out unto a utility method. Tweaks were mostly to make sure we can provide decent values all the time and don't contribute to fee bloating when fee prices are high:

- Keep the tip within a `gwei` range of: 1 < fee < 1.5
- Use the 5th percentile and only the last 10 blocks for the calculation (yielded more accurate results)

One last thing I'm a bit unsure of is turning on the eth-tester `eth_feeHistory` method with a static return like that. That would mean even when requesting the last 10 or 20 blocks we still only return lists with static lengths of 2 for those values. Thoughts? I lean towards turning this back off for now and rethinking it for another PR perhaps if this feature is requested.

Thanks again!

pacrob (approved):
lgtm!

marcgarreau (approved):
thanks for the walkthrough! 🚢

PR #2286, opened 2022-01-03
Add 'get_text' method to look up ENS text record values

smk762:
### What was wrong?

Noticed this at https://docs.ens.domains/dapp-developer-guide/resolving-names

image

### How was it fixed?

- added get_text method
- added related ABI entry

Usage: `ns.get_text('dragonhound.eth', "avatar")`
Returns: `https://i.imgur.com/gAD7BxX.jpg`

### Todo:
- [ ] Add entry to the [docs](https://web3py.readthedocs.io/en/latest/ens_overview.html)

#### Cute Animal Picture

image

### DIFF
--- a/ens/abis.py
+++ b/ens/abis.py
@@ -1173,6 +1173,28 @@ RESOLVER = [
     "type": "function"
   },
   {
+    "constant": True,
+    "inputs": [
+      {
+        "name": "node",
+        "type": "bytes32"
+      },
+      {
+        "name": "key",
+        "type": "string"
+      }
+    ],
+    "name": "text",
+    "outputs": [
+      {
+        "name": "",
+        "type": "string"
+      }
+    ],
+    "payable": False,
+    "type": "function"
+  },
+  {
     "inputs": [
       {
         "name": "ensAddr",
--- a/ens/main.py
+++ b/ens/main.py
@@ -273,6 +273,21 @@ class ENS:
         node = raw_name_to_hash(name)
         return self.ens.caller.owner(node)
 
+    def get_text(self, name: str, key: str) -> str:
+        """
+        Get the value of a text record by key from an ENS name.
+
+        :param str name: ENS name to look up
+        :param str key: ENS name's text record key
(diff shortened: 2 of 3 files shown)


kclowes (commented):
Thank you for this PR! The code looks solid, but it will need a test before we merge. Let me know if you want some direction in that regard!

smk762:
I can add tests, no problem. Would this be the correct file to do so? https://github.com/ethereum/web3.py/blob/master/tests/ens/test_ens.py

I noticed in other places in your tests code it was using `TESTER.eth` but there are no text record entries I can see for that at https://app.ens.domains/name/TESTER.eth/details

Can use an alternative, just concerned that if the text record is changed in future it may result in failed test.

payton:
@kclowes just bumping this for @smk762 's question about test location / strategy.

I was looking to use / add this functionality and came across the PR :)

kclowes:
Thanks for the bump @payton!

@smk762 sorry, I don't know how I missed this! We get a lot of issues so feel free to bump if you haven't heard in a few days.

> Would this be the correct file to do so? https://github.com/ethereum/web3.py/blob/master/tests/ens/test_ens.py

It looks like the public API methods all have their own file in tests/ens, so I would just make a new file and add tests there. 

> I noticed in other places in your tests code it was using TESTER.eth but there are no text record entries I can see for that at https://app.ens.domains/name/TESTER.eth/details

> Can use an alternative, just concerned that if the text record is changed in future it may result in failed test.

I would just use an alternative if there is one, maybe one that seems like it has staying power :) if tests fail, we can re-evaluate. 

Thank you!

kclowes:
I started working on tests for this, and realized that we probably need a `set_text` method as well. I'll PR that separately and then I'll add a quick test here and merge. Thanks @smk762!

kclowes:
Just kidding about the separate PR, they're more inter-related than I thought. I'll just add to this PR.

kclowes:
This just got merged in #2413 and #2395. Thanks for getting this rolling @smk762!

PR #2288, opened 2022-01-04
Add support for external modules to web3 instance

### LINKED ISSUE #2231: Difficult to attach one extra module to a web3 instance..
### What was wrong?

There is no simple way to include a custom module in a web3 instance while still also including all of the default modules.

### How can it be fixed?

Ideally it should be a one-liner

```python

class Foo(ModuleV2):
    ...

w3 = Web3(...)
w3.attach('foo', Foo)

# or

w3 = Web3(..., extra_modules=(('foo', Foo),)
```

fselmo:
### What was wrong?

- We don't have an easy way to attach modules to the `Web3` instance. This will help users customize their web3 experience but can also help to support things like attaching layer 2 modules from a separate library or any other modules that can prove useful.

- Closes #2231

### How was it fixed?

- Added support for attaching external modules at `Web3` instantiation. This is separate from the internal default modules already attached at instantiation.
- Added support for attaching a single module via a new `attach_module()` method
- Added support for passing in a module as the class itself if there are no submodules, rather than having to pass in a tuple with a single module class.
- Tests added

### Todo:
- [x] Add documentation
- [x] Add cute animal picture

### DIFF
--- a/docs/web3.main.rst
+++ b/docs/web3.main.rst
@@ -11,7 +11,7 @@ Web3 API
 
 .. py:class:: Web3(provider)
 
-Each ``web3`` instance exposes the following APIs.
+Each ``Web3`` instance exposes the following APIs.
 
 Providers
 ~~~~~~~~~
@@ -388,7 +388,7 @@ Check Encodability
 RPC APIS
 ~~~~~~~~
 
-Each ``web3`` instance also exposes these namespaced APIs.
+Each ``Web3`` instance also exposes these namespaced APIs.
 
 
 .. py:attribute:: Web3.eth
@@ -410,3 +410,68 @@ Each ``web3`` instance also exposes these namespaced APIs.
 .. py:attribute:: Web3.parity
 
     See :doc:`./web3.parity`
+
+
+Attaching Modules
+~~~~~~~~~~~~~~~~~
+
+Modules that inherit from the ``web3.module.Module`` class may be attached to the ``Web3`` instance either at
+instantiation or by making use of the ``attach_module()`` method.
+
+To instantiate the ``Web3`` instance with external modules:
+
+.. code-block:: python
+
+    >>> from web3 import Web3, EthereumTesterProvider
+    >>> w3 = Web3(
+    ...     EthereumTesterProvider(),
+    ...     external_modules={
+    ...         # ModuleClass objects in this example inherit from the `web3.module.Module` class
+    ...         'module1': ModuleClass1,
+    ...         'module2': (ModuleClass2, {
+    ...             'submodule1': ModuleClass3,
+    ...             'submodule2': (ModuleClass4, {
+    ...                 'submodule2a': ModuleClass5,  # submodule children may be nested further if necessary
+    ...             })
+    ...         })
+    ...     }
+    ... )
+
+    # `return_zero`, in this case, is an example attribute of the `ModuleClass1` object
+    >>> w3.module1.return_zero
+    0
+    >>> w3.module2.submodule1.return_one
+    1
+    >>> w3.module2.submodule2.submodule2a.return_two
+    2
+
+
+.. py:method:: w3.attach_module(module_name, module)
+
+    The ``attach_module()`` method can be used to attach an external module after the ``Web3`` instance has been
+    instantiated.
+
+    .. code-block:: python
+
+        >>> from web3 import Web3, EthereumTesterProvider
+        >>> w3 = Web3(EthereumTesterProvider())
+
+        # attaching a single module - in this case, one with the attribute `return_zero`
+        >>> w3.attach_module('module1', ModuleClass1)
+        >>> w3.module1.return_zero
+        0
+
+        # attaching a module with submodules
+        >>> w3.attach_module(
+        ...     'module2',
+        ...     (ModuleClass2, {
+        ...         'submodule1': ModuleClass3,
+        ...         'submodule2': (ModuleClass4, {
+        ...             'submodule2a': ModuleClass5,
+        ...         })
+        ...     })
+        ... )
+        >>> w3.module2.submodule1.return_one
+        1
+        >>> w3.module2.submodule2.submodule2a.return_two
+        2
--- a/docs/web3.main.rst
+++ b/docs/web3.main.rst
@@ -416,7 +416,7 @@ Attaching Modules
 ~~~~~~~~~~~~~~~~~
 
 Modules that inherit from the ``web3.module.Module`` class may be attached to the ``Web3`` instance either at
-instantiation or by making use of the ``attach_module()`` method.
+instantiation or by making use of the ``attach_modules()`` method.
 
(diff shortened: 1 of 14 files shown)


web3/eth.py
```diff
@@ -458,7 +458,7 @@ async def call(
         return await self._call(transaction, block_identifier, state_override)
 
 
-class Eth(BaseEth, Module):
```
fselmo, line 461: `BaseEth` already inherits from `Module` so this didn't seem necessary.

web3/_utils/module.py
```diff
@@ -18,11 +18,13 @@
 
 def attach_modules(
     parent_module: Union["Web3", "Module"],
-    module_definitions: Dict[str, Sequence[Any]],
+    module_definitions: Dict[str, Any],
```
fselmo, line 21: The recursive nature of this method makes it really hard to add any useful typing here so this has to be `Any` unfortunately. But I am new to typing in python, maybe there's a way to make this stricter 🤔 . At least all the uses of this method are able to be a tad more strict.

web3/main.py
```diff
@@ -323,6 +331,35 @@ def solidityKeccak(cls, abi_types: List[TypeStr], values: List[Any]) -> bytes:
...
+    def attach_module(
+        self,
+        module_name: str,
+        module: Union[Type[Module], Sequence[Any]]
+    ) -> None:
+        """
+        Attach a module to the `Web3` instance. Modules should inherit from the `web3.module.Module`
+        class.
+
+        Attaching a simple module:
```
fselmo, line 343: I added this before the docs. Perhaps this is a bit redundant? Thoughts?

web3/_utils/module.py
kclowes: I found this documentation: https://mypy.readthedocs.io/en/stable/protocols.html#recursive-protocols, but I would have to play with it some to figure out how exactly we'd use it. If digging into that seems out of the scope of this PR, I'm fine leaving this as-is.

web3/eth.py
kclowes: Nice catch!

docs/web3.main.rst
```diff
@@ -410,3 +410,68 @@ Each ``web3`` instance also exposes these namespaced APIs.
...
+    0
+    >>> w3.module2.submodule1.return_one
+    1
+    >>> w3.module2.submodule2.submodule2a.return_two
+    2
+
+
+.. py:method:: w3.attach_module(module_name, module)
+
+    The ``attach_module()`` method can be used to attach an external module after the ``Web3`` instance has been
```
kclowes, line 451: It looks like you can only attach one top-level module here? It may be nice to be able to pass a few if that's not too difficult of a refactor. For example: 

`w3.attach_module({'eth': Eth, 'geth': Geth})`

web3/main.py
kclowes: I'd vote to only maintain these docs in one place so it's less likely that they get out of sync, but if you feel strongly about leaving them I'm fine with that too.

web3/_utils/module.py
```diff
@@ -18,11 +18,13 @@
...
     w3: Optional[Union["Web3", "Module"]] = None
 ) -> None:
     for module_name, module_info in module_definitions.items():
-        module_class = module_info[0]
+        module_info_is_list_like = isinstance(module_info, Sequence)
```
kclowes, line 25: I think we still need to allow the tuple form: `{'eth': (Eth,)}` so that we're not introducing a breaking change here, which definitely makes this method uglier. But we can remove the top-level tuple allowance in v6; we just need to add a comment to the v6 breaking changes issue or create an issue and link it in the v6 breaking changes issue.

docs/web3.main.rst
```diff
@@ -11,7 +11,7 @@ Web3 API
 
 .. py:class:: Web3(provider)
 
-Each ``web3`` instance exposes the following APIs.
+Each ``Web3`` instance exposes the following APIs.
```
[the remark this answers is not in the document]
fselmo: Yeah I guess what I think this means is the instance of the `Web3` class which we designate into a `w3` var by convention. So technically it is the instance of the `Web3` class, right? That's how I read this.

docs/web3.main.rst
fselmo: I think this could be useful, I was just going by the original issue raised which was to add a single module by just passing in a simple name and class. I'm open to making this more complex though if we see the benefit in it. I also don't know how many modules we will be attaching in the future and if passing in a whole dict after instantiation is really a desired thing?

I liked the simplicity of the issue raised to just passing in (`module_name`, `ModuleClass`). But yeah, I'm open to the idea of passing in a dict.

docs/web3.main.rst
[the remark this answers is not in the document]
kclowes: Okay, yeah that makes sense to me. lol semantics

web3/_utils/module.py
fselmo: For sure, did not want to introduce a breaking change. The tuple can still be passed in. This just makes it so both can be passed in. And since it can be simplified, I just changed it in the docs.

`{'eth': (Eth,)}` works but just seems like it is more complicated than needs to be if the module doesn't have submodules. So these changes also allow `{'eth': Eth}` if it is a simple module.

docs/web3.main.rst
kclowes: I'm fine leaving this as attaching a single module. We can always re-think if we get an issue/feature request about it.

web3/main.py
fselmo: Don't feel strongly. I'm not even sure how many users really use the `__doc__` as an instruction. We can keep the examples in the actual documentation then 👌 . I think if we start adding this to every method, our files just get way too lengthy. I will take this out 👍

web3/_utils/module.py
kclowes: Oh sweet. Do you mind adding a test then to make sure we don't regress there?

docs/web3.main.rst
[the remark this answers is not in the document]
fselmo: lol true. Between `web3`, `Web3`, and `w3` it is difficult to pick out what makes sense across the code base.

web3/_utils/module.py
fselmo: > Oh sweet. Do you mind adding a test then to make sure we don't regress there?
> 
> Edit: Or maybe there is a test and I just missed it

Good point... I think that can be useful. Though in the tests we do pass in tuples for the modules that do have submodules so in theory it should work with a tuple with a single class. But this isn't very obvious so I can add a simple test.

kclowes (approved):

docs/web3.main.rst
fselmo: Just noting here what we all discussed... it probably does makes more sense to just open this up and use a `dict` as the argument. This change was pushed in the last commit.

PR #2293, opened 2022-01-07
explicitly add output_values to compile_source example

pacrob:
### What was wrong?

A change from solc (see [here](https://github.com/iamdefinitelyahuman/py-solc-x/pull/140)) caused our contracts example to break.

### How was it fixed?

Explicitly defining the `output_values` fixes our example without waiting for `py-solc-x` to fix it on their end and makes it more clear where we get those values later in the example.

### Todo:

#### Cute Animal Picture

image


kclowes (approved):
LGTM! Thank you!

PR #2304, opened 2022-01-14
Don't require modules to inherit from the Module class

fselmo:
### What was wrong?

- All attached modules needed to inherit from the `web3.module.Module` class

Related to Issues #2231, #2288

### How was it fixed?

- Remove requirement for all modules to inherit from the `Module` class

### Todo:
- [x] Add a cute animal picture
- [x] @marcgarreau, take a look at the doc descriptions / changes around modules in general. I added a bit to try to make more clear what modules are but I think this could use some love. Maybe in a separate PR, but if you have suggestions feel free to comment here or push commits to this PR.

### DIFF
--- a/docs/web3.main.rst
+++ b/docs/web3.main.rst
@@ -385,10 +385,10 @@ Check Encodability
         False
 
 
-RPC APIS
-~~~~~~~~
+RPC API Modules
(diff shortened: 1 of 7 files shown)


marcgarreau (approved):
🚀

PR #2312, opened 2022-01-19
Align NamedTuples

calina-c:
### What was wrong?
Using the Python NamedTuple structure, the ABI is not recognised.

### How was it fixed?
Converted named tuples to tuples. Given a NamedTuple with name "MyTuple", `isinstance(t, tuple)` is True for both a regular tuple and a MyTuple object, but type(t) will be "tuple" for regular tuples and "MyTuple" respectively. This means the typing should be adapted.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2312.feature.rst
+++ b/newsfragments/2312.feature.rst
@@ -1,0 +1,1 @@
+Allow NamedTuples in ABI inputs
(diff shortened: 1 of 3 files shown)


kclowes:
Thanks @calina-c! Do you mind adding a test to this? It will probably be fastest if you can add it, otherwise I'll put it in our queue

calina-c:
@kclowes Added the test.

kclowes (approved):
Thanks @calina-c! Will merge once tests pass.

Issue #2318, opened 2022-01-25
bug ens contract function called twice


dbfreem:
I was looking at what it would take to async resolver in ENS so that it could be used in the name_to_address_middleware when I noticed the below. Feel free to close this issue if I am overlooking something but I am putting in a PR to remove the second call.

In the [main.py](https://github.com/ethereum/web3.py/blob/master/ens/main.py#L245-L248) module for ens the contract function is called twice. Once on line 245 and again on line 248.

fselmo:
closed by #2319

PR #2319, opened 2022-01-25
Feature/ens request

dbfreem:
### What was wrong?

Related Issue #2318 

I was looking at what it would take to async the resolver in ens so it can be used with the name_to_address_middleware when I noticed that the ens contract function to resolve an address was called twice

On line 245 in the [main.py](https://github.com/ethereum/web3.py/blob/master/ens/main.py#L245-L248) for ens the contract function is called to resolve the ChecksumAddress. It is called again in the return statement on line 248

### How was it fixed?

removed the second call and just returned `address`

### Todo:

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -245,7 +245,7 @@ class ENS:
             address = lookup_function(namehash).call()
             if is_none_or_zero_address(address):
                 return None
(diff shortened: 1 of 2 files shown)


kclowes (approved):
Nice find! Thanks!

PR #2328, opened 2022-02-02
Properly initialize external modules

fselmo:
### What was wrong?

We weren't properly initializing external modules that do not inherit from the ``web3.module.Module`` class. This led to class methods not working as there was no `self` to reference.

Also, based on our docs, we expected to be able to initialize an external module with passing in a ``Web3`` instance as the first argument. This was not the case and wasn't tested.

### How was it fixed?

- Properly initialize modules that do not inherit from the `Module` class by adding `()`
- Allow for modules to accept a `Web3` instance as the first argument of their `__init__()` if it requires use of the web3 instance

### Todo:
- [x] Add cute animal picture
- [x] Update docs

#### Cute Animal Picture

GOPR4249

### DIFF
--- a/docs/web3.main.rst
+++ b/docs/web3.main.rst
@@ -419,17 +419,15 @@ web3.py library.
 External Modules
 ~~~~~~~~~~~~~~~~
 
-External modules can be used to introduce custom or third-party APIs to your ``Web3`` instance. Adding external modules
-can occur either at instantiation of the ``Web3`` instance or by making use of the ``attach_modules()`` method.
-
-Unlike the native modules, external modules need not inherit from the ``web3.module.Module`` class. The only requirement
-is that a Module must be a class and, if you'd like to make use of the parent ``Web3`` instance, it must be passed into
-the ``__init__`` function. For example:
+External modules can be used to introduce custom or third-party APIs to your ``Web3`` instance. They are not required to
+utilize the parent ``Web3`` instance; if no reference is required, the external module need only be a standard class.
+If, however, you do want to reference the parent ``Web3`` object, the external module must inherit from the
+``web3.module.Module`` class and handle the instance as an argument within the ``__init__`` function:
 
 .. code-block:: python
 
-    >>> class ExampleModule():
-    ...
+    >>> from web3.module import Module
+    >>> class ExampleModule(Module):
     ...     def __init__(self, w3):
     ...         self.w3 = w3
     ...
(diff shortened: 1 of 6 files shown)


docs/web3.main.rst
```diff
@@ -419,17 +419,13 @@ web3.py library.
...
 ~~~~~~~~~~~~~~~~
 
-External modules can be used to introduce custom or third-party APIs to your ``Web3`` instance. Adding external modules
-can occur either at instantiation of the ``Web3`` instance or by making use of the ``attach_modules()`` method.
-
-Unlike the native modules, external modules need not inherit from the ``web3.module.Module`` class. The only requirement
-is that a Module must be a class and, if you'd like to make use of the parent ``Web3`` instance, it must be passed into
-the ``__init__`` function. For example:
+External modules can be used to introduce custom or third-party APIs to your ``Web3`` instance. External modules are simply
+classes whose methods and properties can be made available within the ``Web3`` instance. Optionally, the external module make
```
kclowes, line 423:
```suggestion
classes whose methods and properties can be made available within the ``Web3`` instance. Optionally, the external module may make
```

marcgarreau (approved):

web3/_utils/module.py
```diff
@@ -34,18 +34,19 @@ def attach_modules(
...
+        if type(parent_module).__name__ == 'Web3':
+            w3 = parent_module
+
+        try:
+            # Modules that need access to the ``Web3`` instance may accept the instance as the first
+            # arg in their ``__init__()`` method. This is the case for any module that inherits from
+            # ``web3.module.Module``.
+            # e.g. def __init__(self, w3):
+            setattr(parent_module, module_name, module_class(w3))
+        except TypeError:
```
kclowes, line 47: Just from my own ignorance, when does this `TypeError` get raised?

kclowes (approved):
LGTM!

web3/_utils/module.py
fselmo: This was a bit tough to navigate but this is the current flow:

1. Assume the first argument to a module's `__init__()`, if there is one, is a web3 instance
2. If it is, we pass it in and the module may make use of the instance in case it needs to make internal calls
3. If there is no argument to be accepted in the `__init__()`, it throws a type error and we catch it here and, instead, initialize the module without any arguments. This is a module that makes no use of the web3 instance.

I had some trouble with trying to read the first argument on the class because the class isn't initialized yet at this point. We are passing in a static class and these lines are actually where they get initialized. I think a separate (and maybe better approach?) is to create a utility method that reads the static class, more specifically the `__init__()` of the static class, and sees if there is an argument passed in. If there is, we can initialize it with `module_class(w3)`, if there is not then we initialize it as `module_class()`.

100% open to changing this if you see a better way. Thoughts?

PR #2329, opened 2022-02-02
Feature/async geth admin

dbfreem:
### What was wrong?
Added async functions to GethAdmin
Related to Issue  #1413 

### Todo:

A couple items to note about this PR

1. take note that I moved `datadir` and `peers` above the deprecated comment in `geth.py` (lines 269 and 271). I believe they were put their in error.
2. I removed` setSolc` from the docs since it is not in the code.
3. I did not test add_peers, start/stop ws, and start/stop rpc. This appeared as if it would have side effects on the rest of the integration test. Let me know if there is a safe way to test these.


pacrob:
Thanks for all your help! I'm looking into your testing question and will let you know.

tests/integration/go_ethereum/test_goethereum_http.py
```diff
@@ -129,6 +134,10 @@ def test_admin_start_stop_ws(self, web3: "Web3") -> None:
         super().test_admin_start_stop_ws(web3)
 
 
+class TestGoEthereumAsyncAdminModuleTest(GoEthereumAsyncAdminModuleTest):
+    pass
```
pacrob, line 138: You can copy the tests w/ xfails from the above `TestGoEthereumAdminModuleTest` class into here

web3/_utils/module_testing/go_ethereum_admin_module.py
```diff
@@ -98,3 +99,21 @@ def test_admin_nodeInfo(self, web3: "Web3") -> None:
...
+
+    @pytest.mark.asyncio
+    async def test_async_nodeinfo(self, async_w3: "Web3") -> None:
+        node_info = await async_w3.geth.admin.node_info()  # type: ignore
+        assert "Geth" in node_info["name"]
+
+    @pytest.mark.asyncio
+    async def test_async_nodes(self, async_w3: "Web3") -> None:
+        nodes = await async_w3.geth.admin.peers()  # type: ignore
+        assert isinstance(nodes, List)
```
pacrob, line 119: You can add the async versions of the `peers` and `start_stop` tests here.

pacrob:
We would like those tests added, and with the noted xfails in `test_goethereum_http.py` they shouldn't mess up the rest of the testing. Otherwise looks good!

web3/_utils/module_testing/go_ethereum_admin_module.py
dbfreem: copied the test and modified to async

tests/integration/go_ethereum/test_goethereum_http.py
dbfreem: copied the test and modified for async

Issue #2334, opened 2022-02-03
contract.call default block_identifier is 'latest' instead of web3.eth.default_block


newtectonics:
* Version: 5.27.0
* Python: 3.9
* OS: osx
* `pip freeze` output

```
aiohttp==3.8.1
aiosignal==1.2.0
async-timeout==4.0.2
asynctest==0.13.0
attrs==21.4.0
autopep8==1.6.0
base58==2.1.1
bitarray==1.2.2
cachetools==5.0.0
certifi==2021.10.8
charset-normalizer==2.0.10
click==8.0.3
cytoolz==0.11.2
eth-abi==2.1.1
eth-account==0.5.7
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.3.4
eth-rlp==0.2.1
eth-typing==2.2.2
eth-utils==1.10.0
frozenlist==1.3.0
google-api-core==2.4.0
google-auth==2.6.0
google-cloud==0.34.0
google-cloud-bigquery==2.32.0
google-cloud-core==2.2.2
google-crc32c==1.3.0
google-resumable-media==2.1.0
googleapis-common-protos==1.54.0
grpcio==1.43.0
grpcio-status==1.43.0
hexbytes==0.2.2
idna==3.3
importlib-metadata==4.10.1
ipfshttpclient==0.8.0a2
jsonschema==3.2.0
lru-dict==1.1.7
multiaddr==0.0.9
multidict==5.2.0
netaddr==0.8.0
numpy==1.22.1
packaging==21.3
pandas==1.4.0
parsimonious==0.8.1
proto-plus==1.19.9
protobuf==3.19.3
pyasn1==0.4.8
pyasn1-modules==0.2.8
pycodestyle==2.8.0
pycryptodome==3.12.0
pyparsing==3.0.6
pyrsistent==0.18.1
python-dateutil==2.8.2
pytz==2021.3
PyYAML==6.0
requests==2.27.1
rlp==2.0.1
rsa==4.8
six==1.16.0
toml==0.10.2
toolz==0.11.2
typing-extensions==3.10.0.2
urllib3==1.26.8
varint==1.0.2
web3==5.27.0
websockets==9.1
yarl==1.7.2
zipp==3.7.0
```

### What was wrong?

The call() functions for eth and contract have different default block_numbers. 

This difference is reflected in the documentation here: 
[...]

PR #2335, opened 2022-02-03
contract.call() block_identifier default change

newtectonics:
Changed default block_identifier in contract.call() to None.
Changed parse_block_identifier to use web3.eth.defaultBlock if None is passed in

### What was wrong?

https://github.com/ethereum/web3.py/issues/2334

### How was it fixed?

changed block_identifier default to None in contract.call
changed parse_block_identifier to parse None to web3.eth.default_block

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -899,7 +899,7 @@ class ContractFunction:
 
     def call(
         self, transaction: Optional[TxParams] = None,
-        block_identifier: BlockIdentifier = 'latest',
+        block_identifier: BlockIdentifier = None,
         state_override: Optional[CallOverrideParams] = None,
     ) -> Any:
         """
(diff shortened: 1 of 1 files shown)


kclowes:
Thanks for the PR! This looks like a good change to me, but we'll need to wait until v6 to pull it in since it's breaking like you mentioned in the issue. We're planning on cutting our v6 branch soon, so will add this to be pulled in then. It will need a test before merging. If you have time to add one that would be great, otherwise I should be able to get to it soon:tm:.

kclowes:
Sorry @newtectonics that I wasn't able to keep this PR. I opened one at #2777 that has your initial commit, and I added some tests and things after that. You should still get credit on that initial commit. 

Closing in favor of #2777.

PR #2337, opened 2022-02-08
WIP moved the existing test for txpool content and inspect

dbfreem:
### What was wrong?

In PR #2273 I realized that the existing test for `geth.txpool.content()` and `geth.txpool.inspect()` were always being skipped because they were using the Web3 instance backed by the `EthereumTesterProvider`. 

### How was it fixed?

I copied those test over to the goethereum http integration test. Also, had to make some modifications to them to get them to run. 

One thing I found interesting was that on line 48 and 49 of the `go_ethereum_txpool_module` I had to cast the nonce to a string in order to compare the nonce received from the `get_transaction` to the nonce received from `txpool.content`. I assume this is because an immutable type is needed since the nonce is used as a dictionary key on the `txpool.content` side. 

There wasn't an issue for this so I used the PR# in the newsfragment instead of and issue #.

### Todo:


dbfreem:
Don't merge this yet. Looks like the integration test are timing out now waiting on the miner to stop. It runs locally but not after pushed here. I still need to look into this.

dbfreem:
So I was looking into this some more. It looks like all the test that relate to mining are skipped in the core test. See screenshot below. Also, there are not currently any test around mining in the module_testing.

image

It appears to me based on my second screenshot that the hashrate is non-zero even after geth is showing that it stopped mining. This is causing the integration test to timeout because it is waiting for the web3.eth.mining and the web3.eth.hashrate to evaluate to false to exit the while loop.

Issue #1984, opened 2021-05-07
[feature request] web3.ens.resolve should take a block number


WyseNynja:
* Version: 5.18.0
* Python: 3.9
* OS: osx

### What was wrong?

I am doing some analysis of some of my old transactions and noticed that `web3.ens.resolve` was returning the current address and not the address at the time of the transaction.

### How can it be fixed?

I think adding a kwarg to `web3.ens.resolve` that gets passed to the `call` functions inside would be a simple addition that would be backwards compatible.

kclowes:
Good idea, thanks for the request! The fastest route to get this in would be to open up a PR that we can review.

PR #2338, opened 2022-02-09
allow resolving specific blocks

BlinkyStitt:
### What was wrong?

Related to Issue #1984

### How was it fixed?

Add a "block" kwarg that gets passed through to the call.

### Todo:

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -131,16 +131,16 @@ class ENS:
         middlewares = web3.middleware_onion.middlewares
         return cls(provider, addr=addr, middlewares=middlewares)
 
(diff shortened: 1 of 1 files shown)


WyseNynja:
Looks like the default should be 'latest' instead of 'None'

fselmo (changes requested):
Hey @WyseNynja... we had some changes to ENS recently that were pretty important to get in. Is there a chance this PR can be rebased against the current `main` and updated? Also adding some relevant tests for these changes would be necessary before merging! You can either try to add them here or let us know when this PR is rebased and we can add tests when we get a chance.

Thanks for the PR!

PR #2340, opened 2022-02-09
Add 'Breaking Changes' and 'Deprecation' to our valid newsfragment types

kclowes:
### What was wrong?
Release notes didn't have a breaking changes category. Also added a Deprecation section

### How was it fixed?
Added them! Could take or leave the Deprecation section if anyone feels strongly. There's already a removal category, which is sort of the same thing. Could also remove `removal` since that would count in the wider Breaking Changes category. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2340.feature.rst
+++ b/newsfragments/2340.feature.rst
@@ -1,0 +1,1 @@
+Added 'Breaking Changes' and 'Deprecations' categories to our release notes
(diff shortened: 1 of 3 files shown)


fselmo (approved):
Yeah I think removing `removal` (hah) would probably make sense. I'm not sure of a case where `removal` isn't a breaking change... Thoughts? 🤔 

We can think about that for this PR too if you'd like. This lgtm 👍

Issue #1416, opened 2019-08-09
v6 Breaking Changes


kclowes:
Now that v5 stable is out, it's time to start collecting a list of breaking changes that we might like to see in v6. I'll update this issue as they come up.

Outstanding breaking tasks:
- [x] Python 3.11 Support - #2699 
- [x] Remove Infura auto provider logic from within web3 ~~and make a template for new auto providers to use. See issue https://github.com/ethereum/web3.py/issues/1736~~ (#2706)
- [x] Remove `SolidityError` in favor of the more generic `ContractLogicError`. See issue #1901 
- [x] Remove camelCase function names in favor of snake_case. See #2598
- [x] Remove Parity module (via #2718)
- [x] Remove version module (see #2729)
- [x] Make ipfshttpclient opt-in. (See #2730 which closes #2629)
- [x] Remove `start_rpc` and `stop_rpc` from `geth.admin` module (see #2731)
- [x] Standardize `apply_formatter_to_array` functions. See #1902
- [x] Custom web3py exceptions inherit from Web3Exception. See #1478 
- [x] Enable strict bytes length checking by default. See issue #1419 and #1154.
- [x] Don't allow users to pass bytestrings to contract functions if the bytestring is the wrong length. As it stands, if a bytestring is shorter than the required length, we pad it. If it is longer, we throw an error. See issue #1903.
- [x] Raise specific web3 error rather than `ValueError` when method isn't available on the node: https://github.com/ethereum/web3.py/issues/2448
- [x] Standardize default value for block_identifier in eth.call and ContractFunctions.call. Change made in https://github.com/ethereum/web3.py/pull/2335 already, just needs tests.
- [x] Opt in or out of `dict` to `AttributeDict` conversion via middleware choice. See #1656.
- [x] Rich tuple decoder #1353 
- [x] Clean up Async typing

_shouldn't_ be breaking: 
- [x] Clean up AsyncHTTPProvider instantiation. - See #2736
- [x] Rearrange `web3/eth.py` into separate file. See #2753 
- [x] Remove dependency upper limits - #2648
- [x] Rearrange `contract.py` into separate files. See #2567
- [x] v6 Migration Guide

kclowes:
All custom web3 exceptions should inherit from Web3Exception. See #1478

kclowes:
[...]

PR #2345, opened 2022-02-11
Add black to ethpm

kclowes:
### What was wrong?

Black is finally stable, so pulling it in. One module at a time. 

Related to Issue #1416 

### How was it fixed?
added the Black dependency, let it make any fixes.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/ethpm/_utils/backend.py
+++ b/ethpm/_utils/backend.py
@@ -43,7 +43,7 @@ ALL_URI_BACKENDS = [
 
 @to_tuple
 def get_translatable_backends_for_uri(
-    uri: URI
+    uri: URI,
 ) -> Generator[Type[BaseURIBackend], None, None]:
     # type ignored because of conflict with instantiating BaseURIBackend
     for backend in ALL_URI_BACKENDS:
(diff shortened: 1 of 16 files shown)


Harmouch101:
How about using [pre-commit](https://pre-commit.com/) to automate formatting stuff?

kclowes:
> How about using [pre-commit](https://pre-commit.com/) to automate formatting stuff?

@Harmouch101 That looks like a good option, but I'm going to hold off for now. If we decide running black is too clunky then I may pull it in. Thanks for the suggestion!

pacrob (approved):
lgtm!

fselmo:
I was originally thinking we could set the line-length to `100` since that's been our standard for some time but I don't feel strongly about that. It still produces a lot of diff (even more I think). Maybe we should just let black do its thing?

This change would do that within `pyproject.toml`:

```toml
[tool.black]
line-length = 100

fselmo (approved):
lgtm! Excited for this addition. I just left one comment to think about before we start changing everything to black across the code base. But maybe we should just leave it all default black settings since that's kind of the point 😅 .

pacrob:
>... Maybe we should just let black do its thing?
+1 for default line length 88

kclowes:
Okay, I'll leave line length at 88. Thanks!

kclowes:
Okay, I'll leave line length at 88. Thanks!

Issue #2344, opened 2022-02-11
Deprecate loop parameter in v5 WebsocketProvider


kclowes:
* Version: 5.28+

### What was wrong?

The websockets library dropped the `loop`parameter in v10+, so we did the same in v6. It would probably be nice to show v5 users a `DeprecationWarning`. 

### How can it be fixed?

In the websocket provider, check for the loop parameter getting passed in, and show a `DeprecationWarning` that it will be removed in v5

kclowes:
We don't expose the loop parameter to users, it's only used internally. No need for a deprecation warning here.

PR #2347, opened 2022-02-16
Loop depr

Harmouch101:
### What was wrong?

Related to Issue #2344

### How was it fixed?
Adds deprecation warnings.
### Todo:

#### Cute Animal Picture

:cat2:

### DIFF
--- a/newsfragments/2344.feature.rst
+++ b/newsfragments/2344.feature.rst
@@ -1,0 +1,1 @@
+Adds a depreaction warning for the optional ``loop`` parameter of ``web3.providers.websocket``.
--- a/tests/core/providers/test_websocket_provider.py
+++ b/tests/core/providers/test_websocket_provider.py
@@ -16,6 +16,7 @@ from web3.exceptions import (
 )
 from web3.providers.websocket import (
     WebsocketProvider,
+    PersistentWebSocket,
 )
 
 if sys.version_info >= (3, 8):
@@ -39,7 +40,7 @@ def start_websocket_server(open_port):
             await websocket.send(data)
 
         asyncio.set_event_loop(event_loop)
-        server = websockets.serve(empty_server, '127.0.0.1', open_port)
+        server = websockets.serve(empty_server, "127.0.0.1", open_port)
         event_loop.run_until_complete(server)
         event_loop.run_forever()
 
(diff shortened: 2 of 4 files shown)


Harmouch101:
That's weird!

```python
    async def coro_make_request(self, request_data: bytes) -> RPCResponse:
>       async with self.conn as conn:
E       AttributeError: __aenter_
```

Harmouch101:
Awaiting reviews from the lords @kclowes...

Harmouch101:
Hey, @kclowes @fselmo. I would like to have your opinions regarding the changes introduced by this PR. 
So basically, I have created a class named `DeprecationMetaClass` that searches upon keyword arguments being passed at the moment of instantiation and checks if `loop` exists(assuming it is passed as a keyword arg `loop=something`). And after checking whether or not the `loop` has been passed, call the `__call__` and `__init__` methods of that class to create an instance. 
With that noted, the class(the one intended to deprecate the `loop` param, say `WebsocketProvider`) needs to inherit from this class like: `WebsocketProvider(JSONBaseProvider, metaclass=DeprecationMetaClass)`.  I don't think there are any drawbacks introduced by doing so. In terms of performance, just a linear runtime(O(n), for a loop).
Besides, I have added this class under the `decorators` module(not sure if it is the right place to add.). Let's forget for a moment about the changes introduced by `black.`
Awaiting your reviews on this matter!

Regards.

Harmouch101:
I forgot to mention there is another approach for fixing this issue is by defining a new decorator that wraps the `__init__`,

```python
def deprecated_parameter(param, version):
    def decorator(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            new_kwargs = {key: val for key, val in kwargs.items() if param not in kwargs}
            if param in kwargs:
                warnings.warn(
                    f"The {param} parameter is deprecated and was removed from "
                    f"websocket provider as of web3 {version}. Consider instantiating "
                    "this class without passing this argument instead.",
                    category=DeprecationWarning,
                    stacklevel=2,
                )
            return func(*args, **new_kwargs)
        return wrapper
    return decorator
```

And use it like:

```python
@deprecated_parameter(param="loop", version="v5")
def __init__(...):
```

What do you prefer, the OO style or this approach? For me, I would prefer the OO style of doing things, as it appears that this project is highly involved in OOP(e.g. abstracts classes and such).

kclowes:
Thanks for the PR @Harmouch101! Taking a deeper look at this PR in the context of v5, I don't think we actually expose the `loop` to users, it's solely something we use internally. The `WebsocketProvider` doesn't allow this keyword to be passed in the `websocket_kwargs` param, so I don't think we actually need to show a `DeprecationWarning` here. Sorry about that. I'm going to close this and the corresponding issue I opened.

Harmouch101:
Alright. Regardless, it was a fun exercise to work on. I learned something new. Thanks for the reply.

Issue #2013, opened 2021-06-04
Fix AsyncEth module typing


kclowes:
### What was wrong?

Because the eth module is called `eth` regardless of whether or not it refers to the `Eth` module or the `AsyncEth` module, the typing is a little bit funky. Once `AsyncEth` is completely switched over to mirror the functionality of the `Eth` module, we can make the `eth` attr [here](https://github.com/ethereum/web3.py/blob/0afdd87d45ca7c4f64d146579e84c7810fac95de/web3/main.py#L160) be `Union[AsyncEth, Eth]`. Then we can go through and remove `type: ignore`s throughout codebase on all `await w3.eth...` lines.

fselmo:
closed by #2819

PR #2348, opened 2022-02-17
fix: add missing type

ChristianCoenen:
### What was wrong?
I encountered a typing warning when initializing the `eth` module with an `AsyncEth` instance.

### How was it fixed?
Adding an additional type to the `eth` definition of a Web3 instance. Alternatively, one could also create `async_eth: AsyncEth`, but I think this breaks existing code where `web3.eth.xxx` is used.

Another alternative would be to use `BaseEth`. Please let me know if you prefer this approach.

### Todo:

#### Cute Animal Picture
(https://www.pinterest.com/pin/443252788328237881/)

### DIFF
--- a/newsfragments/2348.misc.rst
+++ b/newsfragments/2348.misc.rst
@@ -1,0 +1,1 @@
+Add `AsyncEth` as valid type for the `eth` module.
--- a/web3/main.py
+++ b/web3/main.py
@@ -72,7 +72,8 @@ from web3._utils.normalizers import (
     abi_ens_resolver,
 )
 from web3.eth import (
-    Eth,
+    Eth, 
+    AsyncEth,
 )
 from web3.geth import (
     Geth,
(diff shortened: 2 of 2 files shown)


ChristianCoenen:
Seeing why the CI fails, I realized that it might actually be necessary to introduce `async_eth: AsyncEth`, but this would require quite some changes, right? What do you think is the best approach to solve this?

kclowes:
Thanks for the PR. We haven't figured out the best way to handle this yet. We'd like to keep the API the same, so that if you use an AsyncProvider, you still can do a `w3.eth.whatever` (rather than a `w3.async_eth.whatever`), but that doesn't jive with `mypy`. We've been ignoring async types for now, which isn't ideal. The issue is being tracked in #2013 . `BaseEth` _might_ be the way to type this, but I think we'll have to wait until all of the methods have been ported over. We're open to solutions though if you have ideas!

ChristianCoenen:
Thank you very much! I will keep an eye on #2013 and participate in the discussion if I have a good idea :-)

PR #2368, opened 2022-02-28
Fix test for the web3_api

kclowes:
### What was wrong?
When I bumped web3 to version 6, our web3_api test started failing on master. 

### How was it fixed?
Updated the test. It would be nice to figure out a way to catch this before the new version is released, but deprioritizing for now.

### Todo:

#### Cute Animal Picture


pacrob (approved):
lgtm!

PR #2369, opened 2022-03-01
bump docker base image & rm parity env var!

Harmouch101:
### What was wrong?

Related to PR #1753 & #2343

### How was it fixed?

As python 3.6 was dropped, it would make sense to bump the base image of the docker container. The same would apply to the parity version.

### Todo:

#### Cute Animal Picture

:dog:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,8 +3,6 @@ services:
   sandbox:
     build:
       context: .
-    environment:
-      PARITY_VERSION: v2.5.13
     volumes:
       - .:/code
     command: tail -f /dev/null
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     build:
       context: .
     environment:
-      PARITY_VERSION: v2.3.5
+      PARITY_VERSION: v2.5.13
```
kclowes, line 7: We can get rid of this since we removed the parity tests.

kclowes (commented):
Thanks @Harmouch101! I left a comment about removing the `PARITY_VERSION` since we no longer have the parity tests, but looks good other than that. Will merge once that's gone.

Harmouch101:
@kclowes , Roger that! Next time, i will just use raw text without silly emojis. But, it's ok, you know that i know that you know what i know...

PR #2376, opened 2022-03-03
Add async chain_id setter

kclowes:
### What was wrong?
The new chain_id flow wasn't available on async eth.

Related to Issue #2207 

### How was it fixed?
Added it!

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2376.bugfix.rst
+++ b/newsfragments/2376.bugfix.rst
@@ -1,0 +1,1 @@
+Add async default_chain_id and chain_id setter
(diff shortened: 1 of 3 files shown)


pacrob (approved):
lgtm!

fselmo (approved):
lgtm!

fselmo:
I wonder if we should have a validation that it can only be set to the `eth_chainId` value. But we can have this be a separate discussion I think. This completes the current flow 👌

kclowes:
I am going to merge this, but like we talked about this morning, I'll check on the original issue to see if there is any reason to allow a different `chain_id` to be set.

PR #2370, opened 2022-03-02
AyncENS/Contract

dbfreem:
### What was wrong?
Build an AsnycENS and implement the needed async parts in Contract

Related to Issue #1990 

### How was it fixed?

Sorry, ahead of time that this is a gigantic PR. As I started trying to AsyncENS I realized there were three main parts and I found it hard to separate them into smaller PRs.

1. Changes made to AsyncEthereumTestProvider. There were changes that were needed to enable testing of Async methods.
2. Changes made to Contract. Added several Async methods and classes. This is not a fully implemented AsyncContract yet but it is enough to get AsyncENS working.
3. Changes made to ENS. Created the AsyncENS and BaseENS classes.

There was a fare amount of copy and paste in this PR. This was due to the fact of duplicate code needed when separating the Async and non Async methods/classes. I was able to separate out some of the logic and the data access in `contract.py`, so this cut down on "some" of the copy and paste in `contract.py`. I didn't see a good way to do that in ENS or changes to the AsyncEthereumTestProvider so there was a fare amount of copy and paste of existing code there.

### Todo:
- [ ] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)
- [ ] Finish adding test
- [ ] Fix linting errores

PR #2377, opened 2022-03-04
init commit of async contract

dbfreem:
init commit of async contract see PR #2370

### DIFF
--- a/tests/core/contracts/conftest.py
+++ b/tests/core/contracts/conftest.py
@@ -5,7 +5,14 @@ import pytest
 from eth_utils import (
     event_signature_to_log_topic,
 )
+from eth_utils.toolz import (
+    identity,
+)
 
+from ens import (
+    AsyncENS,
+)
+from web3 import Web3
 from web3._utils.module_testing.emitter_contract import (
     CONTRACT_EMITTER_ABI,
     CONTRACT_EMITTER_CODE,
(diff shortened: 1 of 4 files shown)


dbfreem:
@pacrob Here is the AsyncEthereumTester and Contract parts of my other PR. I closed that PR. 

One problem we will need to overcome with removing the ENS work is that contract is reliant on [ENS](https://github.com/ethereum/web3.py/blob/asyncify-contract/web3/contract.py#L338) there are other examples in there too. The problem I ran into while testing was that the w3 instance that is being used to create the AsyncContract has a AsyncEth Module in it. This web3 instance also has a standard ENS in it. When ens is called from AsyncContract the AsyncEth module is then used to call web3.eth.call which is async but ens is expecting it not to be async. Hope that all makes sense. 

There may be a way around this like maybe not allowing ens addresses to be passed into AsyncContract until ENS is asyncified. 

Thoughts??

Also, let me know what parts of async contract you want to work on and I can pick some of the left over parts.

Thanks, for setting this feature branch of for collaborating.

dbfreem:
Just FYI, somewhere along the lines of moving the ENS out and fixing linting errors, a few of the test broke, Will look at these this evening as well as finish up the linting

pacrob:
I've been working on trying to resolve the linting but not much success today. Will continue tomorrow.

I think because this PR has the main class restructuring, we should get it in before I start changing anything further.

On the ENS dependency, is it just the calls to normalize_address that require the ENS object? If that's the only place, we could add a non-ens version of the normalizer or something similar.

dbfreem:
So I almost have the Main BaseContract/Contract/AsyncContract done with the exception of a couple methods. I think it is just be the normalizer that was using ens so that's would probably work making a non ens version of it.

dbfreem:
Here is where I am at. I have that one weird lint error where it is complaining about the web3_module but besides that I "think" contract is for the "most part" done. 

The next thing I was going to look at is ContractConstructor. It has some pieces that need to be async.

pacrob:
Looks good. With that structure in, should be easier to work on in chunks.

I pulled it in with the lint failing, but it looks like reverting that ignore type line in web3/_utils/module_testing/web3_module.py fixes it. Is there another reason for removing it?

dbfreem:
So one time I would run lint and it would tell me the typing wasn't right on that line and another time it would say that the type ignore was wrong. So probably just my local env being flaky. If removing the type ignore works then that is fine with me.

Issue #2258, opened 2021-12-15
Support for `eth_createAccessList`


pipermerriam:
### What was wrong?

Unsure if this is standardized but, support for `eth_createAccessList` would be useful

https://geth.ethereum.org/docs/rpc/ns-eth#eth_createaccesslist

### How can it be fixed?

Add support for that JSON-RPC endpoint either in the `w3.geth` namespace or under `w3.eth` if it is standardized (I think it is)

Harmouch101:
Hey there, I think I can work on dis issue. However, after digging a bit deeper, I realized there is a bit of struggle in implementing the formatter for the response. From what I understood from the [eip-2930 example](https://eips.ethereum.org/EIPS/eip-2930), I think it should be something along these lines:

```python
ACCESS_LIST_FORMATTER = apply_list_to_array_formatter({
    'address': to_checksum_address,
    'storageKeys': apply_list_to_array_formatter(to_hexbytes(32))
})

ACCESS_LIST_RESPONCE_FORMATTER = apply_formatters_to_dict({
    "accessList": ACCESS_LIST_FORMATTER,
    "gasUsed": to_integer_if_hex,
})
```

And under the Eth namespace formatters, we need to add the following line:

```python
    RPC.eth_createAccessList: ACCESS_LIST_RESPONCE_FORMATTER,
```

Similarly, for the request, the formatter might look like:

```python
    RPC.eth_createAccessList: apply_formatter_at_index(transaction_param_formatter, 0),
```

Lemme know whether or not I misinterpreted the format of the req/res to continue working on dis issue.

Regards.

kclowes:
closed via #2381

PR #2381, opened 2022-03-06
[WIP] Implement `web3.eth.Eth.createAccessList`

wiseaidev:
Signed-off-by: Harmouch101 <eng.mahmoudharmouch@gmail.com>

### What was wrong?

Related to Issue #2258

### How was it fixed?

Ref: [Lord Vitalik](https://youtu.be/qQpvkxKso2E?t=175).

### Todo:

#### Cute Animal Picture

:dog:


web3/eth.py
```diff
@@ -868,6 +868,11 @@ def sign_munger(
         mungers=[BaseEth.call_munger]
     )
 
+    createAccessList: Method[Callable[..., Union[bytes, bytearray]]] = Method(
```
kclowes, line 871: We've been moving toward snake_casing our method names, so this should be: 
```suggestion
    create_access_list: Method[Callable[..., Union[bytes, bytearray]]] = Method(
```

web3/eth.py
```diff
@@ -868,6 +868,11 @@ def sign_munger(
...
+        RPC.eth_createAccessList,
+        mungers=[BaseEth.call_munger]
```
kclowes, line 873: `createAccessList` doesn't take a `state_override` variable like `eth_call` does, so you'll need to create a new munger instead of using `call_munger`

kclowes (commented):
I saw your comment on the issue, and your approach looks to be about right on a high level. Thanks for adding such detailed docs. I added a few comments on the code. One thing that may help with the formatters is to write the integration test, and see what you get back. Let me know if you need direction on that front. Thanks @Harmouch101!

web3/eth.py
Harmouch101: My bad, I totally forgot about that. I remember that the other day, I read a common standard you are using as a team for writing code, I forgot which repo was it, probably carver or pipermerriam created it. If you know what I mean, please lemme know by sharing the repo link.

web3/eth.py
Harmouch101: I think we can use call_munger because of this if statement:

https://github.com/ethereum/web3.py/blob/81427e69388a4a900ca1260fbb82bef91129367c/web3/eth.py#L311-L312

Harmouch101: But my initial thought was code reuse. That's why I used this method.

web3/eth.py
Harmouch101: Not sure, but probably this one: [pipermerriam/python-style-guide](https://github.com/pipermerriam/python-style-guide).

ytrezq:
https://github.com/wiseaidev/web3.py/commit/afece45397d229946a07c59473da3f74a9099585#r85299196. How to add the up for grabs tag ?

docs/web3.eth.rst
```diff
@@ -987,6 +987,40 @@ The following methods are available on the ``web3.eth`` namespace.
     As of v6.3.0, the raw data is also returned and
     can be accessed via the ``data`` attribute on ``ContractLogicError``.
 
+
+.. py:method:: Eth.create_access_list(transaction, block_identifier=web3.eth.default_block)
+
+    * Delegates to ``eth_createAccessList`` RPC Method
+
+    This method creates an `EIP2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction, except for the sender account and the precompiles. This method uses the same ``transaction`` call object and ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to unstuck contracts that became inaccessible due to gas cost increases.
```
pacrob:
```suggestion
    This method creates an `EIP-2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction, except for the sender account and the precompiles. This method uses the same ``transaction`` call object and ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to unstick contracts that became inaccessible due to gas cost increases.
```

docs/web3.eth.rst
```diff
@@ -987,6 +987,40 @@ The following methods are available on the ``web3.eth`` namespace.
...
+                },
+                {
+                    'address': '0xbb9bc244d798123fde783fcc1c72d3bb8c189413',
+                    'storageKeys': ()
+                },
+            ),
+            "gas": "21000"
+        }
+
+    The method ``eth_createAccessList`` returns list of addresses and storage keys used by the transaction, plus the gas consumed when the access list is added.
```
pacrob:
```suggestion
    The method ``eth_createAccessList`` returns a list of addresses and storage keys used by the transaction, plus the gas consumed when the ``accessList`` is added.
```

docs/web3.eth.rst
```diff
@@ -987,6 +987,40 @@ The following methods are available on the ``web3.eth`` namespace.
...
+
+    That is, it gives you the list of addresses and storage keys that will be used by that transaction, plus the gas consumed if the access list is included. Like ``eth_estimateGas``, this is an estimation; the list could change when the transaction is actually mined. Adding an ``accessList`` to your transaction does not necessary result in lower gas usage compared to a transaction without an access list.
```
pacrob: The first sentence here is just a very slight rewording of the previous line, doesn't seem to add any new information.

We should use "``accessList``" instead of "access list" whenever referring to it.

docs/web3.eth.rst
```diff
@@ -992,7 +992,11 @@ The following methods are available on the ``web3.eth`` namespace.
 
     * Delegates to ``eth_createAccessList`` RPC Method
 
-    This method creates an `EIP2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction, except for the sender account and the precompiles. This method uses the same ``transaction`` call object and ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to unstuck contracts that became inaccessible due to gas cost increases.
+    This method creates an `EIP-2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given
+    ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction,
+    except for the sender account and the precompiles. This method uses the same ``transaction`` call object and
+    ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to unstick contracts that
```
fselmo: Just a nit. I see they use `"unstuck"` in quotes in the EIP, pertaining to one of the past EIPs that gave rise to these issues. Even in the eip it's sort of ambiguous language. Take or leave.

```suggestion
    ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to access contracts that
```

docs/web3.eth.rst
```diff
@@ -1017,9 +1021,10 @@ The following methods are available on the ``web3.eth`` namespace.
             "gas": "21000"
         }
 
-    The method ``eth_createAccessList`` returns list of addresses and storage keys used by the transaction, plus the gas consumed when the access list is added.
-
-    That is, it gives you the list of addresses and storage keys that will be used by that transaction, plus the gas consumed if the access list is included. Like ``eth_estimateGas``, this is an estimation; the list could change when the transaction is actually mined. Adding an ``accessList`` to your transaction does not necessary result in lower gas usage compared to a transaction without an access list.
+    The method ``eth_createAccessList`` returns a list of addresses and storage keys used by the transaction, plus the gas
+    consumed when the ``accessList`` is included. Like ``eth_estimateGas``, this is an estimation; the list could change when
+    the transaction is actually mined. Adding an ``accessList`` to your transaction does not necessary result in lower
```
fselmo: 1. Let's steer away from "mining". Blocks are considered "finalized" when they earn the highest degree of confidence in consensus. Perhaps we adopt that word here? Or "... included in a block."?

2. necessary -> necessarily

```suggestion
    the transaction is finalized. Adding an ``accessList`` to your transaction does not necessarily result in lower
```

web3/types.py
```diff
@@ -368,6 +368,7 @@ class FeeHistory(TypedDict):
 TxReceipt = TypedDict(
     "TxReceipt",
     {
+        "accessList": AccessList,
```
fselmo: I don't see `accessList` coming through on transaction receipts. Is there a reason we added it here?

web3/_utils/method_formatters.py
```diff
@@ -462,6 +462,19 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
 
 signed_tx_formatter = type_aware_apply_formatters_to_dict(SIGNED_TX_FORMATTER)
 
+ACCESS_LIST_FORMATTER = type_aware_apply_formatters_to_dict(
```
fselmo: We should use this same formatter for `TRANSACTION_RESULT_FORMATTERS` in this file for the `accessList` field for consistency. 

One question that I had when I added it there is whether or not we want `to_int_if_hex` on the response for the `storageKeys` and some version of `to_hex_if_int` with zero padding on the request formatter. It may very well be easier to read / write and they are slot locations so they will represent uint256 as far as I know? Thoughts there? I left it alone until now but now we need to make that decision if we add it since it will change the formatting. I'm inclined to keep what you have here since it's an advanced user that is using this anyways but just curious.

web3/_utils/module_testing/eth_module.py
```diff
@@ -2672,6 +2694,37 @@ def test_eth_get_code_with_block_identifier(
...
+                "from": unlocked_account_dual_type,
+                "to": math_contract.address,
+                "data": txn["data"],
+            }
+        )
+
+        assert is_dict(response)
+        access_list = response["accessList"]
+        assert len(access_list) > 0
+        assert access_list[0]["address"] is not None
```
fselmo, line 2724: We should check the `storageKeys` here too and that they are formatted appropriately.

fselmo (commented):
This mostly lgtm, just had some thoughts I wanted to get opinions on and some nits and minor comments.

web3/_utils/method_formatters.py
fselmo: On second thought, I'm not sure anyone will be plugging these values in by hand. This will be a value that is spit back from this request and they can use it in their transaction so I think we should probably keep `to_hexbytes(64)`.

web3/types.py
reedsa: I wasnt sure what type it belonged to. Seemed like the typing in my `eth_module` test was messed up when I tried integrating the `create_access_list` function. `TxData` doesnt have a `gasUsed` param, which seemed appropriate since the `gasUsed` should be in the response of the `create_access_list` function.

Any recommendations on the typings?

web3/_utils/module_testing/eth_module.py
reedsa: Formatting of the address I think is handled by the `is_checksum_address` on the next line. Is there another function I can use to check the formatting here?

fselmo: Well, we check the address but not the storage keys. We should check the storage keys formatting / values. This will help keep a check on the formatters, if they change, etc.

docs/web3.eth.rst
```diff
@@ -987,6 +987,45 @@ The following methods are available on the ``web3.eth`` namespace.
...
+
+    * Delegates to ``eth_createAccessList`` RPC Method
+
+    This method creates an `EIP-2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given
+    ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction,
+    except for the sender account and the precompiles. This method uses the same ``transaction`` call object and
+    ``block_identifier`` object as :meth:`~web3.eth.Eth.call()`. An ``accessList`` can be used to unstick contracts that
+    became inaccessible due to gas cost increases.
+
+    :param transaction: ``TransactionCall`` object.
```
pacrob: Is a `TransactionCall` object defined somewhere? I'm not finding anything in our docs, and geth docs only have broken links - https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-createaccesslist

web3/_utils/module_testing/eth_module.py
reedsa: Ah missed that, thanks for clarifying. 👍

web3/types.py
fselmo: Oh yeah I missed this entirely. We should not return a `TxReceipt` as that is not the return value for `eth_createAccessList`. `TxReceipt` is meant for `getTransactionReceipt`, among some other cases where a defined transaction receipt is returned. 

It looks like the `result` object for `eth_createAccessList` only returns `accessList` and `gasUsed`. Therefore, we should define a `TypedDict` named `CreateAccessListResponse` with only these two fields and probably keep `total=True` (which is default so no need to override it) since it looks like that will be returned every time.

docs/web3.eth.rst
reedsa: I didnt see that, must be something old or they made it up. I assume this is supposed to point to a class, but maybe the type `TxParams` is appropriate?

fselmo: Yeah let's change this up. For `Eth.call` we say: 

"The ``transaction`` parameter is handled in the same manner as the :meth:`~web3.eth.Eth.send_transaction()` method."

Perhaps that's good enough here too since we go into depth there on each param. Thoughts?

tests/core/eth-module/test_transactions.py
```diff
@@ -293,5 +297,8 @@ def test_get_transaction_formatters(w3):
             "data": HexBytes(unformatted_transaction["data"]),
         }
     )
+
+    print(received_tx)
```
fselmo: These two print statements snuck in here.

web3/_utils/method_formatters.py
```diff
@@ -210,13 +228,7 @@ def type_aware_apply_formatters_to_dict_keys_and_values(
...
     "accessList": apply_formatter_if(
         is_not_null,
-        apply_formatter_to_array(
-            type_aware_apply_formatters_to_dict(
-                {
-                    "address": to_checksum_address,
-                }
-            )
-        ),
+        apply_formatter_to_array(ACCESS_LIST_FORMATTER),
```
fselmo, line 231: 👌🏼

web3/eth/eth.py
```diff
@@ -281,6 +281,22 @@ def _durin_call(
 
         raise TooManyRequests("Too many CCIP read redirects")
 
+    # eth_createAccessList
+
+    _create_access_list: Method[
+        Callable[
+            [TxParams, Optional[BlockIdentifier]],
+            TxReceipt,
```
fselmo: We should use an appropriate return type here as per [this comment](https://github.com/ethereum/web3.py/pull/2381/files#r1416119226).

web3/eth/eth.py
```diff
@@ -281,6 +281,22 @@ def _durin_call(
...
+        ]
+    ] = Method(RPC.eth_createAccessList, mungers=[BaseEth.create_access_list_munger])
+
+    def create_access_list(
+        self,
+        transaction: TxParams,
+        block_identifier: Optional[BlockIdentifier] = None,
+    ) -> TxReceipt:
```
fselmo: Same here `TxReceipt` -> `CreateAccessListResponse`

docs/web3.eth.rst
reedsa: Yeah, directing to one place sounds good to me!

docs/web3.eth.rst
```diff
@@ -987,6 +987,46 @@ The following methods are available on the ``web3.eth`` namespace.
     As of v6.3.0, the raw data is also returned and
     can be accessed via the ``data`` attribute on ``ContractLogicError``.
 
+
+.. py:method:: Eth.create_access_list(transaction, block_identifier=web3.eth.default_block)
+
+    * Delegates to ``eth_createAccessList`` RPC Method
+
+    This method creates an `EIP-2930 <https://eips.ethereum.org/EIPS/eip-2930>`_ type ``accessList`` based on a given
+    ``Transaction``. The ``accessList`` contains all storage slots and addresses read and written by the transaction,
```
fselmo: Sorry, another tiny nit we should've caught. This is implying the first param which we named `transaction`, lowercase, in the method definition. It should remain lowercase here.

```suggestion
    transaction. The ``accessList`` contains all storage slots and addresses read and written by the transaction,
```

fselmo (approved):
One last nit but nothing blocking. Lgtm 👍🏼 

We should add the async call to `AsyncEth` as well but let's do that in a separate PR

docs/web3.eth.rst
reedsa: No trouble at all! Fixing this and then I'll merge.

PR #2387, opened 2022-03-10
Feature/asyncify contract

dbfreem:
In the ContractConstructor I deprecated the buildTransacion since I had noticed that naming convention had been deprecated for a more pythonic naming convention in other parts of web3.py. I took a note to add this to the docs later when we do the docs updates

Finished up AsyncContractConstructor
Added a few new test

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -262,7 +262,8 @@ class ENS:
         resolver_addr = self.ens.caller.resolver(normal_name_to_hash(normal_name))
         if is_none_or_zero_address(resolver_addr):
             return None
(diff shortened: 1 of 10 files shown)


dbfreem:
I plan to work on the `web3.eth.contract` async methods next

dbfreem:
Finished up `Web3.eth.contract` and some other small work in contract. My next plan was to look at `prepare_transaction` and the methods that use it at the bottom of the `contract.py`

pacrob:
Thanks! I'll be going through tests/core/contracts tests, figuring out what should be duplicated for async

PR #2395, opened 2022-03-18
Implement ENS set_text and get_text

kclowes:
### What was wrong?

See PR #2286. @smk762 implemented get_text, and to test I needed to implement set_text.

### How was it fixed?

I didn't have permissions to push directly to @smk762 's branch, so I cherry-picked them to this one and added `set_text`. There wasn't really a clean way to split the two PRs. 

I had to add a new resolver ABI, which I took from the current ENS public resolver. 

### Todo:

#### Cute Animal Picture

See #2286 :)

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -282,6 +282,21 @@ class ENS:
         node = raw_name_to_hash(name)
         return self.ens.caller.owner(node)
 
+    def get_text(self, name: str, key: str) -> str:
+        """
+        Get the value of a text record by key from an ENS name.
+
+        :param str name: ENS name to look up
+        :param str key: ENS name's text record key
+        :return:  ENS name's text record value
+        :rtype: str
+        """
+        node = normal_name_to_hash(name)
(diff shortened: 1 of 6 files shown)


ens/main.py
```diff
@@ -282,6 +282,54 @@ def owner(self, name: str) -> ChecksumAddress:
...
+            :meth:`~web3.eth.Eth.send_transaction`
+        :return: Transaction hash
+        :rtype: HexBytes
+        :raises UnownedName: if no one owns `name`
+        """
+        owner = self.owner(name)
+        node = raw_name_to_hash(name)
+
+        if transact == {}:
+            transact['from'] = owner
```
kclowes, line 325: Note to self: I think I actually want to do a merge here, so that if a user passes in a transaction dict without a `from` key, it will also default to `owner`

fselmo (approved):
LGTM pending docs addition as we discussed! 👍

PR #2394, opened 2022-03-18
Feature/asyncify contract

dbfreem:
More Async, I think the only thing left before test and docs is build_transaction_for_function.

I plan to work on build_transaction_for_function tomorrow.

PR #2404, opened 2022-03-24
Feature/asyncify contract

dbfreem:
More AsyncContract

Set default contract factory to AsyncContract in AsyncEth

Unfortunately, this required copy and pasting those three contract functions and the property from the BaseEth class to the Eth and AsyncEth classes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
-FROM python:3.6
+FROM python:3.9
 
 # Set up code directory
-RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 # Install Linux dependencies
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,8 +3,6 @@ services:
   sandbox:
     build:
       context: .
-    environment:
-      PARITY_VERSION: v2.3.5
     volumes:
       - .:/code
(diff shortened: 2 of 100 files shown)


dbfreem:
@ChristianCoenen this is a change that should fix what you reported in #2394. Let me know what you think.

@pacrob Setting the defaultContractFactory to AsyncContract in AsyncEth

ChristianCoenen:
Thanks, lgtm! 

> Unfortunately, this required copy and pasting those three contract functions and the property from the BaseEth class to the Eth and AsyncEth classes

Was there a reason for this like failing tests or some edge cases? Wouldn't just adding 
```python
defaultContractFactory: Type[Union[AsyncContract, AsyncContractCaller]] = AsyncContract
``` 
to `AsyncEth` be sufficient? That's what I did temporarily to make my code work.

dbfreem:
@ChristianCoenen I don't know what happened but when I made the change you suggested it broke some test. To be honest I didn't look into it and just moved the methods down to AsyncEth and Eth. It was probably something simple I did. 

I read your message above and thought "yea that should have worked" and it did when I made the changes you suggested again this evening. So I am opening another PR to revert some of those changes I made last night and implement your suggestion.

Thanks so much for looking into this!

Issue #2406, opened 2022-03-25
Support for new ENS features


Arachnid:
ENS has recently added new functionality in the form of two major upgrades:

- Wildcard resolution ([ENSIP 10](https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution)) makes it possible to define wildcard subdomains, so that one resolver can provide records for all subdomains of a name.
- CCIP Read ([EIP 3668](https://eips.ethereum.org/EIPS/eip-3668)) makes it possible for names to be resolved using offchain data.

The combination of these two features mean it's now possible to have ENS names that are stored on L2s or on other offchain data stores such as private databases.

We'd love to see web3.py support this new functionality (and note, EIP 3668 applies to contract interactions more generally) - please let us know if there's anything we can do to facilitate getting it added!

fselmo:
Hey @Arachnid, thanks for the heads up. I started looking into this Friday. Are you aware of any contracts that can be used for testing this functionality?

Arachnid:
offchainresolver.eth and hatch.eth are both set up with both features on Ropsten. @ricmoo and @makoto can provide advice on how to use them to test; we're working on official test vectors now.

makoto:
Hi @fselmo 

Here is the test scenario I can think of.

```
Scenario: When offchainexample.eth has offchain resolver set 
Then:
- offchainexample.eth returns resolver address
- offchainexample.eth has ETH record 
- offchainexample.eth has email record
- 1.offchainexample.eth returns resolver
- 1.offchainexample.eth has ETH record
- 1.offchainexample.eth has email record

Scenario: When hatch.eth has wildcard resolver set (but no offchain resolver). ROPSTEN only
Then:
- hatch.eth returns resolver address
- hatch.eth has ETH record 
- makoto.hatch.eth returns resolver
- makoto.hatch.eth has ETH record

Scenario: When resolver.eth have non offchain/ non wildcard resolver set
Then:
- resolver.eth returns a resolver address
- resolver.eth has ETH record
- 1.resolver.eth dose not returns a resolver
- 1.resolver.eth does not have ETH record
```

NOTES: 
- For the current integration with ethers.js you can check with https://offchainexample2.surge.sh
- Test should work on both mainnet and testnet (ropsten) except hatch.eth
- If your lib works on browser, please also test via Metamask and Wallet connect (WalletConnect is not working on ethers. 5.6.2 yet)

makoto:
[...]

PR #2411, opened 2022-04-01
Add support for ENSIP-10

fselmo:
### What was wrong?

- [ENSIP-10](https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution) was introduced and we should implement support for this.

Related to issue #2406 

### How was it fixed?

- Added extended resolver ABI and support for wildcard resolution
- Added related tests including creating some simple resolver contracts for the purpose of testing some of the basic functionality. Hopefully we can test against more robust test vectors soon. Also manually tested against `hatch.eth` ens domain on Ropsten as recommended in issue #2406.
- ``ens.utils.ens_encode_name()`` and ``ENS.parent()`` methods added to provide support for this implementation

### Todo:
- [x] Add tests for resolving contracts that support the ``extendedResolver`` interface and functionality
- [x] Add cute animal picture

#### Cute Animal Picture


fselmo:
This is ready for review / discussion. All that's missing from what my To Do list was to add testing around the `decodeABI()` method and I'll try to get that in here in the next few days or so.

Of importance here is, because this is such a new functionality, there aren't official test vectors for wildcard resolution. I did manually test on Ropsten as was recommended but for the purposes of our test suite I just added a `SimpleExtendedResolver.sol` that makes sure to test the parent ens domain for `extended-resolver.eth` and its subdomains with different logic appropriate for each case and "resolves" different addresses in each case. It's not an ideal test but it's _something_ to have in our test suite for now until hopefully we have some official test vectors to test against.

makoto:
Hi. I am not so familiar with python, but I tried `python3 -m pip install .` to install it locally with the branch but seems not working.

```
~/.../ethereum/web3.py (ensip10-issue-2406)$python3
Python 3.9.5 (default, May  4 2021, 03:36:27) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from web3.auto.infura.ropsten import w3
>>> w3.isConnected()
True
>>> w3.ens.address('matoken.eth')
'0xfFD1Ac3e8818AdCbe5C597ea076E8D3210B45df5'
>>> w3.ens.address('1.offchainexample.eth')
  File "/Users/makoto/work/ethereum/web3.py/ens/main.py", line 155, in address
    return cast(ChecksumAddress, self.resolve(name, 'addr'))
  File "/Users/makoto/work/ethereum/web3.py/ens/main.py", line 306, in resolve
    contract_call_result = extended_resolver.caller.resolve(
  File "/Users/makoto/work/ethereum/web3.py/web3/contract.py", line 1453, in call_function
    return fn(*args, **kwargs).call(transaction, block_identifier)
  File "/Users/makoto/work/ethereum/web3.py/web3/contract.py", line 957, in call
    return call_contract_function(
  File "/Users/makoto/work/ethereum/web3.py/web3/contract.py", line 1501, in call_contract_function
    return_data = w3.eth.call(
  File "/Users/makoto/work/ethereum/web3.py/web3/module.py", line 58, in caller
    result = w3.manager.request_blocking(method_str,
  File "/Users/makoto/work/ethereum/web3.py/web3/manager.py", line 198, in request_blocking
    return self.formatted_response(response,
  File "/Users/makoto/work/ethereum/web3.py/web3/manager.py", line 170, in formatted_response
    apply_error_formatters(error_formatters, response)
  File "/Users/makoto/work/ethereum/web3.py/web3/manager.py", line 70, in apply_error_formatters
    formatted_resp = pipe(response, error_formatters)
  File "cytoolz/functoolz.pyx", line 667, in cytoolz.functoolz.pipe
  File "cytoolz/functoolz.pyx", line 642, in cytoolz.functoolz.c_pipe
  File "/Users/makoto/work/ethereum/web3.py/web3/_utils/method_formatters.py", line 574, in raise_solidity_error_on_revert
    raise ContractLogicError(response['error']['message'])
web3.exceptions.ContractLogicError: execution reverted
```

Let me know if I am doing it wrong.

makoto:
FYI,

`makoto.hatch.eth` is working so looks like wildcard is working but not offchain resolver.

```
>>> w3.ens.address('makoto.hatch.eth')
'0x0Bc0035cE036984749Fd22978394734B89690b5e'
```

ens/utils.py
```diff
@@ -102,11 +113,41 @@ def normalize_name(name: str) -> str:
...
+        return b'\x00'
+
+    normalized_name = normalize_name(name)
+
+    labels = normalized_name.split('.')
+    labels_as_bytes = [to_bytes(text=label) for label in labels]
+
+    # raises if len(label) > 63:
+    for index, label in enumerate(labels):
+        if len(label) > 63:
```
Arachnid, line 141: DNS has a limit of 63 octets per label, but the limit in ENS is 255.

ens/utils.py
```diff
@@ -102,11 +113,41 @@ def normalize_name(name: str) -> str:
...
+    """
+    Encode a name according to DNS standards specified in section 3.1 of RFC1035 with the following
+    exceptions:
+
+        - There is no limit on the total length of the encoded name while still keeping the length
+        limit on labels as the DNS-standard of 63.
+
+        - Return a single 0-octet, b'\x00', if empty name.
+    """
+    if is_empty_name(name):
```
Arachnid, line 131: Won't the function work the same without this?

Arachnid (approved):

ens/utils.py
fselmo: From what I saw in ENSIP-10, the only deviations from the DNS encoding is that the zero octet is returned for the empty string and that "there is no limit on the total length of the encoded name". Does the 63 size limit not matter here then? Should it be 255 for labels and no limit on total length?

I just checked out the [ethers.js](https://github.com/ethers-io/ethers.js/blob/33a029e457320a226c68a01f4cfa2d110125a8b8/packages/hash/src.ts/namehash.ts#L49-L57) implementation and it looks like they may be using a limit of 63 as well.

ens/utils.py
Arachnid: Good point, `''.split('.') = ['']`, not `[]`.

ens/utils.py
Arachnid: DNS encoding also specifies the zero octet for the empty string. The 63 byte limit is a limitation of the DNS protocol, not the encoding mechanism, and we don't enforce the same limit. 255 should be the limit for labels, with no limit on total length.

Arachnid (approved):

ens/utils.py
fselmo: Yep, makes sense. I'll push an update. Thanks again 👌

fselmo:
> FYI,
> 
> Tested makoto.hatch.eth and seems working

@makoto Thanks for testing. Yeah... it was easy to test manually with the examples that were given but I wanted to add some in-house tests that were a bit more difficult for us without official test vectors. I ended up just writing a simple resolver for now. Not ideal but it ends up testing that the functionality works at least.

I appreciated the test examples though. That definitely helped manually test everything 👍

kclowes (approved):
LGTM! Thanks for taking this on! I added some comments but nothing major. Some things we should check before we merge if you haven't already:
- did you check to make sure the doc generation works? Do you think it's worth adding docs to the ENS overview too?
- I can't remember what we decided about backporting this to v5? If we do want to backport, the `fn_name` needs to be changed back to `get` for the v5 PR, and the error classes should remain the same.

ens/main.py
```diff
@@ -252,24 +285,38 @@ def setup_name(
                 self.setup_address(name, address, transact=transact)
             return self._setup_reverse(name, address, transact=transact)
 
-    def resolve(self, name: str, get: str = 'addr') -> Optional[Union[ChecksumAddress, str]]:
+    def resolver(self, normal_name: str) -> Optional['Contract']:
+        return self._get_resolver(normal_name)[0]
+
+    def resolve(self, name: str, fn_name: str = 'addr') -> Optional[Union[ChecksumAddress, str]]:
```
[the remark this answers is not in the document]
fselmo: Yeah... changing the signature should def be `v6` only. I think it makes more sense for anyone reading the code coming in. `v5` would have to remain the same so as to not be a breaking change.

ens/main.py
```diff
@@ -335,16 +395,24 @@ def set_text(
...
+                return r.functions.setText(node, key, value).transact(transaction_dict)
+            else:
+                raise UnsupportedFunction(
+                    f"Resolver for name `{name}` does not support `text` function"
+                )
         else:
-            raise UnownedName("claim domain using setup_address() first")
+            raise ResolverNotFound(
+                f"No resolver found for name `{name}`. It is likely the name contains an "
+                "unsupported top level domain (tld)."
```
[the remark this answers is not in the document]
fselmo: I think that any domain with an `eth` tld should at least return the resolver that is set on `eth` unless the subdomain has been claimed and a resolver is set there, then it would return that resolver.

So I think this should always be a bad `tld` unless for some reason `eth` parent doesn't have a resolver set which shouldn't happen. Unless I'm missing an edge case somewhere?

ens/main.py
```diff
@@ -449,21 +517,61 @@ def _set_resolver(
                 namehash,
                 resolver_addr
             ).transact(transact)
-        return self._resolverContract(address=resolver_addr)
+        return self._resolver_contract(address=resolver_addr)
+
+    def _get_resolver(
+        self,
+        normal_name: str,
+        func: str = 'addr'
```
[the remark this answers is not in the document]
fselmo: yes!

tests/ens/test_setup_address.py
```diff
@@ -23,53 +26,45 @@
...
+def test_set_address(ens, name, namehash_hex, TEST_ADDRESS):
     assert ens.address(name) is None
     owner = ens.owner('tester.eth')
 
     ens.setup_address(name, TEST_ADDRESS)
     assert is_same_address(ens.address(name), TEST_ADDRESS)
 
-    namehash = Web3.toBytes(hexstr=namehash_hex)
-    normal_name = ens.nameprep(full_name)
-    assert is_same_address(ens.address(name), TEST_ADDRESS)
```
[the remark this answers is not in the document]
fselmo: Line 71 is the exact same as line 67. And I removed `full_name` entirely because in every parametrized case it was the exact same as `name`.

tests/ens/test_get_text.py
```diff
@@ -57,10 +39,39 @@ def test_set_text_pass_in_transaction_dict(ens):
...
+    ('description', 'a test'),
+    ('notice', 'this contract is a test contract'),
+),)
+def test_get_text(ens, key, expected):
+    address = ens.w3.eth.accounts[2]
+    ens.setup_address('tester.eth', address)
+    owner = ens.owner('tester.eth')
+    assert address == owner
+    ens.set_text('tester.eth', key, expected)
+    assert ens.get_text('tester.eth', key) == expected
```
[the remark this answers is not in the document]
fselmo: Oops... no I don't think I meant to remove it, nice catch. But we do set it every time and then get it so it overwrites with every test case. I think we should reset it though. Thanks for catching that!

fselmo:
> did you check to make sure the doc generation works? Do you think it's worth adding docs to the ENS overview too?

I think the ENS module could use a big improvement in the docs in general... was thinking about updating the docs in a separate PR but I could tack it on here. Thoughts? I did not check the doc generation... I'll check that out too.

kclowes (approved):
This looks good to me! Nice work!

PR #2416, opened 2022-04-08
Add optional strict mode for ENS reverse resolution

Slokh:
### What was wrong?

We started using the ens library of web3.py at OpenSea and noticed an issue when doing reverse resolutions. In order to accurately determine the reverse resolution, a forward resolution must be checked agains as stated in the ENS docs:
```
ENS does not enforce the accuracy of reverse records - for instance, anyone may claim that the name for their address is 'alice.eth'. To be certain that the claim is accurate, you must always perform a forward resolution for the returned name and check it matches the original address.
```

I don't know what the ideal solution here is. For now, in this PR I have made this an optional lookup using the `strict` param. I'm assuming everyone would want the correct resolution so maybe we should make it the default behavior? Open to discussion.

### How was it fixed?

Added an optional forward resolution check to validate the lookup is correct.

#### Cute Animal Picture

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -142,7 +142,7 @@ class ENS:
         :raises InvalidName: if `name` has invalid syntax
         """
         return cast(ChecksumAddress, self.resolve(name, 'addr'))
-        
+
     def name(
         self, address: ChecksumAddress, strict: Optional[bool] = False
     ) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


fselmo:
Hey @Slokh! Thanks for the PR. That's a nice catch. We're actually in the middle of some ENS support right now so that was great timing too. I would say we should probably PR this against the `v5` branch so that it's available without any other breaking changes as the `master` (v6) branch will be in beta for a bit until we can verify it's in a stable state.

For `v6` we will probably want to drop the flag altogether and make this check a part of the method itself. But I can update that soon unless you want to create another PR against `master` for that change as well.

Also, ignore the flaky timeouts on the test suite... I can re-run from failed when it's ready to go, which it looks like it is?

So, if you can, please change the merge branch to the `v5` branch and I'll take a look. Thanks!

Slokh:
@fselmo Thanks for the quick response! Will open this up against v5

Slokh:
Update: PR for v5 https://github.com/ethereum/web3.py/pull/2417

PR #2417, opened 2022-04-08
Add optional strict mode for ENS reverse resolution

Slokh:
### What was wrong?

We started using the ens library of web3.py at OpenSea and noticed an issue when doing reverse resolutions. In order to accurately determine the reverse resolution, a forward resolution must be checked agains as stated in the ENS docs:
```
ENS does not enforce the accuracy of reverse records - for instance, anyone may claim that the name for their address is 'alice.eth'. To be certain that the claim is accurate, you must always perform a forward resolution for the returned name and check it matches the original address.
```

I don't know what the ideal solution here is. For now, in this PR I have made this an optional lookup using the `strict` param. I'm assuming everyone would want the correct resolution so maybe we should make it the default behavior? Open to discussion.

### How was it fixed?

Added an optional forward resolution check to validate the lookup is correct.

#### Cute Animal Picture

### DIFF
--- a/docs/ens_overview.rst
+++ b/docs/ens_overview.rst
@@ -162,3 +162,37 @@ Wait for the transaction to be mined, then:
 ::
 
     assert ns.name('0x5B2063246F2191f18F2675ceDB8b28102e957458') == 'jasoncarver.eth'
+
+Set Text Metadata for an ENS Record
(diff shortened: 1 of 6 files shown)


fselmo:
Tried to use `gh` for this... haven't had issues with it but ran into some funk. I created yet another PR for this and will merge as soon as tests pass there. Thanks again 🙂 👍

PR #2418, opened 2022-04-08
Add flaky decorator to geth estimate_gas

kclowes:
### What was wrong?
Geth integration tests started failing an annoying amount again.

Related to Issue #

### How was it fixed?
Added a `flaky` decorator with a default of 3 runs. If that doesn't fix it, I think the next step would be do something like: 

`@pytest.mark.xfail(strict=False)`

I ran the suite a bunch of times and didn't see any failures, so we'll see how it goes. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2418.misc.rst
+++ b/newsfragments/2418.misc.rst
@@ -1,0 +1,1 @@
+Add flaky decorator to Geth gas estimation tests.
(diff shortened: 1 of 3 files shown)


pacrob (approved):
lgtm!

fselmo (approved):
Lgtm! Thanks for this!

PR #2419, opened 2022-04-08
Ens name validation

fselmo:
This is what #2417 should have been but got funky results trying to update branch via `gh`. Reference that PR for all info.


Slokh:
Thanks for following up @fselmo

PR #2420, opened 2022-04-08
Validate forward resolution for reverse resolution by default

fselmo:
### What was wrong?

As outlined in the warning in the ENS docs, for the reverse resolution of a name for an address to be accurate, the forward resolution address must also match. Source at the time of this commit is here: https://docs.ens.domains/dapp-developer-guide/resolving-names#reverse-resolution.

Related to Issue #2417, #2419 (`v5` implementation of this same concept. Use this as default logic for `v6`)

### Todo:
- [x] Add cute animal picture

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -152,7 +152,11 @@ class ENS:
         :type address: hex-string
         """
         reversed_domain = address_to_reverse_domain(address)
-        return self.resolve(reversed_domain, get='name')
+        name = self.resolve(reversed_domain, get='name')
+
+        # To be absolutely certain of the name, via reverse resolution, the address must match in
+        # the forward resolution
+        return name if to_checksum_address(address) == self.address(name) else None
 
     def setup_address(
         self,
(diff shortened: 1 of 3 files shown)


kclowes (approved):
LGTM! 🚀

Slokh:
@fselmo When can we get this released? We're currently using 5.24.0 so I'll need to upgrade

fselmo:
> @fselmo When can we get this released? We're currently using 5.24.0 so I'll need to upgrade

Hey @Slokh, thanks for the patience. We're getting one more PR approved that's important for us pre-release but it should be just around the corner. The `v5` release will be coming out sooner than `v6` / `master` but that's also what I'd recommend since `v6` isn't technically "stable" yet. I will try to update once a new release is out.

Slokh:
Thanks for the update @fselmo

kclowes:
@Slokh this feature (with the flag) just got released in web3 v5.29.0!

Slokh:
Thanks both of you!

PR #2425, opened 2022-04-13
Cache eth.chain_id in simple_cache_middleware

kclowes:
### What was wrong?
It was not ideal to have to set `_default_chain_id=None` to make the `chain_id` request again. Added the sync `chain_id` to the `simple_cache_middleware` instead. Our benchmarking suite shows times of ~.003s after `chain_id` is cached, which is down from ~.017s.

### How was it fixed?

Added `eth_chainId` to the `simple_cache_middleware`. Had to take out the async `_default_chain_id` flow for consistency between the two modules, and will add caching for async `chain_id` after the `simple_cache_middleware` gets asyncified.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -216,6 +216,18 @@ The following properties are available on the ``web3.eth`` namespace.
        >>> web3.eth.chain_id
        61
 
+   .. note::
+
+      This property gets called frequently in validation middleware,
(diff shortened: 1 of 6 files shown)


fselmo (approved):
lgtm! I think it just needs a `2425` newsfragment but this is great. Good to know on the performance improvement too 👌

Issue #2378, opened 2022-03-04
ipfshttpclient version is incompatible with python version 3.7


unparalleled-js:
* Version:5.27.0
* python 3.7 and 3.7.1

### What was wrong?

In the `setup.py` file, there is `python_requires=">=3.7,<3.11",` but there is also `ipfshttpclient==0.8.0a2"`.
But, when looking at `ipfshttpclient`, it does not support python 3.7 and 3.7.1: https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/master/pyproject.toml#L15-L23

Thus, web3.py 5.27.0 does not support python 3.7 and 3.7.1

This is causing downstream dependencies such as https://github.com/software-mansion/starknet.py to be stuck on older versions of web3.py.

### How can it be fixed?

* Officially stop supporting python 3.7 and 3.7.1 e.g. `python_requires=">=3.7.2,<3.11",`

---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
please refer to our [documentation](https://web3py.readthedocs.io/en/latest/) and/or join the conversation 
on [discord](https://discord.gg/GHryRvPB84).

unparalleled-js:
My experience is that when I install or use a python version, it just about always is the latest minor in that range, like `3.7.12` or something, so no no one probably experiences this bug... But, tools like `poetry` are pickier when it comes to validating dependencies and creating lock files and all that, which is part of the issue I am seeing.

Perhaps there is an alternative solution that I am not thinking of either!

kclowes:
Ah, interesting. Thanks for pointing this out! You're seeing this when installing with poetry? Without thinking about it too hard, I think this should be fine to update in v5, especially since we're also using `typing.NoReturn` (which as you pointed out is why `ipfshttpclient` doesn't support `3.7` and `3.7.1`).

unparalleled-js:
> You're seeing this when installing with poetry?

A simple way to reproduce is to create a new project with just a `pyproject.toml` in it with the following contents:

```yaml
[tool.poetry]
name = "my_web3_projet"
authors = ["jules"]
version = "0.1.0"
description = "hi!"

[tool.poetry.dependencies]
python = ">=3.7,<3.10"
web3 = "5.27.0"
```

Notice we are selecting the same python versions that `web.py` has in its `setup.py` file.
When you run `poetry lock` on this project, you get the following error:

```
[...]

PR #2428, opened 2022-04-15
chaged to 3.7+ support in readme

dbfreem:
updated the readme to show that web3.py only supports 3.7+ now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A Python library for interacting with Ethereum, inspired by [web3.js](https://github.com/ethereum/web3.js).
 
-* Python 3.6+ support
+* Python 3.7+ support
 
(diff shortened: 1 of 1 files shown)


dbfreem:
@kclowes I can definitely change this. Should I alsI set the setup.py python version to be  `>=3.7.2, <3.11` like #2378 suggest or just leave it with the readme update.

kclowes:
Yeah, let's update in `setup.py` too. Thank`s!

dbfreem:
That py37 failure looks like it was because of rate limiting and not due to this change.

dbfreem:
this one is good to go

kclowes (approved):
Thank you!

PR #2435, opened 2022-04-20
Remove ignored mypy comments

BoboTiG:
### What was wrong?

On my machine, mypy never pass.

### How was it fixed?

I updated the version, and fixed issues.
Let's see if it is enough.

### Todo:

#### Cute Animal Picture


BoboTiG:
OK, Python 3.9 does not like the PR. Closing it.

PR #2450, opened 2022-04-29
fix: add chainId to txn params

unparalleled-js:
### What was wrong?

I noticed even when caching `chain_id`, the request occurs multiple times. I figured out this happens because I am calling `estimate_gas()` and passing it a transaction and the RPC method's defined txn schema does not inclue `chainId` so it makes the request againt to add the chainId at some point (this is only deduced).. By adding `chainId` to the params, I can avoid an extra RPC call to get the chain ID`

### How was it fixed?

Add `'chainId': 'uint',` in `TRANSACTION_PARAMS_ABIS`.

### Todo:

#### Cute Animal Picture
baby_bats

### DIFF
--- a/web3/_utils/rpc_abi.py
+++ b/web3/_utils/rpc_abi.py
@@ -163,6 +163,7 @@ TRANSACTION_PARAMS_ABIS = {
     'nonce': 'uint',
     'to': 'address',
     'value': 'uint',
+    'chainId': 'uint',
 }
 
 FILTER_PARAMS_ABIS = {
--- a/newsfragments/2450.misc.rst
+++ b/newsfragments/2450.misc.rst
@@ -1,0 +1,1 @@
+Add ``chainId`` to tx params to ``TRANSACTION_PARAMS_ABI``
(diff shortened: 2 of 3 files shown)


web3/_utils/rpc_abi.py
```diff
@@ -163,6 +163,7 @@ class RPC:
     'nonce': 'uint',
     'to': 'address',
     'value': 'uint',
+    'chainId': 'uint',
```
unparalleled-js, line 166: Is there a way to make this optional?

kclowes:
Nice catch! I don't think there is a way to indicate that it's optional, but I don't think we do any checks on it. Lmk if you want help debugging the tests. It looks like they're failing because a hex-to-int conversion isn't happening for the error message in the validation middleware.

unparalleled-js:
> It looks like they're failing because a hex-to-int conversion isn't happening for the error message in the validation middleware.

I believe I got it!

kclowes (approved):
LGTM! Thanks!

Issue #2446, opened 2022-04-26
Async concurrency regression with 5.29.0


DefiDebauchery:
* Version: 5.29.0
* Python: 3.9.7
* OS: macos

### What was wrong?

Async code with the AsyncHTTPProvider hangs in 5.29.0; previously worked fine in 5.28.0. 

(Shoddy) code sample:
```py
import asyncio
from web3 import Web3, AsyncHTTPProvider
from web3.eth import AsyncEth
from web3.net import AsyncNet
from web3.middleware import async_geth_poa_middleware

networks = {
    'ARBITRUM' : 'https://arb1.arbitrum.io/rpc',
    'AURORA'   : 'https://mainnet.aurora.dev',
    'AVAX'     : 'https://rpc.ankr.com/avalanche/',
    'BSC'      : 'https://bscrpc.com',
    'CELO'     : 'https://rpc.ankr.com/celo/',
    'CRONOS'   : 'https://evm-cronos.crypto.org',
    'FANTOM'   : 'https://rpc.ankr.com/fantom/',
    'FUSE'     : 'https://rpc.fuse.io/',
    'HARMONY'  : 'https://rpc.ankr.com/harmony/',
    'HECO'     : 'https://http-mainnet.hecochain.com',
    'METIS'    : 'https://andromeda.metis.io/?owner=1088',
    'MOONBEAM' : 'https://rpc.api.moonbeam.network',
    'MOONRIVER': 'https://rpc.api.moonriver.moonbeam.network/',
    'POLYGON'  : 'https://polygon-rpc.com/'
}

async def load_networks():
    async def connect_rpc(network, uri):
        rpc = Web3(
            AsyncHTTPProvider(
                uri,
                request_kwargs={'timeout': 10},
            ),
            modules={'eth': AsyncEth, 'net': AsyncNet},
            middlewares=[]
        ) 
        rpc.middleware_onion.inject(async_geth_poa_middleware, layer=0)

        try:
            chain_id = await rpc.eth.chain_id
            print(network, chain_id)
        except Exception as e:
            print(network, e)

    tasks = [connect_rpc(chain, info) for chain, info in networks.items()]
    await asyncio.gather(*tasks)

if __name__ == "__main__":
    asyncio.run(load_networks())
```

With 5.28.0, the output is as expected.
```
BSC 56
AURORA 1313161554
POLYGON 137
HECO 128
MOONBEAM 1284
MOONRIVER 1285
ARBITRUM 42161
METIS 1088
[...]

PR #2451, opened 2022-04-29
[v5] Increase session cache size

kclowes:
### What was wrong?
With the new SessionCache, users were limited to 8 items in the cache before it hung.

Related to Issue #2446

### How was it fixed?
Bumped the cache size. This is a very quick fix solution. With this solution, if there are more than 20 items in the cache it will still hang. I'm going to leave the initial issue open to find an actual solution.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2451.bugfix.rst
+++ b/newsfragments/2451.bugfix.rst
@@ -1,0 +1,1 @@
+Increase SessionCache size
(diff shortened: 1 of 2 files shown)


pacrob (approved):
lgtm!

PR #2452, opened 2022-04-29
Add v6 warnings to ENS changes from pr #2411

fselmo:
### What was wrong?

Method signature changes and method deprecation for some ENS methods in `v6`.

Related to Issue #2411 

### How was it fixed?

Add warnings for method deprecation and method signature changes for ENS methods in `v6`.

### Todo:

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->/wallaby-GettyImages-590486136-58e667df5f9b58ef7ec76598.jpg)

### DIFF
--- a/ens/main.py
+++ b/ens/main.py
@@ -12,6 +12,7 @@ from typing import (
     Union,
     cast,
 )
+import warnings
 
 from eth_typing import (
     Address,
@@ -259,6 +260,12 @@ class ENS:
             return self._setup_reverse(name, address, transact=transact)
 
     def resolve(self, name: str, get: str = 'addr') -> Optional[Union[ChecksumAddress, str]]:
+        warnings.warn(
+            "In v6, the resolve() method will be made internal and renamed to _resolve(). The "
+            "'get' param name will also change to 'fn_name'.",
+            category=DeprecationWarning,
+        )
+
         normal_name = normalize_name(name)
         resolver = self.resolver(normal_name)
         if resolver:
@@ -272,6 +279,12 @@ class ENS:
             return None
 
     def resolver(self, normal_name: str) -> Optional['Contract']:
+        warnings.warn(
+            "The function signature for resolver() will change in v6 to accept 'name' as a param, "
+            "over 'normalized_name', and the method will normalize the name internally.",
+            category=FutureWarning,
+        )
+
         resolver_addr = self.ens.caller.resolver(normal_name_to_hash(normal_name))
         if is_none_or_zero_address(resolver_addr):
             return None
--- a/ens/main.py
+++ b/ens/main.py
@@ -261,8 +261,10 @@ class ENS:
 
     def resolve(self, name: str, get: str = 'addr') -> Optional[Union[ChecksumAddress, str]]:
         warnings.warn(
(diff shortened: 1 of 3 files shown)


ens/main.py
```diff
@@ -259,6 +260,12 @@ def setup_name(
             return self._setup_reverse(name, address, transact=transact)
 
     def resolve(self, name: str, get: str = 'addr') -> Optional[Union[ChecksumAddress, str]]:
+        warnings.warn(
```
kclowes, line 263: I don't feel super strongly about it, but I noticed that we have 681 new deprecation/future warnings in our tests, so circling back to the conversation we had this morning, I think it may be worth adding the `fn_name` argument to the signature here and then issuing a warning if it's not present. As a user it would drive me nuts to get a whole bunch of warnings and not have an upgrade path. So something like: 
```
def resolve(self, name: str, get: str = 'addr', fn_name=None) -> Optional[Union[ChecksumAddress, str]]:
    if fn_name is None:
        warnings.warn(In v6, the resolve() method will be made internal..., to suppress this warning in the meantime, use the `fn_name` keyword argument. In v6 you'll be able to pass it positionally.)
    ...
```
But also if we're pretty sure that no one is using this method in practice, then I guess it's just a personal problem :)

kclowes (approved):
Looks good! I think I found a few typos in the error message and added a longer comment that you can feel free to take or leave!

ens/main.py
fselmo: Haha, no I think that is a great idea for `v5` I'll push this up.

fselmo: I'm curious though now that I'm taking a look... since the ``resolve()`` method will be made internal and changed to ``_resolve()``, I don't think this is the ideal migration route. Instead, I think the appropriate path(s) would be to change any uses of ``ns.resolve()`` to either ``ns.address()`` for the forward resolution, or ``ns.name()`` for the reverse resolution. That was kind of the point of getting rid of this as a public API - to avoid confusion and abstract the resolutions to their more specific, appropriate methods.

Thoughts?

ens/main.py
```diff
@@ -271,7 +280,16 @@ def resolve(self, name: str, get: str = 'addr') -> Optional[Union[ChecksumAddres
         else:
             return None
 
-    def resolver(self, normal_name: str) -> Optional['Contract']:
+    def resolver(self, normal_name: str, name: str = None) -> Optional['Contract']:
```
kclowes, line 283: What do you think about setting `normal_name = None` here as well, and then the interim API that satisfies the warning looks like: `resolver(name='test.eth')` instead of `resolver('test.eth', name='test.eth')`? We'd probably want to do some validation too, to make sure both aren't passed in.

kclowes (approved):
Sorry it's taken me so long to get to this! I added one comment, but looks good otherwise!

ens/main.py
fselmo: Yeah, I like that way better. I also forgot about this PR before last release... I'll clean this up and run it by for another review 👌

fselmo: @kclowes this should be ready to review again. lmk if the latest commit is good here 👀

kclowes (approved):
:shipit: ! LGTM!

PR #2457, opened 2022-05-02
CCIP Read support

fselmo:
### What was wrong?

- Need to implement [CCIP Read / EIP-3668](https://eips.ethereum.org/EIPS/eip-3668) support.
- Requested in issue #2406

### How was it fixed?

- CCIP Read / EIP-3668 support was added:
  - New `OffchainLookup` exception added with `payload` property to be filled with appropriate data from the contract revert.
  - Sync and async implementations added via `ccip_read_enabled` flags on contract call / `eth_call`.
  - The bulk of offchain lookup payload handling was abstracted to new public utils apis for exception handling: `web3/utils/exception_handling.py` and `web3/utils/async_exception_handling.py`.
  - Test contracts created and added to geth fixture.
  - Test fixtures for `offchain_lookup_contract` added where setup was needed.
  - ENS offchain resolver support tested separately via a new contract setup within `tests/ens`.
  - `GET` request support added and general, non-json-rpc, `POST` request support added.
- Some refactoring:
  - Refactored `requests.py` to account for `json` payloads for both sync and async POST and GET requests.

### Todo:
- [x] Rebase with `master` once the ENSIP-10 PR is merged in
- [x] Add cute animal picture


web3/_utils/method_formatters.py
```diff
@@ -569,6 +586,14 @@ def raise_solidity_error_on_revert(response: RPCResponse) -> RPCResponse:
         reason = data[len(prefix) + 64:len(prefix) + 64 + reason_length * 2]
         raise ContractLogicError(f'execution reverted: {bytes.fromhex(reason).decode("utf8")}')
 
+    # --- EIP-3668 | CCIP Read --- #
+    # 0x556f1830 is the function selector for OffchainLookup(address,string[],bytes,bytes4,bytes)
+    if data[:10] == '0x556f1830':
```
kclowes, line 591: This feels like it should be in it's own function since it's not raising a solidity error (i.e. a `ContractLogicError`). I propose we either rename this function or pull lines 591-596 out into their own function.

kclowes (commented):
Looking good! I got most of the way through (38/42 files), and will pick back up again on Wednesday. A few things I noticed:

- We should probably add some docs
- It doesn't look like this has been implemented for `transact` or `estimate_gas`. I think the spec mentions both of those methods. Do you think those belong in a new PR or should they go in this one?
- I'm probably missing something, but I'm curious about the decision to eliminate the explicit`data` param that gets passed in to `make_post_request` [here](https://github.com/ethereum/web3.py/pull/2457/files#diff-da1c04440e0604ceb9d314eec7151604f68036eaf4e6206f78f7aa38e36b0a65L115)

web3/_utils/method_formatters.py
fselmo: Good point. Alternatively, should this maybe extend from `ContractLogicError`? It's technically an `error` in solidity, even though it's a purposeful error.

web3/eth.py
```diff
@@ -369,6 +375,41 @@ async def fee_history(
...
+
+        return await self._call(transaction, block_identifier, state_override)
+
+    async def durin_call(
+        self,
+        transaction: TxParams,
+        block_identifier: Optional[BlockIdentifier] = None,
+        state_override: Optional[CallOverride] = None,
+    ) -> Union[bytes, bytearray]:
+        for _ in range(4):
```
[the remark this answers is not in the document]
fselmo: I would be good with that... we can add some validation that prevents this number from being less than `4` 👍

.circleci/config.yml
```diff
@@ -244,28 +244,62 @@ jobs:
       # Please don't use this key for any shenanigans
       WEB3_INFURA_PROJECT_ID: 7707850c2fb7465ebe6f150d67182e22
 
-  py37-integration-goethereum-ipc:
+  py37-integration-goethereum-ipc_eth:
     <<: *geth_steps
     docker:
       - image: cimg/python:3.7
+      - image: cimg/python:3.7
```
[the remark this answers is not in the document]
fselmo: Ah, no I think this wasn't rebased correctly. Thanks for catching it 👌

web3/contract.py
```diff
@@ -900,7 +900,8 @@ def _set_function_info(self) -> None:
     def call(
         self, transaction: Optional[TxParams] = None,
         block_identifier: BlockIdentifier = 'latest',
-        state_override: Optional[CallOverrideParams] = None,
+        state_override: Optional[CallOverride] = None,
+        ccip_read_enabled: bool = False,
```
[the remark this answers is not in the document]
fselmo: I don't think ``transact`` should have the flag. The idea behind having it ``false`` by default for ``eth_call`` is that there seems to be some security considerations that make this safer as being an explicit choice to a method you are certain does what you want it to do.

web3/_utils/request.py
```diff
@@ -63,56 +64,88 @@ def cache_session(endpoint_uri: URI, session: requests.Session) -> None:
...
     session.close()
 
 
 _session_cache = lru.LRU(8, callback=_remove_session)
-_async_session_cache_lock = threading.Lock()
-_async_session_cache = SessionCache(size=20)
 
 
-def _get_session(endpoint_uri: URI) -> requests.Session:
+def get_session(endpoint_uri: URI) -> requests.Session:
```
kclowes, line 74: total nit: did you mean to change this from `_get_session` to `get_session`?

web3/_utils/transactions.py
```diff
@@ -239,3 +256,67 @@ def replace_transaction(
...
+                ['bytes', 'bytes'],
+                [
+                    to_bytes_if_hex(result['data']),
+                    to_bytes_if_hex(offchain_lookup_payload['extraData']),
+                ]
+            )
+        ])
+
+        return encoded_data_with_function_selector
+    raise Exception("Offchain lookup failed for supplied urls.")
```
kclowes, line 322: Do you think a more specific exception here would be useful?

web3/eth.py
```diff
@@ -749,11 +782,43 @@ def sign_munger(
...
+
+        if not max_redirects or max_redirects < 4:
+            raise ValueError("ccip_read_max_redirects property on provider must be at least 4.")
+
+        for _ in range(max_redirects):
+            try:
+                return self._call(transaction, block_identifier, state_override)
+            except OffchainLookup as offchain_lookup:
+                durin_calldata = handle_offchain_lookup(offchain_lookup.payload, transaction)
+                transaction['data'] = durin_calldata
```
kclowes, line 818: I assume we don't, but just to be sure, do we need to return something here?

web3/_utils/method_formatters.py
kclowes: Yeah, extending from `ContractLogicError` makes sense to me since it stems from a `revert` 👍

web3/_utils/module_testing/eth_module.py
```diff
@@ -42,13 +40,23 @@
 from web3._utils.method_formatters import (
     to_hex_if_integer,
 )
+from web3._utils.module_testing.utils import (
```
kclowes, line 43: another nit: I think I like `web3._utils.module_testing_utils.py` a little bit better, but am open to having it in the module testing folder if you think that makes more sense.

kclowes (commented):
Okay, I got through the rest! Just a few nitpicky comments

web3/_utils/module_testing/eth_module.py
fselmo: I like that better too 👌🏼

web3/eth.py
fselmo: Yeah we don't. All we're doing is replacing the `data` in the transaction which was built from the offchain REST request and the `OffchainLookup` information and calling that in the `try` block again.

web3/_utils/transactions.py
fselmo: I was thinking about that but wasn't sure what would make it better from our arsenal of exceptions. I just raised what I thought to be the equivalent of like a `500` / generic internal server error. Feels like this should be a faulty offchain lookup contract if we ever get here. Open to suggestions.

web3/_utils/request.py
fselmo: Yeah. I ised this in another internal file... I thought the fact that it's already in `_utils` should detract from anyone using this (technically) and thought it made more sense since it's no longer only used internally in this file, but now also in my mocking methods in module testing utils.

kclowes (approved):
:shipit:!

PR #2438, opened 2022-04-22
deprecated estimateGas and buildTransaction in v5 contract

dbfreem:
### What was wrong?

Deprecating buildTransaction and estimateGas in the v5 branch. I will remove them in the asyncify-contract branch so they are effectively removed in v6

Related to Issue #1416

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

PR #2459, opened 2022-05-04
deprecated estimateGas and buildTransaction in v5 contract #2438

dbfreem:
Deprecating buildTransaction and estimateGas in the v5 branch. I will remove them in the asyncify-contract branch so they are effectively removed in v6

Related to Issue https://github.com/ethereum/web3.py/issues/1416

Todo:

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -244,7 +244,12 @@ Each Contract Factory exposes the following methods.
         >>> txn_receipt['contractAddress']
         '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
 
-.. py:classmethod:: Contract.constructor(*args, **kwargs).estimateGas(transaction=None, block_identifier=None)
+  .. py:classmethod:: Contract.constructor(*args, **kwargs).estimateGas(transaction=None, block_identifier=None)
+    :noindex:
+
+    .. warning:: Deprecated: This method is deprecated in favor of :py:meth:`Contract.constructor(*args, **kwargs).estimate_gas`
+
+.. py:classmethod:: Contract.constructor(*args, **kwargs).estimate_gas(transaction=None, block_identifier=None)
     :noindex:
 
     Estimate gas for constructing and deploying the contract.
@@ -264,10 +269,16 @@ Each Contract Factory exposes the following methods.
 
     .. code-block:: python
 
-        >>> token_contract.constructor(web3.eth.coinbase, 12345).estimateGas()
+        >>> token_contract.constructor(web3.eth.coinbase, 12345).estimate_gas()
         12563
 
 .. py:classmethod:: Contract.constructor(*args, **kwargs).buildTransaction(transaction=None)
+    :noindex:
+
+    .. warning:: Deprecated: This method is deprecated in favor of :py:meth:`Contract.constructor(*args, **kwargs).build_transaction`
+
+
+.. py:classmethod:: Contract.constructor(*args, **kwargs).build_transaction(transaction=None)
     :noindex:
 
     Construct the contract deploy transaction bytecode data.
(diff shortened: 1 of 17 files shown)


dbfreem:
@kclowes take a look at this PR. I closed my other PR because I accidently pulled from master and REALLY messed it up.

dbfreem:
I fixed the issue with the combomethod annotation. Also, @kclowes I took care of the warnings in the test in the asyncify-contract branch already so I was going to leave these. thoughts??

kclowes:
> Also, @kclowes I took care of the warnings in the test in the asyncify-contract branch already so I was going to leave these. thoughts??

For deprecations, we prefer to change all of the instances of our code over to the new function name and then have one test to be sure the deprecation warning gets shown just so that we don't have a bunch of noise in our test output. For example something like:

```
def test_that_deprecation_warning_is_shown():
    ...
    with pytest.warns("DeprecationWarning", match="estimateGas has been deprecated in favor of estimate_gas")
        contract.functions.foo().estimateGas()
```

kclowes:
> Also, @kclowes I took care of the warnings in the test in the asyncify-contract branch already so I was going to leave these. thoughts??

For deprecations, we prefer to change all of the instances of our code over to the new function name and then have one test to be sure the deprecation warning gets shown just so that we don't have a bunch of noise in our test output. For example something like:

```
def test_that_deprecation_warning_is_shown():
    ...
    with pytest.warns("DeprecationWarning", match="estimateGas has been deprecated in favor of estimate_gas")
        contract.functions.foo().estimateGas()
```

dbfreem:
@kclowes  makes sense I will take a look at this.

dbfreem:
@kclowes this one should be good to go now.

docs/contracts.rst
```diff
@@ -834,6 +845,9 @@ Methods
     `More information about archival nodes here <https://ethereum.stackexchange.com/a/84200/620>`_.
 
 .. py:method:: ContractFunction.estimateGas(transaction, block_identifier=None)
+    .. warning:: Deprecated: This method is deprecated in favor of :class:`~estimate_gas`
```
kclowes, line 848: I think this needs one more space in between 847 and 848. It's showing up like this: 
image

docs/contracts.rst
```diff
@@ -862,14 +876,17 @@ Methods
         nodes would result in an error like:  ``ValueError: {'code': -32602, 'message': 'too many arguments, want at most 1'}``
 
 .. py:method:: ContractFunction.buildTransaction(transaction)
+    .. warning:: Deprecated: This method is deprecated in favor of :class:`~build_transaction`
```
kclowes, line 879: I think there needs another line between 878 and 879 as well.

tests/core/contracts/conftest.py
```diff
@@ -1008,7 +1008,7 @@ def invoke_contract(api_call_desig='call',
                     func_args=[],
                     func_kwargs={},
                     tx_params={}):
-    allowable_call_desig = ['call', 'transact', 'estimateGas', 'buildTransaction']
+    allowable_call_desig = ['call', 'transact', 'estimate_gas', 'build_transaction']
```
kclowes, line 1011: I didn't actually look to see, but do we need to keep 'estimateGas' and `buildTransaction` here until `estimateGas` and `buildTransaction` are removed?

web3/contract.py
```diff
@@ -695,7 +708,7 @@ def check_forbidden_keys_in_transaction(
 
 
 class ConciseMethod:
-    ALLOWED_MODIFIERS = {'call', 'estimateGas', 'transact', 'buildTransaction'}
+    ALLOWED_MODIFIERS = {'call', 'estimate_gas', 'transact', 'build_transaction'}
```
kclowes, line 711: Same question here about keeping `estimateGas` and `buildTransaction` here.

kclowes (commented):
Looks good to me! I left a few questions, but after those are addressed, should be good to go!

tests/core/contracts/conftest.py
dbfreem: This one is ok to change since I changed the names of all the methods in the test themselves. This is the list of methods that four methods that come below this method.

web3/contract.py
dbfreem: These turn out to be args passed in to the `ConciseMethod.__call__` method so they should have stayed the same since it would be a breaking change. I will change these back.

docs/contracts.rst
dbfreem: fixed

docs/contracts.rst
dbfreem: fixed

PR #2466, opened 2022-05-11
Feature/asyncify contract

dbfreem:
changes suggested in pr to master


web3/contract.py
```diff
@@ -540,6 +540,38 @@ def find_functions_by_identifier(cls,
...
+    ) -> Union['ContractFunction', 'AsyncContractFunction']:
+        if abi and fallback_func_abi_exists(abi):
+            return function_type.factory(
+                'fallback',
+                w3=w3,
+                contract_abi=abi,
+                address=address,
+                function_identifier=FallbackFn)()
+
+        return cast('ContractFunction', NonExistentFallbackFunction())
```
pacrob, line 557: Should this use `function_type` instead of `'ContractFunction'`?

dbfreem: good catch @pacrob. I fixed it but had to add the type ignore because mypy didn't like that type as a variable

web3/contract.py
```diff
@@ -540,6 +540,38 @@ def find_functions_by_identifier(cls,
...
+    ) -> Union['ContractFunction', 'AsyncContractFunction']:
+        if abi and receive_func_abi_exists(abi):
+            return function_type.factory(
+                'receive',
+                w3=w3,
+                contract_abi=abi,
+                address=address,
+                function_identifier=ReceiveFn)()
+
+        return cast('ContractFunction', NonExistentReceiveFunction())
```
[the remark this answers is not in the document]
dbfreem: good catch @pacrob. I fixed it but had to add the type ignore because mypy didn't like that type as a variable

PR #2469, opened 2022-05-12
Fix type hint for dict in the example doc

kaiix:
### What was wrong?

[Advanced example: Fetching all token transfer events](https://web3py.readthedocs.io/en/stable/examples.html#id23) use `{}` as the dict type hint, which is wrong.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -790,7 +790,7 @@ The script can be run with: ``python ./eventscanner.py <your JSON-RPC API URL>``
     import time
     import logging
     from abc import ABC, abstractmethod
-    from typing import Tuple, Optional, Callable, List, Iterable
+    from typing import Tuple, Optional, Callable, List, Iterable, Dict, Any
 
     from web3 import Web3
     from web3.contract import Contract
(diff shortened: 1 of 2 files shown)


pacrob:
Thanks, @kaiix ! You are correct, we should not be using `{}`, but it looks like we prefer to use `Dict` instead of `dict` most everywhere else. Would you mind making that small change? If you're especially motivated, it looks like we have a few more uses of `{}` that could be updated too, but not necessary.

kaiix:
> Thanks, @kaiix ! You are correct, we should not be using `{}`, but it looks like we prefer to use `Dict` instead of `dict` most everywhere else.

Everywhere else in `example.rst` uses `dict` as type hint, should it all be changed to `Dict`?

>  it looks like we have a few more uses of `{}` that could be updated too, but not necessary.

I haven't found anywhere else that uses `{}` as type hint, so if you do, let me know and I'll be willing to help.

pacrob:
> Everywhere else in `example.rst` uses `dict` as type hint, should it all be changed to `Dict`?
Yes please!

> I haven't found anywhere else that uses `{}` as type hint, so if you do, let me know and I'll be willing to help.
You're right, my mistake. Just the `dict`s then.

pacrob:
Looks good! No worry about the failing tests, we're working on the timeouts. Thank you!

PR #2476, opened 2022-05-16
Pin Jinja2<=3.0.3

kclowes:
### What was wrong?
Jinja v3.1.0 started throwing errors during the release process. I think it has something to do with the fact that they dropped python 3.6 support in version 3.1.0, but I'm not sure.

### How was it fixed?
Pinned Jinja2 to the version before where they dropped python 3.6

### Todo:

#### Cute Animal Picture

image


pacrob (approved):
lgtm!

PR #2477, opened 2022-05-16
Increase cache size

kclowes:
### What was wrong?
Our async cache was limited to 8 entries and would hang after that number was reached. This is a temporary fix so that users can use more than 8. Will solve the root cause in another PR.

Related to Issue #2451

### How was it fixed?
Bumped the cache size.

### Todo:

#### Cute Animal Picture


pacrob (approved):
lgtm!

PR #2478, opened 2022-05-17
removed verison check in __init__

dbfreem:
These version checks are out of data and probably not needed since the setup.py restricts the ability to install in any python version that is not supported.


pacrob:
Thanks!

PR #2479, opened 2022-05-17
removed deprecate sha3 methods

dbfreem:
The sha3 methods have been deprecated in main.py for some time now. Since the v6 branch is now on main it probably makes sense to remove these.


dbfreem:
fixed the broken test

kclowes (approved):
LGTM! Thank you!

PR #2432, opened 2022-04-20
removed deprecated methods from eth and geth

dbfreem:
### What was wrong?

Removed deprecated methods from eth and geth

Related to Issue #1416 

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

PR #2480, opened 2022-05-18
removed more depecrated methods in geth and parity

dbfreem:
### What was wrong?

Related to Issue #2432 

I missed some of the methods in the #2432 PR so here is the removal of the rest of the geth deprecated methods. Also, since some of them were also exposed in parity I had to update them there too.


kclowes (approved):
Thank you! I figured it was going to be easier to go through and take out the strays myself rather than listing them one by one for you to do 😆 Once tests pass, I'll merge!

Issue #2486, opened 2022-05-26
EventScanner example yields SyntaxError


0xRadiant:
* Version: 6.0.0-beta.2
* Python: 3.8.10
* OS: linux

### What was wrong?
When working with the `eventscanner.py` example, I noticed we can get a `SyntaxError` when running the code as-is. Here is an example:
```
0xRadiant@ubuntu-linux-desktop:~/workspace/playground$ python3 eventscanner.py https://arb-mainnet.g.alchemy.com/v2/API_KEY
  File "eventscanner.py", line 213
    logger.debug(f"Processing event {evt["event"]}, block: {evt["blockNumber"]} count: {evt["blockNumber"]}")
                                          ^
SyntaxError: invalid syntax
```

### How can it be fixed?
Assign the event, block number, and count to their own variables and pass them in to the `logger.debug` message.

fselmo:
closed by #2487

PR #2487, opened 2022-05-26
fix: Fixing syntax within EventScanner's scan_chunk method

0xRadiant:
### What was wrong?

Related to Issue #2486

### How was it fixed?
Assigning the event, block number, and count to their own variables and pass them in to the logger.debug message.

#### Cute Animal Picture

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -991,8 +991,12 @@ The script can be run with: ``python ./eventscanner.py <your JSON-RPC API URL>``
                     # Get UTC time when this event happened (block mined timestamp)
                     # from our in-memory cache
                     block_when = get_block_when(block_number)
-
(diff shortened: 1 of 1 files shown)


docs/examples.rst
```diff
@@ -991,8 +991,12 @@ The script can be run with: ``python ./eventscanner.py <your JSON-RPC API URL>``
                     # Get UTC time when this event happened (block mined timestamp)
                     # from our in-memory cache
                     block_when = get_block_when(block_number)
-
-                    logger.debug(f"Processing event {evt["event"]}, block: {evt["blockNumber"]} count: {evt["blockNumber"]}")
```
0xRadiant, line 995: I also noticed that `count` and `block` both refer to `evt["blockNumber"]`. I'm not yet sure if this was intentional. If this is not intended, we can address this in the same pull request.

kclowes: I don't think that was intentional, please feel free to update! Thanks!

PR #2496, opened 2022-06-01
Add get_abi_output_names utility method

feulf:
### What was wrong?
Very simply, there isn't a utility method to get the output names from a call, and we needed that.

#### Cute Animal Picture
image

### DIFF
--- a/web3/_utils/abi.py
+++ b/web3/_utils/abi.py
@@ -108,6 +108,13 @@ def get_abi_input_types(abi: ABIFunction) -> List[str]:
         return [collapse_if_tuple(cast(Dict[str, Any], arg)) for arg in abi['inputs']]
 
 
+def get_abi_input_names(abi: Union[ABIFunction, ABIEvent]) -> List[str]:
+    if 'inputs' not in abi and abi['type'] == 'fallback':
+        return []
+    else:
+        return [arg['name'] for arg in abi['inputs']]
+
+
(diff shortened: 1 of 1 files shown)


fselmo:
Hey @feulf 👋 ... thanks for the PR! I'd be happy to have this added. One not-so-ideal pattern I'm seeing is that the internal utility methods found at ``web3._utils`` seem to be useful for users even though they aren't technically meant to be made public. What that means is they could technically break across minor / patch versions.

In practice, I don't see that happening too often since they perform a particular useful functionality that is core to how the library works but things like method signatures, typing, etc... could technically "break" these internal utility methods and anyone using these methods would then have breaking changes across minor / patch versions.

I'm thinking since this new method isn't even used internally, adding them to the newly created ``web3.utils`` module for public API utility methods, would be best. If the `input` types are also useful to you, perhaps that could be entirely moved over there as well. We should also add documentation for it in the appropriate area for the `utils` module in the docs.

I'm thinking we can move all of that logic over to the external API module (``web3.utils.abi``, perhaps) and if any of the internal logic ever needs to deviate from these public utility methods, it can be created at that time to replace the public one, internally.

Thoughts?

cc: @kclowes

kclowes:
@fselmo great ideas! I think a lot of the abi utility functions that we have do get used fairly frequently even though they are in `_utils` so it would be nice to expose a public `utils` folder 👍

fselmo:
closed in favor of #2596

PR #2502, opened 2022-06-06
Pin eth-rlp<0.3

### LINKED ISSUE #2424: pip install web3, errors due to incompatible version of eth-rlp
* Version: 5.28.0
* Python: 3.8
* OS: linux
* `pip freeze` output

```
appdirs==1.4.3
certifi==2019.11.28
chardet==3.0.4
dbus-python==1.2.16
distlib==0.3.0
distro-info===0.23ubuntu1
filelock==3.0.12
idna==2.8
importlib-metadata==1.5.0
more-itertools==4.2.0
netifaces==0.10.4
PyGObject==3.36.0
pymacaroons==0.13.0
PyNaCl==1.3.0
python-apt==2.0.0+ubuntu0.20.4.6
python-debian===0.1.36ubuntu1
PyYAML==5.3.1
requests==2.22.0
requests-unixsocket==0.2.0
six==1.14.0
ubuntu-advantage-tools==27.4
urllib3==1.25.8
virtualenv==20.0.17
zipp==1.0.0

```

### What was wrong?

When trying to install web3.py in a new virtualenv, on a new installation of Ubuntu 20.04, the following error is thrown:

```

virtualenv -p python3.8 venv
source venv/bin/activate
pip install web3
...
ERROR: eth-rlp 0.3.0 has requirement eth-utils<3,>=2.0.0, but you'll have eth-utils 1.10.0 which is incompatible.
...

```

kclowes:
### What was wrong?
`eth-account` has a typo (I believe) [here](https://github.com/ethereum/eth-account/blob/v0.5.x/setup.py#L67), which causes dependency resolution errors for us on older versions of pip, before they changed their dependency resolution.

I could be convinced that this change belongs in `eth-account` if someone feels strongly, but I think it belongs here because it's breaking here.

Closes #2424 

### How was it fixed?
Pinned `eth-rlp` to `<0.3` to support older versions of pip.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2502.misc.rst
+++ b/newsfragments/2502.misc.rst
@@ -1,0 +1,1 @@
+Pin eth-rlp to <0.3 so that dependencies will resolve with older versions of pip.
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "eth-abi>=2.0.0b6,<3.0.0",
         "eth-account>=0.5.7,<0.6.0",
         "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
+        "eth-rlp<0.3",
         "eth-typing>=2.0.0,<3.0.0",
         "eth-utils>=1.9.5,<2.0.0",
         "hexbytes>=0.1.0,<1.0.0",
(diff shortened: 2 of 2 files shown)


fselmo (approved):
LGTM 👍. Based on our discussion earlier, I think this helps avoid a round of breaking changes. If we could track this for the next round of major updates to eventually pin it to ``eth-account`` that would be nice. We could add a comment on this line or just make a note somewhere else maybe?

kclowes:
Comment added! I'll also add an issue in eth-account to tighten up the eth-rlp dependency requirement over there and we can pull it into the next breaking release on eth-account.

PR #2503, opened 2022-06-06
Add a global ccip read support flag on the provider.

### LINKED ISSUE #2499: [v6] Allow for CCIP Read to be disabled for calls at the provider level
### What was wrong?

EIP-3668 recommends the following for CCIP Read support:

> Client libraries should disable CCIP read for transactions **(but not for calls)** by default, and require the caller to explicitly enable this functionality. Enablement should be possible both on a per-contract, per-domain, or global basis.

We currently enable it by default for calls and don't handle this at all automatically for transactions but do provide documentation on how to properly handle offchain lookup for transactions. However, if a user is not interested in enabling CCIP Read at all, even for calls, we could allow this to happen on the provider level and they wouldn't have to worry about disabling it at the `eth_call` level with each call.

### How can it be fixed?

- Add a provider `ccip_read_calls_enabled` flag and account for this value at the `w3.eth.call()` level.

This should be implemented so that the flag on `eth.call()` can override whatever is set on the global provider flag. This would allow for call-specific control for CCIP read.

i.e. for `Eth` (and similarly `AsyncEth`):
```py
def call(self, ..., ccip_read_enabled: Optional[bool] = None):
    ccip_read_enabled_on_provider = self.w3.provider.ccip_read_calls_enabled
    if (
        # default conditions:
        ccip_read_enabled_on_provider and ccip_read_enabled is not False
        # explicit call flag overrides provider flag, enabling ccip read for specific calls:
        or not ccip_read_enabled_on_provider and ccip_read_enabled is True
    ):
        return self._durin_call(...)

    return self._call(...)
```

fselmo:
### What was wrong?

closes #2499

### How was it fixed?

- Global flag for CCIP Read support for calls added to provider: ``ccip_read_calls_enabled``
- Add relevant tests including tests making sure the ``eth.call()`` flag can override the provider flag for explicit call-specific control
- Tweak older offchain lookup tests due to timeouts after 40 calls to the contract
- Some document edits including wrong header type for CCIP Read section

### Todo:
- [x] Update docs

#### Cute Animal Picture

Image from iOS (3)

### DIFF
--- a/tests/core/contracts/test_offchain_lookup.py
+++ b/tests/core/contracts/test_offchain_lookup.py
@@ -16,6 +16,7 @@ from web3._utils.type_conversion import (
     to_hex_if_bytes,
 )
 from web3.exceptions import (
+    OffchainLookup as _OffchainLookup,
     TooManyRequests,
     ValidationError,
 )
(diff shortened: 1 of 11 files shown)


tests/core/contracts/test_offchain_lookup.py
```diff
@@ -65,6 +66,53 @@ def test_offchain_lookup_functionality(
...
+        )
+
+    # test global flag on the provider
+    w3.provider.ccip_read_calls_enabled = False
+
+    with pytest.raises(_OffchainLookup):
+        offchain_lookup_contract.caller.testOffchainLookup(OFFCHAIN_LOOKUP_TEST_DATA)
+
+    with pytest.raises(_OffchainLookup):
+        offchain_lookup_contract.caller(ccip_read_enabled=False).testOffchainLookup(
```
pacrob, line 89: the `ccip_read_enabled=False` shouldn't be there, because you're testing the global flag, right?

fselmo: In this case I'm testing that `None` (default) and `False` values for the `call` flag should yield the same result when the global flag is set to `False`.

pacrob: dig it

pacrob (approved):
lgtm!

web3/providers/async_base.py
```diff
@@ -36,6 +36,7 @@ class AsyncBaseProvider:
     # a tuple of (all_middlewares, request_func)
     _request_func_cache: Tuple[Tuple[Middleware, ...], Callable[..., RPCResponse]] = (None, None)
 
+    ccip_read_calls_enabled: bool = True
```
kclowes, line 39: semantics nit to take or leave: I know we talked about this some on Discord, but seeing `ccip_read_calls_enabled` next to `ccip_read_enabled` doesn't give me a good idea of the difference between the two. Maybe something like `global_ccip_read_enabled` would be more descriptive? /cc @marcgarreau :)

kclowes (approved):
I think this is a nice improvement! I left one comment around the flag naming, but looks good to me overall!

web3/providers/async_base.py
fselmo: No expectation as of now. We provide a pretty easy way to handle [preflighting](https://web3py.readthedocs.io/en/latest/examples.html#ccip-read-support-for-offchain-lookup) if need be at the moment.

marcgarreau: down the future-proofing semantics rabbit hole:

i think it comes down to: if we were to add ccip read support beyond call, would you prefer
a) more granularity, e.g., more flags - `ccip_read_calls_enabled`, `ccip_read_tx_enabled`, etc. or
b) everything covered by a single flag, e.g., `ccip_read_enabled` or `global_ccip_read_enabled`

with limited context, my instinct is `b`.

pacrob: If it is expected to be the master on/off switch for CCIP Read functionality, then I would agree adding `global` helps clarify, and the inclusion of `call` in the name is less important.

fselmo: For sure, I'd prefer that too of we went that route. We talked about not handling transactions automatically since it gets webbed up by our middleware triggering `eth_estimateGas` if there is no `gas` field on the original transaction, etc... and this mainly a read functionality atm.

I think the main goal here is to distinguish the provider flag from the `eth_call` flag since we don't have a flag that is all-encompassing.

fselmo: I think global works as long as the documentation clarifies we only turn this on for calls and that seems to be the consensus 👍🏼.

PR #2504, opened 2022-06-07
Update examples.rst

brucx:
### What was wrong?

The example code does not work properly.

### How was it fixed?

Replace double quotes with single quotes.


pacrob:
Thanks for catching that!

PR #2518, opened 2022-06-16
Declutter git blame

kclowes:
### What was wrong?
Git blame gets clobbered with the black updates.

### How was it fixed?
Added a .git-blame-ignore-revs file, and added docs so developers can use the file to ignore certain commits when viewing git blame locally.

More info can be found here: 
- https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
- https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

### Todo:

#### Cute Animal Picture


fselmo (approved):
🔥 🔥 lgtm!

PR #2501, opened 2022-06-03
Feature/ens asyncify

dbfreem:
Here is AsyncENS. Still needs docs and cleanup

There was A LOT of copy and paste here from ENS since ENS is so dependent on Eth, and Contract there wasn't much code that could be put in BaseENS.

1. There is a possible breaking change in the way the modules are loaded in ens utils.py. To be consistent between AsyncENS and ENS I only loaded the Eth module when the Web3 instance is created
2. I would like to create a ENSFactory like I did AsyncENSFactory so that there is consistency. The reason that AsyncENSFactory was created was to support the async creation of the web3 with the stalecheck middleware. If I create an ENSFactory then I would remove the init and fromweb3 from ENS which would be breaking but drive consistency between AsyncENS and ENS

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)
- [x] Need to add docs stalecheck and AsyncENS

PR #2526, opened 2022-06-17
Cleaned Up Async ENS

fselmo:
Cherry picked and squashed relevant commits, updated commit messages from PR #2501. Merge commits can't be squashed easily so this is just the easiest way to clean up the commit history. Refer to that PR for all details and conversations around the changes.

Use `git diff cleanup-async-ens dbfreem/feature/ens_asyncify` to see there are minor re-organizing differences between these two PRs / branches

closes #2501

### DIFF
--- a/docs/ens.rst
+++ b/docs/ens.rst
@@ -7,10 +7,16 @@ ENS API
 
 Continue below for the detailed specs on each method and class in the ens module.
 
-ens\.main module
+ens\.ens module
 ----------------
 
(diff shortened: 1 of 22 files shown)


fselmo:
just kidding... we effectively cleaned up #2501 with a force push after cherry picking only the non-merge commits and rebasing locally against `upstream/master`.

Closing this so the conversation can just stay in one place.

Issue #2535, opened 2022-06-24
Failing to parse revert string


BlazeWasHere:
* Version: 5.26.0
* Python: 3.9
* OS: linux
* `pip freeze` output

```
aiohttp==3.8.1
aiosignal==1.2.0
APScheduler==3.9.1
async-timeout==4.0.2
attrs==21.4.0
base58==2.1.1
bech32==1.2.0
bidict==0.22.0
bitarray==1.2.2
certifi==2021.10.8
charset-normalizer==2.0.12
click==8.1.2
cytoolz==0.11.2
Deprecated==1.2.13
eth-abi==2.1.1
eth-account==0.5.7
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.3.4
eth-rlp==0.2.1
eth-typing==2.3.0
eth-utils==1.10.0
Flask==2.0.2
Flask-APScheduler==1.12.3
Flask-Caching==1.10.1
Flask-SocketIO==5.1.1
frozenlist==1.3.0
gevent==21.12.0
gevent-websocket==0.10.1
greenlet==1.1.2
gunicorn==20.1.0
hexbytes==0.2.2
idna==3.3
ipfshttpclient==0.8.0a2
itsdangerous==2.1.2
Jinja2==3.1.1
jsonschema==3.2.0
lru-dict==1.1.7
MarkupSafe==2.1.1
multiaddr==0.0.9
multidict==6.0.2
netaddr==0.8.0
packaging==21.3
parsimonious==0.8.1
protobuf==3.20.0
pycryptodome==3.14.1
pyparsing==3.0.8
pyrsistent==0.18.1
python-dateutil==2.8.2
python-dotenv==0.19.2
python-engineio==4.3.1
python-redis-lock==3.7.0
python-socketio==5.5.2
pytz==2022.1
pytz-deprecation-shim==0.1.0.post0
redis==4.1.1
requests==2.27.1
rlp==2.0.1
simplejson==3.17.6
six==1.16.0
toolz==0.11.2
tzdata==2022.1
tzlocal==4.2
urllib3==1.26.9
varint==1.0.2
web3==5.26.0
websockets==9.1
Werkzeug==2.1.1
wrapt==1.14.0
yarl==1.7.2
zope.event==4.5.0
zope.interface==5.4.0
```

### What was wrong?

```
[...]
  File "/usr/local/lib/python3.9/site-packages/web3/contract.py", line 957, in call
    return call_contract_function(
[...]

PR #2536, opened 2022-06-24
Handle Revert() strings

BlazeWasHere:
### What was wrong?

Related to Issue #2535 

### How was it fixed?

Handle response errors that start with `Revert(`.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/2535.bugfix.rst
+++ b/newsfragments/2535.bugfix.rst
@@ -1,0 +1,1 @@
+Handle json-rpc response errors in the format Revert(0x...)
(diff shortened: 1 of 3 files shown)


kclowes:
Thanks for the PR! However, instead of trying to catch every single provider's edge case error format, we're going to move towards only supporting an error response formatted per the official [JSON-RPC spec](https://www.jsonrpc.org/specification#error_object). You may try and reach out to the Aurora project to get them to return errors like the spec specifies. Closing as won't fix.

PR #2538, opened 2022-06-25
Add some additional chain info for POA middleware

miohtama:
### What was wrong?

Some more info regarding POA middleware in docs.

I also just wanted to post a cute animal picture.

### How was it fixed?

- Noted that this may be needed for non-Rinkeby chains
- Added colour on the error the user is going to receive

#### Cute Animal Picture

image

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -361,10 +361,19 @@ All of the caching middlewares accept these common arguments.
 
 .. _geth-poa:
 
-Geth-style Proof of Authority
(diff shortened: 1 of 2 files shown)


fselmo (approved):
Just left a small nit. I'll take cute animal pictures all day (it's a great one btw 👌). Also I think this could use a doc newsfragment even if it's a small change. Let me know if you want me to just add that in and merge @miohtama.

We should also port it over to `v5` branch which I can do as well if you prefer. Let me know!

PR #2545, opened 2022-06-28
Fix TypeError in ContractFunction.call when args and kwargs properties are None

### LINKED ISSUE #2541: ContractFunction.call may fail as args and kwargs properties may be None
* Version: 5.29.2
* Python: 3.10.4
* OS: linux

### What was wrong?

Depending on the way how a `ContractFunction` is instantiated, its `args` and `kwargs` may be `None`, which occurs errors when invoking `call`.

#### Example code

```python
import json
from web3.auto.infura import w3

USDC = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
EIP20_ABI = json.loads('[{"constant":true,"inputs":[],"name":"name","outputs":[{"name":"","type":"string"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"name":"_spender","type":"address"},{"name":"_value","type":"uint256"}],"name":"approve","outputs":[{"name":"","type":"bool"}],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[],"name":"totalSupply","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"name":"_from","type":"address"},{"name":"_to","type":"address"},{"name":"_value","type":"uint256"}],"name":"transferFrom","outputs":[{"name":"","type":"bool"}],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[],"name":"decimals","outputs":[{"name":"","type":"uint8"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"name":"_owner","type":"address"}],"name":"balanceOf","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"symbol","outputs":[{"name":"","type":"string"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"name":"_to","type":"address"},{"name":"_value","type":"uint256"}],"name":"transfer","outputs":[{"name":"","type":"bool"}],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[{"name":"_owner","type":"address"},{"name":"_spender","type":"address"}],"name":"allowance","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"anonymous":false,"inputs":[{"indexed":true,"name":"_from","type":"address"},{"indexed":true,"name":"_to","type":"address"},{"indexed":false,"name":"_value","type":"uint256"}],"name":"Transfer","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"_owner","type":"address"},{"indexed":true,"name":"_spender","type":"address"},{"indexed":false,"name":"_value","type":"uint256"}],"name":"Approval","type":"event"}]')

contract = w3.eth.contract(USDC, abi=EIP20_ABI)
f = contract.get_function_by_signature("decimals()")
f.call()
```

#### Result
```
  File "...", line 14, in <module>
    f.call()
  File ".../venv/lib/python3.10/site-packages/web3/contract.py", line 957, in call
    return call_contract_function(
TypeError: Value after * must be an iterable, not NoneType

```

### How can it be fixed?

The minimal fix may be changing `*self.args` and  `**self.kwargs` to `*(self.args or [])` and `**(self.kwargs or {})` in the following lines: 

https://github.com/ethereum/web3.py/blob/c68f6564ef85c17c5162ce2493d6db16e755516c/web3/contract.py#L1222-L1223

I can work on creating a PR.

iwiwi:
Fix #2541.

I confirmed that the example code in #2541 passes with this change.

If you want me to add related tests in the same PR, as I am new to this project, I would appreciate if you could give some hints, e.g., similar existing tests, where to put tests (a new file or existing file?), etc.

#### Cute Animal Picture
My cats 😺

### DIFF
--- a/web3/contract.py
+++ b/web3/contract.py
@@ -1219,8 +1219,8 @@ class ContractFunction(BaseContractFunction):
             self.abi,
             state_override,
             ccip_read_enabled,
-            *self.args,
-            **self.kwargs,
+            *(self.args or []),
+            **(self.kwargs or {}),
         )
 
     def transact(self, transaction: Optional[TxParams] = None) -> HexBytes:


iwiwi:
Thanks for useful pointers!

https://github.com/ethereum/web3.py/blob/2bf7e9e21879359834decc2551aa9e1388401152/tests/core/contracts/test_contract_ambiguous_functions.py#L179-L183

The difference between my snippet and this test is that, in this test, `ContractFunction.__call__` is called before calling `ContractFunction.call`. That is, if we change L183 from `assert get_value_func().call() == 'Hello'` to `assert get_value_func.call() == 'Hello'`, it will cause the same type errors.

What do you think?

* My opinion is that, even before calling `ContractFunction.__call__`, as it is indeed a `ContractFunction` instance, I think it should be `call`able. So I think we should apply my fix.
* Another option is to consider this as a design. That is, we distinguish two kinds of `ContractFunction`: (1) instances before invoking `__call__` (which we cannot `call`), and (2) instances after invoking `__call__` (which we can `call`). We  may improve error messages and documents.

kclowes:
@iwiwi Thank you for the PR, and sorry it's taken me so long to get to it. I've dug in here a little bit more, and since there is other logic in the `__call__` method that needs to be executed, I think it makes more sense to validate that users have to go through that path. Closing in favor of #3444.

PR #2547, opened 2022-06-28
Async ens no factory

fselmo:
### What was wrong?

Builds on #2501 by @dbfreem 

This time I felt the need to branch off into a separate PR because it is more than just adding tests and organizing. I went a different route and was able to remove the need for `AsyncENSFactory`. This does build on the initial main changes for async ENS but we can keep the discussion on this alternate PR separated.

### How was it fixed?

- Removes the need for `AsyncENSFactory` from #2501 

### Todo:
- [x] Update the newfragment name for the entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

moogcato


ens/utils.py
```diff
@@ -267,32 +270,38 @@ def get_abi_output_types(abi: 'ABIFunction') -> List[str]:
 # -- async -- #
 
 
-async def async_init_web3(
-    provider: "BaseProvider" = cast("BaseProvider", default),
-    middlewares: Optional[Sequence[Tuple["Middleware", str]]] = None
+def init_async_web3(
```
fselmo, line 273: Note: I did not start the method name with `async` on purpose since this method _is not_ `async`... but rather it initializes an async-compatible web3 instance. I'm a bit more used to anything that starts with `async` having to be awaited, which is not the case here.

tests/ens/test_wildcard_resolution.py
```diff
@@ -10,9 +10,37 @@ def test_wildcard_resolution_with_extended_resolver_for_subdomains(ens, subdomai
     assert resolved_child_address == "0x000000000000000000000000000000000000dEaD"
 
 
-def test_wildcard_resolution_with_extended_resolver_for_parent_ens_domain(ens):
+def test_async_wildcard_resolution_with_extended_resolver_for_parent_ens_domain(ens):
```
pacrob, line 13: This should be the regular version and `async` added to the test name on line 39, yes?

fselmo: 💯 ty 👍

pacrob (approved):
one naming nit, otherwise lg!

web3/main.py
```diff
@@ -237,13 +243,13 @@ def toChecksumAddress(value: Union[AnyAddress, str, bytes]) -> ChecksumAddress:
...
         self,
-        provider: Optional[BaseProvider] = None,
+        provider: Optional[Union[BaseProvider, AsyncBaseProvider]] = None,
         middlewares: Optional[Sequence[Any]] = None,
         modules: Optional[Dict[str, Union[Type[Module], Sequence[Any]]]] = None,
         external_modules: Optional[
             Dict[str, Union[Type[Module], Sequence[Any]]]
         ] = None,
-        ens: ENS = cast(ENS, empty),
+        ens: Optional[Union[ENS, AsyncENS]] = None,
```
kclowes, line 252: Should this still be `empty`? or does it get set somewhere else?

kclowes (approved):
LGTM! Just left a few nits. Thank you @dbfreem and @fselmo!

web3/main.py
```diff
@@ -337,21 +343,23 @@ def attach_modules(
...
-        else:
-            return self._ens
+        elif self._ens is cast(AsyncENS, empty) or self.eth.is_async:
+            return AsyncENS.fromWeb3(self)
+
+        return self._ens
 
     @ens.setter
-    def ens(self, new_ens: ENS) -> None:
+    def ens(self, new_ens: Optional[Union[ENS, AsyncENS]]) -> None:
```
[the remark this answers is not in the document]
fselmo: It's a bit tricky to track all the ways this works but this [line leads](https://github.com/ethereum/web3.py/blob/master/ens/utils.py#L78) me to believe that it can be basically `None`, though only internally. Though I also think we can just not explicitly set it to `None` there and it will just be set to `cast(ENS, empty)`.

Any opinion on what you think would be the ideal setup here? 🤔

web3/main.py
fselmo: Thanks for catching this string of typing. This is the main part I struggled to make the linting work then focused on something else and I admit this isn't ideal at all.

I switched a few times between `empty` and `None` but `None` can sometimes get into recursion... again not really sure what the right setup is here. In the end I went with `cast(ENS, empty)` and we can return `AsyncENS` when it's appropriate (when the eth module is async). Thoughts?

web3/main.py
```diff
@@ -352,14 +352,12 @@ def is_encodable(self, _type: TypeStr, value: Any) -> bool:
     @property
     def ens(self) -> Union[ENS, AsyncENS]:
         if self._ens is cast(ENS, empty):
-            return ENS.fromWeb3(self)
-        elif self._ens is cast(AsyncENS, empty) or self.eth.is_async:
-            return AsyncENS.fromWeb3(self)
+            return AsyncENS.fromWeb3(self) if self.eth.is_async else ENS.fromWeb3(self)
```
kclowes, line 355: I like this better, nice!

PR #2562, opened 2022-07-08
Fix is_dynamic_fee_transaction when gas price is zero

VladLupashevskyi:
### What was wrong?

In case gas price strategy returns zero, the `is_dynamic_fee_transaction` is set to True which breaks the logic and makes web3py call dynamic fee tx RPC calls that are not necessary and might not be supported for client.

### How was it fixed?

Explicit check `is None` has been added.

### Todo:


docs/releases.rst
```diff
@@ -16,6 +16,8 @@ Breaking Changes
 Bugfixes
 ~~~~~~~~
 
+- Fix is_dynamic_fee_transaction when gas price is zero (`#2562
```
fselmo, line 19: We don't actually manually add release notes and this change would assume it happened under a beta release that already happened (beta.3), not the next release.

You actually just need to add a brief description of the fix, such as "Fix is_dynamic_fee_transaction when gas price is zero", to a `newsfragments/2562.bugfix.rst` and during the release process this gets added to the notes correctly.

web3/_utils/async_transactions.py
```diff
@@ -115,7 +115,7 @@ async def fill_transaction_defaults(w3: "Web3", transaction: TxParams) -> TxPara
     strategy_based_gas_price = await w3.eth.generate_gas_price(  # type: ignore
         transaction
     )
-    is_dynamic_fee_transaction = not strategy_based_gas_price and (
```
fselmo, line 118: Eesh, good catch, thanks for fixing!

fselmo (changes requested):
Hey @VladLupashevskyi, thanks for catching this. I made a small note about the newsfragment. It would also be nice to add a test. I can take care of the test if you'd like, just let me know.

web3/_utils/async_transactions.py
VladLupashevskyi: No problem :) Also found the same bug in `TRANSACTION_DEFAULTS`

docs/releases.rst
VladLupashevskyi: Done

VladLupashevskyi:
@fselmo 

So after a number of lint fixes, it's ready :)

After introducing a test I noticed that TRANSACTION DEFAULTS also has this `None` issue, so fixed that as well.

web3/_utils/async_transactions.py
```diff
@@ -52,7 +52,10 @@ async def _estimate_gas(w3: "Web3", tx: TxParams) -> Awaitable[int]:
 
 
 async def _gas_price(w3: "Web3", tx: TxParams) -> Awaitable[Optional[Wei]]:
-    return await w3.eth.generate_gas_price(tx) or w3.eth.gas_price  # type: ignore
+    gas_price = await w3.eth.generate_gas_price(tx)  # type: ignore
+    if gas_price is None:
```
fselmo, line 56: See my comment below, I think this can probably be back to a one-liner:

```py
return await w3.eth.generate_gas_price(tx)
```

Looking a bit further I think we really only support `gas_price_strategy` as static values for Async so I also think that we can remove the need to `await` here. I did some quick testing and that does seem to be the case locally but we can create a separate issue to track that since that deviates from this PR. I would be content with just the suggested change from above here for now.

fselmo (changes requested):
Thanks! I've got some more minor changes but I think this can be even simpler unless I'm missing something. See my comments, let me know what you think.

web3/_utils/transactions.py
```diff
@@ -65,11 +66,22 @@
...
+    if gas_price is None:
+        return w3.eth.gas_price
+    return gas_price
+
+
 TRANSACTION_DEFAULTS = {
     "value": 0,
     "data": b"",
     "gas": lambda w3, tx: w3.eth.estimate_gas(tx),
-    "gasPrice": lambda w3, tx: w3.eth.generate_gas_price(tx) or w3.eth.gas_price,
```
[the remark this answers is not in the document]
VladLupashevskyi: Nice catch 👍 I think it would make sense before EIP1559. Right now given the logic in `fill_transaction_defaults` it does not. So I will remove that and convert it to simple lambda expression without `w3.eth.gas_price`

web3/_utils/async_transactions.py
VladLupashevskyi: Not sure about removing `async`, since we are passing tx object inside which might get used in some async calls. So I'm not sure whether it worth removing this convenience, given how simple it is.

P.S.: These are just my thoughts, I have not experienced async part of web3 yet

fselmo (approved):
This looks great, thanks again 👍

Since this is a bugfix we will probably want to port these changes to the `v5` branch as well. I can take care of that if you want unless you want to create a PR against the `v5` branch. Let us know 👀 .

VladLupashevskyi:
Hi @fselmo, happy to help :)

Just created PR with port: #2570

Issue #2537, opened 2022-06-24
Deprecate web3.iban module


pacrob:
### What was wrong?

Ethereum IBAN is a dead standard and can be removed.

### How can it be fixed?

Add deprecation notices to the module so it can be removed in a future version of web3.py.

dbfreem:
could this be deprecated in the v5 branch and removed in v6?

fselmo:
> could this be deprecated in the v5 branch and removed in v6?

That would be ideal 👌

dbfreem:
I couldn't find any documentation saying that IBAN is deprecated. Was thinking about picking this issue up but wanted to make sure I got the correct wording in the deprecation warning. Can you point me to anything that says what replaced IBAN  or is it just  a complete deprecation without a replacement.

fselmo:
> I couldn't find any documentation saying that IBAN is deprecated. Was thinking about picking this issue up but wanted to make sure I got the correct wording in the deprecation warning. Can you point me to anything that says what replaced IBAN or is it just a complete deprecation without a replacement.

I haven't looked much into this. Deferring to @pacrob here... 👀

pacrob:
I think it is a feature that was created but never caught on. It is per @pipermerriam  that it is a "a dead standard that is largely if not entirely unused". I spent a while combing through github and couldn't find any recent code using it or similar features in web3.js, ethers, etc.

It will be complete deprecation without replacement, but I think that it would be better to add deprecation notices in v6 and completely remove later. That will give any folks who might still be using it time to change and/or make noise about it.

fselmo:
> That will give any folks who might still be using it time to change and/or make noise about it.

I lean towards adding deprecation in `v5` and removing in `v6` since we're getting breaking changes in now. I think the buffer between `v5` and `v6` should be enough. We are porting all major changes that aren't breaking over to `v5`... I think that gives users enough time to hang out in `v5` until all warnings are addressed.

pacrob:
I defer to your leanings, deprecate in 5 & eliminate in 6 it is!

fselmo:
closed by #2563 (`v5`) and #2565 (`v6`)... thanks once again @dbfreem 👌

PR #2563, opened 2022-07-09
Deprecate iban v5

dbfreem:
### What was wrong?

Related to Issue #2537 

### How was it fixed?

Deprecated IBAN and added deprecation to test.

### Todo:

### DIFF
--- a/newsfragments/2537.misc.rst
+++ b/newsfragments/2537.misc.rst
@@ -1,0 +1,1 @@
+Deprecated IBAN since it is an unused feature.
--- a/tests/core/eth-module/test_iban.py
+++ b/tests/core/eth-module/test_iban.py
@@ -15,8 +15,9 @@ from web3.exceptions import (
     )
 )
 def test_createIndirect(value, expected, web3):
-    actual = web3.eth.iban.createIndirect(value).toString()
-    assert actual == expected
+    with pytest.warns(DeprecationWarning, match="The IBAN feature has been deprecated in web3.py"):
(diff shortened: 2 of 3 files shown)


dbfreem:
That lint failure is a new one that I haven't seen before in these checks. I got it in my other PR the first time it ran and then on the next run it went away. Looks like it is associated to ipfs_file_pb2 and I couldn't find a relationship to it and my changes. I am thinking this is unrelated. Thoughts??

fselmo:
> That lint failure is a new one that I haven't seen before in these checks. I got it in my other PR the first time it ran and then on the next run it went away. Looks like it is associated to ipfs_file_pb2 and I couldn't find a relationship to it and my changes. I am thinking this is unrelated. Thoughts??

Looks like your branch we pretty far behind `upstream/v5`. I rebased your branch against it and it should use the updated python3.9 image for the lint job image now. Those were errors related to the older python 3.6 image.

fselmo (changes requested):
I do think we can improve the messaging here so users can know when to expect the removal. Maybe something along the lines of:

"The IBAN feature is deprecated and will be removed in v6"

fselmo:
A nit that might've helped before but not sure now since they are all in place is creating the warning once as an internal method, say `_issue_deprecation_warning()`, and calling to it in the different places since the messaging is the same. Take or leave though.

dbfreem:
@fselmo I made the changes you suggested. I did not add the extra method since this is being removed in v6 branch.

Thanks, for noticing I was behind on v5.

PR #2567, opened 2022-07-10
reorganized contract

dbfreem:
### How was it fixed?

Reorganized contract to align with the format discussed in #2441 

### Todo:


web3/contract/__init__.py
```diff
@@ -0,0 +1,10 @@
+from web3.contract.async_contract import (  # noqa: F401
+    AsyncContract,
+    AsyncContractCaller,
+)
+
+from web3.contract.contract import (  # noqa: F401
+    Contract,
+    ContractConstructor,
+    ContractCaller,
+)
```
kclowes, line 10: These are the only classes that get exported to other files in our codebase. I _think_ that our public API all stems from either `Contract` or `AsyncContract`, but I wonder if it's worth exposing other classes like `ContractFunctions` or `ContractEvents` here too?

kclowes (commented):
@fselmo / @pacrob  This is ready for a once-over. I'll go through tomorrow and double check that all methods from the old `contract.py` file were copied over, but I'm reasonably confident that they did since the tests are passing.

web3/contract/__init__.py
pacrob: I would lean towards only exposing what is necessary and add more if when it's called for. Adding things here wouldn't be breaking, right?

pacrob (approved):
LGTM! Nit/question - I've been grouping async functions together in the lower half of a file, separated with `# --- async --- #`. Do we want to continue that theme here? I think it's cleaner in an overall organization sense, but there's merit to having sync and async versions visually next to each other if they need changes in the future. :shrug:

kclowes:
> I've been grouping async functions together in the lower half of a file, separated with # --- async --- #. Do we want to continue that theme here?

oops, yes! Thanks for the reminder!

fselmo (approved):
This lgtm! ~~I'd piggy back what @pacrob said about splitting out the async functions and say that might be useful to do inside the `utils.py` file as well for better readability~~ [oops that's the only file I see for that hah]. But that's a nit.

PR #2569, opened 2022-07-12
Fix false negative for ``fill_transaction_defaults``

fselmo:
### What was wrong?

Back-port of #2562

### How was it fixed?

see #2562 

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #2570, opened 2022-07-13
Fix `is_dynamic_fee_transaction` and `TRANSACTION_DEFAULTS` when `gas_price_strategy` returns zero

VladLupashevskyi:
Port of #2562 to v5

### DIFF
--- a/newsfragments/2570.bugfix.rst
+++ b/newsfragments/2570.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``is_dynamic_fee_transaction`` and ``TRANSACTION_DEFAULTS`` when ``gas_price_strategy`` returns zero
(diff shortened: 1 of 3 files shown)


newsfragments/2562.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix `is_dynamic_fee_transaction` and `TRANSACTION_DEFAULTS` when `gas_price_strategy` returns zero
```
fselmo, line 1: Hey @VladLupashevskyi... this file should be named `2570.bugfix.rst` as it should relate to the PR number for the `v5` branch.

fselmo (changes requested):
Thanks @VladLupashevskyi! I had already created one as #2569 but I'll close that in favor of this one 😎. Just left one comment on the newsfragment.

Issue #2578, opened 2022-07-18
Please restore chain_id setter (#2376)


DefiDebauchery:
In #2376, `chain_id` was added as a setter. This was a cheap but effective cache to avoid constant lookups during txn building, as `simple_cache` is not currently async.

There was a discussion about adding some validation, but it may have been an accident that @dbfreem removed it in his subsequent commits in #2432 (as I didn't see any comments about it in that PR). To me, it seems better to have this, even if temporary in a v6.0 branch, and improve upon it later (or become obsoleted by an async version of simple_cache)

dbfreem:
After looking into this for a little while, looks like @kclowes removed this [here](https://github.com/ethereum/web3.py/commit/3e8f07c14f10f97071de2f93180e0be07a575ffa).

fselmo:
The simple cache middleware has been asynced, via #2579, and will be out in next release.

PR #2579, opened 2022-07-18
Add support for ``async_simple_cache_middleware``

fselmo:
### What was wrong?

Related to Issue #2578

Hard-coding an internal `chain_id` was considered at some point but ultimately we landed on adding `eth_chainId` to the list of whitelisted endpoints for `simple_cache_middleware`. The `chain_id` setter had made its way into the library already and some users had gotten around the multiple `eth_chainId` checks by using this setter, especially with `AsyncEth`.

Rather than introduce this setter back, this PR adds support to cache `eth_chainId` calls using the `async_simple_cache_middleware` with `AsyncEth`.

### How was it fixed?

- Async request caching support via `async_simple_cache_middleware`
- Supports creating custom simple caching middleware via `async_construct_simple_cache_middleware`
- Refactor ``SessionCache`` to a more generic ``SimpleCache`` within the caching `_utils` and use this cache class in the simple cache middleware as the default cache (for sync and async). Leave support for a simple dictionary cache as well.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/request.py
+++ b/web3/_utils/request.py
@@ -1,17 +1,12 @@
 import asyncio
-from collections import (
-    OrderedDict,
-)
 from concurrent.futures import (
     ThreadPoolExecutor,
 )
-import contextlib
 import logging
 import os
 import threading
 from typing import (
     Any,
-    AsyncGenerator,
     Dict,
     List,
     Optional,
@@ -28,7 +23,11 @@ from eth_typing import (
 )
 import requests
 
+from web3._utils.async_caching import (
+    async_lock,
+)
 from web3._utils.caching import (
+    SimpleCache,
     generate_cache_key,
 )
 
@@ -37,44 +36,11 @@ logger = logging.getLogger(__name__)
 DEFAULT_TIMEOUT = 10
 
 
-class SessionCache:
-    def __init__(self, size: int):
-        self._size = size
-        self._data: OrderedDict[str, Any] = OrderedDict()
-
-    def cache(self, key: str, value: Any) -> Dict[str, Any]:
-        evicted_items = None
-        # If the key is already in the OrderedDict just update it
-        # and don't evict any values. Ideally, we could still check to see
-        # if there are too many items in the OrderedDict but that may rearrange
-        # the order it should be unlikely that the size could grow over the limit
-        if key not in self._data:
-            while len(self._data) >= self._size:
-                if evicted_items is None:
-                    evicted_items = {}
-                k, v = self._data.popitem(last=False)
-                evicted_items[k] = v
-        self._data[key] = value
-        return evicted_items
-
-    def get_cache_entry(self, key: str) -> Any:
-        return self._data[key]
-
-    def clear(self) -> None:
-        self._data.clear()
-
-    def __contains__(self, item: str) -> bool:
-        return item in self._data
-
-    def __len__(self) -> int:
-        return len(self._data)
-
-
 def get_default_http_endpoint() -> URI:
     return URI(os.environ.get("WEB3_HTTP_PROVIDER_URI", "http://localhost:8545"))
 
 
-_session_cache = SessionCache(size=100)
+_session_cache = SimpleCache(size=100)
 _session_cache_lock = threading.Lock()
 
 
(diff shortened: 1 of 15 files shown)


web3/middleware/async_middleware/async_cache.py
```diff
@@ -0,0 +1,104 @@
...
+        return middleware
+
+    return async_simple_cache_middleware
+
+
+async def _async_simple_cache_middleware(
+    make_request: Callable[[RPCEndpoint, Any], Any], async_w3: "Web3"
+):
+    middleware = await async_construct_simple_cache_middleware(
+        cache_class=cast(Type[Dict[Any, Any]], functools.partial(lru.LRU, 256)),
```
fselmo, line 102: Interested in feedback on using LRU here vs anything else

web3/middleware/async_middleware/async_cache.py
```diff
@@ -0,0 +1,104 @@
...
+
+    async def async_simple_cache_middleware(
+        make_request: Callable[[RPCEndpoint, Any], Any], async_w3: "Web3"
+    ) -> AsyncMiddleware:
+        cache = cache_class()
+        lock = threading.Lock()
+
+        async def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
+            if method in rpc_whitelist:
+                with lock:
```
fselmo, line 82: Interested on feedback in this main section of the thread locking

web3/middleware/async_middleware/async_cache.py
kclowes: The problem with LRU is that it's not thread safe, IIRC.

kclowes: We do use it with sync, but I believe the async implementation is currently thread safe. We do have some outstanding work to do with the session cache, though to get it to work just right. See #2446 and the corresponding PR (#2409 I think)

dbfreem: The other issue with using LRU with an async method is that it will not await the callback that it calls when it evicts an item. That was really the main reason that I wrote `SessionCache` in request.py for what it is worth. So if you are going to be using a callback with the above it will just return a coroutine (if I remember correctly) instead of executing the async callback. 

If you really want to you could pull SessionCache out of request.py and put it somewhere in _utils then use that, but that would be a decent amount of work. The SessionCache itself is solid, the main issue still outstanding in request.py is how and where we await the close of an evicted async item.

web3/middleware/async_middleware/async_cache.py
dbfreem: This is the exact issue I currently have in request.py with async and thread locking. The below await will release the execution of this method for something else to work while you are still in the lock. That will eventually cause a deadlock.

fselmo: Yeah I think I get it a bit more now. Thanks for the explanation. I think the first step here is probably writing a good test that fails, with the situation you described, and work to make it not fail. I'll see if I can whip something up this week but I'm out for half of the week. Realistically probably won't get to this until next week.

dbfreem: I was able to use the POC in #2446 to pretty reliably reproduce the request.py issue, so maybe you could use something similar.

fselmo:
Factored `SessionCache` out as a the more generic `SimpleCache` internal cache class. Used this as the default cache for the simple cache middleware while keeping dictionary-based cache support.

web3/_utils/request.py
```diff
@@ -37,44 +36,11 @@
...
-    def __len__(self) -> int:
-        return len(self._data)
-
-
 def get_default_http_endpoint() -> URI:
     return URI(os.environ.get("WEB3_HTTP_PROVIDER_URI", "http://localhost:8545"))
 
 
-_session_cache = SessionCache(size=100)
+_session_cache = SimpleCache(size=100)
```
kclowes, line 43: nit: it defaults to 100 already

web3/_utils/request.py
```diff
@@ -145,9 +111,9 @@ def _close_evicted_sessions(evicted_sessions: List[requests.Session]) -> None:
 # --- async --- #
 
 
-_async_session_cache = SessionCache(size=100)
+_async_session_cache = SimpleCache(size=100)
```
kclowes, line 114: nit: same here

kclowes (approved):
I had a couple nits, but looks good once we pull out the dictionary cache! :shipit:

pacrob (approved):
lgtm!

fselmo:
> I had a couple nits, but looks good once we pull out the dictionary cache! :shipit:

Updated... I'm not quite sure how to add bullets to the feature newsfragment 🤔. Other than that this should be ready for a quick re-review w/ removal of the dictionary-based cache support.

kclowes (approved):
:shipit: ! Sweet!

PR #2587, opened 2022-07-26
Classes inheriting from ``Web3`` did not attach modules appropriately

### LINKED ISSUE #2586: Web3 instantiation from subclass throws error
* Version: 5.30.0
* Python: 3.8
* OS: osx
* pip freeze output is intentionally omitted.

### What was wrong?
We use a sub-class of Web3, for the purpose of this example, call it SubclassWeb3, as follows:

```
class SubclassWeb3(Web3):
def __init__(self, ....):
  ... do bunch of stuff...
  super().__init__(*args, **kwargs)
  ... do bunch more stuff...
```

This setup worked fine with web3 version == 5.23.1. However, switching to web3==5.30.0 breaks this, with error

```
[rest of callstack omitted]
 File “[redacted]/web3/main.py”, line 251, in __init__
    self.attach_modules(modules)
  File “[redacted]/web3/main.py”, line 340, in attach_modules
    _attach_modules(self, modules)
  File “[redacted]/web3/_utils/module.py”, line 68, in attach_modules
    setattr(parent_module, module_name, module_class(w3))
  File “[redacted]/web3/module.py”, line 93, in __init__
    self.codec: ABICodec = web3.codec
AttributeError: ‘NoneType’ object has no attribute ‘codec’
```

This used to work in older versions because the parent_module is assumed to be a Web3 object and stored accordingly - [see logic](https://github.com/ethereum/web3.py/blob/2b229d2e16792e52bb47108e665c6a8937c188a2/web3/_utils/module.py#L33-L35).

Compare this to the somewhat brittle name checking that happens in 5.30.0 - [see logic](https://github.com/ethereum/web3.py/blob/ee5f0543ad91e4ba991997e35f50719ceb58135d/web3/_utils/module.py#L58-L60). This doesn't quite work because if this logic is being executed from a subclass of Web3, the name of the parent module appears as SubclassWeb3 (and not Web3).

### How can it be fixed?

Fill this section in if you know how this could or should be fixed.

fselmo:
### What was wrong?

closes #2586 

### How was it fixed?

* Import ``Web3`` locally in order to avoid circular import. This allows us to make the check we wanted to in the first place which is to check that the ``parent_module`` is an instance of the ``Web3`` class (or inherits from it) when ``w3`` is ``None``.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_implicit_contract.py
+++ b/tests/core/contracts/test_implicit_contract.py
@@ -70,13 +70,13 @@ def test_implicitcontract_transact_default(web3, math_contract, get_transaction_
     assert is_integer(start_count)  # Verify correct type
     # When a function is called that defaults to transact
     blocknum, starting_txns = get_transaction_count("pending")
-    with pytest.warns(DeprecationWarning,
-                      match='deprecated in favor of classic contract syntax') as warnings:
+    with pytest.warns(
+        DeprecationWarning,
+        match='deprecated in favor of classic contract syntax'
+    ):
         math_contract.increment(transact={})
         # Check that a transaction was made and not a call
         assert math_contract.counter() - start_count == 1
-        # Check that the correct number of warnings are raised
-        assert len(warnings) == 2
     # (Auto-mining is enabled, so query by block number)
     assert get_transaction_count(blocknum) == starting_txns + 1
     # Check that only one block was mined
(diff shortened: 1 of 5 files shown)


tests/core/contracts/test_implicit_contract.py
```diff
@@ -70,13 +70,13 @@ def test_implicitcontract_transact_default(web3, math_contract, get_transaction_
     assert is_integer(start_count)  # Verify correct type
     # When a function is called that defaults to transact
     blocknum, starting_txns = get_transaction_count("pending")
-    with pytest.warns(DeprecationWarning,
-                      match='deprecated in favor of classic contract syntax') as warnings:
+    with pytest.warns(
```
fselmo, line 73: These tests were looking for a particular number of warnings which can't be sustainable as our dependencies will eventually issue warnings (as `eth-abi` did which broke these tests).

pacrob (approved):
lgtm!

PR #2592, opened 2022-08-01
 Classes inheriting from Web3 did not attach modules appropriately

fselmo:
### What was wrong?

`v6` version of PR #2587

### How was it fixed?

* Bug fix: On the first run of ``attach_modules()``, ``w3`` should be ``None`` and the ``parent_module`` is the ``Web3`` class. We should look for this condition and set ``w3 = parent_module``. We couldn't use a simple ``isinstance()`` due to a circular import, but we can import it locally if ``w3`` is ``None``. This allows classes to inherit from ``Web3`` without needing to validate that the class name is ``Web3``. This also allows modules to not need to set a ``w3`` if they don't need one, as the older code was intended to do.

### Todo:

#### Cute Animal Picture

Screen Shot 2022-08-01 at 09 39 43

### DIFF
--- a/newsfragments/2592.bugfix.rst
+++ b/newsfragments/2592.bugfix.rst
@@ -1,0 +1,1 @@
+Allow classes to inherit from the ``Web3`` class by attaching modules appropriately.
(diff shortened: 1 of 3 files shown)


pacrob (approved):
lgtm!

kclowes (approved):
LGTM!

PR #2600, opened 2022-08-08
fix bug in how eth_tester_middleware filled default fields

pacrob:
### What was wrong?

function `async_default_transaction_fields_middleware` was using some non-asynced helper functions, so it wouldn't fill properly

### How was it fixed?

added async versions of the helper functions

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/middleware/test_eth_tester_middleware.py
+++ b/tests/core/middleware/test_eth_tester_middleware.py
@@ -1,8 +1,22 @@
 import pytest
+from unittest.mock import (
+    Mock,
+)
 
+from web3.providers.eth_tester.middleware import (
+    async_default_transaction_fields_middleware,
+    default_transaction_fields_middleware,
+)
 from web3.types import (
     BlockData,
 )
+
+SAMPLE_ADDRESS_LIST = [
+    "0x0000000000000000000000000000000000000001",
+    "0x0000000000000000000000000000000000000002",
+    "0x0000000000000000000000000000000000000003",
+]
+SAMPLE_ADDRESS = "0x0000000000000000000000000000000000000004"
 
 
 @pytest.mark.parametrize("block_number", {0, "0x0", "earliest"})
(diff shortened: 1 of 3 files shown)


fselmo (commented):
I haven't looked too deep here yet but right away I returned all of the code in `web3/providers/eth_tester/middleware.py` to what it is in `master` and all the tests pass, including the ones added in this PR.

Let's try to create a test (or a few) that fail(s) _without_ the changes in the PR but passes with the changes.

pacrob:
Updated tests so they now fail on the old code. I tried various ways of patching the w3 instance but couldn't make anything work, so went with Mock instead.

One question/point for discussion - the `guess_from` functions try to fill the `from` field, but don't make any noise if they can't. Should they?

web3/providers/eth_tester/middleware.py
```diff
@@ -340,6 +346,18 @@ def fill_default(
         return assoc(transaction, field, guess_val)
 
 
+@curry
+async def async_fill_default(
+    field: str, guess_func: Callable[..., Any], w3: "Web3", transaction: TxParams
```
fselmo, line 351: nit: `w3` -> `async_w3` for consistency

web3/providers/eth_tester/middleware.py
```diff
@@ -363,15 +381,17 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
 
 
 async def async_default_transaction_fields_middleware(
-    make_request: Callable[[RPCEndpoint, Any], Any], web3: "Web3"
+    make_request: Callable[[RPCEndpoint, Any], Any], w3: "Web3"
```
fselmo, line 384: nit: `w3` -> `async_w3` for consistency

tests/core/middleware/test_eth_tester_middleware.py
```diff
@@ -20,3 +27,157 @@ def test_get_block_formatters(w3):
...
+    filled_params = filled_transaction[0]
+
+    assert ("from" in filled_params.keys()) == from_field_added
+    if "from" in filled_params.keys():
+        assert filled_params["from"] == from_field_value
+
+    filled_transaction[0].pop("from", None)
+    assert filled_transaction[0] == base_params
+
+
```
fselmo, line 102: nit: I've been separating async from sync tests with:

```py
def test_last_sync_test_here():
    ...
    

# -- async -- #


@pytest.mark.asyncio
async def test_async_first_async_test_here():
    ...
```

With sync and async tests that exist in the same file I'd like to organize them somehow for quicker / easier readability. Since that convention has been used in quite a few places already I'm partial to it 😄 .

tests/core/middleware/test_eth_tester_middleware.py
```diff
@@ -20,3 +27,157 @@ def test_get_block_formatters(w3):
     keys_diff = all_block_keys.difference(latest_block_keys)
     assert len(keys_diff) == 1
     assert keys_diff.pop() == "mixHash"  # mixHash is not implemented in eth-tester
+
+
+sample_address_1 = "0x0000000000000000000000000000000000000001"
```
fselmo, line 32: nit: Let's have these as capitalized and near the top of the file just to stick with python convention, especially since sync and async both use these values.

```py
from some_module import last_import_statement

SAMPLE_ADDRESS_1 = "0x0000000000000000000000000000000000000001"
SAMPLE_ADDRESS_2 = "0x0000000000000000000000000000000000000002"
```

tests/core/middleware/test_eth_tester_middleware.py
```diff
@@ -20,3 +27,157 @@ def test_get_block_formatters(w3):
...
+        (None, None, "eth_sendTransaction", True, None),
+        (None, None, "eth_gasPrice", False, None),
+        (None, None, "eth_blockNumber", False, None),
+        (None, None, "meow", False, None),
+    ),
+)
+def test_default_transaction_fields_middleware(
+    w3_accounts, w3_coinbase, method, from_field_added, from_field_value
+):
+    w3_accounts = [w3_accounts]
```
fselmo, line 80: Instead of re-defining the variable here, I think we should either make the mock return the list with a variable more appropriately named `w3_account` (singular):

```py
mock_w3.eth.accounts = [w3_account]  # and the variable is renamed to w3_account (singular)

# async:
async def mock_async_accounts():
    return [w3_account]  # singular
```

or... we can just pass in a list of accounts into the tests themselves and keep the variable as `w3_accounts` (plural). I vote this option and we can change one of the tests to return more than one account in the list to make sure that works as intended as well... I believe it always returns `accounts[0]`.

fselmo (approved):
Nice catch and fix 👌🏼 ... I had a few nits, mostly just on conventions. I'll approve since the fix is in there though 👍🏼. I'm good with the mocking too.

As far as not making any noise when it cannot fill the fields... it looks like we purposefully return `None` if that's the case. I'd leave it for now as this seems more of a helper middleware, to make communicating with `eth-tester` a bit easier, than it seems a validation type middleware.

web3/providers/eth_tester/middleware.py
pacrob: done

web3/providers/eth_tester/middleware.py
pacrob: done

tests/core/middleware/test_eth_tester_middleware.py
pacrob: check

tests/core/middleware/test_eth_tester_middleware.py
pacrob: check

tests/core/middleware/test_eth_tester_middleware.py
pacrob: Makes sense to have the mock return what the function normally would. I've updated the sample address constants accordingly. It will still only ever choose index 0 of the list of addresses though, as that's how the `guess_from` function is written.

kclowes (approved):
LGTM! 🚢

PR #2602, opened 2022-08-10
Change validate_payable to allow zero hexadecimal transaction["value"]

mireynolds:
### What was wrong?

validate_payable was raising ValidationError exception on passing a transaction["value"] of zero in hexadecimal format.

### How was it fixed?

Changed if statement to check if transaction["value"] is not in [0, "0x0] instead of checking for inequality with zero.

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #2609, opened 2022-08-12
[v5] Allow hex for ``value`` when validating via ``validate_payable()``

fselmo:
### What was wrong?

`v5` port of `v6` PR #2602 

### How was it fixed?

- Use `to_integer_if_hex()` to appropriately convert hex values before comparison

### Todo:


pacrob (approved):
lgtm!

PR #2612, opened 2022-08-16
asynced Filter and EventFilterBuilder classes

pacrob:
### What did you do?

Async versions of Filter and EventFilterBuilder classes

#### Cute Animal Picture

image

### DIFF
--- a/web3/_utils/filters.py
+++ b/web3/_utils/filters.py
@@ -41,6 +41,7 @@ from hexbytes import (
 )
 
 from web3._utils.events import (
+    AsyncEventFilterBuilder,
     EventFilterBuilder,
     construct_event_data_set,
     construct_event_topic_set,
@@ -62,6 +63,7 @@ from web3.types import (
 if TYPE_CHECKING:
     from web3 import Web3  # noqa: F401
     from web3.eth import Eth  # noqa: F401
+    from web3.eth import AsyncEth  # noqa: F401
 
 
 def construct_event_filter_params(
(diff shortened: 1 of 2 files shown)


web3/_utils/filters.py
```diff
@@ -233,6 +264,48 @@ def is_valid_entry(self, entry: LogReceipt) -> bool:
         return bool(self.data_filter_set_function(entry["data"]))
 
 
+class AsyncLogFilter(AsyncFilter):
```
kclowes, line 267: Since this is a direct copy/paste from the `LogFilter` class, I wonder if we can get condense it into one class? I see the inheritance there for `AsyncFilter`. The other thing I was thinking that I'm not sure will work is to have an `is_async` flag (or something) on the class that calls the appropriate async or sync method, depending on the flag. We can also figure that out later in this asyncify-filters process

kclowes (commented):
Sweet! The ball is rolling! Are there any tests that should be added here or is it too early? I can't remember if we have tests that test the class separately or if they all go through `w3.eth.filter`.

pacrob:
> Are there any tests that should be added here or is it too early?

I found no tests for these classes directly. I tried to set them up, but it involved using `w3.eth.filter` for `Filter` and classes from `eth-abi` for `EventFilterBuilder`. It appears that they'll be involved in tests for `eth` and `contract` modules. I'll make a note to verify when I get there.

web3/_utils/filters.py
pacrob: I originally tried to compress them into one (like `class LogFilter(Filter, AsyncFilter) and fell down a mypy wormhole. Maybe having a `BaseLogFilter`? Will try

pacrob: As discussed, no immediate solutions present themselves. Will keep it in mind as asyncing filters moves forward

kclowes (approved):
Thanks for looking into it!

PR #2615, opened 2022-08-22
remove docs reference to removed eth-account recoverHash method

pacrob:
### What was wrong?

Deprecated `eth-account` methods are being removed. This removes a reference to the `recoverHash` method that is among them.

Related to https://github.com/ethereum/eth-account/pull/195

### How was it fixed?

Deleted related docs section

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/docs/web3.eth.account.rst
+++ b/docs/web3.eth.account.rst
@@ -168,23 +168,6 @@ With the original message text and a signature:
     >>> w3.eth.account.recover_message(message, signature=signed_message.signature)
     '0x5ce9454909639D2D17A3F753ce7d93fa0b9aB12E'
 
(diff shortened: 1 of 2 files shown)


kclowes (approved):
LGTM!

0xfffffffa:
What was recoverHash replaced with? Ugpraded library, everything is breaking now.

pacrob:
> What was recoverHash replaced with? Ugpraded library, everything is breaking now.

The removal in `eth-account` was done here: https://github.com/ethereum/eth-account/pull/195/files

Looks like `recover_message` is the replacement. I should work similarly, but requires a SignableMessage instead of just a string.

PR #2616, opened 2022-08-23
update beacon api to 2.3.0

wolovim:
### What was wrong?

Web3.py's Beacon API is lagging behind the latest spec.

### How was it fixed?

- Smol fixes: a `v1`->`v2` bump and testing always-in-range values.
- These smoke/sanity tests are still not integrated into CI. Separate effort required to first decide how we'd want to introduce a Beacon node into the test infrastructure. Local testing performed with a lighthouse node on the `sepolia` network, e.g. by running `lighthouse b --http --http-port 5051 --network sepolia`, then executing the tests.
- Removed some `eth2` references and replaced with consensus layer (`cl`)
- Removed the 'experimental' warning.

### Todo:

### DIFF
--- a/docs/web3.beacon.rst
+++ b/docs/web3.beacon.rst
@@ -1,5 +1,5 @@
-Eth 2.0 Beacon API
-=======================
+Beacon API
+==========
 
 .. warning:: This API Is experimental. Client support is incomplete and the API itself is still evolving.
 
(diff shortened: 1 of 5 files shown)


kclowes (approved):
Woohoo! v2! Like we talked about, we can kick the testing down the road for now.

PR #2626, opened 2022-08-31
update all references to deprecated eth_abi.encode_abi to eth_abi.encode

pacrob:
### What was wrong?

We were still calling the deprecated method `eth_abi.encode_abi`.

Related to Issue #2621 
Closes #2621

### How was it fixed?

 Changed all calls to the new `eth_abi.encode`.

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

image

Issue #2621, opened 2022-08-25
eth_abi deprecation warning for abi.encode_abi


Polsaker:
* Version: 6.0.0b4
* Python: 3.10
* OS: osx

### What was wrong?

Contract calls raise a `DeprecationWarning` due to `encode_abi` being renamed to `encode` in `eth_abi`:

```
    return self.contract.functions.decimals().call()
[snip]/lib/python3.10/site-packages/web3/contract.py:1228: in call
    return call_contract_function(
[snip]/lib/python3.10/site-packages/web3/contract.py:1982: in call_contract_function
    call_transaction = prepare_transaction(
[snip]/lib/python3.10/site-packages/web3/_utils/contracts.py:263: in prepare_transaction
    prepared_transaction["data"] = encode_transaction_data(
[snip]/lib/python3.10/site-packages/web3/_utils/contracts.py:303: in encode_transaction_data
    return add_0x_prefix(encode_abi(w3, fn_abi, fn_arguments, fn_selector))
[snip]/lib/python3.10/site-packages/web3/_utils/contracts.py:218: in encode_abi
    encoded_arguments = w3.codec.encode_abi(
-------
    def encode_abi(self, types: Iterable[TypeStr], args: Iterable[Any]) -> bytes:
        """
        Encodes the python values in ``args`` as a sequence of binary values of
        the ABI types in ``types`` via the head-tail mechanism.

        :param types: An iterable of string representations of the ABI types
            that will be used for encoding e.g.  ``('uint256', 'bytes[]',
            '(int,int)')``
        :param args: An iterable of python values to be encoded.

        :returns: The head-tail encoded binary representation of the python
            values in ``args`` as values of the ABI types in ``types``.
        """
>       warnings.warn(
            "abi.encode_abi() and abi.encode_abi_packed() are deprecated and will be "
            "removed in version 4.0.0 in favor of abi.encode() and "
            "abi.encode_packed(), respectively",
            category=DeprecationWarning,
        )
E       DeprecationWarning: abi.encode_abi() and abi.encode_abi_packed() are deprecated and will be removed in version 4.0.0 in favor of abi.encode() and abi.encode_packed(), respectively
```

### How can it be fixed?

Call `encode` instead of `encode_abi` as directed by the `eth_abi` warning.

---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
[...]

PR #2636, opened 2022-09-08
Renaming leftover references to deprecated `eth_abi.decode_abi` to `eth_abi.decode`

Polsaker:
### What was wrong?

`eth_abi.decode_abi` was renamed to `eth_abi.decode` and the original method was deprecated. #2626 already addressed the calls to `eth_abi.encode_abi`

Related to Issue #2621

### How was it fixed?

Replaced all instances of `eth_abi.decode_abi` with `eth_abi.decode`

### Todo:

#### Cute Animal Picture

### DIFF
--- a/ens/base_ens.py
+++ b/ens/base_ens.py
@@ -93,7 +93,7 @@ class BaseENS:
     ) -> Any:
         func = extended_resolver.get_function_by_name(fn_name)
         output_types = get_abi_output_types(func.abi)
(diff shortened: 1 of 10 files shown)


pacrob:
Woo, thanks for catching those! Not sure why I missed them. I'll read it over and look at those test timeouts, but looks good so far.

kclowes:
@pacrob the timeouts are unrelated to this PR. I'm going to do some digging in. We think it may be from the new geth fixture.

PR #2638, opened 2022-09-08
Chore: address test timeouts

fselmo:
### What was wrong?

- We have inconsistent test timeouts that are leading to many failed test runs. Tests using `unlock_account` and `unlock_account_dual_type` pytest fixtures are a known source of timeout issues.

### How was it fixed?

- Split flaky tests, starting with `unlocked_account` and `unlocked_account_dual_type` fixture tests into their own test run / shorter CI job. It does seem to have many continuous successful runs.
- Add some notes around integration test config so we don't start to clutter those conditionals and risk skipping some tests.

### Todo:


pacrob (approved):
lgtm!

kclowes (approved):
🚀 !

PR #2645, opened 2022-09-12
allow eth.filter to return appropriate sync/async filter types

pacrob:
Calls to AsyncEth.filter were returning sync filter types

### How was it fixed?

Updated associated plumbing, added testing

#### Cute Animal Picture

image

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -769,19 +769,21 @@ def filter_wrapper(
 ]:
     if method == RPC.eth_newBlockFilter:
         if module.is_async:
-            return AsyncBlockFilter(filter_id, eth_module=cast(AsyncEth, module))
+            return AsyncBlockFilter(filter_id, eth_module=cast("AsyncEth", module))
         else:
-            return BlockFilter(filter_id, eth_module=cast(Eth, module))
+            return BlockFilter(filter_id, eth_module=cast("Eth", module))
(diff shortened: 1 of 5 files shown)


kclowes (commented):
`Module` has a property called `is_async` already. I _think_ we could use that to differentiate instead of adding another `is_async` property on `Method`. 

Also, I haven't tested this, but it looks like you could also combine the `filter_wrapper` with `async_filter_wrapper` if you wanted to. That's more of a stylistic thing though :) Otherwise, this looks good!

web3/_utils/method_formatters.py
```diff
@@ -754,42 +754,87 @@ def raise_transaction_not_found_with_index(
...
+) -> Union[
+    AsyncBlockFilter,
+    AsyncTransactionFilter,
+    AsyncLogFilter,
+    BlockFilter,
+    TransactionFilter,
+    LogFilter,
+]:
+    
+    if type(module).__name__ == "Eth":
```
kclowes, line 789: Sorry if this is just a wip and you don't want feedback, but could you check for `module.is_async` here?

pacrob: No problem - feedback always welcome, especially if I've asked for review. Will try it.

pacrob: Ok, updated, ready for re-review.

kclowes (approved):
Nice! LGTM!

PR #2648, opened 2022-09-14
Remove dep upper pins

kclowes:
### What was wrong?
We're moving to allow a wider set of dependencies. This gives the power to the user to specify the versions they need without getting in the way. See [this article](https://iscinumpy.dev/post/bound-version-constraints/) with the referenced links for justification.

### How was it fixed?
Removed all upper pins on dependencies that won't break. 

### Todo:

#### Cute Animal Picture


setup.py
```diff
@@ -75,22 +75,22 @@
...
     install_requires=[
-        "aiohttp>=3.7.4.post0,<4",
-        "eth-abi>=3.0.0,<4.0.0",
-        "eth-account>=0.7.0,<0.8.0",
-        "eth-hash[pycryptodome]>=0.2.0,<1.0.0",
-        "eth-typing>=3.0.0,<4.0.0",
-        "eth-utils>=2.0.0,<3.0.0",
-        "hexbytes>=0.1.0,<1.0.0",
+        "aiohttp>=3.7.4.post0",
+        "eth-abi>=3.0.0",
```
fselmo, line 79: We can probably stick this to `3.0.1` already maybe?

kclowes: Since `web3.py` doesn't use any new features in `eth-abi 3.0.1` I think it might be better to leave it as open as possible.

fselmo: sgtm 👍🏼

pacrob (approved):
lgtm!

PR #2652, opened 2022-09-14
Add new predefined block identifiers ``safe`` and ``finalized``.

### LINKED ISSUE #2641: Add new predefined block identifiers
### What was wrong?

There are [new predefined block identifiers](https://github.com/ethereum/execution-apis/blob/830fcd8768ef46353c8a0e3b8df85af421026a98/src/schemas/block.yaml#L97-L98), "finalized" and "safe" that should be added to the current {"latest", "pending", "earliest"} set.

### How can it be fixed?

Search the codebase and add them in where the other predefined block numbers are. Should also add tests to make sure everything works as expected.

fselmo:
### What was wrong?

Related to #2641

### How was it fixed?

- Added new block identifiers `safe` and `finalized` as options. `xfail`ed tests for now until we migrate our test suite to PoS.
- For filters, added appropriate documentation stating that, for `eth_newFilter`, the new block identifiers are not yet in the specifications so they may or may not yield expected results across nodes.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -15,7 +15,7 @@ Filtering
 
         from web3.auto import w3
 
-The :meth:`web3.eth.Eth.filter` method can be used to setup filters for:
+The :meth:`web3.eth.Eth.filter` method can be used to set up filters for:
 
 * Pending Transactions: ``web3.eth.filter('pending')``
 
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -107,6 +107,11 @@ will return a new :class:`BlockFilter` object.
         new_block_filter = w3.eth.filter('latest')
         new_block_filter.get_new_entries()
 
+    .. note::
+
+        ``"safe"`` and ``"finalized"`` block identifiers are not yet supported for
+        ``eth_newBlockFilter``.
+
 .. py:class:: TransactionFilter(...)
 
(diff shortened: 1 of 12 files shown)


pacrob (approved):
lgtm

kclowes (commented):
I'm not sure if you want feedback on this right at the moment, but it looks like there is a predefined block number test [here](https://github.com/ethereum/web3.py/blob/master/tests/core/utilities/test_is_predefined_block_number.py#L18) that could use the new predefined block names. I also dug in a little bit more, and I'm pretty sure that we do want to accept these block names to the filter method. Looks good otherwise!

fselmo:
> I'm not sure if you want feedback on this right at the moment, but it looks like there is a predefined block number test [here](https://github.com/ethereum/web3.py/blob/master/tests/core/utilities/test_is_predefined_block_number.py#L18) that could use the new predefined block names. I also dug in a little bit more, and I'm pretty sure that we do want to accept these block names to the filter method. Looks good otherwise!

Yeah it looks like event filters may or may not take `safe` and `finalized` in the `fromBlock` and `toBlock`. Since they aren't yet in the specifications, I added support for them but I updated the documentation with notes that they may or may not yield expected results.

wolovim (approved):
looks sane to me, based on current support for these filters in clients 👍

PR #2668, opened 2022-10-06
Clean up old eth abi references

fselmo:
### What was wrong?

Related to Issue ethereum/eth-abi#161

### How was it fixed?

- Use ``abi.decode()`` over the deprecated ``abi.decode_single()``
- Since "encode" and "decode" are common across many libraries, import ``abi`` and use ``abi.encode()`` and ``abi.decode()`` so a reader can have more context when reading the code.

### Todo:

### DIFF
--- a/newsfragments/README.md
+++ b/newsfragments/README.md
@@ -5,15 +5,17 @@ release notes. This should be a description of aspects of the change
 commit message and PR description, which are a description of the change as
 relevant to people working on the code itself.)
 
- Each file should be named like `<ISSUE>.<TYPE>.rst`, where
+Each file should be named like `<ISSUE>.<TYPE>.rst`, where
 `<ISSUE>` is an issue number, and `<TYPE>` is one of:
 
 * `feature`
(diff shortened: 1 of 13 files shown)


newsfragments/README.md
```diff
@@ -5,15 +5,17 @@ release notes. This should be a description of aspects of the change
...
 `<ISSUE>` is an issue number, and `<TYPE>` is one of:
 
 * `feature`
 * `bugfix`
+* `performance`
 * `doc`
+* `internal`
+* `removal`
 * `misc`
-* `breaking-change`
```
fselmo, line 15: @kclowes any reason we'd want to keep these different for `web3.py`? I needed the `internal` and so I just copy / pasted from the template project. Makes things easier.

kclowes: Nope, just hadn't gotten around to changing it. Thanks!

kclowes (approved):
LGTM. I like the changes from `encode` to `abi.encode`!

Issue #2598, opened 2022-08-08
Snake case remaining main API methods and properties


wolovim:
### What was wrong?

In `web3/main.py` there are a few properties and methods that have managed to avoid snake casing so far.

### How can it be fixed?

- [x] encoding/decoding: `toBytes`, etc.
- [x] currency utilities: `toWei`, `fromWei`
- [x] address utilities: `isAddress`, etc.
- [x] `clientVersion`
- [x] `solidityKeccak`
- [x] `isConnected`
- [x] ENS `fromWeb3`
- [x] contract.py: 
   - [x] `BaseContractEvent`: `processReceipt`, `processLog`, `createFilter` + kwargs in `createFilter`
   - [x] `ContractEvent` + `AsyncContractEvent`: `getLogs`

linda-le1:
I can get going on snake casing `clientVersion`.

linda-le1:
Will get going on `solidityKeccak` next.

linda-le1:
Working on `fromWeb3` next

linda-le1:
Working on snakecasing the encoding/decoding helpers (toBytes, toHex, toInt, toJSON, and toText)

linda-le1:
Working on snakecasing address helpers (`isAddress`, `isChecksumAddress`,  and`toChecksumAddress`) next

linda-le1:
Working on snakecasing the `contract.py` methods listed

kclowes:
I believe this has been completed! 🎉 Thanks @linda-le1!

PR #2686, opened 2022-10-20
Snakecase clientVersion

linda-le1:
### What was wrong?
Inherited javascript syntax. This PR does not include a deprecation step, intended for v6.
Related to https://github.com/ethereum/web3.py/issues/2598

### How was it fixed?
Changed clientVersion to client_version where needed

### Todo:

#### Cute Animal Picture

shiba


tests/core/web3-module/test_clientVersion.py
```diff
@@ -1,2 +0,0 @@
-def test_web3_clientVersion(w3):
-    assert w3.clientVersion.startswith("EthereumTester/")
```
kclowes, line 2: I may have just missed it, but it looks like this file got deleted instead of moved?

kclowes (commented):
Nice work! 💪 🎉  Looks like there is a quick lint check that needs to be addressed, and a newsfragment is missing. I think most of us have `black` running on save, which _should_ take care of most lint errors. 

The newsfragment will be a new file called `newsfragments/2686.breaking.rst` and can just contain a sentence with what changed. 

Thank you!

tests/core/web3-module/test_clientVersion.py
linda-le1: Thanks for catching this! Just forgot to push up my latest branch where I renamed this file.

linda-le1:
@kclowes question on the linter: I ran black over the files in my PR and nothing seemed to change. Did I maybe miss something?

kclowes:
Ah, sorry. `black` doesn't do anything with strings. You can either add a `#noqa: E501` at the end of the line, or split the string into two lines. Then we should be good to go!

kclowes (approved):
Looks great! Thank you!

PR #2687, opened 2022-10-20
Add async get_block_tx_count

kclowes:
### What was wrong?
Missing async `w3.eth.get_block_transaction_count`

### How was it fixed?
Added it!

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/eth.py
+++ b/web3/eth.py
@@ -310,6 +310,19 @@ class BaseEth(Module):
         RPC.eth_getTransactionReceipt, mungers=[default_root_munger]
     )
 
+    """
+    `eth_getBlockTransactionCountByHash`
+    `eth_getBlockTransactionCountByNumber`
+    """
+    get_block_transaction_count: Method[Callable[[BlockIdentifier], int]] = Method(
+        method_choice_depends_on_args=select_method_for_block_identifier(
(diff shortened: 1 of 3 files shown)


pacrob (approved):
lgtm!

web3/eth.py
```diff
@@ -530,6 +530,15 @@ async def get_code(
     ) -> HexBytes:
         return await self._get_code(account, block_identifier)
 
+    get_block_transaction_count: Method[Callable[[BlockIdentifier], int]] = Method(
```
fselmo, line 533: This looks like a copy of the sync property with the same name... is there a reason we can't move this to `BaseEth` and keep one copy? 🤔

kclowes: I was just looking at that 👀 this PR triggered the thought I had about standardizing `AsyncEth` and `Eth`, which led to the typing conversation... We can just put this in `BaseEth` for now, but I think it will need to be moved/expanded on when we figure out what we want to do about our types.

fselmo: Yeah that makes sense 👍🏼. I'd be good with moving it to `BaseEth` for now while we explore the refactoring a bit more.

PR #2706, opened 2022-11-02
Remove auto provider except for local geth dev

### LINKED ISSUE #1736: Autoprovider repo template
### What was wrong?

Currently, you can access an Infura provider in web3 using `from web3.auto.infura import w3`. This was good at the time because Infura was the only viable provider. However, a while back someone submitted a PR to add Nodesmith as an auto provider. After some internal discussion, we decided that we didn't want to be the gatekeeper for every new node provider, and decided that we would use Infura as an example to make a new template that people can follow to implement their own auto node providers. 

### How can it be fixed?

On a high level, pull out the logic for Infura's auto provider, put it in a new repo, and provide a web3 API for people to use their own auto provider package.

fselmo:
### What was wrong?

Related to issue #1416
closes #1736?

### How was it fixed?

- We decided the overhead of setting up your own provider is not much and is ultimately more transparent for the user. 
- Keep around ``geth --dev`` auto provider as that connects locally and not to any external provider service.
 
### Todo:
- [x] ~~Consider dropping the ``geth --dev`` auto provider altogether?~~ [keeping it]


pacrob:
As discussed, `docs/web3.auto.infura.rst` should be removed.

pacrob:
> As discussed, `docs/web3.auto.infura.rst` should be removed.

nm, was a local issue. GTG!

pacrob (approved):
lgtm!

PR #2709, opened 2022-11-03
Snakecase contract event methods

linda-le1:
### What was wrong?
Inherited javascript syntax. This PR does not include a deprecation step, intended for v6.
Related to Issue #2598 

### How was it fixed?
Changed `processReceipt` to `process_receipt`, `processLog` to `process_log`, `createFilter` to `create_filter` in `BaseContractEvent`. Change `getLogs` to `getlogs` in `ContractEvent` and `AsyncContractEvent`. 

### Todo:

#### Cute Animal Picture
Screen Shot 2022-11-03 at 12 23 39 AM


web3/contract.py
```diff
@@ -1681,7 +1681,7 @@ def getLogs(
 
 class AsyncContractEvent(BaseContractEvent):
     @combomethod
-    async def getLogs(
+    async def get_logs(
         self,
         argument_filters: Optional[Dict[str, Any]] = None,
         fromBlock: Optional[BlockIdentifier] = None,
```
fselmo, line 1687: See comment on the `ContractEvent` class. These can be snake cased.

web3/types.py
```diff
@@ -149,8 +149,8 @@ class FormattersDict(TypedDict, total=False):
 class FilterParams(TypedDict, total=False):
     address: Union[Address, ChecksumAddress, List[Address], List[ChecksumAddress]]
     blockHash: HexBytes
-    from_block: BlockIdentifier
-    to_block: BlockIdentifier
+    fromBlock: BlockIdentifier
```
fselmo, line 152: I think ultimately we make these changes here to camel case and everything else that we can leave snake cased should probably be snake cased. My reasoning is that the method arguments to python functions are expected to be snake cased, by convention. They are just variables up until we construct this `TypedDict`. Once we start building the `TypedDict`, this is the one translation step that basically says "Ok, we passed everything around using python conventions but now we need to speak to a service that does not use python conventions". The way I see it behaving is these `TypedDict` request param classes are the translation layer between python and provider and I think ultimately this is the only place we need the explicit change to camelCase.

I'm curious though if by doing this we run into complications elsewhere. Thoughts @linda-le1 @kclowes @pacrob?

docs/contracts.rst
```diff
@@ -922,25 +922,25 @@ For example:
         myContract = web3.eth.contract(address=contract_address, abi=contract_abi)
         tx_hash = myContract.functions.myFunction().transact()
         receipt = web3.eth.get_transaction_receipt(tx_hash)
-        myContract.events.myEvent().processReceipt(receipt)
+        myContract.events.myEvent().process_receipt(receipt)
 
 :py:class:`ContractEvent` provides methods to interact with contract events. Positional and keyword arguments supplied to the contract event subclass will be used to find the contract event by signature.
 
 .. _processReceipt:
```
pacrob, line 929: Can update the anchor title too? Or was there a reason not to?

docs/contracts.rst
```diff
@@ -993,11 +993,11 @@ For example:
...
        )
-       >>> processed_logs = contract.events.myEvent().processReceipt(tx_receipt, errors=DISCARD)
+       >>> processed_logs = contract.events.myEvent().process_receipt(tx_receipt, errors=DISCARD)
        >>> assert processed_logs == ()
        True
 
-.. py:method:: ContractEvents.myEvent(*args, **kwargs).processLog(log)
+.. py:method:: ContractEvents.myEvent(*args, **kwargs).process_log(log)
 
    Similar to processReceipt_, but only processes one log at a time, instead of a whole transaction receipt.
```
pacrob, line 1002: Same question as above re: anchor title

docs/contracts.rst
fselmo: +1 for updating anchor title

pacrob:
I left one note for the docs anchor title. It also looks like you could remove the `Eth.getLogs` deprecation notice in `docs/web3.eth.rst`. Otherwise lookin' good, thanks!

web3/types.py
kclowes: Yeah, I think I'm in the same boat. Basically anything that goes to/from the node should be camelCase, and everything else should be snaked.

kclowes (commented):
Awesome! 🚀  I just left one nit pick. Let me/us know if you have questions about when to `snake_case` vs. `camelCase` based on @fselmo's comment.

web3/contract.py
```diff
@@ -1606,11 +1606,11 @@ def factory(cls, class_name: str, **kwargs: Any) -> PropertyCheckingFactory:
 
 class ContractEvent(BaseContractEvent):
     @combomethod
-    def getLogs(
+    def get_logs(
         self,
         argument_filters: Optional[Dict[str, Any]] = None,
-        fromBlock: Optional[BlockIdentifier] = None,
-        toBlock: Optional[BlockIdentifier] = None,
+        from_block: Optional[BlockIdentifier] = None,
```
linda-le1, line 1612: @fselmo just confirming that you meant I should snakecase these keyword args up through line 1674:
`self._get_event_filter_params(
                abi, argument_filters, blockHash, fromBlock=from_block, toBlock=to_block)`

Is this right?

fselmo: Yeah. I think ultimately the `FilterParams` type in _types.py_ should be the only translation layer between snake_case and camelCase. So all the actual python function arguments should be snake-cased, as is convention. This includes `blockHash` in here though which looks like we missed.

docs/examples.rst
```diff
@@ -1180,7 +1180,7 @@ The script can be run with: ``python ./eventscanner.py <your JSON-RPC API URL>``
         """
 
         if from_block is None:
-            raise TypeError("Missing mandatory keyword argument to getLogs: fromBlock")
+            raise TypeError("Missing mandatory keyword argument to get_logs: from_block")
```
linda-le1, line 1183: Not sure if the code example here needs additional updates based on the changes in the PR here. I think it does, but would someone be able to take a look and double check me? @fselmo @pacrob @kclowes 

A few lines below in this file, `event_filter_params` is passed into `get_logs` like so:  
```
    data_filter_set, event_filter_params = construct_event_filter_params(
        abi,
        codec,
        address=argument_filters.get("address"),
        argument_filters=argument_filters,
        fromBlock=from_block,
        toBlock=to_block
    )

    logger.debug(f"Querying eth_getLogs with the following parameters: {event_filter_params}")

    # Call JSON-RPC API on your Ethereum node.
    # get_logs() returns raw AttributedDict entries
    logs = w3.eth.get_logs(event_filter_params)
```

pacrob: It looks like the `construct_event_filter_params` function def still has it's args camelCased.  But you could update there and then the docs too.

linda-le1: Okay - sounds good. I'll leave this where it's at then, request another review, and have a separate PR to update `construct_event_filter_params` in the docs. 

Thanks for double checking me!

pacrob (approved):
lgtm!

kclowes (approved):
LGTM too! Thanks @linda-le1!

fselmo (approved):
lgtm!

linda-le1:
@kclowes I've consolidated commits - let me know if that looks okay to you. Thanks!

Issue #2620, opened 2022-08-25
TypeError from `validate_abi_type(abi_type)` with `decimal.Decimal`


Akkarine:
* Version: 5.30.0
* Python: 3.9
* OS: docker desktop on WSL2

### What was wrong?

I'm passing to `w3.solidityKeccak(abi_types, values)` `uint256` values as `decimal.Decimal`, supposing, that `web3.py` aware of this type, like it does in docs [here](https://web3py.readthedocs.io/en/stable/examples.html#converting-currency-denominations).

But I get `TypeError: The following abi value is not a 'uint256': 1000000000000000000000000000000000000`:
```log
File /usr/local/lib/python3.9/site-packages/eth_utils/decorators.py:18, in combomethod.__get__.<locals>._wrapper(*args, **kwargs)
     15 @functools.wraps(self.method)
     16 def _wrapper(*args: Any, **kwargs: Any) -> Any:
     17     if obj is not None:
---> 18         return self.method(obj, *args, **kwargs)
     19     else:
     20         return self.method(objtype, *args, **kwargs)

File /usr/local/lib/python3.9/site-packages/web3/main.py:327, in Web3.solidityKeccak(cls, abi_types, values)
    324     w3 = cls
    325 normalized_values = map_abi_data([abi_ens_resolver(w3)], abi_types, values)
--> 327 hex_string = add_0x_prefix(HexStr(''.join(
    328     remove_0x_prefix(hex_encode_abi_type(abi_type, value))
    329     for abi_type, value
    330     in zip(abi_types, normalized_values)
    331 )))
    332 return cls.keccak(hexstr=hex_string)

File /usr/local/lib/python3.9/site-packages/web3/main.py:328, in <genexpr>(.0)
    324     w3 = cls
    325 normalized_values = map_abi_data([abi_ens_resolver(w3)], abi_types, values)
    327 hex_string = add_0x_prefix(HexStr(''.join(
--> 328     remove_0x_prefix(hex_encode_abi_type(abi_type, value))
    329     for abi_type, value
    330     in zip(abi_types, normalized_values)
    331 )))
    332 return cls.keccak(hexstr=hex_string)

File /usr/local/lib/python3.9/site-packages/web3/_utils/encoding.py:66, in hex_encode_abi_type(abi_type, value, force_size)
     62 """
     63 Encodes value into a hex string in format of abi_type
     64 """
     65 validate_abi_type(abi_type)
---> 66 validate_abi_value(abi_type, value)
[...]

PR #2712, opened 2022-11-08
Add isinstance decimal to validate_abi_value

ryo-ma:
### What was wrong?
Does not support Decimal type on the solidity_keccak

```
w3.solidity_keccak(['uint256'], [Decimal('1000000000000000000000000000000000000')])
```

Related to Issue #
#2620 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/validation.py
+++ b/web3/_utils/validation.py
@@ -1,3 +1,4 @@
+import decimal
 import itertools
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


pacrob:
Thank you for your contribution! In this case, as I wrote in Issue #2620, we don't want to accept Decimal values for a `uint` as that would require implicit conversion of the value. I would recommend using the `Web3.to_wei` function to explicitly convert `Decimal`s to `int`s first.

PR #2713, opened 2022-11-09
Async session cache improvements

### LINKED ISSUE #2597: Async request session caching incompatible with Flask
Flask is not natively async, so it creates a new thread loop for every async request.

When async web3py does an aiohttp call, it caches the session. So the next time you make a request it pulls the stale session out of the cache and throws an "Event loop is closed" exception.

To fix the issue we made the following code changes in request.py as a hack.

```
async def get_async_session(endpoint_uri: URI) -> ClientSession:
    cache_key = generate_cache_key(endpoint_uri)
    ########### TEMP CODE ###########
    if True:# cache_key not in _async_session_cache: # no caching for you!
    ########### TEMP CODE ###########
        await cache_async_session(endpoint_uri, ClientSession(raise_for_status=True))
    return _async_session_cache.get_cache_entry(cache_key)


async def async_get_response_from_get_request(
    endpoint_uri: URI, *args: Any, **kwargs: Any
) -> ClientResponse:
    kwargs.setdefault("timeout", ClientTimeout(10))
    session = await get_async_session(endpoint_uri)
    response = await session.get(endpoint_uri, *args, **kwargs)
    ########### TEMP CODE ###########
    await session.close()
    ########### TEMP CODE ###########
    return response


async def async_get_response_from_post_request(
    endpoint_uri: URI, *args: Any, **kwargs: Any
) -> ClientResponse:
    kwargs.setdefault("timeout", ClientTimeout(10))
    session = await get_async_session(endpoint_uri)
    response = await session.post(endpoint_uri, *args, **kwargs)
    ########### TEMP CODE ###########
    await session.close()
    ########### TEMP CODE ###########
    return response
```

Would be nice to have a setting to disable session caching.

---

fselmo:
### What was wrong?

I believe this closes #2597. That issue is a bit stale so we'll see if it gets re-opened after these changes are in.

### How was it fixed?

- If the loop for a cached async session is closed, or the session itself was closed, create a new session at that cache key and properly close and evict the stale session.

### Todo:

### DIFF
--- a/newsfragments/2713.feature.rst
+++ b/newsfragments/2713.feature.rst
@@ -1,0 +1,1 @@
+If the loop for a cached async session is closed, or the session itself was closed, create a new session at that cache key and properly close and evict the stale session.
--- a/tests/core/utilities/test_request.py
+++ b/tests/core/utilities/test_request.py
@@ -396,3 +396,73 @@ async def test_async_unique_cache_keys_created_per_thread_with_same_uri():
 
     # appropriately close the test sessions
     [await session.close() for session in test_sessions]
+
+
+@pytest.mark.asyncio
+async def test_async_use_new_session_if_loop_closed_for_cached_session():
+    # create new loop, cache a session wihin the loop, close the loop
+    loop1 = asyncio.new_event_loop()
+
+    session1 = ClientSession(raise_for_status=True)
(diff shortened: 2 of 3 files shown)


kclowes (approved):
LGTM! :shipit:

PR #2719, opened 2022-11-17
Performance updates to ``SimpleCache``, session cache, cache middlewares

fselmo:
### What was wrong?

Some performance considerations for when to lock the cache.

- Return the cached `value` at the same time as the `evicted_items` rather than reaching back into the cache for it. There's no reason to reach back into the cache for it.
- Only lock the cache when necessary. I think we were locking the cache a bit too early in almost all cases here. One exception is the async session cache since a session that is cached can be considered stale. That case was left alone as it should be under lock the whole time.

### How was it fixed?

- Lock cache only when necessary.
- Return the cached value from `SimpleCache` at the same time as the evicted items.

### Todo:

#### Cute Animal Picture

My beloved former pigmy goat, Bootsy, as a kid:

### DIFF
--- a/newsfragments/2719.feature.rst
+++ b/newsfragments/2719.feature.rst
@@ -1,0 +1,1 @@
+Some minor performance improvements to the ``SimpleCache`` class and simple cache middlewares (sync and async).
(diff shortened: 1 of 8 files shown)


pacrob (approved):
lgtm!

PR #2727, opened 2022-11-22
Update ``black`` command to use regex pattern

fselmo:
### What was wrong?

- Running `black` locally via `tox` or `make lint` has never really worked for me. This regex way to set up the `black` command via `tox` seems to work a bit better locally and regex seems to be the recommended way to run the command. 

### Todo:

### DIFF
--- a/newsfragments/2727.internal.rst
+++ b/newsfragments/2727.internal.rst
@@ -1,0 +1,1 @@
+Use regex pattern for ``black`` command for ``tox`` / ``make lint`` linting commands.
--- a/tox.ini
+++ b/tox.ini
@@ -64,7 +64,7 @@ basepython=python
 extras=linter
 commands=
     flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/ethpm {toxinidir}/tests --exclude {toxinidir}/ethpm/ethpm-spec,{toxinidir}/**/*_pb2.py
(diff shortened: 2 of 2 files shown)


pacrob:
> regex seems to be the recommended way to run the command.

Could you add a link to your reference? Thanks!

fselmo:
> Could you add a link to your reference? Thanks!

Clicking the `CLI reference` down arrow [here](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options):

```
Options:
    ...
    
    --exclude TEXT                A regular expression that matches files and
                                  directories that should be excluded on
                                  recursive searches. An empty value means no
                                  paths are excluded. Use forward slashes for
                                  directories on all platforms (Windows, too).
                                  Exclusions are calculated first, inclusions
                                  later. [default: /(\.direnv|\.eggs|\.git|\.h
                                  g|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn
                                  |\.ipynb_checkpoints|_build|buck-
                                  out|build|dist|__pypackages__)/]
```

fselmo:
@pacrob come to think of it, we could probably just use `--exclude` like so:

```
black {toxinidir}/ens {toxinidir}/ethpm {toxinidir}/web3 {toxinidir}/tests {toxinidir}/setup.py --exclude /ethpm/ethpm-spec/|/ethpm/_utils/protobuf/ipfs_file_pb2\.py --check
```

pacrob:
> Any preference between the two?

Slight preference for the latter

pacrob (approved):
lgtm!

Issue #2460, opened 2022-05-05
Core Test Cleanups


kclowes:
### What was wrong?

There are some test cleanups that we'd like to look into. 

### How can it be fixed?

Some ideas which should probably be separate PRs if implemented: 

- explore the possibility of sharing contracts between `async_w3` and `w3`. 
- explore `skip_if_testrpc`. Decide if they add enough documentation to keep around, or if documentation should be moved to the docs.
- explore which tests (if any) can be removed from the async core tests
- split async core tests into a separate tox run

PR #2732, opened 2022-11-22
async more filter tests

pacrob:
### What was done?

Async more filter test files

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/filtering/conftest.py
+++ b/tests/core/filtering/conftest.py
@@ -196,6 +196,14 @@ async def async_emitter(
     )
 
 
-@pytest.fixture(scope="module")
-def async_create_filter(request):
-    return async_partial(return_filter)
+async def async_return_filter(contract=None, args=[]):
+    event_name = args[0]
+    kwargs = apply_key_map({"filter": "argument_filters"}, args[1])
+    if "fromBlock" not in kwargs:
+        kwargs["fromBlock"] = "latest"
+    return await contract.events[event_name].create_filter(**kwargs)
+
+
+@pytest_asyncio.fixture(scope="module")
+async def async_create_filter(request):
+    return async_partial(async_return_filter)
--- a/tests/core/filtering/conftest.py
+++ b/tests/core/filtering/conftest.py
@@ -149,7 +149,7 @@ def emitter_log_topics():
     return LogTopics
 
 
-def return_filter(contract=None, args=[]):
+def return_filter(contract, args):
     event_name = args[0]
     kwargs = apply_key_map({"filter": "argument_filters"}, args[1])
     if "fromBlock" not in kwargs:
@@ -196,7 +196,7 @@ async def async_emitter(
     )
 
 
-async def async_return_filter(contract=None, args=[]):
+async def async_return_filter(contract, args):
     event_name = args[0]
     kwargs = apply_key_map({"filter": "argument_filters"}, args[1])
     if "fromBlock" not in kwargs:
--- a/tests/core/filtering/conftest.py
+++ b/tests/core/filtering/conftest.py
@@ -204,6 +204,6 @@ async def async_return_filter(contract=None, args=[]):
     return await contract.events[event_name].create_filter(**kwargs)
 
 
-@pytest_asyncio.fixture()
+@pytest_asyncio.fixture(scope="module")
 async def async_create_filter(request):
     return async_partial(async_return_filter)
--- a/tests/core/filtering/conftest.py
+++ b/tests/core/filtering/conftest.py
@@ -204,6 +204,6 @@ async def async_return_filter(contract=None, args=[]):
     return await contract.events[event_name].create_filter(**kwargs)
 
 
-@pytest_asyncio.fixture(scope="module")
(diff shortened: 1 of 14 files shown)


tests/core/filtering/test_contract_topic_filters.py
```diff
@@ -135,6 +102,30 @@ def array_values(draw):
...
+
+
+@pytest.fixture(scope="module")
+def Emitter(w3, EMITTER):
+    return w3.eth.contract(**EMITTER)
+
+
+@pytest.fixture(scope="module")
+def emitter(w3, Emitter, wait_for_transaction, wait_for_block, address_conversion_func):
+    return _emitter_fixture_logic(
```
kclowes, line 124: nit: I'm not sure I see the value in having a function just pass on it's args to another function. Can we move the logic back here and avoid some indirection? Alternatively, if this is used in more places than it looks like it is in this PR, can we pull the fixture up a few levels?

tests/core/filtering/test_existing_filter_instance.py
```diff
@@ -1,18 +1,14 @@
...
     Timeout,
 )
-from web3.providers.eth_tester import (
-    EthereumTesterProvider,
-)
 
 
 @pytest.fixture()
 def filter_id(w3):
-    if not isinstance(w3.provider, EthereumTesterProvider):
```
kclowes, line 13: Do you know why this was here in the first place? Does git blame give any insight? 🤔

kclowes (approved):
This looks generally good to me! I left some comments, mostly out of curiosity since all the tests look like they're running fine.

tests/core/filtering/test_basic_filter_tests.py
```diff
@@ -51,7 +47,7 @@ async def test_async_filtering_sequential_blocks_with_bounded_range(
     initial_block_number = await async_w3.eth.block_number
     builder.toBlock = initial_block_number + 100
     filter_ = await builder.deploy(async_w3)
-    async for i in async_range(100):
```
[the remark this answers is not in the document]
pacrob: I created the `async_range` helper in a previous commit on this branch because I misunderstood how `for` vs `async for` worked. I thought `async for` was needed if there  were any `await`s in the loop, but it is not.

tests/core/filtering/test_contract_topic_filters.py
pacrob: The other use is in the local `conftest.py`, which came through in the previous PR. The fixture is redefined here due to scope. I had pulled the logic of the longer fixtures out to `utils.py` to avoid duplication. It is only used twice though, so I'm happy to move the logic back if you would prefer.

pacrob: And I'll aim to have these sorts of related changes come through in a single PR in the future.

tests/core/filtering/conftest.py
```diff
@@ -196,6 +196,14 @@ async def async_emitter(
     )
 
 
-@pytest.fixture(scope="module")
-def async_create_filter(request):
-    return async_partial(return_filter)
+async def async_return_filter(contract=None, args=[]):
```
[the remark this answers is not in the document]
fselmo: +1 for this

tests/core/filtering/test_existing_filter_instance.py
pacrob: I looked at the history of that line and couldn't tell.  Were there other providers used for running tests pre-2018?

tests/core/filtering/test_basic_filter_tests.py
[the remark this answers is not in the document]
fselmo: Yeah I can't really figure it out either. It *is* pretty old... I'm assuming there were other test providers running our core test suite in the past. I'm OK taking this out since the whole core suite runs with `EthereumTesterProvider()`.

fselmo (approved):
lgtm 👍🏼. Don't have anything else to add here.

tests/core/filtering/test_existing_filter_instance.py
kclowes: I'm not sure. Looks like it is running fine now though 🤷‍♀️

tests/core/filtering/test_contract_topic_filters.py
kclowes: Ah, right. I forgot about those warnings. I don't have a strong preference either way. I think the real fix is to figure out why we can't use this fixture on the higher levels without warnings. I just added another bullet point to issue #2460 to look into this.

tests/core/filtering/conftest.py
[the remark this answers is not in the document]
pacrob: I went with first option, changed in the sync version I copied it from too.

tests/core/filtering/test_contract_topic_filters.py
pacrob: In this case, it's hypothesis that has issues with the fixture scope. Not sure about other cases.

PR #2734, opened 2022-11-22
Update ``mypy`` file exclusion to use regex search

fselmo:
### What was wrong?

- Due to similar issues as was happening in #2727, ``mypy`` looks for regular expressions which led to issues running ``make lint`` locally with some versions of `tox`.

### Todo:

#### Cute Animal Picture

278611_10150693933715613_722865612_19422754_4741222_o


pacrob (approved):
lgtm!

PR #2735, opened 2022-11-22
Remove unnecessary await for ``generate_gas_price()``

fselmo:
### What was wrong?

- ``generate_gas_price()`` for the ``AsyncEth`` module was unnecessarily awaited. Since we don't use the ``w3`` instance directly, we can move this method to ``BaseEth`` and remove the awaits when called via ``AsyncEth``.

### Todo:

#### Cute Animal Picture

4980528495_64ee30707b_b


pacrob (approved):
lgtm!

PR #2736, opened 2022-11-23
Clean up ``AsyncHTTPProvider`` instantiation

fselmo:
### What was wrong?

Related to #1416. This is listed as non-breaking but instead of having an `async_net` module, it seems more appropriate to set the `net` module up as the `AsyncNet` class. This will break things.

### How was it fixed?

- Add ``is_async`` flag to base provider classes and appropriately set them to ``True`` or ``False``.
- Use these flags to determine whether to load sync or async default middlewares and default modules.
- Refactor the ``Web3`` property ``async_net`` as ``net`` mapped to ``AsyncNet`` class when dealing with async and ``Net`` when dealing with sync.

### Todo:
- [x] Update docs

#### Cute Animal Picture

 ___ O>
/\ /\

### DIFF
--- a/tests/core/providers/test_async_http_provider.py
+++ b/tests/core/providers/test_async_http_provider.py
@@ -51,8 +51,8 @@ def test_web3_with_async_http_provider_has_default_middlewares_and_modules() -> 
     # assert default modules
 
     assert isinstance(async_w3.eth, AsyncEth)
-    assert isinstance(async_w3.geth, Geth)
     assert isinstance(async_w3.net, AsyncNet)
+    assert isinstance(async_w3.geth, Geth)
     assert isinstance(async_w3.geth.admin, AsyncGethAdmin)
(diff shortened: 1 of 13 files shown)


tests/core/providers/test_async_http_provider.py
```diff
@@ -4,18 +4,80 @@
...
+
+
+def test_init_kwargs():
+    provider = AsyncHTTPProvider(endpoint_uri=URI, request_kwargs={"timeout": 60})
+    w3 = Web3(provider)
+    assert w3.manager.provider == provider
+
+
+def test_web3_with_async_http_provider_has_default_middlewares_and_modules() -> None:
+    async_w3 = Web3(AsyncHTTPProvider(endpoint_uri="http://mynode.local:8545"))
```
pacrob, line 49: can use `URI` here

pacrob (approved):
one small update noted, otherwise lgtm! thanks!

PR #2737, opened 2022-11-23
Remove internal ``apply_formatters_to_array``

### LINKED ISSUE #1902: Standardize `apply_formatter_to_array`
* Version: v6

### What was wrong?

The eth-utils and web3 `apply_formatter_to_array` functions are slightly different. We should standardize on one or the other (probably the eth-utils version). The `apply_formatter_to_array` function in web3 converts the return value to an array and the eth-utils version does not.

fselmo:
### What was wrong?

- Closes #1902 
- This is in #1416 but I don't think this is breaking since the method itself wasn't a public method (it's under `_utils`).

### How was it fixed?

- Remove web3.py's `apply_formatter_to_array` and use the eth-utils method with the same name instead.

### Todo:

#### Cute Animal Picture

n621963065_1604526_454

### DIFF
--- a/newsfragments/2737.internal.rst
+++ b/newsfragments/2737.internal.rst
@@ -1,0 +1,1 @@
+Remove internal method ``apply_formatter_to_array`` and use the method with the same name from the ``eth-utils`` library.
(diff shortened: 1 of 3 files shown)


kclowes (approved):
LGTM! 🚀

PR #2740, opened 2022-11-29
bump readme python supported version

wolovim:
### What was wrong?

we're not supporting python 3.6 on v5 either, right?

### Todo:

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A Python library for interacting with Ethereum, inspired by [web3.js](https://github.com/ethereum/web3.js).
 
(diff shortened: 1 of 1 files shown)


fselmo (commented):
@wolovim we [are supporting](https://github.com/ethereum/web3.py/blob/v5/tox.ini#L51-L52) it in `v5` actually. We dropped it only in `v6` since it would be breaking to stop supporting python 3.6 partway through `v5`.

wolovim:
@fselmo that checks out! testing a new webhook also accomplished here 😅

PR #2751, opened 2022-12-07
Support latest tox v4

kclowes:
### What was wrong?
A tox upgrade broke our lint CI runs. 

### How was it fixed?
Added an upper pin to the tox install in our CI config

### Todo:

#### Cute Animal Picture


pacrob (approved):
lgtm!

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ common: &common
         command: git submodule update --init --recursive
     - run:
         name: install dependencies
-        command: pip install --user tox
+        command: pip install --user tox==3.*
```
fselmo, line 17: This adds a bit more surface area when updating the tox version. I think that's fine but we should maybe add a comment in `setup.py` to update this as well when we fix the breaking changes. Thoughts?

fselmo (approved):
~~lgtm, just left one nit that might help us track down the different places where we may need to update tox.~~ [outdated]

.circleci/config.yml
kclowes: good call. Will add a comment!

fselmo (approved):
I prefer this route! Consider this re-approved 🚀 ... just needs the newsfragment message updated

kclowes:
You're so fast! Thanks!

PR #2753, opened 2022-12-12
Organize eth module as separate files

fselmo:
### What was wrong?

- Reorganize `eth` module into separate files
- Tighten up typing on `eth` methods

### Todo:


ethpm/package.py
```diff
@@ -115,7 +115,7 @@ def __init__(
         validate_w3_instance(w3)
 
         self.w3 = w3
-        self.w3.eth.defaultContractFactory = cast(Type[Contract], LinkableContract)
+        self.w3.eth._default_contract_factory = cast(Type[Contract], LinkableContract)
```
fselmo, line 118: Was this ever meant to be a public property? 🤔

I don't mind putting this back, but this made a bit more sense. We don't have documentation on this afaik.

web3/eth/async_eth.py
```diff
@@ -0,0 +1,545 @@
...
+class AsyncEth(BaseEth):
+    is_async = True
+
+    _default_contract_factory: Type[
+        Union[AsyncContract, AsyncContractCaller]
+    ] = AsyncContract
+
+    # eth_accounts
+
+    _accounts: Method[Callable[[], Awaitable[Tuple[ChecksumAddress]]]] = Method(
```
kclowes, line 84: Since we're not using these underscore methods as shared variables between sync and async any more, what do you think about removing them and doing something like:

```py
def accounts(self) -> Tuple[ChecksumAddress]:
        return await Method(RPC.eth_accounts, is_property=True,)
```
It would save us some lines, but I don't have a strong preference either way

kclowes (approved):
This looks good to me! I just left a few comments, mostly nits :)

web3/eth/async_eth.py
```diff
@@ -0,0 +1,545 @@
...
+        Callable[[Optional[Union[str, FilterParams, HexStr]]], Awaitable[AsyncFilter]]
+    ] = Method(
+        method_choice_depends_on_args=select_filter_method(
+            if_new_block_filter=RPC.eth_newBlockFilter,
+            if_new_pending_transaction_filter=RPC.eth_newPendingTransactionFilter,
+            if_new_filter=RPC.eth_newFilter,
+        ),
+        mungers=[BaseEth.filter_munger],
+    )
+
```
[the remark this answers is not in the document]
fselmo: > nit: the `# eth_getFilterChanges` comment is missing from here on out

Nice catch 👀

web3/eth/async_eth.py
fselmo: We get rid of the type checking in this scenario, right?

kclowes: Right. You would know better since you're knee deep in it right now, but I don't think it would impact the typing since we still have the correct return typing in the method signature (i.e. `Tuple[ChecksumAddress]`)?

fselmo: I suppose for the properties this doesn't matter all that much since they don't take arguments but would we keep this pattern for the methods?

fselmo: Oops hadn't refreshed so didn't see your last comment. Yeah I think for properties this can make sense. I think keeping it for methods makes sense though.

kclowes: What do we gain from keeping it for the methods?

fselmo: Oh wait I see your point. I think I'm good with this change. I'll put it in a separate commit and we can go from there.

fselmo: @kclowes Ah, the way Method is set up is that the `__get__()` is what actually makes the call. So unless we can think of an elegant way to instantiate and get the `Method` all in the `return` call, I think the old way may still be the best way to go about things. Even when they lived in the same file, we still kept the underscored, typed definitions. Or am I missing something from your original comment?

kclowes: Ah, right. Nope, I had forgotten about the `__get__()` magic. Okay, this LGTM! Thanks for looking into it!

Issue #2762, opened 2022-12-15
Get contract address from `CREATE` and `CREATE2` opcode


tinom9:
### What was wrong?

Unlike `ethers.js` ([5.7.2](https://github.com/ethers-io/ethers.js/blob/v5.7.2/packages/address/src.ts/index.ts#L130)), `web3.py` does not include a utility method to deterministically determine the contract address from the inputs of `CREATE` and `CREATE2` opcodes.

### How can it be fixed?

The address of a `CREATE` opcode deployed contract can be determined as a function of its sender and its respective nonce:

```python
new_address = hash(sender, nonce)
```

And the address of a `CREATE2` opcode deployed contract can be determined as a function of a constant `0xFF`, its sender, an arbitrary value `salt` and the contract's bytecode:

```python
new_address = hash(0xFF, sender, salt, bytecode)
```

PR #2763, opened 2022-12-15
Get contract address from `CREATE` and `CREATE2` opcodes

tinom9:
### What was wrong?
Unlike `ethers.js` ([5.7.2](https://github.com/ethers-io/ethers.js/blob/v5.7.2/packages/address/src.ts/index.ts#L130)), `web3.py` does not include a utility method to deterministically determine the contract address from the inputs of `CREATE` and `CREATE2` opcodes.

Related to Issue #2762

### How was it fixed?
Added public utility methods to determine contract addresses.

### Todo:
- [X] Get contract address from `CREATE` opcode
- [X] Get contract address from `CREATE2` opcode
- [X] Documentation

#### Cute Animal Picture


web3/utils/address.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+def get_create_address(sender: HexAddress, nonce: Nonce) -> ChecksumAddress:
+    """
+    Determine the resulting `CREATE` opcode contract address for a sender and a nonce.
+    """
+    bytes_sender = bytes.fromhex(sender[2:])
+
+    contract_address = keccak(encode([bytes_sender, nonce])).hex()[-40:]
```
fselmo, line 25: nit: Ideally we import `rlp` and use `rlp.encode` here for clarification. We've done the same with `import abi` -> `abi.encode()` and it just makes it easier to read the code.

```suggestion
    contract_address = keccak(rlp.encode([bytes_sender, nonce])).hex()[-40:]
```

web3/utils/address.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_create2_address(
+    sender: HexAddress, salt: HexStr, init_code: HexStr
+) -> ChecksumAddress:
+    """
+    Determine the resulting `CREATE2` opcode contract address for a sender, salt and
+    bytecode.
+    """
+    bytes_prefix = bytes.fromhex("0xFF"[2:])
```
fselmo, line 36: nit: I could go either way here but `to_bytes` from the `eth-utils` dependency would not only make this bit of code more similar to the rest of the code base (so faster to read / process), but it also would accept both 0x-prefixed and non-0x-prefixed hexes.

Something like: 

```py
contract_address = keccak(
    b"\xff"
    + to_bytes(hexstr=sender)
    + to_bytes(hexstr=salt)
    + keccak(to_bytes(hexstr=init_code))
).hex()[-40:]
```

Thoughts on that?

web3/utils/address.py
```diff
@@ -0,0 +1,44 @@
...
+    HexStr,
+    Nonce,
+)
+
+
+def get_create_address(sender: HexAddress, nonce: Nonce) -> ChecksumAddress:
+    """
+    Determine the resulting `CREATE` opcode contract address for a sender and a nonce.
+    """
+    bytes_sender = bytes.fromhex(sender[2:])
```
fselmo, line 23: nit: See comment below, maybe use `to_bytes(hexstr=sender)` here?

fselmo (approved):
@tinom9 this is a great addition to the public utilities, thank you so much. We'd talked about it at some point and just hadn't gotten around to it. This is good to go as far as I'm concerned, I just had some nits I wanted to pick your brain on. Thoughts on those?

web3/utils/address.py
tinom9: It's more readable as `rlp.encode`, agreed!

web3/utils/address.py
tinom9: I slightly prefer `bytes.fromhex` because it's built-in, but code readability is king!

fselmo:
Thanks again @tinom9!

PR #2768, opened 2022-12-27
dependencies: upgrade websockets, drop py3.6

BrendanGraham14:
### What was wrong?

`web3.py` depends on a pretty old version of `websockets` and has dependency collisions with many other packages that also depend on `websockets`.

### How was it fixed?
I upgraded `websockets` to `>=10,<11`. Because `websockets>=10,<11` drops support for python3.6, I did that here as well.

### Todo:


BrendanGraham14:
Ah. Looks like I'll have to do something like https://github.com/ethereum/web3.py/pull/2343/files first.

PR #2769, opened 2022-12-27
v5: Drop python 3.6 support

BrendanGraham14:
### What was wrong?

Python 3.6 has reached [end of support](https://www.python.org/downloads/). Some packages can't be upgraded yet (e.g. `websockets`) without first dropping support for Python 3.6.

### How was it fixed?

Cherry-picked #2343 onto `v5`

### Todo:


fselmo:
Hey @BrendanGraham14. Thanks for the interest here. The reason we didn't backport this to `v5` is this would be a breaking change for those who still haven't migrated from py36. We only planned on dropping py36 for the `v6` major version of web3.py.

Issue #2774, opened 2023-01-10
Installing web3-6.0.0b9 in clean environment is missing base58 dependency


rokm:
* Version: 6.0.0b9
* Python: 3.x
* OS: linux (but applicable to all)
* `pip freeze` output

```
aiohttp==3.8.3
aiosignal==1.3.1
async-timeout==4.0.2
attrs==22.2.0
bitarray==2.6.2
certifi==2022.12.7
charset-normalizer==2.1.1
cytoolz==0.12.1
eth-abi==4.0.0b2
eth-account==0.8.0
eth-hash==0.5.1
eth-keyfile==0.6.0
eth-keys==0.4.0
eth-rlp==0.3.0
eth-typing==3.2.0
eth-utils==2.1.0
frozenlist==1.3.3
hexbytes==0.3.0
idna==3.4
jsonschema==4.17.3
lru-dict==1.1.8
multidict==6.0.4
parsimonious==0.9.0
protobuf==4.21.12
pycryptodome==3.16.0
pyrsistent==0.19.3
regex==2022.10.31
requests==2.28.1
rlp==3.0.0
toolz==0.12.0
urllib3==1.26.13
web3==6.0.0b9
websockets==10.4
yarl==1.8.2
```

### What was wrong?

With the latest beta of `web3` (6.0.0b9), the `ethpm` package cannot be imported anymore due to missing `base58`.

Steps to reproduce (e.g., linux):
```
python3.9 -m venv venv
. venv/bin/activate
pip install web3==6.0.0b9
python -c "import ethpm"

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/rok/tmp/web3-test/venv/lib64/python3.9/site-packages/ethpm/__init__.py", line 19, in <module>
    from .package import Package  # noqa: E402, F401
  File "/home/rok/tmp/web3-test/venv/lib64/python3.9/site-packages/ethpm/package.py", line 47, in <module>
    from ethpm.dependencies import (
  File "/home/rok/tmp/web3-test/venv/lib64/python3.9/site-packages/ethpm/dependencies.py", line 7, in <module>
    from ethpm.validation.package import (
  File "/home/rok/tmp/web3-test/venv/lib64/python3.9/site-packages/ethpm/validation/package.py", line 11, in <module>
    from ethpm._utils.ipfs import (
[...]

PR #2775, opened 2023-01-10
prevent ``ethpm`` import issues without ``ipfshttpclient``

fselmo:
### What was wrong?

Related to Issue #2774

### How was it fixed?

- Don't force imports related to ``ipfshttpclient`` for ``ethpm``

### Todo:

#### Cute Animal Picture

Image from iOS (3) copy

### DIFF
--- a/ethpm/_utils/backend.py
+++ b/ethpm/_utils/backend.py
@@ -10,9 +10,6 @@ from eth_typing import (
 from eth_utils import (
     to_tuple,
 )
-from ipfshttpclient.exceptions import (
-    ConnectionError,
-)
 
 from ethpm.backends.base import (
     BaseURIBackend,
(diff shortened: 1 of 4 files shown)


ethpm/_utils/backend.py
```diff
@@ -30,8 +27,17 @@
     RegistryURIBackend,
 )
 
+try:
```
fselmo, line 30: This isn't the most elegant way to handle this but it will only break if an exception is raised in the below methods. I am open to different ideas here.

ethpm/_utils/backend.py
```diff
@@ -30,6 +27,12 @@
...
+    from ipfshttpclient.exceptions import ConnectionError as IpfsConnectionError
```
fselmo, line 31: This was renamed to `IpfsConnectionError` because there is a builtin `ConenctionError`... so we'd be mistakenly catching an unwanted exception if the `ipfshttpclient` module is not available.

kclowes (approved):
LGTM! 🚢

pacrob (approved):
lgtm!

PR #2783, opened 2023-01-18
Properly handle eth-tester exception messages

fselmo:
### What was wrong?

- After switching to `abi.encode()` across the code base, some `eth-tester` messages were not properly decoded.

context: `eth-tester` [wraps some exceptions in other exceptions](https://github.com/ethereum/eth-tester/blob/cd07e441c64879a13b27af88afd6f777dcd81512/eth_tester/backends/pyevm/main.py#L611-L613). This parsing needs to be accounted for as it was broken when we switched from `eth-abi`'s `encode_single` to `abi.encode`.

### How was it fixed?

Properly parse encoded messages from wrapped exceptions coming from `eth-tester`

### Todo:


web3/providers/eth_tester/main.py
```diff
@@ -133,29 +111,46 @@ def __init__(
...
+            {"error": f"RPC Endpoint has not been implemented: {method}"}
+        )
+    except TransactionFailed as e:
+        first_arg = e.args[0]
+        try:
+            # sometimes eth-tester wraps an exception in another exception
+            raw_error_msg = (
+                first_arg if not isinstance(first_arg, Exception) else first_arg.args[0]
+            )
+            reason = abi.decode(["string"], raw_error_msg[4:])[0]
```
fselmo, line 149: The only change, since this refactor doesn't show it too well, is this line. Parentheses removed here:

```["(string)"] -> ["string"]```

File: web3/providers/eth_tester/main.py [hunk 2]
```diff
@@ -133,29 +114,50 @@ def __init__(
...
+            {"error": f"RPC Endpoint has not been implemented: {method}"}
+        )
+    except TransactionFailed as e:
+        first_arg = e.args[0]
+        try:
+            # sometimes eth-tester wraps an exception in another exception
+            raw_error_msg = (
+                first_arg if not isinstance(first_arg, Exception) else first_arg.args[0]
+            )
+            reason = (
```
fselmo, line 152: Also new... if the reason is bytes, decode as string, else use the raw message.

web3/providers/eth_tester/main.py
```diff
@@ -133,29 +114,50 @@ def __init__(
...
+        response = delegator(ethereum_tester_instance, params)
+    except NotImplementedError:
+        return RPCResponse(
+            {"error": f"RPC Endpoint has not been implemented: {method}"}
+        )
+    except TransactionFailed as e:
+        first_arg = e.args[0]
+        try:
+            # sometimes eth-tester wraps an exception in another exception
+            raw_error_msg = (
```
fselmo, line 149: In case `eth-tester` stops wrapping one exception in another, check to see if we have a wrapped exception, otherwise take the first arg as the raw error message.

kclowes (approved):
This PR looks good to me. Is the wrapping something we should fix in `eth-tester`? It seems less than ideal to wrap an exception in an exception, but I'm not sure why it was added in the first place.

fselmo:
> Is the wrapping something we should fix in `eth-tester`? It seems less than ideal to wrap an exception in an exception, but I'm not sure why it was added in the first place.

Yeah, specifically [here](https://github.com/ethereum/eth-tester/blob/cd07e441c64879a13b27af88afd6f777dcd81512/eth_tester/backends/pyevm/main.py#L611-L613).

PR #2799, opened 2023-01-27
Decode tuples per abi

pacrob:
### What was wrong?

Named objects can already be passed to contracts where tuples are expected as input, but there is no way to decode outputs as named objects despite having the names available in the ABI.

Original PR #1353
Related to Issue #1267

### How was it fixed?

A `decode_tuples` option is added when instantiating a Contract. When set, tuples returned from a contract function call will be a new `NamedTuple` of the name `ABIDecodedNamedTuple`. NamedTuples are compatible wherever regular tuples are used, but fields can be accessed with `tuple_name.key` syntax. I set the NamedTuple option `rename=True` to handle restricted field names, rather than add a custom renaming scheme as in the original PR. More [info](https://docs.python.org/3/library/collections.html#collections.namedtuple) on NamedTuples.

Contract event data was previously returned as an `AttributeDict` with any nested info as unnamed tuples. It is now `AttributeDict`s all the way down.

### Todo:

#### Cute Animal Picture

image


docs/web3.contract.rst
```diff
@@ -146,12 +146,29 @@ Each Contract Factory exposes the following properties.
     The runtime part of the contract bytecode string.  May be ``None`` if not
     provided during factory creation.
 
+
+.. py:attribute:: Contract.decode_tuples
+
+    If a Tuple/Struct is returned by a contract function, this flag defines whether
+    to apply the field names from the ABI to the returned data.
+    If False, the returned value will be a normal Python `Tuple`. If True, the returned
+    value will be a Python `NamedTuple` of the name `ABIDecodedNamedTuple`.
```
kclowes, line 155:
```suggestion
    value will be a Python `NamedTuple` of the class `ABIDecodedNamedTuple`.
```

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -902,6 +928,115 @@ def test_call_tuple_contract(tuple_contract, method_input, expected):
...
+                ],
+            ),
+        ),
+    ),
+)
+def test_call_tuple_contract_with_decode_tuples_set(
+    tuple_contract_with_decode_tuples, method_input, expected
+):
+    result = tuple_contract_with_decode_tuples.functions.method(method_input).call()
+    assert result == expected
```
kclowes, line 1036: In general, I think these tests would be clearer if we were also checking for that the `type(expected)` is equal to the namedtuple class name, and maybe even checking the `__repr__` to make sure everything gets assigned like it should. It took me a while to figure out why an apparently "normal" tuple was returned here ha.

web3/contract/base_contract.py
```diff
@@ -436,6 +437,7 @@ def __init__(
         w3: "Web3",
         contract_function_class: Type["BaseContractFunction"],
         address: Optional[ChecksumAddress] = None,
+        decode_tuples: Optional[bool] = None,
```
kclowes, line 440: Is there a variant where we're explicitly using `None` somewhere? If not, I think I'd prefer to just have this default to `False`, but feel free to take or leave.

web3/contract/utils.py
```diff
@@ -129,6 +132,10 @@ def call_contract_function(
     )
     normalized_data = map_abi_data(_normalizers, output_types, output_data)
 
+    if decode_tuples:
+        decoded = named_tree(fn_abi["outputs"], normalized_data)
+        normalized_data = dict_to_namedtuple(decoded)
```
kclowes, line 137: nit: since we're using these three lines in two different places, I'd slightly prefer to pull them out into their own function. But feel free to take or leave that too. I know some people consider 3 uses the magic number :)

web3/_utils/abi.py
```diff
@@ -906,3 +907,80 @@ def build_strict_registry() -> ABIRegistry:
...
+            return dict(super()._asdict())
+
+    return ABIDecodedNamedTuple  # type: ignore
+
+
+def generate_namedtuple_from_dict(**kwargs: Dict[str, Any]) -> Tuple[Any, ...]:
+    """
+    Literal namedtuple constructor such that:
+    `generate_namedtuple_from_dict(x=1, y=2)`
+    returns `generate_namedtuple_from_dict(x=1, y=2)`.
```
kclowes, line 982: nit: this comment doesn't make sense to me. Should it be something like: 
```suggestion
    returns `ABIDecodedNamedTuple(x=1, y=2)`.
```

web3/_utils/abi.py
```diff
@@ -906,3 +907,80 @@ def build_strict_registry() -> ABIRegistry:
...
+    """
+
+    class ABIDecodedNamedTuple(namedtuple("ABIDecodedNamedTuple", fields, rename=True)):  # type: ignore # noqa: E501
+        def __new__(self, args: Any) -> "ABIDecodedNamedTuple":
+            return super().__new__(self, *args)
+
+        def _asdict(self) -> Dict[str, Any]:
+            return dict(super()._asdict())
+
+    return ABIDecodedNamedTuple  # type: ignore
```
kclowes, line 975: Any chance we can get rid of these `type: ignores` in this function?

kclowes (commented):
I looked through this on a pretty high level, and it's looking good to me. I left some nits, and then a few thoughts I had are: 
- Do we need to pass `decode_tuples` to ContractEvents as well?
- Can we also add a test to the contract caller tests to make sure that's working okay?
- I think I'm okay with the `rename` option, since if you are opting in to this functionality, you should have some idea of how `namedtuples` work, but I'm curious to get feedback from @fselmo here too. I think originally we had talked about returning a regular tuple as a fallback if there were fields that aren't allowed as namedtuple field names. 
- It might be nice to give banteg some credit here, either by cherry-picking a PR or two, or maybe by tagging him in the notes when you merge or something.

web3/contract/base_contract.py
fselmo: +1 for this... the `ccip_read_enabled` flag was `None` for a reason since there are two competing flags and only if the contract flag was explicitly set to a value would it be considered, having the global flag override its value otherwise. If there isn't a more complex situation here, the `False` default would be ideal.

fselmo (commented):
I did a first pass. This looks great. I don't think I like the internal renaming of keywords to `_0`, `_1`, `_2`... etc. I wonder if we can come up with a better solution there? 🤔

If we kept track of them somehow and remapped them to `from_`, `class_`, etc... would that be better? I haven't looked but is that an option in `namedtuple`?

If this seems like the best approach I think I'm OK with it since this is just an enhancement / nice-to-have flag. But ideally, and for a better user experience, I would like to know what those fields were called by having the name from the key in one way or another.

Thoughts?

fselmo:
>     * I think I'm okay with the `rename` option, since if you are opting in to this functionality, you should have some idea of how `namedtuples` work, but I'm curious to get feedback from @fselmo here too. I think originally we had talked about returning a regular tuple as a fallback if there were fields that aren't allowed as namedtuple field names.

Yeah I had some thoughts on this [here](https://github.com/ethereum/web3.py/pull/2799#pullrequestreview-1279952903). Not sure what the best approach is but I'd like to have a better UX than just `_0`, `_1`, etc... worst case, I don't think I'm against just inherent `namedtuple` handling.

>     * It might be nice to give banteg some credit here, either by cherry-picking a PR or two, or maybe by tagging him in the notes when you merge or something.

Agreed!

docs/web3.contract.rst
pacrob: fixed

web3/contract/base_contract.py
pacrob: set to default to `False` in all contract files and in docs.

docs/web3.contract.rst
```diff
@@ -146,12 +146,29 @@ Each Contract Factory exposes the following properties.
...
+    
+    NamedTuples have some restrictions regarding field names.
+    Web3.py sets `NamedTuple`'s `rename=True`, so disallowed field names my be
```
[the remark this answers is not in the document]
pacrob: fixed

pacrob:
I chose to go with namedtuple's `rename=True` for predictability. They have a set of rules for renaming that we can point to and say "this is how it behaves". If a user has previous experience with namedtuples, they would be expecting this renaming pattern. If we roll our own renaming scheme, we need to explain it in detail and still run the risk of missing edge cases.

Regarding returning just a plain tuple instead: I feel it would be more confusing, having set `decode_tuples=True`, to just get a plain tuple back with no explanation. If the user sees a `ABIDecodedNamedTuple` with renamed fields, they at least know they set the flag correctly and can do more research.

web3/_utils/abi.py
```diff
@@ -906,3 +907,80 @@ def build_strict_registry() -> ABIRegistry:
...
+        item: Union[Dict[str, Any], List[Any]]
+    ) -> Union[Tuple[Any, ...], List[Any]]:
+        if isinstance(item, dict):
+            return generate_namedtuple_from_dict(**item)
+        return item
+
+    return recursive_map(to_tuple, data)
+
+
+def foldable_namedtuple(fields: Tuple[Any, ...]) -> Tuple[Any, ...]:
```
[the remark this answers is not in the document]
pacrob: works!

web3/_utils/abi.py
```diff
@@ -906,3 +907,80 @@ def build_strict_registry() -> ABIRegistry:
...
+        def __new__(self, args: Any) -> "ABIDecodedNamedTuple":
+            return super().__new__(self, *args)
+
+        def _asdict(self) -> Dict[str, Any]:
+            return dict(super()._asdict())
+
+    return ABIDecodedNamedTuple  # type: ignore
+
+
+def generate_namedtuple_from_dict(**kwargs: Dict[str, Any]) -> Tuple[Any, ...]:
```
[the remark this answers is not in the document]
pacrob: I like it. Refactored.

web3/_utils/abi.py
```diff
@@ -906,3 +907,80 @@ def build_strict_registry() -> ABIRegistry:
...
+def foldable_namedtuple(fields: Tuple[Any, ...]) -> Tuple[Any, ...]:
+    """
+    Customized namedtuple such that `type(x)(x) == x`.
+    """
+
+    class ABIDecodedNamedTuple(namedtuple("ABIDecodedNamedTuple", fields, rename=True)):  # type: ignore # noqa: E501
+        def __new__(self, args: Any) -> "ABIDecodedNamedTuple":
+            return super().__new__(self, *args)
+
+        def _asdict(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
pacrob: You are correct. Removed `_asdict` here and added to tests.

web3/contract/utils.py
pacrob: I think that because it's once for sync and once for async, I'll leave it as is this time. I can't imagine where it might split again, but :shrug:

web3/_utils/abi.py
pacrob: removed in refactor

tests/core/contracts/test_contract_call_interface.py
pacrob: I agree. Changed some naming and added several more checks. I wasn't sure how to check against the full namedtuple form, but using `__repr__` works great!

web3/_utils/abi.py
```diff
@@ -906,3 +907,70 @@ def build_strict_registry() -> ABIRegistry:
...
+    Convert function inputs/outputs or event data tuple to dict with names from ABI.
+    """
+    names = [item["name"] for item in abi]
+    items = [named_subtree(*item) for item in zip(abi, data)]
+
+    return dict(zip(names, items))
+
+
+def named_subtree(
+    # TODO make this a better type
```
pacrob, line 929: I'm having trouble figuring out typing here. @fselmo - do you have any insight?

fselmo (commented):
I answered a bit on the typing. Let me know if I missed something there but that all does work locally for me. I think this all looks good to me since the tests are passing but I'd like a bit more time to look carefully at it. I can make another pass Monday 👀.

web3/_utils/abi.py
```diff
@@ -906,3 +907,70 @@ def build_strict_registry() -> ABIRegistry:
...
+    abi_type = parse(collapse_if_tuple(dict(abi)))
+
+    if abi_type.is_array:
+        item_type = abi_type.item_type.to_type_str()
+        item_abi = {**abi, "type": item_type, "name": ""}
+        items = [named_subtree(item_abi, item) for item in data]
+        return items
+
+    if isinstance(abi_type, TupleType):
+        names = [item["name"] for item in abi["components"]]
```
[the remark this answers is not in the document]
pacrob: In ABIs (at least all of our sample ones), if a `tuple` is defined it always has a `components` key. Gotta know what's inside the tuple!

web3/_utils/abi.py
```diff
@@ -906,3 +907,70 @@ def build_strict_registry() -> ABIRegistry:
...
+) -> Union[Dict[str, Any], Tuple[Any, ...], Any]:
+    abi_type = parse(collapse_if_tuple(dict(abi)))
+
+    if abi_type.is_array:
+        item_type = abi_type.item_type.to_type_str()
+        item_abi = {**abi, "type": item_type, "name": ""}
+        items = [named_subtree(item_abi, item) for item in data]
+        return items
+
+    if isinstance(abi_type, TupleType):
```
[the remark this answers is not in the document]
pacrob: done

web3/_utils/contracts.py
```diff
@@ -312,6 +319,21 @@ def encode_transaction_data(
...
+    normalizers: Sequence[Callable[[TypeStr, Any], Tuple[TypeStr, Any]]] = None,
+) -> Dict[str, Any]:
+    # type ignored b/c expects data arg to be HexBytes
+    data = HexBytes(data)  # type: ignore
+    types = get_abi_input_types(fn_abi)
+    abi_codec = ABICodec(default_registry)
+    decoded = abi_codec.decode(types, HexBytes(data[4:]))
+    if normalizers:
+        decoded = map_abi_data(normalizers, types, decoded)
+    return named_tree(fn_abi["inputs"], decoded)
```
fselmo, line 334: I understand the reasoning for wanting all regular tuples to be named tuples but I think it's a bit confusing / misleading to convert it when the flag isn't set on the contract instance. My biggest worry is that as a user I don't know what's going to come out as a tuple type and what will be a `ABIDecodedNamedTuple`. I think it would be ideal if the flag was an on/off switch. Am I missing a key idea of this PR or does that seem like a reasonable assumption?

This would apply for `get_event_data()` as well as it looks like this PR makes it default behavior there as well.

web3/_utils/events.py
```diff
@@ -255,6 +256,10 @@ def get_event_data(
     normalized_log_data = map_abi_data(
         BASE_RETURN_NORMALIZERS, log_data_types, decoded_log_data
     )
+    named_log_data = named_tree(
```
fselmo, line 259: See my comment on `decode_function_input()`... I think it may be ideal to created `ABIDecodedNamedTuples` only if requested by the user. This probably looks like a flag on this method which, when called internally in `contract.py` and `async_contract.py` would be just passing along the `contract.decode_tuples` flag.

web3/contract/utils.py
```diff
@@ -129,6 +132,10 @@ def call_contract_function(
     )
     normalized_data = map_abi_data(_normalizers, output_types, output_data)
 
+    if decode_tuples:
```
fselmo, line 135: This seems like the ideal behavior for all cases. If requested, convert to `ABIDecodedNamedTuple` types.

fselmo (commented):
This is looking good. I had some final thoughts I wanted to discuss before approving. I think the best approach would be a user choice approach where in call cases the `ABIDeocdedNamedTuple` is an opted for choice. Curious on your input here and from @kclowes

docs/web3.contract.rst
```diff
@@ -146,12 +146,29 @@ Each Contract Factory exposes the following properties.
...
+
+.. py:attribute:: Contract.decode_tuples
+
+    If a Tuple/Struct is returned by a contract function, this flag defines whether
+    to apply the field names from the ABI to the returned data.
+    If False, the returned value will be a normal Python `Tuple`. If True, the returned
+    value will be a Python `NamedTuple` of the class `ABIDecodedNamedTuple`.
+    
+    NamedTuples have some restrictions regarding field names.
+    Web3.py sets `NamedTuple`'s `rename=True`, so disallowed field names may be
```
fselmo, line 158:
```suggestion
    web3.py sets `NamedTuple`'s `rename=True`, so disallowed field names may be 
```

re: rebranding, heh

web3/_utils/abi.py
```diff
@@ -906,3 +907,68 @@ def build_strict_registry() -> ABIRegistry:
...
+    names = [item["name"] for item in abi]
+    items = [named_subtree(*item) for item in zip(abi, data)]
+
+    return dict(zip(names, items))
+
+
+def named_subtree(
+    abi: Union[ABIFunctionParams, ABIFunction, ABIEvent, Dict[TypeStr, Any]],
+    data: Tuple[Any, ...],
+) -> Union[Dict[str, Any], Tuple[Any, ...], Any]:
```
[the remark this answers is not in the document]
pacrob: Looks like if/when we get to `mypy==0.990`, [recursive typing](https://stackoverflow.com/a/53639221/11672028) works. For now, I changed the type to  `List[Any]` and made it `_named_subtree`.

I do think having them separate does make for better readability.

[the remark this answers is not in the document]
fselmo: Ah yeah, nice catch on the recursive typing. `List[Any]` works well 👌🏼

fselmo (approved):
Ignore my previous ping, this lgtm in current state... 🎉  I'll resolve those conversations

docs/web3.contract.rst
pacrob: done

PR #2805, opened 2023-02-03
Remove ``AttributeDict`` method formatters and opt for middleware

### LINKED ISSUE #1656: Typing of AttributeDict return values is broken
* Version: 5.10.0
* Python: 3.6
* OS: linux

### What was wrong?
Many data like events are returned as an `AttributeDict`, so that `event['event']` and `event.event` both work. The typings however inherit from `TypedDict`, for example for `EventData`, so that only `event['event']` correctly typechecks, but not `event.event`. This is annoying, as perfectly valid code does not pass typechecking. 

### How can it be fixed?
Somehow allow (similar to the AttributeDict for runtime) for every key to in the `TypedDict` to bypass typechecking also for the attribute. I sadly do not know how to do this.

fselmo:
### What was wrong?

- ``AttributeDict`` leads to typing complications when properties are accessed via attribute, rather than key / value. This doesn't complicate too much since a user can opt for key / value every time but it isn't ideal to provide an option that breaks typing.

closes #1656

### How was it fixed?

The approach suggested in #1656 does not feel like a great resolution since `NamedTuple` will rename any reserved python keywords to `_0`, `_1`, `_2`, etc... Seeing as JSON-RPC responses commonly have properties like `from`, for example, this would not lead to a very good user experience.

- ``AttributeDict`` conversion was removed from the result formatters and moved back to being only in the middleware. A more inclusive recursive approach was taken to make sure all results get passed through the recursive checks for conversion of `dict` to `AttributeDict`. Users who wish to turn off all recursive conversion to ``AttributeDict`` can simply remove the ``attrdict_middleware`` (or ``async_attrdict_middleware``) and not worry about possible typing complications.

- In order for this to work for async as well, support for ``async_attrdict_middleware`` was introduced. This is also now one of the default middlewares for the ``EthereumTesterProvider`` (and `AsyncEthereumTesterProvider`) as `dict` -> `AttributeDict` was the default behavior, via result formatters, for `EthereumTesterProvider`.

### Todo:
- [x] Fix some broken tests that expect the attrdict result formatter to be present
- [x] Update docs / add new documentation around these changes.
- [x] Add new tests for ``attrdict_middleware`` + ``async_attrdict_middleware``

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -40,9 +40,11 @@ you can find the latest block number in these two ways:
         TypeError: This data is immutable -- create a copy instead of modifying
 
 This feature is available via the ``attrdict_middleware`` which is a default middleware.
-Of note, accessing a property via attribute will break type hinting. If typing is
-crucial for your application, accessing via key / value, as well as removing the
-``attrdict_middleware`` altogether, may be desired.
+
+.. note::
+    Accessing an ``AttributeDict`` property via attribute will break type hinting. If
+    typing is crucial for your application, accessing via key / value, as well as
+    removing the ``attrdict_middleware`` altogether, may be desired.
 
 
 Properties
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -16,7 +16,7 @@ web3.eth API
 The ``web3.eth`` object exposes the following properties and methods to
 interact with the RPC APIs under the ``eth_`` namespace.
 
-Often, when a property or method returns a mapping of keys to values, it
+By default, when a property or method returns a mapping of keys to values, it
 will return an ``AttributeDict`` which acts like a ``dict`` but you can
 access the keys as attributes and cannot modify its fields. For example,
 you can find the latest block number in these two ways:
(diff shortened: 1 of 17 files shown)


docs/web3.eth.rst
```diff
@@ -39,6 +39,11 @@ you can find the latest block number in these two ways:
         Traceback # ... etc ...
         TypeError: This data is immutable -- create a copy instead of modifying
 
+This feature is available via the ``attrdict_middleware`` which is a default middleware.
+Of note, accessing a property via attribute will break type hinting. If typing is
```
kclowes, line 43: This might be better as a `..note ::` box thing. Feel free to take or leave though.

web3/providers/eth_tester/middleware.py
```diff
@@ -368,14 +370,13 @@ async def async_fill_default(
...
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
-    fill_default_from = fill_default("from", guess_from, w3)
-
     def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
         if method in (
             "eth_call",
             "eth_estimateGas",
             "eth_sendTransaction",
         ):
+            fill_default_from = fill_default("from", guess_from, w3)
```
kclowes, line 379: That's wild that these are the only methods that need a default `from` 🤔

kclowes (approved):
LGTM! Just had one nit in the docs, feel free to take or leave.

pacrob (approved):
lgtm!

PR #2819, opened 2023-02-16
Async web3

fselmo:
### What was wrong?

Related to Issue #1416 

- Separates ``AsyncWeb3`` from ``Web3`` in an effort to make async a first class citizen of the library. Ideally we refactor the structure of the files so they are organized as such for developers. I feel like there are enough changes here where re-organization should be a separate PR, which will be quite large as well.
- Create ``AsyncMiddleware`` type that expects ``AsyncWeb3`` rather than ``Web3``. Account for this change down the tree through async middlewares and middleware onion.
- Remove all cases of ``type: ignore`` that were fixed by the tighter typing for async web3 and related modules.

Things to note:

- ``AsyncContractCaller`` inherits from ``BaseContractCaller`` which used the synchronous method ``parse_block_identifier`` in its ``__init__()`` method. This leads me to think we weren't testing all cases for initialization of the contract caller. This came up as an issue with typing and was addressed using the simpler ``parse_block_identifier_no_extra_call()`` method which technically shouldn't break anything since the previous structure would return a coroutine and also not work. This may need to be addressed in the future if we want to try to account for all elif cases that ``ContractCaller`` accounts for.

### Todo:
- [x] ~~Update documentation + document new async flow~~ (separate PR, see #2821)


web3/eth/async_eth.py
```diff
@@ -545,3 +554,35 @@ async def get_filter_logs(self, filter_id: HexStr) -> List[LogReceipt]:
 
     async def uninstall_filter(self, filter_id: HexStr) -> bool:
         return await self._uninstall_filter(filter_id)
+
+    @overload
+    def contract(self, address: None = None, **kwargs: Any) -> Type[AsyncContract]:
+        ...  # noqa: E704,E501
```
kclowes, line 560: I think you might be able go get rid of all of these `noqa`s

web3/main.py
```diff
@@ -321,6 +280,12 @@ def keccak(
             "keccak(b'\\x74\\x78\\x74'), or keccak(0x747874)."
         )
 
+    @classmethod
+    def normalize_values(
+        cls, w3: "BaseWeb3", abi_types: List[TypeStr], values: List[Any]
```
kclowes, line 285: I don't think you need this 'w3' arg here?

kclowes (commented):
Hooray! :rocket: I left some comments around adding a new `TypeVar` instead of `Union["ContractFunction", "AsyncContractFunction"]`, that I think can help avoid some of the `cast`ing we're doing in the `contract`/`async_contract` files. It was sort of hard to figure out where to put the comments, so lmk if they don't make sense and we can hop on a call!

ens/ens.py
```diff
@@ -87,6 +87,9 @@ class ENS(BaseENS):
     like: ``"0x314159265dD8dbb310642f98f50C066173C1259b"``
     """
 
+    # mypy types
+    w3: "Web3"
```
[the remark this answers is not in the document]
fselmo: We do since `BaseENS` defines `Union["ENS", "AsyncENS"]`. I can either be here or it can be in the `__init__()` method when we assign it `self.w3: "Web3" = init_web3(...)`. It seemed consistent with the other `Module` to set it this way but I could go either way. Any preference?

fselmo (commented):
@kclowes I addressed your comments in the latest commit. ~~It looks like some typing broke for `filter_builder.args` when I added the `TypeVar` suggestion. I ignored those for now. It's not immediately obvious to me where that broke.~~ [nevermind, fixed 🎉]

ens/ens.py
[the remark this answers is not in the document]
kclowes: Got it. No preference, I was mostly just wondering!

web3/_utils/async_transactions.py
```diff
@@ -43,25 +42,28 @@
 )
 
 if TYPE_CHECKING:
-    from web3 import Web3  # noqa: F401
     from web3.eth import AsyncEth  # noqa: F401
+    from web3.main import (  # noqa: F401
```
pacrob, line 46: Here it's `from web3.main import`, but we also use `from web3 import` all over the place and I can't find a pattern. Can we just use one or the other?

web3/geth.py
```diff
@@ -68,253 +53,407 @@
...
-    async def new_account(self, passphrase: str) -> Awaitable[ChecksumAddress]:
-        return await self._new_account(passphrase)  # type: ignore
 
-    async def send_transaction(
-        self, transaction: TxParams, passphrase: str
-    ) -> Awaitable[HexBytes]:
-        return await self._send_transaction(transaction, passphrase)  # type: ignore
+class GethMiner(Module):
+    """
+    https://github.com/ethereum/go-ethereum/wiki/Management-APIs#miner
```
pacrob, line 208: This goes to a redirect page, which then 404s. New link, or just remove?

web3/contract/async_contract.py
```diff
@@ -471,26 +558,51 @@ def build_filter(self) -> AsyncEventFilterBuilder:
...
+                    func["name"],
+                    w3=self.w3,
+                    contract_abi=self.abi,
+                    address=self.address,
+                    function_identifier=func["name"],
+                    decode_tuples=decode_tuples,
+                )
+
+                # TODO: The no_extra_call method gets around the fact that we can't call
+                #  the full async method from within a class's __init__ method. We need
```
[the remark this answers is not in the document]
pacrob: The `__await__` magic method allows for things to be awaited at creation. There are a few examples of it in the codebase.

pacrob (commented):
I left a couple comments, but overall lookin' good! I love all those `# type: ignore`s going away.

web3/_utils/async_transactions.py
fselmo: Yeah. I guess `from web3.main` might be "best" since the `web3/__init__.py` imports it from `web3.main` so it's a bit of a hand-off. I don't think it matters too much. There may be some things that are defined in `web3.main` that we don't expose via `web3/__init__.py` at some point too. For example, I'm not sure that it makes much sense to expose `BaseWeb3` there since that's not a complete class that is ready for use.

web3/geth.py
fselmo: Good catch. Updated 👍🏼

web3/contract/utils.py
```diff
@@ -52,11 +50,17 @@
...
 )
 
 if TYPE_CHECKING:
-    from web3 import Web3  # noqa: F401
+    from web3 import (  # noqa: F401
+        AsyncWeb3,
+        Web3,
+    )
+    from web3.contract.async_contract import AsyncContractFunction  # noqa: F401
+    from web3.contract.contract import ContractFunction  # noqa: F401
```
kclowes, line 63: nit: I don't think these are getting used.

kclowes (approved):
LGTM! 🚢 ! Just had a tiny nit.

Issue #2826, opened 2023-02-17
Add async eth.sign


kclowes:
### What was wrong?

Some methods are still unavailable on the `AsyncEth` class.

### How can it be fixed?

Following the pattern already established, add the `sign` method to the `AsyncEth` class, and add corresponding tests. The `AsyncHTTPProvider` docs in `docs/provider.rst` will also need to be updated.

e3243eric:
I would like to work on this issue. I will make a PR for it this week.

sujantkumarkv:
I guess the changes were merged. I then suggest closing this issue. Thanks :)
@kclowes

fselmo:
closed by #2833

PR #2833, opened 2023-02-21
Add async eth.sign

e3243eric:
### What was wrong?

Related to Issue #2826. The `sign` method is unavailable on the `AsyncEth` class.

### How was it fixed?

Add the `sign` method to the `AsyncEth` class, and add corresponding tests.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -1454,6 +1454,54 @@ class AsyncEthModuleTest:
         assert is_integer(transaction_count)
         assert transaction_count >= 1
 
+    async def test_async_eth_sign(
+        self, async_w3: "AsyncWeb3", unlocked_account_dual_type: ChecksumAddress
+    ) -> None:
+        signature = await async_w3.eth.sign(
+            unlocked_account_dual_type, text="Message tö sign. Longer than hash!"
+        )
+        assert is_bytes(signature)
+        assert len(signature) == 32 + 32 + 1
+
+        # test other formats
+        hexsign = await async_w3.eth.sign(
+            unlocked_account_dual_type,
+            hexstr=HexStr(
+                "0x4d6573736167652074c3b6207369676e2e204c6f6e676572207468616e206861736821"  # noqa: E501
+            ),
+        )
+        assert hexsign == signature
+
+        intsign = await async_w3.eth.sign(
+            unlocked_account_dual_type,
+            0x4D6573736167652074C3B6207369676E2E204C6F6E676572207468616E206861736821,
+        )
+        assert intsign == signature
+
+        bytessign = await async_w3.eth.sign(
+            unlocked_account_dual_type, b"Message t\xc3\xb6 sign. Longer than hash!"
+        )
+        assert bytessign == signature
+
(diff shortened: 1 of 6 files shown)


web3/_utils/module_testing/eth_module.py
```diff
@@ -1465,6 +1465,52 @@ async def test_eth_getBlockTransactionCountByNumber_block_with_txn(
...
+            unlocked_account_dual_type, text="different message is different"
+        )
+        assert new_signature != signature
+
+    @pytest.mark.asyncio
+    async def test_async_eth_sign_ens_names(
+        self, async_w3: "Web3", unlocked_account_dual_type: ChecksumAddress
+    ) -> None:
+        with ens_addresses(async_w3, {"unlocked-acct.eth": unlocked_account_dual_type}):
+            signature = async_w3.eth.sign(  # type: ignore
```
kclowes, line 1508: It looks like the integration tests are failing on this line because it's missing an `await`. And it looks like the lint job is failing here too because this `# type: ignore` isn't being used, so you can probably delete and be good to go!

kclowes (commented):
Thanks for the contribution! I left a comment on the line where the tests are failing. It should be a quick fix, but I'm happy to do it if you don't want to/have time. Just let me know. Looks good otherwise!

web3/_utils/module_testing/eth_module.py
e3243eric: I tried these commits before, but the integration tests failed because the ENS "unlocked-acct.eth" didn't seem to convert to a hex string.
I tried to replace `ens_addresses` with an asynchronous generator, but I still got the same error.

web3/eth/async_eth.py
```diff
@@ -518,6 +521,31 @@ async def get_storage_at(
     ) -> HexBytes:
         return await self._get_storage_at(account, position, block_identifier)
 
+    # eth_sign
+
+    def sign_munger(
```
kclowes, line 526: This method can get moved to BaseEth and then called on line 537 as: `BaseEth.sign_munger`

kclowes (commented):
This looks good, thank you! I left an inline comment saying this same thing, but the `sign_munger` can get moved into the `BaseEth` class and then just called like: `mungers=[BaseEth.sign_munger]` in the `Eth` and `AsyncEth` classes. Once that's updated, we should be good to merge!

e3243eric:
Ok, I moved the `sign_munger` to the `BaseEth` class and updated it in the `Eth` and `AsyncEth` classes.

kclowes (approved):
Looks great! Thanks @e3243eric!

web3/_utils/module_testing/eth_module.py
```diff
@@ -1454,6 +1454,54 @@ async def test_eth_getBlockTransactionCountByNumber_block_with_txn(
         assert is_integer(transaction_count)
         assert transaction_count >= 1
 
+    async def test_async_eth_sign(
```
e3243eric, line 1457: Ah, sorry, I found out that `test_async_eth_sign` is missing `@pytest.mark.asyncio`, I didn't check its CI warnings.

PR #2842, opened 2023-02-23
Change ``eth_call``  return type to reflect pythonic formatter conver…

### LINKED ISSUE #2836: The return type of eth_call should be HexBytes instead of bytes
* Version: 6.0.0-beta.10
* Python: 3.10
* OS: Linux(WSL)

### What was wrong?

The return type annotation of `eth_call` is `bytes | bytearray`, but actually uses `hexbytes.HexBytes` as the return type. But since the hexbytes library has compatibility issues(ethereum/hexbytes#14), it can bother users developing and create unexpected problems.
It looks like this issue has been reported to the hexbytes library long time ago, but it has never been released. And there is no expectation that this will be fixed. As a project that uses hexbytes as a dependency, it is necessary to explicitly tell users the correct specific return type, so that users can directly code for the type (to fix the compatibility issue).

A minimal reproducible example is as follows:
```python
import asyncio
import typing

from web3 import AsyncHTTPProvider, Web3
from web3.eth.async_eth import AsyncEth
from web3.types import TxParams

addr = Web3.to_checksum_address(f"0x8ad599c3a0ff1de082011efddc58f1908eb6e6d8")

txn: TxParams = {"to": addr, "data": bytes.fromhex("0x3850c7bd"[2:])}


async def main():
    url = "http://localhost:8545/"
    w3 = Web3(AsyncHTTPProvider(url))
    result = await typing.cast(AsyncEth, w3.eth).call(txn, "latest")
    print(type(result), f"0x{result.hex()}")


asyncio.run(main())
```
The output is as follows:
```
<class 'hexbytes.main.HexBytes'> 0x0x000000...00000001
```
It can be observed that `0x` appears twice, due to the inconsistency in `hex` between HexBytes and bytes.

### How can it be fixed?

Explicitly declare the return type annotation of `eth_call` as `HexBytes`, which allows users to find such inconsistencies faster and use the `isinstance` method to solve them. This saves a lot of debugging time.

fselmo:
### What was wrong?

closes #2836

### How was it fixed?

- More accurately define the return type for ``eth_call`` as ``HexBytes``

### Todo:

### DIFF
--- a/newsfragments/2842.bugfix.rst
+++ b/newsfragments/2842.bugfix.rst
@@ -1,0 +1,1 @@
+More accurately define the ``eth_call`` return type as ``HexBytes`` since the response is converted to ``HexBytes`` in the pythonic formatters and there are differences between ``HexBytes`` and ``bytes`` types.
--- a/web3/eth/async_eth.py
+++ b/web3/eth/async_eth.py
@@ -225,7 +225,7 @@ class AsyncEth(BaseEth):
                 Optional[BlockIdentifier],
                 Optional[CallOverride],
             ],
-            Awaitable[Union[bytes, bytearray]],
+            Awaitable[HexBytes],
         ]
     ] = Method(RPC.eth_call, mungers=[BaseEth.call_munger])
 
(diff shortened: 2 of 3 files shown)


kclowes (approved):
LGTM! 🚢

pacrob (approved):
lgtm!

PR #2843, opened 2023-02-23
bugfix: Use correct types and formatters for call overrides

### LINKED ISSUE #2832: Cannot unmarshal non-string into Go struct field OverrideAccount.balance of type *hexutil.Big
* Version: 6.0.0-beta.10
* Python: 3.10
* OS: Linux(WSL)

### What was wrong?
Occurred an exception when calling `eth_call` with `state_override`. The exception message is as follows:
```
ValueError: {'code': -32602, 'message': 'invalid argument 2: json: cannot unmarshal non-string into Go struct field OverrideAccount.balance of type *hexutil.Big'}
```
I found this problem when using `AsyncHTTPProvider` to connect my Geth node. I overrode the balance of an account in the `state_override` parameter, and this exception occurred at runtime.

A minimal example to reproduce the problem is as follows:
```python
import asyncio
import secrets
import typing

from eth_typing import ChecksumAddress
from web3 import AsyncHTTPProvider, Web3
from web3.eth.async_eth import AsyncEth
from web3.types import TxParams, CallOverrideParams, Wei, Nonce

rand: bytes = secrets.token_bytes(20)
addr = Web3.to_checksum_address(f"0x{rand.hex()}")

txn: TxParams = {"to": addr}
over: dict[ChecksumAddress, CallOverrideParams] = {
    addr: {
        "balance": Wei(0)
    }
}


async def main():
    url = "http://localhost:8545/"
    w3 = Web3(AsyncHTTPProvider(url))
    await typing.cast(AsyncEth, w3.eth).call(txn, "latest", over)


asyncio.run(main())

```

Running this example, I get the following error message:
```
  File "/.../lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/.../lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "<stdin>", line 3, in main
  File "/.../lib/python3.10/site-packages/web3/eth/async_eth.py", line 240, in call
    return await self._durin_call(transaction, block_identifier, state_override)
  File "/.../lib/python3.10/site-packages/web3/eth/async_eth.py", line 259, in _durin_call
    return await self._call(transaction, block_identifier, state_override)
  File "/.../lib/python3.10/site-packages/web3/module.py", line 90, in caller
    result = await w3.manager.coro_request(
  File "/.../lib/python3.10/site-packages/web3/manager.py", line 236, in coro_request
    return self.formatted_response(
  File "/.../lib/python3.10/site-packages/web3/manager.py", line 194, in formatted_response
    raise ValueError(response["error"])
ValueError: {'code': -32602, 'message': 'invalid argument 2: json: cannot unmarshal non-string into Go struct field OverrideAccount.balance of type *hexutil.Big'}
```

### How can it be fixed?

According to the error report, a hex string is needed to convert to `*hexutil.Big`, but the corresponding type in web3.py is `Wei`. Web3.py should convert the numbers here into hex string and pass  them to Geth.

It seems that all parameters covered here needed to be converted to hex string. `Nonce` needs to be converted to `*hexutil.Uint64` after my testing.

fselmo:
### What was wrong?

closes #2832

### How was it fixed?

- Fix typing for `CallOverrideParams` based on the [expected types](https://github.com/ethereum/go-ethereum/blob/master/ethclient/gethclient/gethclient.go#L224-L244).
- Use proper formatters for call override params
- Add testing

### Todo:

### DIFF
--- a/newsfragments/2843.breaking.rst
+++ b/newsfragments/2843.breaking.rst
@@ -1,0 +1,1 @@
+Fix typing for ``CallOverrideParams`` and add proper request formatters for call state overrides.
--- a/tests/integration/test_ethereum_tester.py
+++ b/tests/integration/test_ethereum_tester.py
@@ -457,8 +457,14 @@ class TestEthereumTesterEthModule(EthModuleTest):
     @pytest.mark.xfail(
         raises=TypeError, reason="call override param not implemented on eth-tester"
     )
-    def test_eth_call_with_override(self, w3, revert_contract):
-        super().test_eth_call_with_override(w3, revert_contract)
+    def test_eth_call_with_override_code(self, w3, revert_contract):
+        super().test_eth_call_with_override_code(w3, revert_contract)
+
+    @pytest.mark.xfail(
+        raises=TypeError, reason="call override param not implemented on eth-tester"
+    )
+    def test_eth_call_with_override_param_type_check(self, w3, math_contract):
+        super().test_eth_call_with_override_param_type_check(w3, math_contract)
 
     def test_eth_call_revert_with_msg(self, w3, revert_contract, unlocked_account):
         with pytest.raises(
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -85,6 +85,9 @@ from web3._utils.rpc_abi import (
     RPC_ABIS,
     abi_request_formatters,
 )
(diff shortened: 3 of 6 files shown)


kclowes (approved):
:tada: !

PR #2846, opened 2023-02-24
Update contract.py

leafyoung:
### What was wrong?

`contract_function.call()` returns the latest result when `w3.eth.default_block` has been set.

The default value "latest" for `block_identifier` parameter masks `w3.eth.default_block` even when it is set to non-None.

`w3.eth.default_block` is by default set to 'latest'. Changing the default value to None will not change its default behaviour.

If we do not change this, we have to add `block_identifier=past_block_number` to every `.call()`. This breaks existing codes using web3.py and it's very tedious to do so.

Related to Issue #

I found this wrong behaviour when running regression test with an earlier version of web3.py

### How was it fixed?

Change the default value to `None` to make `parse_block_identifier` do the work.

### Todo:

Set default block_identifier in ContractFunction.call() to None

#### Cute Animal Picture

phoenix.


kclowes:
Thanks for raising this issue @leafyoung. This looks like a good change to me. I found a couple other places where we're doing the same thing, so I'll update those and push it up.

kclowes (approved):
Thanks @leafyoung!

leafyoung:
Thanks @kclowes

Issue #2824, opened 2023-02-17
Add async eth.replace_transaction


kclowes:
### What was wrong?

Some methods are still unavailable on the `AsyncEth` class.

### How can it be fixed?

Following the pattern already established, add `replace_transaction` to the `AsyncEth` class, and add corresponding tests.

kclowes:
Closed via #2847

PR #2847, opened 2023-02-24
Add async eth.replace_transaction

e3243eric:
### What was wrong?

Related to Issue #2824. The `replace_transaction` method is unavailable on the `AsyncEth` class.

### How was it fixed?

Add the `replace_transaction` method to the `AsyncEth` class, and add corresponding tests.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -1548,6 +1548,251 @@ class AsyncEthModuleTest:
             assert len(signature) == 32 + 32 + 1
 
     @pytest.mark.asyncio
+    async def test_async_eth_replace_transaction_legacy(
+        self, async_w3: "AsyncWeb3", unlocked_account_dual_type: ChecksumAddress
+    ) -> None:
+        txn_params: TxParams = {
+            "from": unlocked_account_dual_type,
+            "to": unlocked_account_dual_type,
+            "value": Wei(1),
+            "gas": 21000,
+            "gasPrice": async_w3.to_wei(
+                1, "gwei"
+            ),  # must be greater than base_fee post London
+        }
+        txn_hash = await async_w3.eth.send_transaction(txn_params)
+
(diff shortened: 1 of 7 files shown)


web3/_utils/module_testing/eth_module.py
```diff
@@ -1547,6 +1547,251 @@ async def test_async_eth_sign_ens_names(
...
+            "from": unlocked_account_dual_type,
+            "to": unlocked_account_dual_type,
+            "value": Wei(1),
+            "gas": 21000,
+            "maxFeePerGas": async_w3.to_wei(2, "gwei"),
+            "maxPriorityFeePerGas": async_w3.to_wei(1, "gwei"),
+        }
+        txn_hash = await async_w3.eth.send_transaction(txn_params)
+        try:
+            async_w3.geth.miner.start()  # type: ignore
```
e3243eric, line 1669: The AsyncGethMiner is not implemented, so I mark this test as an expected failure.

web3/_utils/transactions.py
```diff
@@ -219,7 +223,7 @@ def assert_valid_transaction_params(transaction_params: TxParams) -> None:
 
 
 def prepare_replacement_transaction(
-    w3: "Web3",
+    w3: Union["Web3", "AsyncWeb3"],
```
e3243eric, line 226: The `prepare_replacement_transaction` doesn't use any async method, so I just reuse the sync version of `prepare_replacement_transaction`. I'm not sure if it is better to reuse it in this way or copy it to async_transactions.py.

web3/_utils/module_testing/eth_module.py
e3243eric: The go-ethereum [`miner`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-miner) namespace is now deprecated, but I can help implement it if needed, just like the other async Geth namespaces.

kclowes (approved):
This looks good to me! Thanks @e3243eric!

PR #2849, opened 2023-02-25
Use TypedDict for **kwargs type checking

e3243eric:
### What was wrong?

https://github.com/ethereum/web3.py/blob/168fceaf5c6829a8edeb510b997940064295ecf8/web3/eth/eth.py#L551-L555
After [Mypy 0.981](https://mypy-lang.blogspot.com/2022/09/mypy-0981-released.html), it supports precise type annotations for `**kwargs` using TypedDict (https://github.com/python/mypy/pull/13471), like `**transaction_params: Any` can be rewritten as `**transaction_params: Unpack[TxParams]`.

### How was it fixed?

- Upgrade Mypy to 0.981.
- Update `**kwarg` type checking with `Unpack[TxParams]`.
- Add `Unpack` type. The `Unpack` is a built-in type after Python 3.11, so I updated the `web3/_utils/compat`.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/compat/__init__.py
+++ b/web3/_utils/compat/__init__.py
@@ -11,4 +11,5 @@
 from typing_extensions import (
     NotRequired,  # py311
     Self,  # py311
+    Unpack,  # py311
 )
--- a/web3/_utils/events.py
+++ b/web3/_utils/events.py
@@ -444,8 +444,8 @@ class EventFilterBuilder(BaseEventFilterBuilder):
         if not isinstance(w3, web3.Web3):
             raise Web3ValueError(f"Invalid web3 argument: got: {w3!r}")
 
-        for arg in AttributeDict.values(self.args):
-            arg._immutable = True  # type: ignore[attr-defined]
+        for arg in self.args.values():
+            arg._immutable = True
         self._immutable = True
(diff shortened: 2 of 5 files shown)


e3243eric:
The remaining lint errors are from the submodule "ethpm-spec".

pacrob:
`Unpack` does look useful, but we can't include something that is labeled "experimental" and "incomplete" in a release. Thanks for bringing it up, though. I'll leave this open and check back on it in a few months.

fselmo:
As far as I know this is still an experimental feature. I'm open to including it, but I'm not certain that it's desirable at this stage. Considering this for the next major release, `v8`, might be best.

pacrob:
`Unpack` is out of experimental in more recent `mypy` versions. I've rebased and applied the changes to the async versions. Can discuss inclusion timeline.

fselmo:
Nice find. I think we should aim to put in `v7` then. Thoughts?

pacrob (approved):

pacrob:
No reason not to include in `v7` from me.

fselmo (approved):
:shipit:

PR #2855, opened 2023-02-27
Drop down resource class in CI

kclowes:
### What was wrong?
Apparently we don't have access to the xlarge resource class any more. I suspect something happened with billing on the devops side of the house. 

### How was it fixed?
Dropped down to large resource_class which seems to fix things. 

### Todo:

#### Cute Animal Picture


kclowes:
I don't think this was a real problem. Closing.

PR #2857, opened 2023-03-02
Deprecate Geth miner namespace

### LINKED ISSUE #2856: Deprecate `miner` namespace support and tests
### What was wrong?

Geth removed support for their `miner` API since it's no longer needed after The Merge ™️ . See: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-miner

### How can it be fixed?

Add deprecation warnings to the methods within the miner namespace, and change the tests to check for the deprecation warnings. We can then either remove the methods as Geth stops supporting these, or wait until v7 to remove.

e3243eric:
### What was wrong?

Closes #2856

Geth removed support for their [miner API](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-miner).

### How was it fixed?

Change `GethMiner` methods to `DeprecatedMethod`.

### Todo:

#### Cute Animal Picture


e3243eric:
I wonder why the tests in `tests/core/mining-module` are always skipped.

kclowes:
Thank you!

> I'm wondering why the tests in tests/core/mining-module are always skipped.

The core tests run with the `eth-tester` provider which doesn't have mining implemented.

I'm hoping I'll be able to review this later in the week!

newsfragments/2857.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecate Geth miner namespace
```
reedsa: Rename this file as `2857.deprecation.rst`

tests/core/mining-module/test_miner_start.py
```diff
@@ -26,7 +28,8 @@ def test_miner_start(w3_empty, wait_for_miner_start):
     assert not w3.eth.mining
     assert not w3.eth.hashrate
 
-    w3.miner.start(1)
+    with pytest.warns(DeprecationWarning):
+        w3.geth.miner.start(1)
```
reedsa: Thanks for updating the namespace!

reedsa (commented):
One nit about the newsfragment filename, otherwise lgtm! @kclowes do you have any feedback here?

web3/_utils/miner.py
```diff
@@ -18,49 +19,62 @@
...
+#
+# The Geth client deprecated the miner namespace.
+#
+
+_make_dag: Method[Callable[[BlockNumber], bool]] = Method(
     RPC.miner_makeDag,
     mungers=[default_root_munger],
 )
 
+make_dag = DeprecatedMethod(_make_dag, "make_dag", "deprecated")
```
kclowes: Nice use of `DeprecatedMethod`! I think since the only place we're using it is in this file, we can change it to better suit our purposes. Because we're not replacing the method with anything, we can update the warning [here](https://github.com/ethereum/web3.py/blob/main/web3/method.py#L250) to say something along the lines of f"{self.old_name} is no longer supported." 

Or, if you really wanted to go for some extra credit, you could change the `DeprecatedMethod` class to take a new, optional argument called `message` and then on `__get__` we could return the message if there is one. So the new DeprecatedMethod class might look something like:

```

class DeprecatedMethod:
    def __init__(
        self,
        method: Method[Callable[..., Any]],
        old_name: str,
        new_name: Optional[str] = None,
        msg: Optional[str] = None,
    ) -> None:
        self.method = method
        self.old_name = old_name
        self.new_name = new_name
        self.msg = msg

    def __get__(
        self, obj: Optional["Module"] = None, obj_type: Optional[Type["Module"]] = None
    ) -> Any:
        if self.msg is not None:
            warnings.warn(self.msg)
        else:
            warnings.warn(
                f"{self.old_name} is deprecated in favor of {self.new_name}",
                category=DeprecationWarning,
            )
        return self.method.__get__(obj, obj_type)
```

And this line would look something like:
```suggestion
make_dag = DeprecatedMethod(_make_dag, "make_dag", msg="All mining methods have been deprecated")
```

kclowes (commented):
Thanks for reviving this PR, @reedsa! 

@e3243eric I made a comment around the messaging in the `DeprecatedMethod` class, let me know what you think. I'm happy to make that update if you don't want/have time to, just let me know!

Thanks for taking it on!

pacrob (approved):
lgtm! - but should be to v6, removed for v7, yes?

fselmo:
> lgtm! - but should be to v6, removed for v7, yes?

Yep, along with #3236, this should close it all out

reedsa:
Rebased this against `main` by mistake, closing in favor of #3238

Issue #2814, opened 2023-02-14
Question: safety of multiple HTTPProviders


menaitm:
* Version: 5.31.1
* Python: 3.9

### Example

A single python process which can connect to multiple RPC endpoints may look like this:
```python
chain_a = Web3(Web3.HTTPProvider(RPC_NODE_ETHEREUM))
chain_b = Web3(Web3.HTTPProvider(RPC_NODE_ARBITRUM))
chain_c = Web3(Web3.WebsocketProvider(RPC_NODE_OPTIMISM))

chain_a.eth.get_balance()
chain_b.eth.get_balance()
chain_c.eth.get_balance()
```

However, [the docs](https://github.com/ethereum/web3.py/blob/master/docs/providers.rst) state you should not create multiple HTTPProviders:

> Note that you should create only one HTTPProvider per python process, as the HTTPProvider recycles underlying TCP/IP network connections, for better performance.

#### Questions

1. Is this still true?
2. Is there a workaround that doesn't require using multiple `WebsocketProvider`s instead?
3. Does this apply to `AsyncHTTPProvider` too?

Thanks

menaitm:
I'm not sure if this relates to the session caching that was recently changed here: https://github.com/ethereum/web3.py/issues/1847#issuecomment-1334046678

kclowes:
> 1. Is this still true? I see the session cache uses the endpoint_uri as the key, so does this comment actually mean that you should only create one HTTPProvider per RPC URI per python process?

Currently it is, but only if you're using the same URL. The PR linked is a fix to allow multiple Web3 instances with the same URL too.

> 2.  Is there a workaround that doesn't require using multiple WebsocketProviders instead?

Not that I know of, but there might be. 

> 3.  Does this apply to AsyncHTTPProvider too?

I believe so. The PR mentioned above should fix that though, and again, only if you're using the same provider URL. 

> 4  Does the request/response model of the WebsocketProvider mean you should not use asyncio.gather() to fire off multiple requests concurrently (to prevent race conditions when receiving messages in a different order back from the node)?

Right. You should not use `asyncio.gather` with the websocket provider.

kclowes:
After thinking further about allowing multiple web3 instances with the same provider type and URL, we decided to highlight and discourage that use case in the docs. The fix introduces a lot of complexity for what seems like a pretty small edge case. 

@menaitm - if you (or anyone else following along) have a use case for this feature though that they'd like to advocate for, we'd be happy to reconsider.

PR #2861, opened 2023-03-06
Add provider id to allow multiple instances of the HTTPProvider

kclowes:
### What was wrong?
There was an issue that came in asking if it was safe to use multiple HTTPProviders. It was safe if they had different URLs, but now it is safe to use even with the same URL. 

I will squash commits before merging!

Related to Issue #2814

### How was it fixed?

Added a provider_id to the `HTTPProvider` (and the `EthTesterProvider`), so that multiple instances of the HTTPProvider will not reuse the same underlying TCP/IP connections. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/request.py
+++ b/web3/_utils/request.py
@@ -147,10 +147,16 @@ _async_session_pool = ThreadPoolExecutor(max_workers=1)
 
 async def async_cache_and_return_session(
     endpoint_uri: URI,
+    provider_id: Optional[int] = None,
     session: Optional[ClientSession] = None,
 ) -> ClientSession:
     # cache key should have a unique thread identifier
-    cache_key = generate_cache_key(f"{threading.get_ident()}:{endpoint_uri}")
+    if provider_id:
+        cache_key = generate_cache_key(
+            f"{threading.get_ident()}:{provider_id}:{endpoint_uri}"
+        )
+    else:
+        cache_key = generate_cache_key(f"{threading.get_ident()}:{endpoint_uri}")
 
     evicted_items = None
     async with async_lock(_async_session_pool, _async_session_cache_lock):
--- a/web3/_utils/request.py
+++ b/web3/_utils/request.py
@@ -47,10 +47,10 @@ _session_cache_lock = threading.Lock()
 
 
 def cache_and_return_session(
-    endpoint_uri: URI, session: requests.Session = None
+    endpoint_uri: URI, provider_id, session: requests.Session = None
 ) -> requests.Session:
     # cache key should have a unique thread identifier
-    cache_key = generate_cache_key(f"{threading.get_ident()}:{endpoint_uri}")
+    cache_key = generate_cache_key(f"{provider_id}:{endpoint_uri}")
 
     cached_session = _session_cache.get_cache_entry(cache_key)
     if cached_session is not None:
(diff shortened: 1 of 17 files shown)


web3/_utils/request.py
```diff
@@ -132,10 +147,16 @@ def _close_evicted_sessions(evicted_sessions: List[requests.Session]) -> None:
 
 async def async_cache_and_return_session(
     endpoint_uri: URI,
+    provider_id: Optional[int] = None,
     session: Optional[ClientSession] = None,
 ) -> ClientSession:
     # cache key should have a unique thread identifier
-    cache_key = generate_cache_key(f"{threading.get_ident()}:{endpoint_uri}")
+    if provider_id:
```
menaitm, line 154: `if provider_id is not None` safer here (in case provider_id == 0)

menaitm (commented):
thanks

web3/_utils/request.py
```diff
@@ -86,7 +94,8 @@ def get_response_from_get_request(
     endpoint_uri: URI, *args: Any, **kwargs: Any
 ) -> requests.Response:
     kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
-    session = cache_and_return_session(endpoint_uri)
+    provider_id = kwargs.get("provider_id")
```
fselmo, line 97: We should update all the async versions of these as well and add a similar integration test for the `AsyncHTTPProvider` to make sure it passes.

web3/_utils/request.py
```diff
@@ -100,16 +109,22 @@ def json_make_get_request(
...
 
 def get_response_from_post_request(
-    endpoint_uri: URI, *args: Any, **kwargs: Any
+    endpoint_uri: URI,
+    *args: Any,
+    **kwargs: Any,
 ) -> requests.Response:
     kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
-    session = cache_and_return_session(endpoint_uri)
+    provider_id = kwargs.get("provider_id")
```
fselmo, line 117: When we pass the `provider_id` along from the `make_request()`, this will likely either need to be a `pop()` or handled in the method signature since it can't be passed on to the `session.post()` call on line 119.

I slightly prefer the explicit expectation / handling in the method signature `provider_id: int = None` but either could work here.

web3/_utils/request.py
```diff
@@ -47,10 +47,18 @@ def get_default_http_endpoint() -> URI:
 
 
 def cache_and_return_session(
-    endpoint_uri: URI, session: requests.Session = None
+    endpoint_uri: URI, provider_id: int = None, session: requests.Session = None
 ) -> requests.Session:
     # cache key should have a unique thread identifier
-    cache_key = generate_cache_key(f"{threading.get_ident()}:{endpoint_uri}")
+    if provider_id:
```
fselmo, line 53: Should we even provide a choice here? It seems like a unique `provider_id`, and in turn, `cache_key`, should be a requirement otherwise we leave the door open to possible session sharing.

web3/_utils/request.py
```diff
@@ -47,10 +47,18 @@ def get_default_http_endpoint() -> URI:
 
 
 def cache_and_return_session(
-    endpoint_uri: URI, session: requests.Session = None
+    endpoint_uri: URI, provider_id: int = None, session: requests.Session = None
```
fselmo, line 50: If we make `provider_id` a positional argument here and keyword argument everywhere else, we can "require" a `provider_id` each time this method is used. I think this is ideal if we remove the conditional logic below. This means closing the door on possible session sharing. Since this is an internal method, we could then make sure every time this is called we have a `provider_id`.

I think the only time this deviates from ideal is when the `handle_offchain_lookup()` utility methods are called, since those are public. I don't think those methods are ideal in their current form after this addition. It would be good to revisit those methods and see if there's a better flow there.

fselmo (dismissed):
We should pass on the `provider_id` in the `make_request()` method within `HTTPProvider`. I suggested we start an integration test suite for broader spectrum HTTPProvider + go-ethereum integration tests. I think that will be just a nice to have in general and this seems like a good time to add something like that. I suggested a test that should currently fail until we pass on the `provider_id` in `make_request()`.

We should also add a similar integration test for the `AsyncHTTPProvider` and pass the `provider_id` along the async methods since the `_async_session_cache` is subjected to the same sort of session sharing at the moment with same provider uris.

I don't think we should make the `provider_id` optional in the cache keys either. I think we should just close the door on any session sharing possibility there. One issue with that is the way the `handle_offchain_lookup()` utility method (and the async version) is set up. I'd rather re-visit those though than leave this optional just for their sake. Thoughts on that?

fselmo (commented):
Just updating my review... this is a good discussion but this is an "issue" that I ended up finding when the same url is used for two separate provider instances. "Resolving" it seems to create more complexity than necessary. As we discussed, two instances of web3 providers with the same url should just be discouraged as bad practice and stated as such in the docs.

kclowes:
Yep, agreed. Thanks for the review @fselmo! I will update the docs accordingly!

kclowes:
I'm going to close this PR and open a new one with just the doc updates.

Issue #2859, opened 2023-03-04
Bump pytest version


pacrob:
Running pytest using python3.10, I see:

`DeprecationWarning: You're using an outdated version of pytest. Newer releases of pytest-asyncio will not be compatible with this pytest version. Please update pytest to version 7 or later.`

`pytest --version` yields `pytest 6.2.5`. It looks like pytest 7 still supports python 3.7.

kclowes:
Closed via #2863

PR #2863, opened 2023-03-07
Bump pytest to version 7

e3243eric:
### What was wrong?

Related to Issue #2859. From CI warning: `DeprecationWarning: You're using an outdated version of pytest. Newer releases of pytest-asyncio will not be compatible with this pytest version. Please update pytest to version 7 or later.`

### How was it fixed?

Bump `pytest` to version 7.

### Todo:

#### Cute Animal Picture


e3243eric:
The plugin [`pytest-pythonpath`](https://pypi.org/project/pytest-pythonpath/) is obsolete after `pytest` 7.0.0. Now `pytest` using the [`pythonpath`](https://docs.pytest.org/en/7.0.x/reference/reference.html#confval-pythonpath) configuration option.

e3243eric:
The `test_eth_get_balance_with_block_identifier` failed because the block was not mined to block 1. This is because the test execution order has been changed by `pytest`. `test_eth_get_balance_with_block_identifier` was executed before some tests which mining blocks. (e.g. `test_eth_get_storage_at` will mine a block because of the `emitter_contract_deploy_txn_hash` fixture)

Should I mine a new block by the `empty_block` fixture in `test_eth_get_balance_with_block_identifier`, or any other suggestions?

fselmo:
> Should I mine a new block by the `block_with_txn` fixture in `test_eth_get_balance_with_block_identifier`? Overwrite `test_eth_get_balance_with_block_identifier` in the `TestEthereumTesterEthModule` class? Or any other suggestions?

I'd rather address it in the `TestEthereumTesterEthModule` since the dynamics are a bit different there. There are different ways to do it I suppose but one of the most explicit ways may be to do something like: 

```py
    def test_eth_get_balance_with_block_identifier(self, w3: "Web3") -> None:
        # send some money to the coinbase account to ensure later balance > genesis
        w3.eth.send_transaction({"to": w3.eth.coinbase, "value": 1, "gas": 21000})

        super().test_eth_get_balance_with_block_identifier(w3)
 ```
 
 thoughts?

**Comment by fselmo** [comment]
> will the block reward to the coinbase address?

Frankly I don't remember if the `eth-tester` + `py-evm` combo handles that appropriately (I imagine it does?) 😅. If it works to just mine a block then that's great too 🙂

**Comment by e3243eric** [comment]
I think it does. The test works until the execution order changes.

Since it is to test get_balance with a block number, I think sending a transaction to the coinbase address is a good idea, no matter where the block reward goes. I was curious about the block reward. 🙂

**Comment by fselmo** [comment]
Yeah it should handle all of that appropriately but the value transfer is more explicit 👍🏼

File: setup.py [hunk 1]
```diff
@@ -28,10 +28,9 @@
         "flaky>=3.7.0",
         "hypothesis>=3.31.2",
         "importlib-metadata<5.0;python_version<'3.8'",
-        "pytest>=6.2.5",
+        "pytest==7.2.2",
```
kclowes, line 31: If possible, I'd rather bump `pytest>=7.0.0` to allow a wider range. Does that work here?

kclowes (commented):
Thanks for the contribution @e3243eric! I left a few comments that I can get to in the next few days if you aren't able/willing.

e3243eric:
Thanks for all the reviews.

kclowes (approved):
Thanks @e3243eric! Looks good to me!

PR #2864, opened 2023-03-08
migration guide updates

pacrob:
### What was wrong?

v6 migration guide needs a few things:
- [x] Web3 is now Web3 and AsyncWeb3
- [x] removal of `dict -> AttributeDict` conversion from the result formatters and leaving that logic to the middleware

#### Cute Animal Picture

image


kclowes (approved):
LGTM! 🎉

Issue #2862, opened 2023-03-06
Alias camelCase toHex, toChecksumAddress, etc. to make Web3 6.0 migration less painful


miohtama:
* Version:6.0

### What was wrong?

Currently there is no migration period for snake casing some of the functions

* toChecksumAddress
* solidityKeccak
* processRecept
* etc.

This will make Web3 5.0 -> 6.0 upgrade unnecessary painful, because it breaks the existing code without allowing gracefully to migrate to  new function naming.

### How can it be fixed?

Add a stub wrapper functions that just 

- Call the snake case function
- Emit deprecation warning

kclowes:
Thanks @miohtama! I think we got them all.

PR #2872, opened 2023-03-09
Deprecate processLog and getLogs

kclowes:
### What was wrong?

Need to deprecate `processLog` and `getLogs`. 

Related to Issue #2862 

### How was it fixed?
Added warnings and aliases. 

### Todo:

#### Cute Animal Picture


pacrob (approved):
needs newsfragment, otherwise lgtm!

fselmo (approved):
This lgtm, there's one more internal use of `getLogs` that is in the `test_updating_greeting_emits_event()` test that should probably use the `snake_case` but it also still works and we know it's aliased to that method so I could go either way.

kclowes:
Got 'em! Thanks for the reviews!

PR #2876, opened 2023-03-10
Deprecate processReceipt and createFilter in favor of snake_case

kclowes:
### What was wrong?
`processReceipt` and `createFilter` needed to be deprecated and snake_cased in v5.

Related to Issue #2862 

### How was it fixed?
Added warnings and snake_cased!

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/filtering/test_contract_on_event_filtering.py
+++ b/tests/core/filtering/test_contract_on_event_filtering.py
@@ -8,9 +8,9 @@ from eth_utils import (
 @pytest.mark.parametrize('call_as_instance', (True, False))
 def test_create_filter_address_parameter(web3, emitter, Emitter, call_as_instance):
     if call_as_instance:
-        event_filter = emitter.events.LogNoArguments.createFilter(fromBlock="latest")
+        event_filter = emitter.events.LogNoArguments.create_filter(fromBlock="latest")
     else:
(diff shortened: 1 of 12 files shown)


tests/core/filtering/test_contract_on_event_filtering.py
```diff
@@ -204,10 +204,21 @@ def test_on_sync_filter_with_topic_filter_options_on_old_apis(
...
     )
 
     old_logs = post_event_filter.get_all_entries()
     assert len(old_logs) == 4
+
+
+#
+# Deprecated
+#
+def test_createFilter_emits_deprecation_warning(web3, emitter):
```
fselmo, line 219: It might be nice, as a gut check, to check that `createFilter()` actually works and / or delegates to `create_filter()`. Since this is a pretty minor addition and we don't expect any future `v5` releases, I think it is easy enough to see that it's working as intended so feel free to take or leave.

fselmo (approved):
lgtm, just left one comment to take or leave

tests/core/filtering/test_contract_on_event_filtering.py
kclowes: That's a good point. Will add the expectations back!

pacrob (approved):
lgtm!

saadaltafofficial:
what parameters does create_filter() takes in?

PR #2878, opened 2023-03-10
snakecase and deprecate toBytes, toWei, fromWei

pacrob:
snakecase and deprecate toBytes, toWei, fromWei

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/docs/contracts.rst
+++ b/docs/contracts.rst
@@ -711,7 +711,7 @@ Taking the following contract code as an example:
 
     >>> array_contract.functions.getBytes2Value().call()
     [b'b\x00']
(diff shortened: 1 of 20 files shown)


fselmo (approved):
Though this will likely be the end of `v5` and we can see that these are implemented appropriately, I think it's still good practice to test that the deprecated methods either delegate to the new method, respectively, or verify some result in these tests.

I prefer the former but we should have at least some gut check here before merging.

kclowes (approved):
I agree with the comments @fselmo left about adding the expectations back to the tests, but looks good otherwise! 🚀

PR #2882, opened 2023-03-16
removed camel deprecation notices from web3.eth docs

pacrob:
### What was wrong?

Camel-cased methods were removed, but their deprecation notices were still in the docs.

I also removed:
 - `Eth.getUncle`, as it was just a note saying such a method never existed.
 - `Eth.getTransactionFromBlock`, which was deprecated going into v5.

### How was it fixed?

Removed from web3.eth docs. I left them in geth.miner as those will be coming in another PR.

### Todo:

#### Cute Animal Picture

image


wolovim (approved):
🔨

PR #2898, opened 2023-03-28
Refactor geth steps to use correct py-geth version

fselmo:
### What was wrong?

- pip install with [minimum requirements](https://pip.pypa.io/en/stable/cli/pip_install/#examples) must use quotes around the library and minimum version.

### How was it fixed?

- Make sure pip is up to date and use quotes around minimum version for pip installing py-geth.

Alternative:

Just use `pip install --user py-geth=={version}`

---
**Note:**

The windows wheel build just started failing on this PR. I'm going to address that in a separate PR, so please ignore that failing CI build and it should still fail on `master` until we get the fix in.

### Todo:


pacrob (approved):
lgtm, based on discussion about separate PR for borked windows build.

PR #2913, opened 2023-04-12
Use the correct type for `trace_filter`

Uxio0:
- Use `TraceFilterParams` for `trace_filter` instead of `FilterParams` (for events)


kclowes (approved):
Thanks @Uxio0!

PR #2916, opened 2023-04-12
Run ci 2795

kclowes:
### What was wrong?

Related to Issue #

### Todo:

#### Cute Animal Picture


fselmo (commented):
This looks good to me. I'm wondering if we can't catch the `ContractCustomError` at the `Contract` level, whether `call` or `transact`, and use the contract's ABI to decode the `data` for the defined custom error too...

That could maybe be an improvement for another time but any thoughts on that?

jluuM2:
Thanks @kclowes for filling-in the test requirements, I have had little time to devote to this recently

kclowes:
@fselmo - I like that idea, and explored doing that for a little bit yesterday. I don't think it would be too complicated, but will take a some plumbing since we don't have the utility functions built out to handle custom error abi's yet. Our `call` function gets passed the function abi but we'd have to pull the error abi from the contract abi and match the 4byte identifier that comes back in the response. The other thing that would be sort of weird is that while it would be straightforward to plug into the contract's `call` method, if someone were to do a raw `w3.eth.call`, we don't have access to the contract at all, so I don't think we could parse the return value into something human-readable. But seeing as you have to call a contract function to get the custom error response and message, I think that would be a pretty small edge case. I'm going to add a separate issue to track returning the raw error data that comes back, because that sounds like it would be a useful thing. See [this comment](https://github.com/ethereum/web3.py/pull/2795#issuecomment-1508323849).

PR #2919, opened 2023-04-14
docs: new send txs guide

wolovim:
### What was wrong?

Current explanation for when and why to reach for `send_transaction` or `send_raw_transaction` is lacking. 

### How was it fixed?

New `Guide` created and referenced within docs. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/transactions.rst
+++ b/docs/transactions.rst
@@ -1,0 +1,187 @@
+Sending Transactions
+====================
+
+.. note::
+
+  Prefer to view this code in a Jupyter Notebook? View the repo `here <https://github.com/wolovim/ethereum-notebooks/blob/master/Sending%20Transactions.ipynb>`_.
+
+There are two methods for sending transactions using web3.py: :meth:`~web3.eth.Eth.send_transaction` and :meth:`~web3.eth.Eth.send_raw_transaction`. A brief guide:
+
+#. Want to sign a transaction offline or send pre-signed transactions?
+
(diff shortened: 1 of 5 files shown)


docs/transactions.rst
```diff
@@ -0,0 +1,187 @@
...
+
+   * load account via eth-account (``w3.eth.account.from_key(pk)``), then
+   * use ``send_transaction``
+
+Interacting with or deploying a contract?
+
+* Option 1: :meth:`~web3.contract.ContractFunction.transact` uses ``send_transaction`` under the hood
+* Option 2: :meth:`~web3.contract.ContractFunction.build_transaction` + ``sign_transaction`` + ``send_raw_transaction``
+
+An example for each can be found below.
```
kclowes, line 29: It looks like this is a mix of links and `code` references. Is that on purpose? I think it would be nice if they all linked to their documentation, but feel free to take it or leave it!

kclowes (approved):
Looks great!

docs/transactions.rst
wolovim: thought process applied here was to link the method only the first time it appears, as not to distract from the content. maybe there's an argument for it being more distracting though, if that's not obvious. i don't mind updating all to links, if that's preferable.

kclowes: Ah, gotcha. That wasn't obvious to me until you pointed it out, but makes sense. I'll defer to your writing expertise! :)

wolovim: opted to just link everything. the red formatting really harshes the vibe, man. ✌️

PR #2922, opened 2023-04-17
Pass raw data with ContractLogicError

### LINKED ISSUE #2793: Add support for custom errors or provide full RPC response if transaction is reverted

kclowes:
### What was wrong?

People frequently request the raw data to be sent back with the ContractLogicError. 

Closes #2793, #2336 

### How was it fixed?

This PR sends back the raw data as an attribute on the error class without breaking the existing API. 

### Todo:
- [x] Documentation

#### Cute Animal Picture


fselmo (approved):
Oh I like this approach in general! The more information the better on exceptions 👌🏼. 

I think we should be able to add this to the ``OffchainLookup`` exception also which would be some nice information to have available.

tests/integration/test_ethereum_tester.py
```diff
@@ -596,6 +609,36 @@ def test_eth_get_balance_with_block_identifier(self, w3: "Web3") -> None:
...
+        self,
+        w3: "Web3",
+        offchain_lookup_contract: "Contract",
+    ) -> None:
+        # test ContractFunction call
+        with pytest.raises(OffchainLookup) as e:
+            offchain_lookup_contract.functions.testOffchainLookup(
+                OFFCHAIN_LOOKUP_TEST_DATA
+            ).call(ccip_read_enabled=False)
+        assert e.value.data == OFFCHAIN_LOOKUP_RETURN_DATA
```
kclowes, line 622: @fselmo this assertion fails if it's run in isolation. the `OFFCHAIN_LOOKUP_RETURN_DATA` changes, but only fails that way on eth-tester, not with the geth integration tests. Do you have any ideas why the `OFFCHAIN_LOOKUP_RETURN_DATA` would be different when the whole file is run vs. run in isolation?

fselmo: Oh, yeah. The very first value that's returned is the [particular contract's address](https://github.com/ethereum/web3.py/blob/c8ef104fd428972567ed9475eb8ac7b28e4be261/web3/_utils/contract_sources/OffchainLookup.sol#L18) which will differ between the contract deployed via the geth fixture and the one for eth-tester. You'll just have to change the first 20 bytes of the `OFFCHAIN_LOOKUP_RETURN_DATA` to match that.

kclowes: Aha, thank you! I'll update the assertion so that it only asserts on `e.value.data[:74]`

fselmo: Or just update [10:74] to match the address that deploys to the `eth-tester` fixture

fselmo: Changing `OFFCHAIN_LOOKUP_RETURN_DATA`... should be:

```py
OFFCHAIN_LOOKUP_RETURN_DATA = "0x556f1830000000000000000000000000f2e246bb76df876cef8b38ae84130f4f55de395b00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000001c0da96d05a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002400000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000002c68747470733a2f2f776562332e70792f676174657761792f7b73656e6465727d2f7b646174617d2e6a736f6e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002568747470733a2f2f776562332e70792f676174657761792f7b73656e6465727d2e6a736f6e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001474657374206f6666636861696e206c6f6f6b757000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001474657374206f6666636861696e206c6f6f6b7570000000000000000000000000"
```

pacrob (approved):
lgtm!

PR #2939, opened 2023-04-29
added function to tupleize lists inside (nested) dicts

### LINKED ISSUE #2908: web3.datastructures.AttributeDict.__hash__ fails if event logs contain list objects
* Version: 5.31.1
* Python: 3.8
* OS: linux
* pip freeze output omitted since not a package specific issue

### What was wrong?

Note: Not exactly an issue but something I had to circumvent myself when using the library and thinking it may be helpful if `web3.py` handled it instead directly.

I was trying to use functions requiring `__hash__` on event logs and the code failed saying lists are not hashable [here](https://github.com/ethereum/web3.py/blob/402d001b4f0a19720d859efe3c9e0ece39168224/web3/datastructures.py#L117). This happens if the event logs contain `list` objects nested inside them (which some smart contracts seem to return), as lists by default are mutable and thus not hashable in Python.

### How can it be fixed?

Would it make sense to (possibly recursively) cast any present lists to tuples inside the hash function to avoid crashes? Would some detection mechanism inside these pseudo-immutable classes be helpful to make sure they are indeed hashable (e.g., raising an error if a list is added)?

Any thoughts?

JustCallMeDavid:
### What was wrong?

Closes #2908

### How was it fixed?

Tupleized (nested) lists to render them hash-able inside `AttributeDict` instances

#### Cute Animal Picture

This is Rosie. Adoption profile here: https://www.petfinder.com/dog/rosie-61590987/ny/canastota/wanderers-rest-humane-association-ny79/

### DIFF
--- a/web3/datastructures.py
+++ b/web3/datastructures.py
@@ -129,7 +129,7 @@ def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValu
     This method converts them to tuples, rendering them hashable.
     """
 
-    def _to_tuple(lst: list) -> tuple:
+    def _to_tuple(lst: List[Any]) -> Any:
         return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
 
     ret = dict()
--- a/web3/datastructures.py
+++ b/web3/datastructures.py
@@ -125,8 +125,9 @@ class AttributeDict(ReadableAttributeDict[TKey, TValue], Hashable):
 
 def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValue]:
     """
-    Lists inside dicts will throw an error if attempted to be hashed.
-    This method converts them to tuples, rendering them hashable.
+    Unhashable types inside dicts will throw an error if attempted to be hashed.
+    This method converts lists to tuples, rendering them hashable.
+    Other unhashable types found will raise a TypeError
     """
 
     def _to_tuple(lst: List[Any]) -> Any:
@@ -134,10 +135,12 @@ def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValu
 
     ret = dict()
     for k, v in d.items():
-        if isinstance(v, list):
+        if isinstance(v, List):
             ret[k] = _to_tuple(v)
-        elif isinstance(v, dict) or isinstance(v, ReadableAttributeDict):
+        elif isinstance(v, Mapping):
             ret[k] = tupleize_lists_nested(v)
+        elif not isinstance(v, Hashable):
+            raise TypeError(f"Found unhashable type '{type(v).__name__}': {v}")
         else:
             ret[k] = v
     return AttributeDict(ret)
--- a/web3/datastructures.py
+++ b/web3/datastructures.py
@@ -114,13 +114,33 @@ class AttributeDict(ReadableAttributeDict[TKey, TValue], Hashable):
         raise TypeError("This data is immutable -- create a copy instead of modifying")
 
     def __hash__(self) -> int:
-        return hash(tuple(sorted(self.items())))
+        return hash(tuple(sorted(tupleize_lists_nested(self).items())))
 
     def __eq__(self, other: Any) -> bool:
         if isinstance(other, Mapping):
             return self.__dict__ == dict(other)
         else:
             return False
+
+
+def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValue]:
+    """
+    Lists inside dicts will throw an error if attempted to be hashed.
+    This method converts them to tuples, rendering them hashable.
+    """
+
+    def _to_tuple(lst: list) -> tuple:
+        return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
+
+    ret = dict()
+    for k, v in d.items():
+        if isinstance(v, list):
+            ret[k] = _to_tuple(v)
+        elif isinstance(v, dict) or isinstance(v, ReadableAttributeDict):
+            ret[k] = tupleize_lists_nested(v)
+        else:
+            ret[k] = v
+    return AttributeDict(ret)
 
 
 class NamedElementOnion(Mapping[TKey, TValue]):
--- a/newsfragments/2908.bugfix.rst
+++ b/newsfragments/2908.bugfix.rst
@@ -1,0 +1,1 @@
+fix AttributeDicts unhashable if they contain lists recursively tupleizing them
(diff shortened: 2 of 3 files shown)


pacrob:
Thanks for this @JustCallMeDavid ! I modified the testing a bit and will get the typing/linting sorted out, but the tupelizing function is great!

JustCallMeDavid:
Hi @pacrob, thanks a lot for the review and fixing up the tests, they look a lot better now.

Linting and typing should pass now, but please have a look - not sure if it could not be done better (was having trouble getting the dict assignments to work with more precise types).

pacrob:
Yep, I worked on the typing for a while and couldn't get anything more precise either. I'll get some other eyes on it just to check, but otherwise I think we're good.

web3/datastructures.py
```diff
@@ -123,6 +123,26 @@ def __eq__(self, other: Any) -> bool:
...
+    """
+
+    def _to_tuple(lst: List[Any]) -> Any:
+        return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
+
+    ret = dict()
+    for k, v in d.items():
+        if isinstance(v, list):
+            ret[k] = _to_tuple(v)
+        elif isinstance(v, dict) or isinstance(v, ReadableAttributeDict):
```
fselmo: I think we may want to be less selective here

```suggestion
        elif isinstance(v, Mapping):
```

web3/datastructures.py
```diff
@@ -123,6 +123,26 @@ def __eq__(self, other: Any) -> bool:
...
 
 
+def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValue]:
+    """
+    Lists inside dicts will throw an error if attempted to be hashed.
+    This method converts them to tuples, rendering them hashable.
+    """
+
+    def _to_tuple(lst: List[Any]) -> Any:
+        return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
```
fselmo, line 134: I wonder if we should be less selective in general. I wouldn't expect too many `set` types to come in but those are also not hashable, as an example. Rather than make this utility method about `list` types exclusively, what do you think about targeting `Iterable` types in general if they are not tuples?

Something like

```py
def _is_non_tuple_iterable(value: Iterable[Any]) -> bool:
    return isinstance(value, Iterable) and not isinstance(value, tuple)

def _recursive_iterable_to_tuple(lst: Iterable[Any]) -> Any:
    return tuple(
        _recursive_iterable_to_tuple(i) if _is_non_tuple_iterable(i) else i
        for i in lst
    )

def dict_to_attrdict_hashable(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValue]:
    """
    This method converts a ``dict`` to an ``AttributeDict`` while converting any
    un-hashable ``Iterable`` dict values to hashable ``tuple`` types.
    """

    ret = dict()
    for k, v in d.items():
        if isinstance(v, Mapping):
            ret[k] = dict_to_attrdict_hashable(v)
        elif _is_non_tuple_iterable(v):
            ret[k] = _recursive_iterable_to_tuple(v)
        else:
            ret[k] = v
    return AttributeDict(ret)
```

We can add some `set` examples in the tests too and the above should pass.

Thoughts?

fselmo (commented):
I added a comment as some food for thought. I'm good with the current implementation but I think we can make this a bit more of a catch-all for non-hashable types... or at least non-tuple Iterable types to make it a bit simpler. Curious on thoughts there.

web3/datastructures.py
JustCallMeDavid: To me the changes look good and would help generalize the code. I also like the idea of checking for the generic `Mapping` type in particular. Thanks a lot @fselmo, learned some things.

My only concern/question: I had chosen the `list`-type approach because some of the logs I was retrieving from deployed Solidity contracts contained lists. Not an expert with Solidity/Vyper (yet), but would it ever be possible for a contract to return any other data type inside logs?
While I agree the structure is better, I can't say if the extensibility materializes in use cases and possibly could create surface for harder to spot issues (maybe increased memory consumption/hidden code execution due to 'auto-unrolling' of `Iterable` types? - just a thought).

Happy to add some tests with `set` or general `Iterable` types if we go with this approach. Thanks a lot for the input @pacrob  and @fselmo.

pacrob: I don't know what other datastructures could show up either, but I don't think it'll hurt to check. I like opening it up, but what about checking if it is iterable and not hashable, like:

```py
def _is_non_hashable_iterable(value: Any) -> bool:
    return isinstance(value, Iterable) and not isinstance(value, Hashable)

def _recursive_iterable_to_tuple(itr: Iterable[Any]) -> Any:
    return tuple(
        _recursive_iterable_to_tuple(i) if _is_non_hashable_iterable(i) else i
        for i in itr
    )

def mapping_to_attrdict_hashable(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValue]:
    """
    This method converts a ``Mapping`` to an ``AttributeDict`` while converting any
    un-hashable ``Iterable`` dict values to hashable ``tuple`` types.
    """

    ret = dict()
    for k, v in d.items():
        if isinstance(v, Mapping):
            ret[k] = mapping_to_attrdict_hashable(v)
        elif _is_non_hashable_iterable(v):
            ret[k] = _recursive_iterable_to_tuple(v)
        else:
            ret[k] = v
    return AttributeDict(ret)
```

I've got a couple tests for `set` and `bytearray` passing locally. What other non-hashable iterables are there?

fselmo: @JustCallMeDavid, good thoughts. I was thinking about the problem more from the perspective of fixing the `__hash__()` method of `AttributeDict` rather than patching `list` types coming from logs. I think either way it would work well.

@pacrob If we decide to open it up, that's definitely an improvement 👌🏼. No other examples come to mind.

pacrob: Changes pushed. I changed the typing a bit, but I think it's an improvement.

pacrob: Nevermind my confidence - calling `tuple(set))` does not guarantee order, and `tuple(sorted(Iterable))` throws an error if the iterable contains non-comparable items (`int` and `tuple` in the test case).

A custom sorting function seems like overkill, but excluding `set`s from being in AttributeDicts also sounds complex.

Any preference or other ideas?

fselmo: ~~Hmm not sure I understand the issue but~~ [I see it now] I'm OK with just sticking with the original `list` -> `tuple` idea for now if it's too complicated. Perhaps that was the best way to go all along 🙃 . I think that way if any issue does come up we can address it based on the use case a bit more.

fselmo (approved):
looks good. Thanks @JustCallMeDavid + @pacrob 👍🏼

PR #2940, opened 2023-05-01
refactored _get_raw_miner_data

mac-lawson:
### What was wrong?

The _get_raw_miner_data function had some readability issues and unnecessary type annotations, which made it harder to understand and maintain.

### How was it fixed?

- Rename the latest variable to latest_block for clarity.
- Explicitly assign miner, block_hash, and gas_price values to variables for clarity.
- Simplify the yield statement to return miner, block_hash, and gas_price as a tuple.
- Initialize the block variable to latest_block before the loop to avoid redundancy.
- Remove the type: ignore comments because the type hint for the transaction argument is now Any, and the HexBytes type hint is not needed.

### Todo:


kclowes:
Thanks for the PR. It looks like the lint job is failing without the `type: ignore` so I'm going to close since this doesn't really add anything. Let me know if you want to continue the conversation though.

Issue #1701, opened 2020-08-06
Cannot set multiple providers


nolash:
* Version: 5.12.0
* Python: 3.8.4
* OS: linux 5.7.0
* `pip freeze` output

```
http://gateway.ethswarm.org/files/aaa5b1f479fa918270eac50c705de8bd21512adffbe22f64ce2765e621721014

```

wtr: https://web3py.readthedocs.io/en/stable/web3.main.html#providers

```
(venv) web3 $ cat <<EOF | python
> from web3 import Web3, WebsocketProvider, IPCProvider
> 
> w3 = Web3()
> w3.setProviders([IPCProvider('/tmp/foo.ipc'), WebsocketProvider('ws://127.0.0.1:8546')])
> EOF
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
AttributeError: 'Web3' object has no attribute 'setProviders'
```

```
$ cat <<EOF | python
> from web3 import Web3, WebsocketProvider, IPCProvider
> 
> w3 = Web3([IPCProvider('/tmp/foo.ipc'), WebsocketProvider('ws://127.0.0.1:8546')])
> print(w3.eth.blockNumber)
> EOF
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/home/lash/src/tests/py/web3/venv/lib/python3.8/site-packages/web3/eth.py", line 149, in blockNumber
    return self.web3.manager.request_blocking(RPC.eth_blockNumber, [])
  File "/home/lash/src/tests/py/web3/venv/lib/python3.8/site-packages/web3/manager.py", line 149, in request_blocking
    response = self._make_request(method, params)
  File "/home/lash/src/tests/py/web3/venv/lib/python3.8/site-packages/web3/manager.py", line 124, in _make_request
    request_func = self.provider.request_func(
AttributeError: 'list' object has no attribute 'request_func'
```

```
$ cat <<EOF | python
> from web3 import Web3, WebsocketProvider, IPCProvider
> 
> w3 = Web3([WebsocketProvider('ws://127.0.0.1:8546'), IPCProvider('/tmp/foo.ipc')])
> print(w3.eth.blockNumber)
> EOF
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/home/lash/src/tests/py/web3/venv/lib/python3.8/site-packages/web3/eth.py", line 149, in blockNumber
    return self.web3.manager.request_blocking(RPC.eth_blockNumber, [])
[...]

PR #2949, opened 2023-05-09
remove ens docs reference to ability to specify a list of providers

pacrob:
### What was wrong?

Dangling reference to the ability to set multiple providers was still in the ENS docs.
Related to Issue #1701 

### How was it fixed?

Removed said reference.

### Todo:

#### Cute Animal Picture

image


kclowes (approved):
Thanks for updating this! 🎉 
Looks like these lines need to be updated too: https://github.com/ethereum/web3.py/blob/main/docs/internals.rst?plain=1#L69-L71. Other than that, this looks good to me!

Issue #2955, opened 2023-05-11
AttributeError: 'functools.partial' object has no attribute 'eth' when attempting to load a contract with a field named "w3"


rolo:
* Version: dev version after 6.3.0 (commit 68cc94769b531149ad33f6c0e5818a1880fc8486)
* Python: 3.10.9
* OS: Ubuntu 22.04

### What was wrong?

I found a contract in the wild that I couldn't instantiate with web3.py  (0x480de54f42b17d459314f9D76cAa1A531B43c240).

It seems that this is because the contract includes a function named "w3", which overwrites the w3 member on the ContractCaller class meaning the class can't use it internally when it tries to create any further contract members.

* The code which produced the error

```
from web3 import Web3

provider = <your provider>
w3 = Web3(Web3.HTTPProvider(provider))

[...]

PR #2956, opened 2023-05-11
Fix assertion error when w3 used as abi function name

rolo:
### What was wrong?

Created an issue #2955 

Creating a contract from an ABI with a function called "w3" caused an exception:
```
  File "testcaseweb3.py", line 33, in <module>
    failing = w3.eth.contract(address=address, abi=failing_abi)
  File "web3.py/web3/eth/eth.py", line 652, in contract
    ContractFactory = ContractFactoryClass.factory(self.w3, **kwargs)
  File "web3.py/web3/contract/contract.py", line 320, in factory
    contract.caller = ContractCaller(
  File "web3.py/web3/contract/contract.py", line 585, in __init__
    block_id = parse_block_identifier(self.w3, block_identifier)
  File "web3.py/web3/_utils/contracts.py", line 417, in parse_block_identifier
    return w3.eth.default_block
AttributeError: 'functools.partial' object has no attribute 'eth'
```

### How was it fixed?

I copied the existing test case which used the MATH_CONTRACT_ABI to show that changing the name of the function from "add" to "w3" caused the test to fail. I also added another test case to show that changing the name to "something_else" was fine.

I'm not familiar with the internals here but it seems that members found in the ABI get added to the class directly and this is breaking any place in the class that uses `self.w3`. I thought it might be a fun fix to use a variable name that was valid in Python but not Solidity, maybe something with extended Unicode like a Japanese word, but in the end it just seemed as though I could pass the original argument variable from the constructor rather than the one in `self`.

I'm not sure if this might manifest itself as a problem elsewhere, and I didn't actually test calling a function named w3 so it would probably be prudent to do so but I figured pointing this out, getting the PR created, and letting the test suite run was probs best to do first.

Let me know if there's anything else I can do on this.

Cheers.

### DIFF
--- a/tests/core/contracts/test_contract_class_construction.py
+++ b/tests/core/contracts/test_contract_class_construction.py
@@ -1,6 +1,6 @@
+import copy
 import json
 import pytest
-import copy
 
 from eth_utils import (
     decode_hex,
--- a/tests/core/contracts/test_contract_class_construction.py
+++ b/tests/core/contracts/test_contract_class_construction.py
@@ -1,5 +1,6 @@
 import json
 import pytest
+import copy
 
 from eth_utils import (
     decode_hex,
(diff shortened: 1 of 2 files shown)


kclowes:
Thanks for this PR! On quick glance, I think the approach you went with is probably the right one. One of us will review soon :tm:

reedsa:
@rolo this is great, thanks! I verified the behavior and added an integration test, however I was unable to push to your fork so instead I opened #3147 from my own fork. Closing this out and will merge in your fix from the other PR.

PR #2958, opened 2023-05-12
Shanghai eth tester changes

fselmo:
- Shanghai support for `eth-tester` + `py-evm` and add appropriate formatters and tests
- Update `BlockData` to include `withdrawals` and `withdrawalsRoot` fields
- Peripheral updates related to above changes

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -1106,7 +1106,7 @@ Event Log Object
      'blockNumber': 2})]
     >>> transfer_filter.get_new_entries()
     []
-    >>> tx_hash = contract.functions.transfer(alice, 10).transact({'gas': 899000, 'gasPrice': 674302241})
+    >>> tx_hash = contract.functions.transfer(alice, 10).transact({'gas': 899000, 'gasPrice': 1000000000})
     >>> tx_receipt = w3.eth.wait_for_transaction_receipt(tx_hash)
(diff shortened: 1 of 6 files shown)


docs/web3.contract.rst
```diff
@@ -1115,7 +1115,7 @@ Event Log Object
      'event': 'Transfer',
      'logIndex': 0,
      'transactionIndex': 0,
-     'transactionHash': HexBytes('0xa23e7ef4d2692c5cf34ee99123c9c73099e9c3b68c7850f91c1cbcb91ac327e0'),
+     'transactionHash': HexBytes('0x81691802dac3c3b44725bf2f1a6e8338a4e81175327228c479f19ae1eecbf23a'),
```
kclowes: If you don't want to keep changing this, I believe we could do:
```suggestion
     'transactionHash': HexBytes('...'),
```
instead, similar to `blockHash` on line 1120 below. But if you think it adds value to have the actual value, then we can leave it.

kclowes (approved):
Looks good to me! 🚢

docs/web3.contract.rst
fselmo: Yeah I had actually done this and thought it would just be better to track the hash... but it's not like we're actually checking that it's hashed correctly so I don't think it matters at all. I will change it to `...`. I think that's better.

PR #2960, opened 2023-05-16
complete docstrings for `ContractFunction` and reorder contract classes

pacrob:
### What was wrong?

`ContractFunction` and `AsyncContractFunction` were missing params in their docstrings.
Contract files have similarly named classes, but all were in different order.

### How was it fixed?
Completed docstrings for above classes and updated associated docs to match.
Ordered related classes similarly across the 3 contract files.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/contract/base_contract.py
+++ b/web3/contract/base_contract.py
@@ -125,6 +125,511 @@ if TYPE_CHECKING:
     from .contract import ContractFunction  # noqa: F401
 
 
+class BaseContractEvent:
+    """Base class for contract events
+
+    An event accessed via the api `contract.events.myEvents(*args, **kwargs)`
+    is a subclass of this class.
+    """
+
+    address: ChecksumAddress = None
+    event_name: str = None
+    w3: Union["Web3", "AsyncWeb3"] = None
+    contract_abi: ABI = None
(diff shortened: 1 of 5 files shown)


web3/contract/base_contract.py
```diff
@@ -125,6 +125,132 @@
     from .contract import ContractFunction  # noqa: F401
 
 
+class BaseContractEvents:
```
kclowes, line 325: I think it would make sense for this to be close in location to the `BaseContractEvent` class, what do you think about that? And similarly, have `BaseContractFunctions` and `BaseContractFunction` close as well.

kclowes (approved):
LGTM! Had a nit around co-locating `ContractEvent`/`ContractEvents` and `ContractFunction`/`ContractFunctions`. Feel free to take or leave!

web3/contract/base_contract.py
pacrob: Yup, I went back and forth on that - `ContractFunctions` and `ContractEvents` are the only ones that *needed* to be on top for dependency ordering, so I went with just floating them up, but no issue with pulling `Function` and `Event` up too. Will do.

PR #2962, opened 2023-05-22
remove list of AsyncHTTPProvider-supported methods

pacrob:
### What was wrong?

`AsyncHTTPProvider` had a list of supported methods, but they're all supported now!

### How was it fixed?

Removed the list and formatted the remaining differences from `HTTPProvider`
Created function headers for the async middleware in docs.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -12,6 +12,12 @@ possible for a middleware to return early from a
 call without the request ever getting to the provider (or even reaching the middlewares
 that are in deeper layers).
 
+When integrating middleware with your provider, please ensure you're choosing the right
+version. For ``AsyncWeb3`` users, select the version prefixed with ``async``, such as
+``async_attrdict_middleware``. On the other hand, ``Web3`` users should opt for versions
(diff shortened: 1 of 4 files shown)


kclowes (approved):
Thank you! It seems like we should have some documentation around the async versions of the middleware that include examples, or at the very least list out what they're called. I'd be okay with adding a new issue for that though if you'd rather.

docs/middleware.rst
```diff
@@ -288,12 +291,13 @@ Web3 ships with non-default middleware, for your custom use. In addition to the
   either use ``middleware_onion.add()`` from above, or add the default middlewares to your list of
   new middlewares.
 
-Below is a list of built-in middleware, which is not enabled by default.
+Below is a list of available middlewares which are not enabled by default.
```
kclowes, line 300: Much better wording :)

kclowes (approved):
LGTM! Do you think it's worth adding a disclaimer up at the top that async middlewares can only be used with the async provider? or is that self-evident? I can see either way.

pacrob:
Good idea for choosing the correct version for your provider documentation. I've added.

PR #2964, opened 2023-05-23
misc(provider): Change User-Agent header signature

BoboTiG:
### What was wrong?

The User-Agent used to include a stringiffied class name like:

    "web3.py/6.4.0/<class 'web3.providers.rpc.HTTPProvider'>"
    "web3.py/6.4.0/<class 'web3.providers.async_rpc.AsyncHTTPProvider'>"

This patch sligthly changes it to:

    "web3.py/6.4.0/web3.providers.rpc.HTTPProvider"
    "web3.py/6.4.0/web3.providers.async_rpc.AsyncHTTPProvider"

### How was it fixed?

Use `type.__module__`, and `type.__qualname__`, rather then `str(type))`.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/providers/test_async_http_provider.py
+++ b/tests/core/providers/test_async_http_provider.py
@@ -6,6 +6,7 @@ from aiohttp import (
 
 from web3 import (
     AsyncWeb3,
+    __version__ as web3py_version,
 )
 from web3._utils import (
     request,
@@ -108,18 +109,12 @@ async def test_async_user_provided_session() -> None:
     assert cached_session == session
 
 
-def test_construct_user_agent():
-    """
-    The User-Agent used to include a stringiffied class name like:
-
-        "web3.py/6.4.0/<class 'web3.providers.async_rpc.AsyncHTTPProvider'>"
-
-    This test does minimal checks to ensure it is more like:
-
-        "web3.py/6.4.0/web3.providers.async_rpc.AsyncHTTPProvider"
-
-    """
+def test_get_request_headers():
     provider = AsyncHTTPProvider()
     headers = provider.get_request_headers()
-    assert "<" not in headers["User-Agent"]
-    assert " " not in headers["User-Agent"]
+    assert len(headers) == 2
+    assert headers["Content-Type"] == "application/json"
+    assert (
+        headers["User-Agent"] == f"web3.py/{web3py_version}/"
+        f"{AsyncHTTPProvider.__module__}.{AsyncHTTPProvider.__qualname__}"
+    )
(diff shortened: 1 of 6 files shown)


kclowes:
Thanks @BoboTiG! I like this change, but it's (arguably) breaking so I'd prefer to wait until v7 to pull it in. I added it to our v7 issue.

tests/core/providers/test_async_http_provider.py
```diff
@@ -106,3 +106,20 @@ async def test_async_user_provided_session() -> None:
...
+        "web3.py/6.4.0/<class 'web3.providers.async_rpc.AsyncHTTPProvider'>"
+
+    This test does minimal checks to ensure it is more like:
+
+        "web3.py/6.4.0/web3.providers.async_rpc.AsyncHTTPProvider"
+
+    """
+    provider = AsyncHTTPProvider()
+    headers = provider.get_request_headers()
+    assert "<" not in headers["User-Agent"]
```
fselmo: These are kinda clunky. 

- We aren't testing `construct_user_agent()`. We are maybe indirectly "testing" it but there's no way to prove that `get_request_headers()` is actually calling it. So let's rename the test to `test_get_request_headers()`. 

- Since we are now testing the request headers method, let's test both fields that should be there. We are certain of what the value should be for the content-type, so let's validate that logic entirely.

- The docstring can go here. I think the test name along with the logic inside it is pretty clear of what this is testing for and if it's not it should be changed to reflect it.

```py
from web3 import (
    __version__ as web3_version,
)

...

def test_get_request_headers():
    provider = AsyncHTTPProvider()

    headers = provider.get_request_headers()
    assert len(headers) == 2
    assert headers["Content-Type"] == "application/json"
    assert (
        headers["User-Agent"] == f"web3.py/{web3py_version}/"
        f"{AsyncHTTPProvider.__module__}.{AsyncHTTPProvider.__qualname__}"
    )
```

tests/core/providers/test_http_provider.py
```diff
@@ -103,3 +103,20 @@ def test_user_provided_session():
     assert isinstance(adapter, HTTPAdapter)
     assert adapter._pool_connections == 20
     assert adapter._pool_maxsize == 20
+
+
+def test_construct_user_agent():
```
fselmo: I think we should do the same with this test. See previous comment.

fselmo (approved):
I think the change here is a good one. I made some comments to clean up the tests around this. @reedsa, it looks like you requested a review here. Did you want to take care of those if you agree with them and merge? Otherwise I'm good with this. Just seems like a more explicit test makes sense here since we are trying to get a very specific outcome out of the user-agent field.

fselmo (approved):
looks good!

PR #2979, opened 2023-06-05
bugfix: Unique caches for simple_cache_middleware instances

### LINKED ISSUE #2977: web3.middlewares.simple_cache_middleware only works correctly for a single web3 instance
* Version: 6.3.0
* Python: 3.10.1
* OS: osx

<details><summary>`pip freeze` output</summary>
<p>

```
aiohttp==3.8.4
aiosignal==1.3.1
alabaster==0.7.13
async-timeout==4.0.2
attrs==23.1.0
Babel==2.12.1
base58==2.1.1
bitarray==2.7.4
black==23.3.0
bleach==6.0.0
build==0.10.0
bump2version==1.0.1
bumpversion==0.6.0
cached-property==1.5.2
certifi==2023.5.7
charset-normalizer==3.1.0
click==8.1.3
click-default-group==1.2.2
colorama==0.4.6
cytoolz==0.12.1
distlib==0.3.6
docopt==0.6.2
docutils==0.18.1
eth-abi==4.0.0
eth-account==0.8.0
eth-bloom==2.0.0
eth-hash==0.5.1
eth-keyfile==0.6.1
eth-keys==0.4.0
eth-rlp==0.3.0
eth-tester==0.9.0b1
eth-typing==3.3.0
eth-utils==2.1.0
exceptiongroup==1.1.1
execnet==1.9.0
filelock==3.12.0
flake8==3.8.3
flaky==3.7.0
frozenlist==1.3.3
hexbytes==0.3.0
hypothesis==6.75.9
idna==3.4
imagesize==1.4.1
importlib-metadata==6.6.0
incremental==22.10.0
iniconfig==2.0.0
ipfshttpclient==0.8.0a2
isort==5.12.0
jaraco.classes==3.2.3
Jinja2==3.1.2
jsonschema==4.17.3
keyring==23.13.1
lru-dict==1.2.0
markdown-it-py==2.2.0
MarkupSafe==2.1.2
mccabe==0.6.1
mdurl==0.1.2
more-itertools==9.1.0
multiaddr==0.0.9
multidict==6.0.4
mypy==0.910
mypy-extensions==0.4.4
netaddr==0.8.0
packaging==23.1
parsimonious==0.9.0
pathspec==0.11.1
pkginfo==1.9.6
platformdirs==3.5.1
pluggy==0.13.1
protobuf==4.23.2
py==1.11.0
py-ecc==6.0.0
py-evm==0.7.0a2
py-geth==3.12.0
pycodestyle==2.6.0
pycryptodome==3.18.0
pyethash==0.1.27
pyflakes==2.2.0
Pygments==2.15.1
pyproject_hooks==1.0.0
pyrsistent==0.19.3
pytest==7.3.1
pytest-asyncio==0.21.0
pytest-mock==3.10.0
pytest-watch==4.2.0
pytest-xdist==3.3.1
readme-renderer==37.3
regex==2023.5.5
requests==2.31.0
requests-toolbelt==1.0.0
rfc3986==2.0.0
rich==13.4.1
rlp==3.0.0
safe-pysha3==1.0.4
semantic-version==2.10.0
six==1.16.0
snowballstemmer==2.2.0
sortedcontainers==2.4.0
Sphinx==6.2.1
sphinx-rtd-theme==1.2.1
sphinxcontrib-applehelp==1.0.4
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==2.0.1
sphinxcontrib-jquery==4.1
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.5
toml==0.10.2
tomli==2.0.1
toolz==0.12.0
towncrier==21.9.0
tox==3.28.0
tqdm==4.65.0
trie==2.1.0
twine==4.0.2
types-protobuf==3.19.13
types-requests==2.31.0.1
types-setuptools==67.8.0.0
types-urllib3==1.26.25.13
typing_extensions==4.6.3
urllib3==2.0.2
varint==1.0.2
virtualenv==20.23.0
watchdog==3.0.0
-e git+ssh://git@github.com/ethereum/web3.py.git@1782f7d033aeb0aa2941b1d5fad950ddbec5b36e#egg=web3
webencodings==0.5.1
websockets==11.0.3
when-changed==0.3.0
yarl==1.9.2
zipp==3.15.0
```

</p>
</details> 

### What was wrong?

In the `nucypher` client, we're currently working with a multi-chain setting, so each node is connected to both ETH and Polygon (and in the future to other EVM compatible chains). We recently noticed that when calling `w3.eth.chain_id` in these web3 instances, we get mixed up results: sometimes both web3 instances return 137 (polygon), sometimes 1 (ethereum mainnet). 

After @kprasch and I pulled our hair out for a few days, we realized the problem is that in both providers we add the available `web3.middlewares.simple_cache_middleware`, which by defaults caches the result for `eth_chainId`. We didn't expect this cache to share state between different web3 instances.

A simple test reproducing the problem can be seen here: https://github.com/cygnusv/web3.py/blob/6c35544e3c50cb42f528f27b7268dbe8a5a17f87/tests/core/middleware/test_simple_cache_middleware.py#L161-L178

Now, the question is: _Is the `web3.middlewares.simple_cache_middleware` object expected to be safe to be shared between different web3 instances?_
- If so, then cache keys should also include the web3 instance (e.g., `id(w3)`)
- Otherwise, it should be clear from the documentation to not reuse that cache middleware object and to construct individual middlewares for each web3 instance.

Additionally, from code inspection it seems that `time_based_cache_middleware` has the same problem (but haven't written a test for that). Since `latest_block_based_cache_middleware` includes the `latest_block_hash` as part of the cache key is extremely unlikely that it would be a problem, but potentially it can happen.

### How can it be fixed?

We have a suggested fix for the above assuming you can use the simple cache middleware with different instances. See this branch here: https://github.com/ethereum/web3.py/compare/main...cygnusv:web3.py:middlewares. Would be happy to open a PR for this.

fselmo:
### What was wrong?

closes #2977 

### How was it fixed?

- For undefined `cache` arguments to `*construct_simple_cache_middleware` methods, only create the cache at the middleware definition, not at the constructor method. This makes sure that a fresh cache is created each time the middleware is called (once for each instance). This was previously the case and I believe I broke it when we refactored the LRU cache out of this middleware.

- Tests added inspired by @cygnusv 

### Todo:

### DIFF
--- a/web3/middleware/cache.py
+++ b/web3/middleware/cache.py
@@ -81,9 +81,6 @@ def construct_simple_cache_middleware(
         ``response`` and returns a boolean as to whether the response should be
         cached.
     """
-    if cache is None:
-        cache = SimpleCache(256)
-
     if rpc_whitelist is None:
         rpc_whitelist = SIMPLE_CACHE_RPC_WHITELIST
 
@@ -92,12 +89,17 @@ def construct_simple_cache_middleware(
     ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
         lock = threading.Lock()
 
+        # Setting the cache here, rather than in ``construct_simple_cache_middleware``,
+        # ensures that each instance of the middleware has its own cache. This is
+        # important for compatibility with multiple ``Web3`` instances.
+        _cache = cache if cache else SimpleCache(256)
+
         def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
             if method in rpc_whitelist:
                 cache_key = generate_cache_key(
                     f"{threading.get_ident()}:{(method, params)}"
                 )
-                cached_request = cache.get_cache_entry(cache_key)
+                cached_request = _cache.get_cache_entry(cache_key)
                 if cached_request is not None:
                     return cached_request
 
@@ -105,7 +107,7 @@ def construct_simple_cache_middleware(
                 if should_cache_fn(method, params, response):
                     if lock.acquire(blocking=False):
                         try:
-                            cache.cache(cache_key, response)
+                            _cache.cache(cache_key, response)
                         finally:
                             lock.release()
                 return response
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -31,7 +31,7 @@ are likely to change regularly, so this list may not include the latest version'
 You can find the latest defaults in the constructor in ``web3/manager.py``
 
 AttributeDict
-~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~
 
 .. py:method:: web3.middleware.attrdict_middleware
                web3.middleware.async_attrdict_middleware
@@ -185,7 +185,7 @@ the order of: ``[2, 1, 0]``.
 See "Internals: :ref:`internals__middlewares`" for a deeper dive to how middlewares work.
 
 Middleware Stack API
-~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~
 
 To add or remove items in different layers, use the following API:
 
@@ -282,7 +282,7 @@ To add or remove items in different layers, use the following API:
 
 
 Optional Middleware
------------------------
+-------------------
 
 Web3 ships with non-default middleware, for your custom use. In addition to the other ways of
 :ref:`Modifying_Middleware`, you can specify a list of middleware when initializing Web3, with:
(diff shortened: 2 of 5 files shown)


web3/middleware/cache.py
```diff
@@ -92,20 +89,25 @@ def simple_cache_middleware(
     ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
         lock = threading.Lock()
 
+        # Setting the cache here, rather than in ``construct_simple_cache_middleware``,
```
reedsa, line 92: Might consider adding a couple words for users about the ability to use multiple web3/cache instances in the ``construct_simple_cache_middleware`` comment.

reedsa (commented):
Not blocking, just one suggestion for docs.

web3/middleware/cache.py
fselmo: I'm always for more clarification. I'm not quite sure what to put here though. Do you mean in the docstring description? Imo, this is fixing a broken case since middleware should never be expected to share state. Maybe the general [middleware docs](https://web3py.readthedocs.io/en/latest/middleware.html#middleware) could use some updating with a sentence or two to reflect multiple instance support? Thoughts?

pacrob: +1 for usage and explanation in docs

pacrob (approved):
Code looks good. An explanation + (maybe usage example?) in the docs would be sweet, but could be a separate PR too.

web3/middleware/cache.py
cygnusv: > I'm always for more clarification. I'm not quite sure what to put here though. Do you mean in the docstring description? Imo, this is fixing a broken case since middleware should never be expected to share state. Maybe the general [middleware docs](https://web3py.readthedocs.io/en/latest/middleware.html#middleware) could use some updating with a sentence or two to reflect multiple instance support? Thoughts?

I agree with @fselmo here. I opened the original issue as I understood that multiple providers would be supported by the cache middleware and I was surprised to find out that state was shared. Also, current behavior seems to be a regression from previous web3py versions, so not sure why docs should be updated now.

fselmo: @reedsa @pacrob I took this time to update the docs (which were pretty outdated). But lmk if you meant something else. If so, maybe it's beyond the scope of this PR.

fselmo:
> Code looks good. An explanation + (maybe usage example?) in the docs would be sweet, but could be a separate PR too.

I updated the documentation for now and made sure it was up to date (it was poor to begin with and not up to date). I think I'm not understanding what makes this stand out from other middlewares and needs its own explanation other than the doc description. 

Maybe we create a ticket to better highlight that request and it becomes as separate PR?

web3/middleware/cache.py
reedsa: Thank you, the docs are a great place to outline this.

reedsa (approved):

Issue #2972, opened 2023-06-03
Middleware + docs update


pacrob:
### What was wrong?

Middleware and related docs need cleanup and organization

### How can it be fixed?

Noted so far:

Clean up middleware

 - Remove middlewares marked `Delete` in `web3/manager.py -> RequestManager -> default_middlewares`
 - Delete those middlewares from `web3/middleware` too
 - Create tests for both regular and async default middlewares somewhere in `tests/core/manager` that verify defaults haven't changed
- Add a comment, something like "If you're changing the default middlewares, remember to update tests and docs"

Clean up middleware docs
in `docs/middlwares.rst`:
 - List default middlewares for sync and async providers towards the top
 - List all middleware descriptions together, with the Configuring Middleware above or below them
 - Any deleted middlewares' docs can be removed.
 - `HTTPRequestRetry` is in the docs but nowhere in the code? Can be removed?

Improve middleware docs
 - Document Validation Middleware, any others that are in `web3/middleware` but not in docs
 - Code examples with async middlewares

reedsa:
`HTTPRequestRetry` is in reference to the middleware defined here: https://github.com/ethereum/web3.py/blob/eabd4852e159007953c343df4ae3b7020a611e9b/web3/middleware/exception_retry_request.py#L121

PR #2982, opened 2023-06-06
Remove middlewares marked for deletion

pacrob:
### What was wrong?
Needed checks for whether default middlewares have changed

Related to Issue #2972 

### How was it fixed?
Added test to check default middlewares

### Todo:

#### Cute Animal Picture

image


tests/core/manager/test_default_middlewares.py
```diff
@@ -0,0 +1,51 @@
...
+    ]
+
+    default_middlewares = RequestManager.default_middlewares(w3)
+
+    for x in range(len(default_middlewares)):
+        assert default_middlewares[x][0].__name__ == expected_middlewares[x][0].__name__
+        assert default_middlewares[x][1] == expected_middlewares[x][1]
+
+
+def test_default_async_middlwares():
```
fselmo, line 35: We have a test for this [here](https://github.com/ethereum/web3.py/blob/e0f17867b316aaf0f329c67c8e2d52fa25e6a6ee/tests/core/providers/test_async_http_provider.py#L68-L95). I only added the async version but perhaps we can make a similar one that is just as comprehensive, checking modules, etc as well?

I like the list approach, maybe a combination of these and that test? Thoughts?

fselmo (commented):
I know this is still draft but just making a quick comment that a similar test exists. Maybe we combine the ideas in both?

tests/core/manager/test_default_middlewares.py
reedsa: What granularity of tests do we want for these? I agree we should bundle this with the provider tests, given the static methods will always return the same middleware lists.

Would a sync test belong in [test_http_provider.py](https://github.com/ethereum/web3.py/blob/main/tests/core/providers/test_http_provider.py)? I think that would still allow us to run through the expected list of middlewares and assert that each tuple contains [middleware, string].

pacrob: I like your testing as more comprehensive, but the location is less intuitive. If I'm making changes to `manager.py`, then I would look to update tests in the corresponding location in tests. Both locations seems overkill. I suppose if someone updates the default middleware without updating the provider test, it'll fail and they'll find it anyway.

fselmo: Happy to house these in a more appropriate place. I think I put it there because most of the test suite runs on `EthereumTesterProvider` and the middleware that is default for that is very different than the middleware that is default for a provider that connects to actual clients.

`test_base_provider.py` and `test_async_base_provider.py` might make more sense but the modules also only exist on the `Web3` instances so maybe these tests need to be separated as well.

web3/manager.py
```diff
@@ -139,13 +137,11 @@ def default_middlewares(w3: "Web3") -> List[Tuple[Middleware, str]]:
...
         """
         return [
-            (request_parameter_normalizer, "request_param_normalizer"),  # Delete
             (gas_price_strategy_middleware, "gas_price_strategy"),
             (name_to_address_middleware(w3), "name_to_address"),  # Add Async
             (attrdict_middleware, "attrdict"),
-            (pythonic_middleware, "pythonic"),  # Delete
             (validation_middleware, "validation"),
-            (abi_middleware, "abi"),  # Delete
+            (abi_middleware, "abi"),
```
reedsa, line 145: Couldnt delete this just yet, removal causes validation errors due to invalid formats (string not converted to hex, etc). 
Tests failing in this build without `abi_middleware`: https://app.circleci.com/pipelines/github/ethereum/web3.py/4559/workflows/8c5fe87f-fe5f-4a45-97ee-8391158c502f/jobs/308436

docs/middleware.rst
```diff
@@ -28,7 +28,8 @@ Default Middleware
 
 Some middlewares are added by default if you do not supply any. The defaults
 are likely to change regularly, so this list may not include the latest version's defaults.
```
pacrob: The defaults aren't actually changing that much anymore. I think that fact, plus the note to update documentation in `manager.py`, is enough to reduce this to something like:

These middlewares are added by default if you don't add any:

docs/internals.rst
```diff
@@ -174,13 +174,6 @@ function be called.  For example, if you were writing a middleware which cached
 responses for certain methods your middleware would likely not call the
 ``make_request`` method, but instead get the response from some local cache.
 
-By default, Web3 will use the ``web3.middleware.pythonic_middleware``.  This
-middleware performs the following translations for requests and responses.
-
-* Numeric request parameters will be converted to their hexadecimal representation
-* Numeric responses will be converted from their hexadecimal representations to
```
kclowes, line 181: Do we have this info somewhere else? Even though we're not using the middleware any more, I think this info is still relevant if we can find a good place for it.

docs/middleware.rst
```diff
@@ -198,9 +213,11 @@ To add or remove items in different layers, use the following API:
     .. code-block:: python
 
         >>> w3 = Web3(...)
-        >>> w3.middleware_onion.add(web3.middleware.pythonic_middleware)
+        >>> w3.middleware_onion.add(web3.middleware.gas_price_strategy_middleware)
+        # or
+        >>> w3.middleware_onion.add(web3.middleware.gas_price_strategy_middleware, 'gas_price_strategy')
         # or
-        >>> w3.middleware_onion.add(web3.middleware.pythonic_middleware, 'pythonic')
+        >>> w3.middleware_onion.add(web3.middleware.async_gas_price_strategy_middleware, 'gas_price_strategy')
```
pacrob: async middleware should only go on an async provider. Adding the following above the async examples would work:

```py
async_w3 = AsyncWeb3(...)
```

docs/middleware.rst
```diff
@@ -26,9 +26,24 @@ More information is available in the "Internals: :ref:`internals__middlewares`"
 Default Middleware
 ------------------
 
-Some middlewares are added by default if you do not supply any. The defaults
-are likely to change regularly, so this list may not include the latest version's defaults.
-You can find the latest defaults in the constructor in ``web3/manager.py``
+Middlewares are added by default if you don't add any.
+
+Sync middlewares include:
```
kclowes, line 31: I don't think this list is showing up quite right: 
image

docs/middleware.rst
reedsa: Good catch, updated!

web3/manager.py
kclowes: ~This can be a separate PR, but since all of the tests are failing on the `eth-tester` inbound validation, I suspect there is a bug with either the inbound validation in `eth-tester`, or with our `eth-tester` logic that does some type conversions before we send to `eth-tester`~. We should make an issue though to track this removal.

Never mind, just saw that the integration tests are failing too.

kclowes (approved):
This looks good to me! Thanks for taking it on! I just left a couple small comments.

docs/internals.rst
fselmo: I could be mistaken but I don't actually see any good detail on formatters in the docs. Since this moved to the formatters we should definitely add a section explaining what they do and how they work, etc. I would say that's beyond the scope of this PR but I can create an issue to track it.

---
*edit: created #2994 to track this*

PR #2983, opened 2023-06-06
move ethpm deprecation warning to only show when enabled

pacrob:
### What was wrong?

The `ethpm` deprecation warning was showing up too much. 

### How was it fixed?

Moved it to only warn when enabling with `enable_unstable_package_management_api`

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/2983.internal.rst
+++ b/newsfragments/2983.internal.rst
@@ -1,0 +1,1 @@
+Moved ``ethpm`` deprecation warning to only show when the module is explicitly enabled
--- a/web3/main.py
+++ b/web3/main.py
@@ -1,4 +1,5 @@
 import decimal
+import warnings
 
 from ens import (
     AsyncENS,
@@ -343,9 +344,14 @@ class BaseWeb3:
             )
 
     def enable_unstable_package_management_api(self) -> None:
-        from web3.pm import PM  # noqa: F811
-
         if not hasattr(self, "_pm"):
+            warnings.warn(
+                "The ``ethPM`` module is no longer being maintained and will be "
+                "deprecated with ``web3.py`` version 7",
+                UserWarning,
+            )
+            from web3.pm import PM  # noqa: F811
+
             self.attach_modules({"_pm": PM})
 
 
--- a/ethpm/__init__.py
+++ b/ethpm/__init__.py
@@ -1,12 +1,5 @@
-import warnings
 from pathlib import Path
 
-
-warnings.warn(
-    "The ``ethPM`` module is no longer being maintained and will be "
-    "deprecated with ``web3.py`` version 7",
-    UserWarning,
-)
 
 ETHPM_DIR = Path(__file__).parent
 ASSETS_DIR = ETHPM_DIR / "assets"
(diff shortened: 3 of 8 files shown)


web3/main.py
```diff
@@ -343,6 +344,12 @@ def pm(self) -> "PM":
             )
 
     def enable_unstable_package_management_api(self) -> None:
+        warnings.warn(
```
fselmo: Tiny nit, take or leave, but we can move this down to where we actually attach the PM module, inside of the if block. 

Unrelated, but it would be nicer to move the local import in there too if we go that route.

```py
def enable_unstable_package_management_api(self) -> None:
    if not hasattr(self, "_pm"):
        # warning here
    
        from web3.pm import PM  # noqa: F811
        self.attach_modules({"_pm": PM})
```

newsfragments/2983.internal.rst
```diff
@@ -0,0 +1 @@
+Moved `ethpm` deprecation warning to only show when the module is explicitly enabled
```
fselmo: Just double checking... wrapping text in single backticks ` will [italicize](https://web3py.readthedocs.io/en/latest/releases.html#deprecations) in ReadTheDocs, wrapping in double backticks `` will create [code text](https://web3py.readthedocs.io/en/latest/releases.html#features). I think either could make sense here but we seem to have varying use of double vs single in the docs. Whatever you mean to do here, do it 😆.

fselmo (approved):
lgtm, just a nit 👍🏼

web3/main.py
```diff
@@ -343,6 +344,12 @@ def pm(self) -> "PM":
...
+            "The ``ethPM`` module is no longer being maintained and will be "
+            "deprecated with ``web3.py`` version 7",
+            UserWarning,
```
fselmo: I'm thinking a `DeprecationWarning` might make more sense here?

pacrob: I had initially tried that, but it didn't show the warning when I imported it. Either way :shrug:

newsfragments/2983.internal.rst
pacrob: Thanks! This PR now includes updates to a few other newsfragments.

web3/main.py
pacrob: moved!

web3/main.py
pacrob: Looks like a `DeprecationWarning` will only show when the offending behavior occurs within a `__main__` module. `UserWarning` will keep it visible for anyone who hits it. Leaving as-is, but if your opinions are strong, tell me about them.

fselmo: `DeprecationWarning` is usually turned off by default. You have to want to see them. I just turned it on in a script and ran it with the change and it shows up nicely. I do lean towards using that since it seems more appropriate in this case.

Try running after this:

```py
import warnings
warnings.simplefilter('always', DeprecationWarning)
```

fselmo: Oh, I see what you mean... in case someone turns this on NOT from a `__main__`... yeah, we can leave it `UserWarning` since now it's in a place that doesn't make as much noise. A `FutureWarning` may make more sense in that case but either works :)

fselmo: @pacrob lgtm 👍🏼

Issue #2973, opened 2023-06-03
add AsyncIPCProvider


pacrob:
### What was wrong?

`IPCProvider` needs async version

PR #2984, opened 2023-06-07
✨ Add AsyncIPCProvider

### LINKED ISSUE #1413: Web3 Async API Tracking Issue

dzhalaevd:
### What was wrong?

IPCProvider needs async version

Related to Issue https://github.com/ethereum/web3.py/issues/2973
Closes #1413 

### How was it fixed?
Add AsyncIPCProvider

### Todo:

#### Cute Animal Picture

image


web3/providers/async_ipc.py
```diff
@@ -0,0 +1,155 @@
...
+    async def _open(self) -> socket.socket:
+        return await async_get_ipc_socket(self.ipc_path)
+
+    async def reset(self) -> socket.socket:
+        self.sock.close()
+        self.sock = await self._open()
+        return self.sock
+
+
+class AsyncIPCProvider(JSONBaseProvider):
```
dzhalaevd: In the `AsyncIPCProvider` class, there is a synchronous context manager. However, the `_make_request` method is asynchronous. Therefore, I added support for asynchronous context managers to the `Timeout` class as part of the current task.

web3/_utils/threads.py
```diff
@@ -99,7 +113,7 @@ def sleep(self, seconds: float) -> None:
         self.check()
 
     async def async_sleep(self, seconds: float) -> None:
-        await asyncio.sleep(seconds)
+        await asyncio.wait_for(asyncio.sleep(seconds), timeout=self.seconds)
```
dzhalaevd: The implementation of the `Timeout.async_sleep` method was modified to improve performance, if the delay exceeds the specified time, an exception will be thrown. This approach prevents program freezes in the event of a delay taking too long, providing improved reliability and responsiveness. However, I'm not sure if this is necessary in `async_sleep` or we can do with the old option.

web3/providers/async_ipc.py
```diff
@@ -133,12 +136,12 @@ async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
                 sock.sendall(request)
 
             raw_response = b""
-            async with Timeout(self.timeout) as timeout:
+            async with asyncio.timeout(self.timeout) as timeout:
```
dzhalaevd: Why did you replace Timeout from `web3._utils.threads` on Timeout from `asyncio`?

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,158 @@
...
+        request = self.encode_rpc_request(method, params)
+
+        # potentially missing a lock here
+        async with self._socket as sock:
+            try:
+                sock.sendall(request)
+            except BrokenPipeError:
+                # one extra attempt, then give up
+                sock = self._socket.reset()
+                sock.sendall(request)
```
dzhalaevd:
```suggestion
                await sock.send(request)
```

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,158 @@
...
+                    elif has_valid_json_rpc_ending(raw_response):
+                        try:
+                            response = self.decode_rpc_response(raw_response)
+                        except JSONDecodeError:
+                            await timeout.async_sleep(0)
+                            continue
+                        else:
+                            return response
+                    else:
+                        await timeout.async_sleep(0)
```
dzhalaevd:
```suggestion
                        await asyncio.sleep(0)
                    elif has_valid_json_rpc_ending(raw_response):
                        try:
                            response = self.decode_rpc_response(raw_response)
                        except JSONDecodeError:
                            await asyncio.sleep(0)
                            continue
                        else:
                            return response
                    else:
                        await asyncio.sleep(0)
```

JustAnotherC0der:
Pardon my ignorance to github work flow but is this implemented yet? Can't seem to import AsyncIPCProvider on newest web3

kclowes:
@JustAnotherC0der It's not yet. We're still working on it.

DavidRomanovizc:
Hello, @reedsa, how can we make progress on this pr? We/I seem to be stuck, and I need assistance to complete work on this pr. If these changes are still relevant and you're not occupied with other tasks

kclowes:
This is ready for review (finally)! I will add docs tomorrow, but wanted to get the code out there for feedback!

SheldonHolmgren:
Thanks for your hard work. I will really appreciate this feature.

Any plans to make AsyncIPCProvider a PersistentConnectionProvider? I think without this it will be impossible to `w3.eth.subscribe()`:
<details>

```

Web3ValidationError                       Traceback (most recent call last)
<ipython-input-12-2fc2efbc61b6> in <module>
----> 1 async with AsyncWeb3.persistent_websocket(AsyncIPCProvider("/mnt/ssd/optimism_data/geth.ipc")) as w3:
      2     await w3.eth.subscribe("newHeads")

~/dev/web3.py/web3/main.py in persistent_websocket(provider, middlewares, modules, external_modules, ens)
    516         a ``PersistentConnectionProvider`` instance.
    517         """
--> 518         return _PersistentConnectionWeb3(
    519             provider,
    520             middlewares,

~/dev/web3.py/web3/main.py in __init__(self, provider, middlewares, modules, external_modules, ens)
    541     ) -> None:
    542         if not isinstance(provider, PersistentConnectionProvider):
--> 543             raise Web3ValidationError(
    544                 "Provider must inherit from PersistentConnectionProvider class."
```
</details>

This is not a big deal since I can still use filters to get similar functionality.

pacrob (approved):
lgtm!

web3/tools/benchmark/main.py
```diff
@@ -139,6 +151,7 @@ def main(logger: logging.Logger, num_calls: int) -> None:
                     "params": {},
                     "exec": lambda: w3_http.eth.gas_price,
                     "async_exec": lambda: async_w3_http.eth.gas_price,
+                    "ipc": lambda: w3_ipc.eth.gas_price,
```
fselmo: Did we want to add `async_ipc` here as well?

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,166 @@
...
+        from web3._utils.windows import (
+            NamedPipe,
+        )
+
+        return NamedPipe(ipc_path)
+    else:
+        return await asyncio.open_unix_connection(ipc_path)
+
+
+class PersistantSocket:
```
fselmo: persistant -> persistent

```suggestion
class PersistentSocket:
```

I think this is copied from the `PersistantSocket` defined in the synchronous `IPCProvider`. Perhaps `AsyncPersistentSocket` (or `PersistentAsyncSocket`?) might be best in case it's ever imported anywhere then we don't have the same class name imported from two separate modules? Maybe we can fix the spelling on that one in `v7` too 😅 .

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,166 @@
...
+            except OSError as e:
+                # Broken pipe
+                if e.errno == errno.EPIPE:
+                    # one extra attempt, then give up
+                    new_sock = await self._socket.reset()
+                    reader, writer = new_sock
+                    writer.write(request)
+                    await writer.drain()
+
+            async def decode_response() -> RPCResponse:
```
fselmo: nit: If we define this method outside of the scope of the async context manager and call it from within the context instead, this might have a bit better performance as well as readability. I believe there's an overhead to redefining the method each time the async context manager is entered. Even if it's slight, it might read a bit better too? Thoughts?

Something like 

```py
async def get_response_from_reader(self, reader: asyncio.StreamReader) -> RPCResponse:
    raw_response = b""
    
    while True:
        ...
        
        return response
```

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,166 @@
...
+
+    async def reset(self) -> Tuple[asyncio.StreamReader, asyncio.StreamWriter]:
+        reader, writer = self.sock
+        writer.close()
+        await writer.wait_closed()
+        self.sock = await self._open()
+        return self.sock
+
+
+_async_session_cache_lock = threading.Lock()
```
fselmo: Is there a session or a session cache that the IPC Provider is utilizing? If we don't plan on doing anything related to any sessions and if it's not sharing anything across provider instances, I wonder if we can't make use of an `asyncio.Lock` by itself here.

We currently use one [here](https://github.com/ethereum/web3.py/blob/main/web3/providers/persistent.py#L32) on the `PersistentConnectionProvider` to keep separate parts of the code base from calling `recv()` on the websocket connection at the same time. Though there's an open PR that would get rid of our need for that lock, it may help here. Thoughts?

web3/providers/async_ipc.py
```diff
@@ -0,0 +1,166 @@
...
+            raise TypeError("ipc_path must be of type string or pathlib.Path")
+
+        self.timeout = timeout
+        self._socket = PersistantSocket(self.ipc_path)
+        super().__init__()
+
+    def __str__(self) -> str:
+        return f"<{self.__class__.__name__} {self.ipc_path}>"
+
+    async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
```
fselmo: This question builds a bit on the comment from @SheldonHolmgren. ~~If we create a reader and writer here each time we make a request, we aren't utilizing the "persistent open socket" paradigm in a way that could be advantageous to state preservation requests like subscriptions.~~ 

[Nope, it's staying alive. User error / I wasn't testing correctly. I stand corrected. I wonder how much it would take then to build subscription support, even if it doesn't inherit from `PersistentConnectionProvider` too well. It may take more work and breaking changes to try to come back later and put that work in. It may take just utilizing the reader in an asynchronous iterator context and trying to make use of the `_PersistentConnectionWeb3` class? 🤔 ]

~~I'm curious how much work it may take to make this a `PersistentConnectionProvider` or something similar to where we can keep an open connection throughout the life of the provider rather than opening / closing each time.~~ 

Not necessarily trying to head in that direction as that may take quite a bit more work, but curious on everyone's thoughts on that.

---
edited to reflect only how we may inherit from `PersistentConnectionProvider` and / or add `eth_subscribe` support 🤔

fselmo (commented):
Thanks for everyone's contribution here. I think this will be a great option to have. I left some initial comments. Mostly curious about the `PersistentConnectionProvider` question.

web3/providers/async_ipc.py
kclowes: Ha thanks. Yes it is 😆

web3/providers/async_ipc.py
kclowes: Oop, yep. I think an `asyncio.Lock` should work by itself. Nice catch!

web3/providers/async_ipc.py
kclowes: Yeah, I was thinking about that too. IMHO it makes sense to allow for `eth_subscribe` support here, but not sure how much work it'll take. I'll timebox it and if I can get something working, I'll add in a new commit for reviewing.

fselmo: Sounds good. In a best case scenario, assuming we go with the message listener task that's currently in PR for the `WebsocketProviderV2`, I think we can probably hijack all of the `RequestProcessor` logic, switch the naming within the more broad `PersistentConnectionProvider` class and the `RequestManager` (from `ws` or `websocket` to `socket` maybe?) while using the appropriate logic for the specific sockets in the implementation classes. We can likely route the `socket.reader` into a message listener background task in a similar way as iterating over new messages on the websocket is done in wsv2. Curious to see what you find out as you timebox it though 👀

web3/providers/async_ipc.py
kclowes: Done! Good call

web3/tools/benchmark/main.py
kclowes: Eventually yes, but currently there is some bug where the `async_ipc` `send_transaction` stalls out at 41 calls via the benchmarking suite. I can make 500 `send_transaction` calls attached to a local dev node super fast, but haven't figured out the problem with benchmarking yet. I didn't want to spend too much time, so decided to add sync ipc calls and we can add async later if we want in a separate PR.

fselmo: Ah, I see. I vaguely remember some chatter about that. Thanks for the info 👍🏼

SheldonHolmgren:
This code won't work if the client receives multiple responses before it reads the first one. I have coded up a dirty fix for my own use. Perhaps it might be useful as a hint: https://github.com/DavidRomanovizc/web3.py/compare/main...SheldonHolmgren:async_ipc_multiple_resps

kclowes:
Thank you @SheldonHolmgren! I'll see what I can do.

kclowes:
All right, I think this is ready again for another look. Thanks @fselmo for the boost!

fselmo (approved):
I caught one minor doc change we missed but that was all I saw (and I went ahead and pushed it). Thanks for all the work here @DavidRomanovizc, @kclowes, and @SheldonHolmgren! lgtm 👍🏼 

@kclowes I'll let you merge this in when you are good with it. Maybe some more squashing if it makes sense too?

web3/manager.py
```diff
@@ -314,25 +314,25 @@ async def coro_request(
...
-    async def _get_next_ws_message(self) -> Any:
-        return await self._ws_message_stream().__anext__()
+    async def _get_next_message(self) -> Any:
+        return await self._message_stream().__anext__()
 
-    async def _ws_message_stream(self) -> AsyncGenerator[RPCResponse, None]:
+    async def _message_stream(self) -> AsyncGenerator[RPCResponse, None]:
         if not isinstance(self._provider, PersistentConnectionProvider):
             raise TypeError(
                 "Only websocket providers that maintain an open, persistent connection "
```
pacrob: It looks like we're removing references to websockets, and either just using "socket" or "persistent connection". Should that happen here too?

kclowes: oop, yep. Nice catch!

docs/providers.rst
```diff
@@ -261,7 +376,8 @@ Persistent Connection Providers
 
 .. py:class:: web3.providers.persistent.PersistentConnectionProvider(endpoint_uri: str, request_timeout: float = 50.0, subscription_response_queue_size: int = 500)
 
-    This is a base provider class, currently inherited by the ``WebsocketProviderV2``.
+    This is a base provider class, currently inherited by the ``WebsocketProviderV2``,
+    and the ``AsyncIPCProvider``.
     It handles interactions with a persistent connection to a JSON-RPC server. Among
     its configuration, it houses all of the
     :class:`~web3.providers.websocket.request_processor.RequestProcessor` logic for
```
pacrob: This moved to `web3.providers.persistent.request_processor` I believe.

PR #2986, opened 2023-06-07
Panic error support

### LINKED ISSUE #2985: Support the Solidity Panic error
* Version: 6.4.0
* Python: 3.11.2
* OS: Ubuntu 22.04.2 LTS
* `pip freeze` output

```
web3==6.4.0
```

### What was wrong?

Currently the `Panic` error is processed as `ContractCustomError` which is confusing.

There is already a `Panic` error support mentioned in this old issue: https://github.com/ethereum/web3.py/issues/1812

### How can it be fixed?

It will be nice to add support to `raise_contract_logic_error_on_revert` so it raises `PanicError` instead of `ContractCustomError` which is kinda confusing.

See https://medium.com/coinmonks/learn-solidity-lesson-26-error-handling-ccf350bc9374 for the hash id of the Panic error and https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require for the list of the error codes.

This is a code to work around the issue, I am using a failed transaction at Polygon Mumbai testnet:

```python
import logging
import web3
from web3 import Web3
from web3.exceptions import ContractCustomError
from web3.gas_strategies.rpc import rpc_gas_price_strategy
from web3.middleware import geth_poa_middleware


RPC_URL = "https://polygon-testnet-archive.allthatnode.com:8545"
PANIC_TX = "0xe15d38e012c7bddaecdfd065583221d671b43d13512759d1b731af2b524fb4ee"

logger = logging.getLogger(__name__)


def get_web3():
    w3 = Web3(web3.HTTPProvider(RPC_URL))
    if w3.eth.chain_id == 80001:  # Polygon Mumbai
        w3.middleware_onion.inject(geth_poa_middleware, "geth_poa", layer=0)
    w3.eth.set_gas_price_strategy(rpc_gas_price_strategy)
    return w3


def replay_reverted_transaction(tx_id):
    w3 = get_web3()
    tx = w3.eth.get_transaction(tx_id)
    replay_tx = {
        "to": tx["to"],
        "from": tx["from"],
        "value": tx["value"],
        "data": tx["input"],
    }
    w3.eth.call(replay_tx, tx.blockNumber - 1)


# See https://medium.com/coinmonks/learn-solidity-lesson-26-error-handling-ccf350bc9374
PANIC_ERROR_HASH = "4e487b71"

# See https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require
PANIC_ERRORS = {
    "00": "Panic error 0x00: Generic compiler inserted panics.",
    "01": "Panic error 0x01: Assert evaluates to false.",
    "11": "Panic error 0x11: Arithmetic operation results in underflow or overflow.",
    "12": "Panic error 0x12: Division by zero.",
    "21": "Panic error 0x21: Cannot convert value into an enum type.",
    "22": "Panic error 0x12: Storage byte array is incorrectly encoded.",
    "31": "Panic error 0x31: Call to 'pop' on an empty array.",
    "32": "Panic error 0x32: Array index is out of bounds.",
    "41": "Panic error 0x41: Allocation of too much memory.",
    "51": "Panic error 0x51: Call to a zero-initialized variable of internal function type.",
}

try:
    replay_reverted_transaction(PANIC_TX)
except ContractCustomError as e:
    logger.error(f"{e.__class__.__name__}: {e.data}")
    error_hash = str(e.data)[2:10].casefold()
    if error_hash == PANIC_ERROR_HASH:
        error_code = str(e.data)[-2:]
        message = PANIC_ERRORS.get(error_code, f"Panic error 0x{error_code}: Unknown error code")
        logger.error(message)
```

Output:
```
ContractCustomError: 0x4e487b710000000000000000000000000000000000000000000000000000000000000011
Panic error 0x11: Arithmetic operation results in underflow or overflow.
```

---

fselmo:
### What was wrong?

closes #2985 

### How was it fixed?

- Refactor contract error handling into its own `_utils` file for better code readability
- Add support for Solidity >= 0.8.0 `Panic` errors
- Create a test contract that tests all known cases directly
- Add core tests testing against `EthereumTesterProvider` and integration tests testing against `geth`
- Update config for go-ethereum integration test suite to turn on Shanghai by specifying a `shanghaiTime` (also added the glacier blocks to the config though those are less important since it was just difficulty bomb timing).
- Regenerate test fixture so the compiled contracts are updated**

_** It turns out we were a few geth versions behind though the filename was misleading. This PR regenerates the test fixture with `1.11.6` and bumps the CI's py-geth version and geth version as well._

### Todo:


tests/core/contracts/test_contract_panic_errors.py
```diff
@@ -0,0 +1,48 @@
...
+    panic_errors_contract_factory = w3.eth.contract(**PANIC_ERRORS_CONTRACT_DATA)
+    return deploy(w3, panic_errors_contract_factory, address_conversion_func)
+
+
+@pytest.mark.parametrize(
+    "panic_error,params",
+    (
+        ("01", []),
+        ("11", []),
+        ("12", [0]),
```
fselmo, line 29: This is kind of annoying but I had to pass these invalid values in, rather than write them into the contract, since Solidity checks these before compilation and doesn't allow it.

web3/_utils/contract_error_handling.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Ganache case:
+    if isinstance(data, dict) and response["error"].get("message"):
+        raise ContractLogicError(
+            f'execution reverted: {response["error"]["message"]}', data=data
+        )
+
+    # Parity/OpenEthereum case:
+    if data.startswith("Reverted "):
+        # "Reverted", function selector and offset are always the same for revert errors
```
fselmo, line 75: This refactor was mostly a copy / paste except for this part... I refactored this since I was already touching the code so that some code is shared and not doubled up and I tired to make the `revert_pattern` / `prefix` a bit more clear, comparing with the docstring notes and using similar terminology.

fselmo:
This should be ready for review now

web3/_utils/contract_error_handling.py
```diff
@@ -0,0 +1,132 @@
...
+        prefix = f"Reverted {SOLIDITY_ERROR_FUNC_SELECTOR}"
+        data_offset = f"{'00' * 31}" + "20"  # 0x0000...0020 (32 bytes)
```
kclowes: nit: you don't need the `f-string` here
```suggestion
        data_offset = "00" * 31 + "20"  # 0x0000...0020 (32 bytes)
```

web3/_utils/method_formatters.py
```diff
@@ -698,95 +691,6 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
     STANDARD_NORMALIZERS, RPC_ABIS
 )
 
-# the first 4 bytes of keccak hash for:
```
kclowes, line 701: So much better. Nice cleanup!

kclowes (approved):
Nice refactor! This looks good to me! I just left one nit that you can definitely ignore.

web3/exceptions.py
```diff
@@ -259,7 +259,7 @@ class ContractLogicError(Web3Exception):
     Raised on a contract revert error
     """
 
-    def __init__(self, message: Optional[str] = None, data: Optional[str] = None):
+    def __init__(self, message: Optional[str] = None, data: Optional[Any] = None):
```
reedsa: Was the type changed anywhere? From what I can tell it's still used as a str, is this Any fixing some compatibility issue?

fselmo: Ah, thanks for pointing this out. It isn't always a string, [sometimes it's a dict](https://github.com/ethereum/web3.py/pull/2986/files#diff-abffe06b9afa9b6544499307493e153d58127dfa36968d55ac4090a526da3e2fR68). But I do feel like this is technically a breaking change 🤔. I'm not sure if we leave it alone for now and address it in the `v7` changes instead.

cc: @kclowes  @pacrob

fselmo: Also, some type narrowing might be more appropriate. Maybe `Union[str, Dict[str, str]]`.

reedsa: Thanks for pointing to that dict instance, that's what I was looking for. Opening it up to Any is technically backwards compatible 😄 though it'd be great to lock this down to str | dict if possible.

kclowes: It shouldn't be breaking if it goes from `str` -> `Any`. If it was `Any` -> `str` or something narrowing, then it would be breaking. But I think a Union would be fine too.

fselmo: Sounds good, I'll narrow it down 👌🏼

kclowes: haha @reedsa beat me to it 😆

web3/exceptions.py
```diff
@@ -259,7 +260,11 @@ class ContractLogicError(Web3Exception):
     Raised on a contract revert error
     """
 
-    def __init__(self, message: Optional[str] = None, data: Optional[Any] = None):
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        data: Optional[Union[str, Dict[str, str]]] = None,
```
reedsa, line 266: 💯

reedsa (approved):
lgtm

PR #2992, opened 2023-06-12
drop pluggy from dev reqs

pacrob:
### What was wrong?

We had the pin `pluggy==0.13.1`.  We don't use it directly - `pytest` and `tox` have it as a dependency.
We'll be bumping to `tox>=4.0.0` soonish, which requires `pluggy>=1.0.0`.

### How was it fixed?
Removed our explicit dev requirement for `pluggy`

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/2992.internal.rst
+++ b/newsfragments/2992.internal.rst
@@ -1,0 +1,1 @@
+Removed `pluggy` from dev requirements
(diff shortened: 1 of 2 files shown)


fselmo (approved):
sgtm 👍🏼

kclowes (approved):

PR #2997, opened 2023-06-18
Uses `Self` Type in contract factory methods

konichuvak:
### What was wrong?

1. Return type of `Contract.factory()` is erroneously annotated as `"Contract"` instead of `Type["Contract"]`

2. Even once we fix the error above, type annotation for subclasses of `Contract` are broken

```python
from typing import Type

from web3 import Web3
from web3.contract import Contract


class MyContract(Contract):
    address = Web3.to_checksum_address("0x0000000000000000000000000000000000000000")

    @classmethod
    def factory(cls, *args, **kwargs) -> Type["Contract"]:
        return super().factory(*args, **kwargs)

    def my_method(self):
        pass

contract_factory = MyContract.factory(Web3())()
contract_factory.my_method()  # error: "Contract" has no attribute "my_method"  [attr-defined]
```

### How was it fixed?

`Self` Type [was added](https://peps.python.org/pep-0673/) in Python 3.11 which solves this problem of underspecified return types of constructor methods for subclasses.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/ethpm/contract.py
+++ b/ethpm/contract.py
@@ -27,6 +27,9 @@ from ethpm.exceptions import (
 from ethpm.validation.misc import (
     validate_empty_bytes,
 )
+from web3._utils.compat import (
+    Self,
+)
 from web3._utils.validation import (
     validate_address,
 )
@@ -59,7 +62,7 @@ class LinkableContract(Contract):
         super().__init__(address=address, **kwargs)  # type: ignore
 
     @classmethod
-    def factory(cls, w3: "Web3", class_name: str = None, **kwargs: Any) -> Contract:
+    def factory(cls, w3: "Web3", class_name: str = None, **kwargs: Any) -> Type[Self]:
         dep_link_refs = kwargs.get("unlinked_references")
         bytecode = kwargs.get("bytecode")
         needs_bytecode_linking = False
--- a/newsfragments/2997.misc.rst
+++ b/newsfragments/2997.misc.rst
@@ -1,0 +1,1 @@
+ Uses `Self` Type in contract factory methods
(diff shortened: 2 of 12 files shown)


pacrob:
`Self` is definitely a helpful addition, but we need to maintain compatibility with all active versions of python. 3.7 is being deprecated soon, but it'll be a while before we can use features that are brand new in 3.11.

I'll leave this open in case you want to try a different solution that maintains compatibility.

konichuvak:
> `Self` is definitely a helpful addition, but we need to maintain compatibility with all active versions of python. 3.7 is being deprecated soon, but it'll be a while before we can use features that are brand new in 3.11.
> 
> I'll leave this open in case you want to try a different solution that maintains compatibility.

Thanks for reviewing the PR.
I am not sure which part is incompatible with Python3.7 in here? We are importing `Self` from `typing_extensions`, which supports 3.7 and above.

pacrob:
OK, dug in a bit further. You are correct that we can use `Self` with earlier python versions, which is great news. It looks like `mypy` doesn't support it until `v1.0.0` though, and when I bump that it opens a number of other typing issues to deal with.

Once we get that in, I'll come back and pull this in. Thank you!

konichuvak:
> It looks like `mypy` doesn't support it until `v1.0.0` though

Great point, I did not catch that.
I bumped it to >=1.0.0 and fixed / ignored all the typing issues that popped up as a result!
For the exception of the ones in `ethpm-spec` which I am not sure how to deal with. It seems like that module is only type-checked when when mypy is ran locally.

pacrob:
`ethpm` is being deprecated, so probably OK to drop it from mypy checking. I'm not sure why the CI insists on running on your account though - it usually switches to our resources once I push a commit. Maybe try deleting the `web3.py` project from your CircleCI account? 

I've branched it in #3001 and CI is passing there, so we can just use that if we can't get it to run here.

pacrob:
Thanks for this, @konichuvak ! Your original commits are merged in #3001

PR #2998, opened 2023-06-18
fix(utils): Missing single quote in arguments count error message

BoboTiG:
### What was wrong?

When calling a function with an inappropriate number of arguments, current error is like:

```
TypeError: Incorrect argument count. Expected '3. Got '2'
```
Note the missing single quote around `'3`.

### How was it fixed?

Add the missing quote.

### Todo:

#### Cute Animal Picture


pacrob:
Looks good, thanks @BoboTiG !

PR #3000, opened 2023-06-20
ENSIP-15 support

### LINKED ISSUE #2967: Update ENS normalization to new standard
The NameHash team is open sourcing a new library that we hope can be useful for the Ethereum Python Community 🚀 

https://github.com/namehash/ens-normalize-python

ENS has been developing a new standard for normalizing ENS names such as the ".eth" names you're familiar with. This library linked above is fully compliant with this new ENS Normalize standard. E.g. ens.domains is using the new standard.

web3py needs to be updated for this new ENS Normalize standard. The existing ENS API in web3py is no longer compliant. For example, names that are ENS Normalized will not be recognized as normalized by the current web3py implementation.

How to approach this? Should we create a pull request?

fselmo:
### What was wrong?

Closes #2967 

### How was it fixed?

- Add ENSIP-15 normalization standard support
- Add tests

See specification here: [ENSIP-15](https://docs.ens.domains/ens-improvement-proposals/ensip-15-normalization-standard#specification)

### Todo:
- [x] Documentation for this support... likely in a separate PR? (separate PR incoming)


fselmo:
This is ready for a first review. The way I implemented it follows pretty true to the spec descriptions through each section. This isn't necessarily the fastest way to implement it but it may be easier to read this way. I don't see it presenting much of an issue but if we want to optimize it, I think we can tackle that in a separate PR.

fselmo:
We should also think about documenting this but I'm thinking also in a separate PR?

tests/ens/test_utils.py
```diff
@@ -95,7 +96,7 @@ def test_ens_encode_name_validating_total_encoded_name_size(name, expected):
     assert ens_encoded == expected
 
 
-@pytest.mark.parametrize("empty_name", ("", "."))
+@pytest.mark.parametrize("empty_name", ("", ".", None, " ", "  "))
```
reedsa, line 99: 💯 Nice improvement!

reedsa (approved):
Looks good as far as I can tell. I like the way you used the flag here, it makes sense to flip for v7. In the long run will this flag be removed or will that be necessary?

fselmo:
> In the long run will this flag be removed or will that be necessary?

The goal for ENS is for ENSIP-15 to entirely replace ENSIP-1 for normalization so I think on our end it would make sense to remove the flag in `v7` since there should only be one way to normalize a name by that point.

ens/_normalization.py
```diff
@@ -0,0 +1,512 @@
...
+class Token:
+    type: Literal[TokenType.TEXT, TokenType.EMOJI]
+    _original_text: str
+    _original_codepoints: List[int]
+    _normalized_codepoints: Optional[List[int]] = None
+
+    restricted: bool = False
+
+    def __init__(self, original_text: str, original_codepoints: List[int]) -> None:
+        self._raw_text = original_text
```
kclowes: nit, but I'm curious about the name change here, `raw_text` vs. `original_text`. And also how this gets used later?

ens/_normalization.py
```diff
@@ -0,0 +1,512 @@
...
+    def codepoints(self) -> List[int]:
+        return (
+            self._normalized_codepoints
+            if self._normalized_codepoints
+            else self._original_codepoints
+        )
+
+    @property
+    def text(self) -> str:
+        return _cps_to_text(self.codepoints)
```
kclowes: another nit: Since the spec uses the word 'codepoints' so often, it might be nice to make this method name `codepoints_to_text`. Feel free to take it or leave it though :)

ens/_normalization.py
```diff
@@ -0,0 +1,512 @@
...
+
+    @property
+    def as_text(self) -> str:
+        return ".".join(label.text for label in self.labels)
+
+
+# -----
+
+GROUP_COMBINED_VALID_CPS = []
+for d in NORMALIZATION_SPEC["groups"]:
```
kclowes, line 147: nit: what does `d` represent here?

ens/_normalization.py
fselmo: Noted 👌🏼

ens/_normalization.py
fselmo: I actually have some cleanups to this file coming in the documentation + ensip15 flag tests PR but I can add it here instead as a commit on top of the current state so we can better review it here.

pacrob (approved):
lgtm!

ens/ens.py
```diff
@@ -345,10 +345,9 @@ def get_text(self, name: str, key: str) -> str:
             the "0x59d1d43c" interface id
         :raises ResolverNotFound: If no resolver is found for the provided name
         """
-        node = raw_name_to_hash(name)
-        normal_name = normalize_name(name)
```
kclowes, line 349: I'm curious about this removal and the one on line 391 - was it redundant with the `raw_name_to_hash`?

ens/_normalization.py
```diff
@@ -0,0 +1,520 @@
...
+
+    if any(len(label) == 0 for label in raw_labels):
+        raise InvalidName("Labels cannot be empty")
+
+    normalized_labels = []
+
+    for label_str in raw_labels:
+        # _input takes the label and breaks it into a list of unicode code points
+        # e.g. "xyz👨🏻" -> [120, 121, 122, 128104, 127995]
+        _input = [ord(c) for c in label_str]
```
kclowes, line 456: TIL about `ord` 🤯

ens/_normalization.py
```diff
@@ -0,0 +1,520 @@
...
+            break
+
+    # check wholes
+    # start with set of all groups with confusables
+    retained_groups = set(VALID_BY_GROUPS.keys())
+    confused_chars = set()
+    buffer = set()
+
+    for char_cp in text_token_cps_set:
+        groups_excluding_ce = WHOLE_CONFUSABLE_MAP.get(char_cp)
```
kclowes, line 363: does `ce` mean confusable extent here?

kclowes (approved):
This ENSIP was a doozy! Thanks for taking it on! Overall this looks really great. Just a few observations:
- Can we catch the warnings that are emitted in the existing ens tests? :)
- It might be more understandable to split some of the big functions (in `ens/_normalization.py` in particular) into smaller functions with names that describe what they do, even though they don't get re-used. Feel free to take or leave this one
- It might also be nice to do some smoke testing to make sure the flag is working on the static methods. For example, pass an ens name where `nameprep('...', ensip15=True)` return a different result than `nameprep('...', ensip15=False)`

ens/ens.py
fselmo: `resolver()` normalizes the name already. So we're normalizing a name to normalize it again. I think we could add some tests to make sure every method that is expected to normalize a name does indeed normalize it just in case it gets taken out at some point then the tests would fail. I'll create an issue for that.

ens/_normalization.py
fselmo: It does.

ens/_normalization.py
fselmo: dict

fselmo:
>  Can we catch the warnings that are emitted in the existing ens tests? :)

Which tests?

> It might also be nice to do some smoke testing to make sure the flag is working on the static methods.

Coming up in a separate PR with documentation on this as well 👍🏼 

> t might be more understandable to split some of the big functions (in ens/_normalization.py in particular) into smaller functions with names that describe what they do, even though they don't get re-used.

These were just helped methods to help me extract information from the json files. I can do a pass-through and try to clarify some more things 👍🏼 

----
edit: Oh I see which tests. All of them basically 😆 ... yeah that's a good point. I'll see what I can do 🤔

ens/ens.py
kclowes: Ah, gotcha. In `async_ens.py` I believe we're doing a `normalize_name` still. Do you mind standardizing that in `async_ens`?

fselmo: Good call, can do 👍🏼

PR #3002, opened 2023-06-22
Fix max_priority_fee when eth_maxPriorityFeePerGas is unavailable

sanekto:
Hardhat anwers `{'code': -32601, 'message': 'Method eth_maxPriorityFeePerGas not found', 'data': {'message': 'Method eth_maxPriorityFeePerGas not found'}}` which raises `MethodUnavailable` instead of `ValueError`. 

It is not clear to me if the answer of other nodes causes ValueError or now only MethodUnavailable (so to replace the exception).

### What was wrong?

Explained above. I did not open an issue because I feel that in this case it is easier to explain the issue showing a possible fix. Sorry if this mess things, also to not provide tests.

### Todo:

### DIFF
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -806,6 +806,38 @@ class AsyncEthModuleTest:
         assert is_integer(max_priority_fee)
 
     @pytest.mark.asyncio
+    async def test_eth_max_priority_fee_with_fee_history_calculation_error_dict(
+        self, w3: "Web3"
+    ) -> None:
+        fail_max_prio_middleware = construct_error_generator_middleware(
+            {
+                RPCEndpoint("eth_maxPriorityFeePerGas"): lambda *_: {
+                    "error": {
+                        "code": -32601,
+                        "message": (
+                            "The method eth_maxPriorityFeePerGas does "
+                            "not exist/is not available"
+                        ),
+                    }
+                }
+            }
+        )
+        w3.middleware_onion.add(
+            fail_max_prio_middleware, name="fail_max_prio_middleware"
+        )
+
+        with pytest.warns(
+            UserWarning,
+            match=(
+                "There was an issue with the method eth_maxPriorityFeePerGas."
+                " Calculating using eth_feeHistory."
+            ),
+        ):
+            w3.eth.max_priority_fee
+
+        w3.middleware_onion.remove("fail_max_prio_middleware")  # clean up
+
+    @pytest.mark.asyncio
     async def test_eth_max_priority_fee_with_fee_history_calculation(
         self, async_w3: "AsyncWeb3"
     ) -> None:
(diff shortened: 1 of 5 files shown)


kclowes:
Thanks for the PR! I added a test and will add a newsfragment shortly.

web3/_utils/module_testing/eth_module.py
```diff
@@ -2315,6 +2342,32 @@ def test_eth_max_priority_fee(self, w3: "Web3") -> None:
...
+                }
+            }
+        )
+        w3.middleware_onion.add(
+            fail_max_prio_middleware, name="fail_max_prio_middleware"
+        )
+
+        with pytest.warns(
+            UserWarning,
+            match="There was an issue with the method eth_maxPriorityFeePerGas."
```
reedsa: I'm not sure if there is a comma missing here?

web3/middleware/fixture.py
```diff
@@ -81,8 +81,20 @@ def error_generator_middleware(
     ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
         def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
             if method in error_generators:
-                error_msg = error_generators[method](method, params)
-                return {"error": error_msg}
+                error = error_generators[method](method, params)
+                if isinstance(error, dict):
+                    msg = error["error"].get("message", "")
+                    code = error_generators[method](method, params).get("code", -32000)
```
reedsa: Nit: use a var to dry up repeat calls to `error_generators[method](method, params)`

fselmo: Agreed. Even tinier nit (take or leave), we can skip creating `msg`, `code`, or `data` vars altogether since we only use them once.

```py
error_obj = error_generators[method](method, params)

# we should probably add a check for the "error" key existing here as well.
# if not, we just return the object below.
if isinstance(error_obj, dict) and "error" in error_obj:
    return {
        "error": {
            "code": error_obj.get("code", -32000),
            "message": error_obj["error"].get("message", ""),
            "data": error_obj.get("data", ""),
        }
    }
else:
    return {"error": error_obj}
```

If we don't have an "error" it doesn't abide by the JSON-RPC spec 2.0 so we just return the object below. But assuming we have an `error` key might lead to ~~some weird error~~ [an obscure `KeyError`] at some point with some apis.

web3/_utils/module_testing/eth_module.py
fselmo: In the actual grammar of the messaging or just between the two match lines?

reedsa: Ah now I understand it, multiline concatenation. Hadn't seen this, was expecting `\` or `+`

fselmo: > Ah now I understand it, multiline concatenation. Hadn't seen this, was expecting `\` or `+`

For sure... sometimes using parentheses helps with readability too. Might be a nice nit.

```py
with pytest.warns(
    UserWarning,
    match=(
        "There was an issue with the method eth_maxPriorityFeePerGas. "
        "Calculating using eth_feeHistory."
    ),
)
```

pacrob (approved):
lgmt!

web3/middleware/fixture.py
kclowes: Nice catches, thanks! Updated!

kclowes:
Thanks @sanekto!

PR #3004, opened 2023-06-23
Use the latest available solidity version for compiling test contracts

fselmo:
### What was wrong?

Nothing was really wrong but we should just use the latest available solidity version for compiling maybe? Seems like a chore to keep changing that static version. I don't think we should ever rely on that lazy-configured version anyways but this seems like a nicer default.

### How was it fixed?

Grab the solidity version from the latest version in the supported versions sorted list.

#### Cute Animal Picture

```
 (\  /)
 ( o o )
   ' '
 \_| |
```


pacrob (approved):
lgtm!

PR #3011, opened 2023-06-28
`EthereumTesterProvider` support for get storage at

fselmo:
### What was wrong?

- *eth_getStorageAt* is supported in the latest *eth-tester*. Turn on support for it in *web3.py* via `EthereumTesterProvider`.

### How was it fixed?

- Bump eth-tester version to latest - `v0.9.0-beta.2`
- Turn on support for *eth_getStorageAt* for `EthereumTesterProvider`
- Add tests
    - Create a *StorageContract.sol* with actual values at storage slots and update all *eth_getStorageAt* tests to use this contract and check against the values received + formatted via the API
    - Turn on these tests for `EthereumTesterProvider`, assert behaves exactly as geth does by passing the same tests

---
~~Note - expected test failures for now:~~

~~This has been tested locally but the eth-tester version will need to be updated after it is released so that the CI runs can install that version and pass the `integration-ethtester-pyevm` tests.~~ [updated]

### Todo:
- [x] ethereum/eth-tester#264 needs to be merged and released and the eth-tester version needs to be updated in web3.py


pacrob (approved):
lgtm! nice drying in test_ethereum_tester

PR #3012, opened 2023-06-30
Async name to address new approach

### LINKED ISSUE #1990: Async name_to_address middleware

### LINKED ISSUE #2583: Async curry?
I was starting to look at ways to async `name_to_address` middleware now that `AsyncENS` is done. I ran into a problem that I wanted to get some feedback on. I ran into this [curry](https://github.com/ethereum/web3.py/blob/master/web3/_utils/normalizers.py#L207). Then I realized that I really need curry in this example because I want to pass in the w3 instance when the middleware is added so it can be used to resolve the name of the address. When I try to use curry on an async method I get and exception.

Has anyone come up with a solution for async curry methods already while writing some of the other async code? If not I will keep thinking about a good way to make this work.

fselmo:
### What was wrong?

ENS name to address needs async support

- closes #1990
- closes #2758
- closes #2583

### How was it fixed?

- Add async support for ENS name to address in a way that doesn't need to utilize functools as that was a messy approach for async (see #2758).
- Make tests more robust for both sync and async name to address middleware

### Todo:
- [x] Update relevant documentation

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -40,6 +40,7 @@ Sync middlewares include:
 Async middlewares include:
 
 * ``gas_price_strategy``
+* ``name_to_address``
 * ``attrdict``
 * ``validation``
 * ``gas_estimate``
@@ -66,6 +67,7 @@ AttributeDict
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. py:method:: web3.middleware.name_to_address_middleware
+               web3.middleware.async_name_to_address_middleware
 
     This middleware converts Ethereum Name Service (ENS) names into the
     address that the name points to. For example :meth:`w3.eth.send_transaction <web3.eth.Eth.send_transaction>` will
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -72,9 +72,9 @@ AttributeDict
     accept .eth names in the 'from' and 'to' fields.
 
     .. note::
-        This middleware only converts ENS names if invoked with the mainnet
-        (where the ENS contract is deployed), for all other cases will result in an
-        ``InvalidAddress`` error
+        This middleware only converts ENS names on chains where the proper ENS
+        contracts are deployed to support this functionality. All other cases will
+        result in a ``NameNotFound`` error.
 
 Gas Price Strategy
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/core/middleware/test_name_to_address_middleware.py
+++ b/tests/core/middleware/test_name_to_address_middleware.py
@@ -7,6 +7,7 @@ from web3 import (
     Web3,
 )
 from web3.exceptions import (
+    InvalidAddress,
     NameNotFound,
 )
 from web3.middleware import (
@@ -96,7 +97,7 @@ def test_pass_name_resolver_send_transaction_dict_args(
 
 
 def test_fail_name_resolver(w3):
-    with pytest.raises(NameNotFound, match=r".*ethereum\.eth.*"):
+    with pytest.raises(InvalidAddress, match=r".*ethereum\.eth.*"):
         w3.eth.get_balance("ethereum.eth")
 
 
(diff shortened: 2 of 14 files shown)


docs/middleware.rst
pacrob, line 1: Add async to the method header

```
.. py:method:: web3.middleware.name_to_address_middleware
               web3.middleware.async_name_to_address_middleware
```

pacrob: Also needs adding in `web3.middleware.__init__.py`

reedsa (approved):
Lgtm

tests/core/middleware/test_name_to_address_middleware.py
```diff
@@ -29,31 +32,150 @@ def address(self, name):
...
 
 
 def test_fail_name_resolver(w3):
-    return_chain_on_mainnet = construct_fixture_middleware(
-        {
-            "net_version": "2",
-        }
-    )
-    w3.middleware_onion.inject(return_chain_on_mainnet, layer=0)
-    with pytest.raises(InvalidAddress, match=r".*ethereum\.eth.*"):
```
kclowes, line 58: For backwards compatibility, I think we need to keep this as `InvalidAddress`? Or have `NameNotFound` inherit from `InvalidAddress`?

web3/_utils/normalizers.py
```diff
@@ -267,3 +265,31 @@ def normalize_bytecode(bytecode: bytes) -> HexBytes:
...
+        if async_w3 is None:
+            raise InvalidAddress(
+                f"Could not look up name {val!r} because no web3"
+                " connection available"
+            )
+
+        _async_ens = cast(AsyncENS, async_w3.ens)
+        if _async_ens is None:
+            raise InvalidAddress(
+                f"Could not look up name {val!r} because ENS is" " set to None"
```
kclowes: nit:
```suggestion
                f"Could not look up name {val!r} because ENS is set to None"
```

kclowes (commented):
This looks good to me! I left one nit and a comment around error backwards compatibility.

As a side note, when I was testing this out, I would see multiple warnings about ensip15_normalization, and it wasn't immediately clear how I could turn this off. I tried doing a `w3.ens.ensip15_normalization = True`, but that didn't work. It doesn't have to be in this PR, but it would be nice to add how to turn off that warning into the error message. I think it's confusing UX because when I was running  `w3.eth.get_balance('ethereum.eth')`, I get a warning about `normalize_name()` which I know is being called under the hood, but it's not super clear how or why IMHO. Here's a screenshot since my comment is sort of confusing:
image

tests/core/middleware/test_name_to_address_middleware.py
fselmo: Maybe we keep the mainnet check for sync for now? ~~This whole PR feels like it walks a gray line of breaking and not breaking.~~ (the ENSIP-15 changes feel like it but there might be a way around it).

fselmo:
@kclowes regarding the ENSIP-15 warnings, I think we talked about this the other day and these are the changes I mentioned might be best in a separate PR. We're going to have to pass around a `Web3` / `AsyncWeb3` instance internally and make sure it can make its way to `validate_address()`. I think this technically constitutes a "bugfix" and would be better as a separate PR.

tests/core/middleware/test_name_to_address_middleware.py
fselmo: I put a commit up that catches the `NameNotFound` and raises `InvalidAddress` with a note to remove this necessity in `v7`. Does that seem like a good compromise? Still remove strictness for mainnet-only and raise the same exception as before.

kclowes: Yep, that makes sense to me!

kclowes:
>  I think this technically constitutes a "bugfix" and would be better as a separate PR.

Yeah, a separate PR is fine by me. I just wanted to make sure it got flagged. I'll add an issue too so we can track.

PR #2700, opened 2022-10-30
Update ENS Resolver ABI

Snedashkovsky:
### What was wrong?

Deprecated [Resolver ABI](https://github.com/ethereum/web3.py/blob/master/ens/abis.py#L521)

Related to Issue https://github.com/ethereum/web3.py/issues/1839, https://github.com/ethereum/web3.py/issues/1397

### How was it fixed?

@Snedashkovsky 
- Update of Resolver ABI to be more generic. Research actual resolver contracts and create a more generic ABI [here](https://github.com/Snedashkovsky/ens-to-cyber/blob/main/web3py_abi_research.ipynb)

@fselmo 
- Match ENS's [Public Resolver 2 ABI](https://etherscan.io/address/0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41#code) as the default along with the Extended Resolver functions. I'd rather not have functions like `renounceOwnership()` available without the proper documentation and the proper UX, with aliased methods that are clear on what they do.

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3030, opened 2023-07-07
Add multicoin support for ENS / AsyncENS `address()` and `setup_address()`

### LINKED ISSUE #1839: Support ENS multicoin EIP-2304
* Version: 5.11.1
* Python: 3.7
* OS: osx/linux/win

[ENS PublicResolver](https://docs.ens.domains/contract-api-reference/publicresolver)
[EIP-2304](https://eips.ethereum.org/EIPS/eip-2304)

`RESOLVER` would require something like this:
```python
  {
      "constant": True,
      "inputs": [
          {
              "name": "node",
              "type": "bytes32"
          },
          {
              "name": "coinType",
              "type": "uint256"
          },
      ],
      "name": "addr",
      "outputs": [
          {
              "name": "ret",
              "type": "bytes"
          }
      ],
      "payable": False,
      "type": "function"
  },
```

[bruno.eth](https://app.ens.domains/name/bruno.eth) is an ENS domain with multiple coins set up.

By calling `addr(bytes32, unit256)` on its public resolver with:
- bytes32: `0x9de0b76f3b87380f0bcb9e41b403c69b88b01fbeaf785e83f38a6879902af71b`
- uint256: `434`

You should expect:
- bytes: `0x0aff6865635ae11013a83835c019d44ec3f865145943f487ae82a8e7bed3a66b`

Thanks

fselmo:
### What was wrong?

Extended support for *ENSIP-9 Multichain Address Resolution* via ENS / AsyncENS `address()` and `setup_address()` methods

closes #1839 

### How was it fixed?

- Add `ns.address(name, coin_type=coin_type)` + `ns.setup_address(name, address, coin_type=coin_type, transact=transact)`
- Minor DRY / refactors along the way in `ens/ens.py` and `ens/async_ens.py`
- Point internal ENS test contracts to the contract data in the dynamically compiled contracts located in `web3/_utils/contract_sources/contract_data` rather than using static values for bytecode, abi, etc.

--
Note: This originally aimed to add more methods but those seem unnecessary when users can already interact with the resolver contract via the updated ABI from #2700 

### Todo:
- [x] Add tests for multicoin support in `address()` and `setup_address()`
- [x] Add relevant documentation

#### Cute Animal Picture

Screenshot 2023-07-07 at 09 08 19


docs/ens_overview.rst
```diff
@@ -226,6 +243,22 @@ You can claim arbitrarily deep subdomains.
...
+introduced multichain address resolution, allowing users to resolve addresses from
+different chains, specified by the coin type index from
+`SLIP44 <https://github.com/satoshilabs/slips/blob/master/slip-0044.md>`_. The
+``setup_address()`` method on the ``ENS`` class supports multichain address setup
+via the ``coin_type`` keyword argument.
+
+.. code-block:: python
+
+    from ens.auto import ns
+    ns.setup_address('ens.eth', coin_type=60)  # ETH is coin_type 60
```
reedsa, line 259: Nit: example above sets a variable and uses that but this line doesnt add anything to this example. Suggest setting a variable here as well.

ens/contract_data.py
```diff
@@ -1,6 +1,12 @@
 # flake8: noqa
 import json
 
+"""
+This contract data is taken form the actual ENS contracts deployed on mainnet.
```
reedsa:
```suggestion
This contract data is taken from the actual ENS contracts deployed on mainnet.
```

reedsa: Recommend including the URL to the etherscan.io contract. I'm a noob and can't seem to find it, but maybe I was looking at the wrong address (https://etherscan.io/token/0xc18360217d8f7ab5e7c516566761ea12ce7f9d72#code)

ens/ens.py
```diff
@@ -65,8 +67,9 @@
 
 if TYPE_CHECKING:
     from web3 import Web3  # noqa: F401
-    from web3.contract import (  # noqa: F401
+    from web3.contract.contract import (  # noqa: F401
         Contract,
+        ContractFunction,
```
reedsa, line 72: Any reason `web3.contract.contract` is needed? Could it be added to the __init__ file and imported just as before?

ens/async_ens.py
```diff
@@ -134,21 +137,40 @@ def from_web3(cls, w3: "AsyncWeb3", addr: ChecksumAddress = None) -> "AsyncENS":
 
         return ns
 
-    async def address(self, name: str) -> Optional[ChecksumAddress]:
+    async def address(
+        self,
+        name: str,
+        coin_type: int = None,
```
reedsa: Should this be:
```suggestion
        coin_type: Optional[int] = None,
```

docs/ens_overview.rst
fselmo: The setter returns the transaction hash for the transaction that sets the address on the name. Do you mean to break it up into two more lines for clarity?

```py
from ens.auto import ns

# set up the address (this is the doc for the setup which is why this is here)
ns.setup_address('ens.eth', coin_type=60)  # ETH is coin_type 60

# assert the address was set
address = ns.address('ens.eth', coin_type=60)  # add this line?
assert address == '0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7'
```

This example is in line with the ones directly above it in the docs for `setup_address()` without the `coin_type` kwarg. Let me know if / how we can make it more clear.

ens/async_ens.py
fselmo: Yes! Thanks for catching.

pacrob (approved):
lgtm!

ens/ens.py
fselmo: I don't see any real utility in adding `ContractFunction` to the init. I don't think anyone would build their own `ContractFunction` object and it would just add to the list that is imported there so we can point to it easier internally. Thoughts?

ens/contract_data.py
fselmo: I added a link to the ENS tag on etherscan so we can see all their tagged contracts. Do you think that works well? 🤔

ens/async_ens.py
```diff
@@ -64,8 +66,9 @@
 )
 
 if TYPE_CHECKING:
-    from web3.contract import (  # noqa: F401
+    from web3.contract.async_contract import (  # noqa: F401
         AsyncContract,
+        AsyncContractFunction,
```
kclowes, line 71: nit: Do we just want to put `AsyncContractFunction` in `web3.contract.__init__.py`?

kclowes (approved):
Looks good to me! :shipit:

ens/async_ens.py
fselmo: Not sure. @reedsa mentioned that as well. Feels like importing it there doesn't have much utility other than making things feel slightly better internally but I could be convinced. Thoughts?

kclowes: whoops, meant to delete this when I saw your response to @reedsa's comment. I agree that it doesn't have much utility and should be internal.

fselmo:
@reedsa, did you get a chance to see my comments here? Let me know if your original comments were all addressed 👀

docs/ens_overview.rst
reedsa: I understand that now, thanks. I do think it makes sense to keep the assert in one line since the highlight here is `setup_address()`

reedsa (approved):
All good, thanks for your responses!

Issue #3040, opened 2023-07-13
Formatter Functions ETH<->Wei


sebastiandine:
I am looking for a feature to easily convert values in ETH to wei and vice versa, similar to these [formatters that are provided by ethers.js](https://docs.ethers.org/v5/api/utils/display-logic/#display-logic--functions).
```[tasklist]
### Tasks
```

fselmo:
I'm not sure I understand how this differs from `Web3.to_wei()` and `Web3.from_wei()`

```py
>>> from web3 import Web3

>>> Web3.from_wei(110000222200000000, "ether")
Decimal('0.1100002222')

>>> Web3.to_wei(1.0, "ether")
1000000000000000000
```

I'm going to close this as this has been implemented for some time. Please ask to re-open if you feel that your request is not covered by these methods.

PR #3041, opened 2023-07-13
Denomination Formatting Utilities (#3040)

sebastiandine:
Formatting Utilities to convert from higher denominations to smaller ones and and vice versa (including functions to specify the actual amount of decimals).
These functions are similar to the [formatting functions of `ethers.js`](https://docs.ethers.org/v5/api/utils/display-logic/#display-logic--functions)

### DIFF
--- a/newsfragments/3040.feature.rst
+++ b/newsfragments/3040.feature.rst
@@ -1,0 +1,5 @@
+========================================
+Denomination Formatter Utility Functions
+========================================
+New functions to convert values from a higher denomination to a smaller one and vice versa.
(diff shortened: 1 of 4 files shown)


reedsa:
Hey @sebastiandine sorry for the delay, but I like the idea here! There's a couple of methods in our utility library for `to_wei` and `from_wei` in the currency utils and another one in the humanize utils for `humanize_wei`. I recommend yours belong with those in the [eth-utils](https://github.com/ethereum/eth-utils/tree/master) library.

Would appreciate if if you opened a new PR and we'll get it across!

Issue #3042, opened 2023-07-13
Unupdated type declarations from #2488


tinom9:
### What was wrong?

#2488 made 2 formatted changes that were not reflected in their respective `web3.types` declarations.

The `LogReceipt` and `TxReceipt` types right now state:

```python
# web3/types.py

class LogReceipt(TypedDict):
    address: ChecksumAddress
    blockHash: HexBytes
    blockNumber: BlockNumber
    data: HexStr
    logIndex: int
    payload: HexBytes
    removed: bool
    topic: HexBytes
    topics: Sequence[HexBytes]
    transactionHash: HexBytes
    transactionIndex: int

TxReceipt = TypedDict(
    "TxReceipt",
    {
        "blockHash": HexBytes,
        "blockNumber": BlockNumber,
        "contractAddress": Optional[ChecksumAddress],
        "cumulativeGasUsed": int,
        "effectiveGasPrice": Wei,
        "gasUsed": int,
        "from": ChecksumAddress,
        "logs": List[LogReceipt],
        "logsBloom": HexBytes,
        "root": HexStr,
        "status": int,
        "to": ChecksumAddress,
        "transactionHash": HexBytes,
        "transactionIndex": int,
        "type": int,
    },
)
```

While their formatters convert them to:

```python
# web3/_utils/method_formatters.py

LOG_ENTRY_FORMATTERS = {
    "blockHash": apply_formatter_if(is_not_null, to_hexbytes(32)),
    "blockNumber": apply_formatter_if(is_not_null, to_integer_if_hex),
    "transactionIndex": apply_formatter_if(is_not_null, to_integer_if_hex),
    "transactionHash": apply_formatter_if(is_not_null, to_hexbytes(32)),
    "logIndex": to_integer_if_hex,
    "address": to_checksum_address,
    "topics": apply_list_to_array_formatter(to_hexbytes(32)),
    "data": HexBytes,
}

RECEIPT_FORMATTERS = {
    "blockHash": apply_formatter_if(is_not_null, to_hexbytes(32)),
    "blockNumber": apply_formatter_if(is_not_null, to_integer_if_hex),
    "transactionIndex": apply_formatter_if(is_not_null, to_integer_if_hex),
    "transactionHash": to_hexbytes(32),
    "cumulativeGasUsed": to_integer_if_hex,
    "status": to_integer_if_hex,
    "gasUsed": to_integer_if_hex,
[...]

PR #3043, opened 2023-07-13
Update type declarations

tinom9:
### What was wrong?

Related to Issue #3042

### How was it fixed?

Updated the `web3.types` module.

### Todo:

#### Cute Animal Picture

Wow

### DIFF
--- a/web3/types.py
+++ b/web3/types.py
@@ -225,21 +225,16 @@ class BlockData(TypedDict, total=False):
     proofOfAuthorityData: HexBytes
 
 
-class LogEntry(TypedDict):
+class LogReceipt(TypedDict):
     address: ChecksumAddress
     blockHash: HexBytes
     blockNumber: BlockNumber
-    data: HexStr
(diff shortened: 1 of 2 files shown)


web3/types.py
```diff
@@ -177,7 +177,7 @@ class LogReceipt(TypedDict):
     address: ChecksumAddress
     blockHash: HexBytes
     blockNumber: BlockNumber
-    data: HexStr
+    data: HexBytes
     logIndex: int
     payload: HexBytes
```
tinom9: Should `payload` and `topic` be removed?

pacrob: I think you are correct. I can't find any reference to expected `payload` or `topic` fields, not sure where those came from. Go ahead and remove, thank you!

tinom9: Done!

pacrob:
Looks good, @tinom9 ! Thanks for catching those details. I've rebased and will merge once tests are passing.

PR #3046, opened 2023-07-13
Test to catch if files are missing from a release

kclowes:
### What was wrong?

`v6.6.0` was missing a few new, required, files in the release. The wheel tests should catch this, but they didn't this time. 

### How was it fixed?

Now the built files get installed in a new temporary directory and a separate virtualenv so that the dist files don't pick up the files that were present after the circleci code check out. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/scripts/release/test_wheel_install.sh
+++ b/web3/scripts/release/test_wheel_install.sh
@@ -1,0 +1,7 @@
+rm -rf build dist
+python -m build
+cd $(mktemp -d)
+python -m venv venv-test
+source venv-test/bin/activate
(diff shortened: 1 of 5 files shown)


web3/scripts/release/test_wheel_install.sh
```diff
@@ -0,0 +1,7 @@
+rm -rf build dist
```
reedsa, line 4: Recommend using #!/bin/bash for the script. Can also use `set -e` to exit on error if that's useful.

Not sure what the equivalent is for windows, maybe it also follows some best practices.

pacrob (approved):
agree on shebang addition, otherwise lgtm!

Issue #2880, opened 2023-03-13
Open discussion on improving the Websocket experience


DefiDebauchery:
Following a short discussion on Discord with @fselmo, I'd like to open up some preliminary discussion on improving Websockets.

As some know, my personal use case for web3.py has been concurrency among 19 (and growing) different blockchains. Additionally, 99% of what I do is within Discord bots, so much of my code is within an async context, which I can appreciate not everyone is bound to.

I ended up hacking together my own Client and Manager classes, using `websockets` (which is what web3.py also utilizes), but I'm certainly missing some of the creature comforts of the library.

In evaluating whether I should have just created my own `Provider`, I came across several potential issues I have with the current implementation, but I don't know enough advanced python to know if they are _actually_ issues. For instance, the fact that each instantiated Websocket is forked to its own Thread. Are there ever any potential issues with adding a large number of Threads? My implementation uses a single Thread, which handles adding Clients; the Clients auto-reconnect based on `websockets`' [reconnection recommendation](https://websockets.readthedocs.io/en/stable/faq/client.html#how-do-i-reconnect-when-the-connection-drops).

One of the biggest drawbacks to the current `WebsocketProvider` is the lack of handling for `eth_subscribe`. I think one useful approach, in my code below, is to require a callback method, which is executed against the `subscription` id provided by `eth_subscribe`. Obviously, we would need to codify callback signatures - if this were turned into a Provider, I think that would probably be sufficient.

Also, when a websocket reconnects, it will re-issue the subscribe messages so that connection handling can be relatively seamless.

Is any of this code (if cleaned up) particularly useful for inclusion with web3py?

```py
class WebsocketClient:
    def __init__(self, network: str, ws_url: str):
        self.network: str = network
        self.ws_url: str = ws_url
        self.futures: dict = {}
        self.callbacks: dict = {}
        self.websocket: websockets.WebSocketClientProtocol = None
        self.connection_count = itertools.count(start=1)  # Some RPCs (notably Ankr) exhibit inconsistent behavior with connection ids of 0
        self.connected = asyncio.Event()

    async def send(self, message: dict, callback: Callable = None):
        message_id = next(self.connection_count)

        message = message | {'jsonrpc': '2.0', 'id': message_id}
        if not message.get('params'):
            message['parms'] = []

        print('>>', self.network, message)
        payload = {'message': message, 'callback': callback}

        future = asyncio.get_event_loop().create_future()
[...]

PR #3048, opened 2023-07-19
WebsocketProviderV2

fselmo:
### What was wrong?

Related to Issue #2880
- The current ``WebsocketProvider`` doesn't keep an open connection and doesn't allow for subscription support
- Missing support for `eth_subscribe`

### How was it fixed?

- Started as an expirement a few months back. I put together a new version of the Websocket Provider that maintains an open connection and processes responses asynchronously using all the middleware and formatters we're used to. 
- I polished up as much as I could of that hacky version and abstracted a lot of the logic into classes, a transient async response processing cache, and kept going along the same route until it passed all the integration tests. Very open to suggestions on improvements.

Features:
  - Persistent connection async websocket provider support
  - Debug logging where appropriate around the transient async response processing cache and websocket connection
  - `eth_subscribe`  (and `eth_unsubscribe`) support

Usage (from the docs added):

```python

>>> import asyncio
>>> from web3 import AsyncWeb3
>>> from web3.providers import WebsocketProviderV2

>>> LOG = True  # toggle debug logging
>>> if LOG:
...     import logging
...     logger = logging.getLogger("web3.providers.WebsocketProviderV2")
...     logger.setLevel(logging.DEBUG)
...     logger.addHandler(logging.StreamHandler())

>>> async def ws_v2_subscription_example():
...     async with AsyncWeb3.persistent_websocket(
...         WebsocketProviderV2(f"ws://127.0.0.1:8546")
...     ) as w3:
...         # subscribe to new block headers
...         subscription_id = await w3.eth.subscribe("newHeads")
...
...         unsubscribed = False
...         while not unsubscribed:
...             async for response in w3.listen_to_websocket():
...                 print(f"{response}\n")
...                 # handle responses here
...
...                 if some_condition:
...                     # unsubscribe from new block headers
...                     unsubscribed = await w3.eth.unsubscribe(subscription_id)
...                     break
...
...         # still an open connection, make any other requests and get
...         # responses via send / receive
...         latest_block = await w3.eth.get_block("latest")
...         print(f"Latest block: {latest_block}")
...
...         # the connection closes automatically when exiting the context
...         # manager / `async with` block

>>> asyncio.run(ws_v2_subscription_example())
```

### Todo:
- [x] Fix failing docs
- [x] Add documentation for WebsocketProviderV2
- [x] Fix failing `EthereumTesterProvider` / eth-tester based core tests


fselmo:
This is ready for an initial review... sorry for the PR size here 😔 

cc: @DefiDebauchery

docs/providers.rst
```diff
@@ -211,6 +218,72 @@ WebsocketProvider
...
+        >>> import asyncio
+        >>> from web3 import Web3
+        >>> from web3.providers import WebsocketProviderV2
+
+        >>> LOG = True  # toggle debug logging
+        >>> if LOG:
+        >>>     import logging
+        >>>     logger = logging.getLogger("web3.providers.WebsocketProviderV2")
+        >>>     logger.setLevel(logging.DEBUG)
+        >>>     logger.addHandler(logging.StreamHandler())
```
pacrob: nit
```suggestion
        ...     import logging
        ...     logger = logging.getLogger("web3.providers.WebsocketProviderV2")
        ...     logger.setLevel(logging.DEBUG)
        ...     logger.addHandler(logging.StreamHandler())
```

fselmo: As far as I've ever used the ellipsis in code blocks it's a continuation of the same statement. These are all separate lines. Is this common usage?

pacrob: I see it as a single statement because it's all within a single `if`.

reedsa (approved):
🚢 Nothing stands out in my mind, lgtm! Will be curious to see adoption feedback and see if there are challenges migrating to v2.

web3/providers/async_base.py
```diff
@@ -46,6 +50,7 @@ class AsyncBaseProvider:
     )
 
     is_async = True
+    has_persistent_connection = False
```
kclowes, line 53: I haven't thought too hard about this, but should this flag get put on the `BaseProvider` too? Or is it only here because of the inheritance chain?

kclowes (commented):
This looks good to me from a code perspective! I'll take some time on Monday to play around with the UI. It'll be awesome to get this in!

docs/providers.rst
fselmo: Gotcha. I'm good with it 👍🏼. I'll pull this change in.

web3/providers/async_base.py
fselmo: Good point. I think it does make sense to add it there for consistency.

docs/providers.rst
```diff
@@ -211,6 +218,72 @@ WebsocketProvider
...
+      ``'ws://localhost:8546'``.
+    * ``websocket_kwargs`` this should be a dictionary of keyword arguments which
+      will be passed onto the ws/wss websocket connection.
+    * ``call_timeout`` is the timeout in seconds, used when receiving or sending data
+      over the connection. Defaults to ``None`` (no timeout).
+
+    .. code-block:: python
+
+        >>> import asyncio
+        >>> from web3 import Web3
```
kclowes: Should be `AsyncWeb3`, I believe

kclowes (approved):
Left a nit on the docs, but looks great otherwise!

docs/providers.rst
fselmo: Yes! Thanks for catching.

PR #3060, opened 2023-07-24
fix default windows ipc provider path

### LINKED ISSUE #3058: Default IPC path is incorrect on Windows with Anaconda 2023.07
* Version: 6.6.1
* Python: 3.11
* OS: win

I updated my Anaconda to the latest version recently, which uses Python 3.11. web3.py is no longer able to set the default IPC path for IPCProvider on Windows. The problem and fix are as follows:

In ipc.py,  line 105
`ipc_path = os.path.join("\\\\", ".", "pipe", "geth.ipc")`
makes the default IPC path ` '\\\\\\.\\pipe\\geth.ipc'`, which cannot be found with `os.path.exists(ipc_path)` in the next line

### How can it be fixed?
In ipc.py, replace line 105
`ipc_path = os.path.join("\\\\", ".", "pipe", "geth.ipc")`
with
`ipc_path = '\\\.\pipe\geth.ipc'` as is described in the [documentation](https://web3py.readthedocs.io/en/latest/providers.html#web3.providers.ipc.IPCProvider).

pacrob:
### What was wrong?

`os.path.join` seems to add additional backslashes in py3.11 on windows. I couldn't find a reason/changelog explanation, but at least found someone with the same problem: https://stackoverflow.com/questions/76107858/python-3-10-3-11-changes-on-os-path-join

Closes #3058 

### How was it fixed?

Changed the default path to `r"\\.\pipe\geth.ipc"`. Tested against py3.9, 10, and 11 on Windows 10.

Updated docs to match [geth](https://geth.ethereum.org/docs/interacting-with-geth/rpc#ipc-server).

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -173,7 +173,7 @@ IPCProvider
 
     - On Linux and FreeBSD: ``~/.ethereum/geth.ipc``
     - On Mac OS: ``~/Library/Ethereum/geth.ipc``
-    - On Windows: ``\\\.\pipe\geth.ipc``
+    - On Windows: ``\\.\pipe\geth.ipc``
 
 
 WebsocketProvider
--- a/newsfragments/3058.bugfix.rst
+++ b/newsfragments/3058.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed default windows IPC provider path to work with python 3.11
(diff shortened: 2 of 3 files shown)


reedsa (approved):
Lgtm

fselmo (approved):
lgtm 👍🏼

PR #3061, opened 2023-07-25
Proper RPC error responses

### LINKED ISSUE #252: Have error formatting comply with the JSON-RPC error format
### What was wrong?

eth-tester (or possibly py-evm) returns errors as strings instead of dicts, which makes downstream libraries have to account for a case in testing that isn't quite what they'd see if they were connected to a node.

### How can it be fixed?

A JSON-RPC error format can be found [here](https://www.jsonrpc.org/specification#error_object). Convert errors coming from `eth-tester` to be compliant.

reedsa:
### What was wrong?

When an invalid request is sent to a endpoint that does not exist or hasn't been implemented, the error returned does not comply with the JSON-RPC spec.

Closes https://github.com/ethereum/eth-tester/issues/252

### How was it fixed?

Return response objects with required fields for the RPC response.
Raise exception from `error.data` response to expose the underlying cause of the issue.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/manager/test_response_formatters.py
+++ b/tests/core/manager/test_response_formatters.py
@@ -39,7 +39,6 @@ METHOD_NOT_FOUND_RESP_FORMAT = {
         "code": -32601,
         "message": "the method eth_getTransactionByHash does not exist/is not "
         "available",
-        "data": MethodUnavailable("method not found"),
     },
 }
 ETH_TESTER_METHOD_NOT_FOUND_RESP_FORMAT = {
--- a/tests/core/manager/test_response_formatters.py
+++ b/tests/core/manager/test_response_formatters.py
@@ -39,6 +39,7 @@ METHOD_NOT_FOUND_RESP_FORMAT = {
         "code": -32601,
         "message": "the method eth_getTransactionByHash does not exist/is not "
         "available",
+        "data": MethodUnavailable("method not found"),
     },
 }
 ETH_TESTER_METHOD_NOT_FOUND_RESP_FORMAT = {
--- a/tests/core/providers/test_tester_provider.py
+++ b/tests/core/providers/test_tester_provider.py
@@ -62,3 +62,31 @@ def test_eth_tester_provider_properly_handles_eth_tester_error_messages(
         TransactionFailed, match="execution reverted: The error message."
     ):
         provider.make_request(RPCEndpoint("eth_blockNumber"), [])
+
+
+def test_eth_tester_provider_properly_handles_eth_tester_key_error_messages():
+    provider = EthereumTesterProvider(api_endpoints={})
+    response = provider.make_request(RPCEndpoint("eth_blockNumber"), [])
+
+    assert response["error"]["code"] == -32601
+    assert isinstance(response["error"]["data"], KeyError)
+    assert response["error"]["message"] == "Unknown RPC Endpoint: eth_blockNumber"
+
+
+def test_eth_tester_provider_properly_handles_eth_tester_not_implmented_error_messages(
+    mocker,
+):
+    mocker.patch(
+        "eth_tester.main.EthereumTester.get_block_by_number",
+        side_effect=NotImplementedError("The error message."),
+    )
+
+    provider = EthereumTesterProvider()
+    response = provider.make_request(RPCEndpoint("eth_blockNumber"), [])
+
+    assert response["error"]["code"] == -32601
+    assert isinstance(response["error"]["data"], NotImplementedError)
+    assert (
+        response["error"]["message"]
+        == "RPC Endpoint has not been implemented: eth_blockNumber"
+    )
--- a/tests/core/providers/test_tester_provider.py
+++ b/tests/core/providers/test_tester_provider.py
@@ -69,7 +69,6 @@ def test_eth_tester_provider_properly_handles_eth_tester_key_error_messages():
     response = provider.make_request(RPCEndpoint("eth_blockNumber"), [])
 
     assert response["error"]["code"] == -32601
-    assert isinstance(response["error"]["data"], KeyError)
     assert response["error"]["message"] == "Unknown RPC Endpoint: eth_blockNumber"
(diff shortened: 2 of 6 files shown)


pacrob (approved):
lgtm!

pacrob (approved):
:+1:

web3/manager.py
```diff
@@ -213,7 +212,9 @@ def formatted_response(
             if isinstance(response["error"], dict):
                 resp_code = response["error"].get("code")
                 if resp_code == -32601:
-                    raise MethodUnavailable(response["error"])
+                    e = response["error"].get("data")
```
kclowes: To go along with my comment above, this method is used across all providers, not just eth-tester. We'll want to keep this `MethodUnavailable` error here for other clients who don't implement a certain method.  Right now, we fall back to raising the `ValueError` on line 218.

tests/integration/test_ethereum_tester.py
```diff
@@ -598,9 +604,3 @@ def test_personal_unlock_account_failure(self, w3, unlockable_account_dual_type)
             unlockable_account_dual_type, "bad-password"
         )
         assert result is False
-
-    @pytest.mark.xfail(
-        raises=ValueError, reason="list_wallets not implemented in eth-tester"
```
kclowes, line 603: Nice catch!

tests/core/manager/test_response_formatters.py
```diff
@@ -39,6 +39,7 @@
         "code": -32601,
         "message": "the method eth_getTransactionByHash does not exist/is not "
         "available",
+        "data": MethodUnavailable("method not found"),
```
kclowes: This is the actual response from a client (in this case Geth) that doesn't implement a method. The `data` key doesn't get returned, so we can't rely on that to determine which error to throw. We'll want to make this test pass without the `data` key.

kclowes (commented):
I left a couple comments around the same thing. Let me know if they don't make sense!

web3/providers/eth_tester/main.py
```diff
@@ -135,6 +136,25 @@ def is_connected(self, show_traceback: bool = False) -> Literal[True]:
         return True
 
 
+def _make_response(result: Any, message: str = "") -> RPCResponse:
+    def _error_response(error: str) -> RPCResponse:
+        return RPCResponse(
+            {
+                "id": 1,
+                "jsonrpc": "2.0",
+                "error": RPCError({"code": -32601, "message": error, "data": None}),
```
reedsa: @kclowes I've set data for an error response to None here. I cant think of anything else that we could send back but open to suggestions.

kclowes: I think that makes sense. It's optional, so we could also just skip the `data` key altogether, but either way works. 

Are we using this error response function for errors besides `MethodUnavailable`? I believe the `-32601` error code is reserved for that case only. I'm not sure why I can't link directly to it, but the codes are in section 5.1 of [this doc](https://www.jsonrpc.org/specification) just in case we want to check for that before building the `_error_response`

tests/core/providers/test_tester_provider.py
```diff
@@ -62,3 +62,29 @@ def test_eth_tester_provider_properly_handles_eth_tester_error_messages(
         TransactionFailed, match="execution reverted: The error message."
     ):
         provider.make_request(RPCEndpoint("eth_blockNumber"), [])
+
+
+def test_eth_tester_provider_properly_handles_eth_tester_key_error_messages():
+    provider = EthereumTesterProvider(api_endpoints={})
+    response = provider.make_request(RPCEndpoint("eth_blockNumber"), [])
```
kclowes, line 69: Does this raise? If it does, it might be nice to make sure the exception class is what we expect, similar to how we do in the test above with a `with pytest.raises: ...`

kclowes (approved):
I left a few more comments, feel free to take them or leave them!

web3/providers/eth_tester/main.py
reedsa: This RPCError response is returned if there is  a `KeyError` or a `NotImplementedError` in `_make_request`.

tests/core/providers/test_tester_provider.py
reedsa: For missing or not implemented errors, it will not raise but instead return a reponse with an RPCError

web3/providers/eth_tester/main.py
fselmo: nit: do we need to define these inner methods? Can this be

```py
def _make_response(result: Any, message: str = "") -> RPCResponse:
    if isinstance(result, Exception):
        return RPCResponse(
            {
                "id 1,
                "jsonrpc": "2.0",
                "error": RPCError({"code": -32601, "message": error, "data": None}),
            }
        )
    # optional `else` here at this point too, could be removed since we either return up top or not
    else:
        return RPCResponse({"id": 1, "jsonrpc": "2.0", "result": result})
```

fselmo: Agreed on the error code here. Maybe we should have a static map of conditions to error codes and apply that based on the exception in some way. Would be good to have the error codes be correct for each case or at least only use this code when the error is due to the method not being available.

reedsa: @fselmo Sounds good to both of your points. As implemented, the endpoint not being defined is treated the same as if the endpoint is there but has not been implemented. As of now, both cases return 32601. Is there a distinction that should be made for each of those cases?

fselmo: Oh, nope, my mistake. Didn't see that `KeyError` in this sense means the method is missing. Looks like the error code is appropriate in both of those cases.

No need to touch any other error code either. If that's requested at some point, we'll cross that bridge. Thanks for the explanation 👌🏼

fselmo (approved):
Looks good 👍🏼 . Closer to homogeneity in the eth-tester responses 🎉

PR #3062, opened 2023-07-25
Corrected Python and rst files

criadoperez:
### What was wrong?

Some typos in python and rst files

### How was it fixed?

Corrected them


pacrob:
looks good @criadoperez ! Thanks!

PR #3065, opened 2023-07-26
Fix rpc gas strategy

fselmo:
### What was wrong?

- `rpc_gas_price_strategy` incorrectly returns the hex value without convesion to `int` even though the defined return type is `Wei`.
- We don't have tests to catch this
- bonus: Let's also not assume the value will always be an `int` and only convert to `hex` if it is an `int` in the `gas_price_strategy_middleware`.

### How was it fixed?

- In the `rpc_gas_price_strategy`, use the `eth` module `gas_price` call which always returns an `int`, rather than `request_blocking` which can be variable depending on the provider.
- Convert `gasPrice` value from strategy to `hex` only if it's an `int`
- Add tests that would've failed before this change. The reason here is we only test on EthereumTesterProvider (core test suite)... but eth-tester already returns an `int`, even though most providers return the `hex` value.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3065.bugfix.rst
+++ b/newsfragments/3065.bugfix.rst
@@ -1,0 +1,1 @@
+Fix return type for ``rpc_gas_price_strategy`` to ``int`` but also only convert the ``strategy_based_gas_price`` to ``hex`` if it is an ``int`` in the ``gas_price_strategy_middleware``.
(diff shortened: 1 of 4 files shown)


kclowes (approved):
LGTM! 🚀

PR #3066, opened 2023-07-26
Add ens name-to-address support for eth_subscribe method

fselmo:
### What was wrong?

- `eth_subscribe` request for "logs" takes an optional dict arg that can include an address param. This case isn't supported by the usual flow for the rpc-abi request formatters because those parse *either* sequence args to methods *or* dict args, but not dict args within list sequences... e.g. `("logs", {"address": "0x...", "topics": ["0x...", "0x..."]})`

### How was it fixed?

- Parse this special case differently in the middleware for async ENS name-to-address.

Note: It isn't ideal to not add a test here. However, we don't yet have a good test flow for `eth_subscribe`. Since this is still in beta, via WebsocketProviderV2, I suggest to test this locally and we can work on `eth_subscribe` tests separately.

### Todo:

#### Cute Animal Picture

MidJourney's interpretation of "a websocket as an animal":

Screenshot 2023-08-01 at 15 52 38

### DIFF
--- a/newsfragments/3066.feature.rst
+++ b/newsfragments/3066.feature.rst
@@ -1,0 +1,1 @@
+ENS name-to-address support for ``eth_subscribe``.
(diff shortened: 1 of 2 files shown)


wolovim (approved):

pacrob (approved):
lgtm!

PR #3070, opened 2023-08-02
Update async http docs

kclowes:
### What was wrong?

- The AsyncHTTPProvider docs still said that you couldn't use the ENS middleware, but you can. 
- Also removed the `data` key from the Eth-tester RPCError response to standardize with other clients. 

### How was it fixed?
Updated docs, and eth-tester error response. 

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -359,11 +359,6 @@ AsyncHTTPProvider
     `aiohttp <https://docs.aiohttp.org/en/stable/>`_ library for making requests.
 
 
-    The ``AsyncHTTPProvider`` has almost all the same functionality available as the ``HTTPProvider``.
-    The only documented exceptions to this are:
-
-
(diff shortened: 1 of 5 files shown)


docs/providers.rst
```diff
@@ -363,7 +363,6 @@ AsyncHTTPProvider
     The only documented exceptions to this are:
```
fselmo: Maybe get rid of this exceptions line too? I don't think there are any exceptions listed after it now that the ENS line is gone.

fselmo (approved):
lgtm 👍🏼 I think we can remove the "exceptions" when comparing to sync HTTPProvider but other than that ship it 🚢

docs/providers.rst
kclowes: Nice catch! Removed!

PR #3074, opened 2023-08-13
fix type of `input` in `TxData`

minaminao:
The type of `input` in `TxData` is actually `HexBytes`(`bytes`), not `HexStr`(`str`).

```
>>> type(tx["input"])
<class 'hexbytes.main.HexBytes'>
```


fselmo:
Thanks @minaminao 👍🏼 . I added a newsfragment and will merge.

PR #3086, opened 2023-08-28
fix bug in websocketProviderV2 docs

Vojtch159:
### What was wrong?
The documentation example doesn't work. 

### How was it fixed?
Changed the documentation. 
### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -313,6 +313,7 @@ and reconnect automatically if the connection is lost. A similar example, using 
     ...         WebsocketProviderV2(f"ws://127.0.0.1:8546")
     ...     ):
     ...         try:
+    ...             await w3.provider.connect()
     ...             ...
     ...         except websockets.ConnectionClosed:
     ...             continue


Vojtch159:
The issue exists because the iterator yields the object itself but the enter method returns the connected object.

fselmo:
I don't believe this addition is necessary. If there is no connection by the time the first call is made, the connection is initiated.

```py
async def wsV2_example():
    async for w3 in AsyncWeb3.persistent_websocket(
        WebsocketProviderV2(f"wss://mainnet.infura.io/ws/v3/{INFURA_PROJECT_ID}")
    ):
        try:
            subscription_id = await w3.eth.subscribe("newHeads")
        
            while True:
                    print(response)
                    ...
         
         except websockets.ConnectionClosed:
             continue

asyncio.run(wsV2_example())
```

Do you have issues running the above example? What is the specific issue you are seeing?

Vojtch159:
Oh, then I apologise for wasting your time here. I investigated, and when I don't connect manually my code gets stuck at creating filters (I'm subscribing to specific "logs"). At the moment I'm not sure why it happens but it seems like a different issue and I'll have to investigate it more deeply. Everything starts working perfectly fine when I do the connection manually right at the start 😕 Anyway, thank you for the response!

PR #3087, opened 2023-08-29
RPC Error Revert Reason Cleanup

### LINKED ISSUE #3053: Re-think Revert Reason Parsing
### What was wrong?

Many nodes/networks return error data in different formats that doesn't comply with the JSON-RPC error spec. `raise_contract_logic_error_on_revert` attempts to parse the errors, and is getting unwieldy. It returns unexpected errors when parsing fails. 

### How can it be fixed?

One option is to check if the error returned matches the JSON-RPC error spec. If it does, parse the error and return the parsed message.  If we go this route, I think we should still keep the `data` attribute on `ContractLogicError` so that users can pass along the raw data to other contracts if they'd like. If the response doesn't match the spec, we'd return some message that indicates that we can't parse it, along with the raw error data. I'm also open to other ideas.

reedsa:
### What was wrong?

Closes #3053 

### Todo:

#### Cute Animal Picture


tests/core/manager/test_response_formatters.py
```diff
@@ -185,6 +201,30 @@ def test_formatted_response_raises_errors(
...
+            INVALID_MESSAGE_RESP_FORMAT,
+            (),
+            identity,
+            identity,
+            BadResponseFormat,
+            re.escape(
+                f"The response was in an unexpected format and unable to be parsed. "
+                f"The error is: error['message'] must be a string, integer or null. "
+                f"The raw response is: {INVALID_MESSAGE_RESP_FORMAT}"
+            ),  # noqa: E501
```
pacrob: can drop the #noqa

web3/_utils/contract_error_handling.py
```diff
@@ -49,67 +49,47 @@
...
-    """
-    Reverts contain a `data` attribute with the following layout:
-        "Reverted "
-        Function selector for Error(string): 08c379a (4 bytes)
-        Data offset: 32 (32 bytes)
-        String length (32 bytes)
-        Reason string (padded, use string length from above to get meaningful part)
 
-    See also https://solidity.readthedocs.io/en/v0.6.3/control-structures.html#revert
+def _parse_openethereum_hex_error(data: str) -> str:
```
pacrob: OpenEthereum/Parity are definitely defunct, but I don't know if any other client uses the `Revert`-prefixed `data` format. No docs I've looked through seem to be of any help. Short of testing against all actively-developed clients and seeing what they spit out, I'd say just change naming here, calling it "revert-prefixed" (or something) error data instead of openethereum.

web3/_utils/contract_error_handling.py
```diff
@@ -49,67 +49,47 @@
...
+                "utf8"
+            )
         except UnicodeDecodeError:
             warnings.warn("Could not decode revert reason as UTF-8", RuntimeWarning)
             raise ContractLogicError("execution reverted", data=data)
 
-    # --- EIP-3668 | CCIP Read --- #
+        raise ContractLogicError(f"execution reverted: {reason_string}", data=data)
+
     if data[:10] == OFFCHAIN_LOOKUP_FUNC_SELECTOR:
```
pacrob: If they are mutually exclusive, I'd say use an `if-elif-else` instead of 3 `if`'s here. Is there a test that catches if none of the conditions are met?

web3/_utils/contract_error_handling.py
```diff
@@ -134,11 +113,43 @@ def raise_contract_logic_error_on_revert(response: RPCResponse) -> RPCResponse:
...
-    if message_present and response["error"].get("code", "") == 3:
-        raise ContractLogicError(message, data=data)
-    # Geth Revert without error message case:
-    if message_present and "execution reverted" in message:
-        raise ContractLogicError("execution reverted", data=data)
+
+def raise_contract_logic_error_on_revert(response: RPCResponse) -> RPCResponse:
+    """
+    Reverts contain a `data` attribute with the following layout:
+        "Reverted "
```
pacrob: Given the above, some note here that the "Reverted" may or may not be present, while the rest of the layout is standard.

web3/manager.py
```diff
@@ -198,43 +208,66 @@ async def _coro_make_request(
...
+    # Responses are inconsistent so allow for missing required JSON-RPC properties
+    #
     @staticmethod
     def formatted_response(
         response: RPCResponse,
         params: Any,
         error_formatters: Optional[Callable[..., Any]] = None,
         null_result_formatters: Optional[Callable[..., Any]] = None,
     ) -> Any:
+        if "jsonrpc" in response and response["jsonrpc"] != "2.0":
```
fselmo, line 231: nit: This can be simplified a bit

```py
if response.get("jsonrpc") != "2.0":
```

---
edit: This may actually change the expected response in the tests. I think we could use `'"jsonrpc" field must be present in response and its value must be "2.0".'`. Or open to suggestions there too. More clarifying info can only be better though, I suppose.

web3/manager.py
```diff
@@ -198,43 +208,66 @@ async def _coro_make_request(
...
+            )
+
+        # Format and validate errors
         if "error" in response:
+            error = response.get("error")
+            if error is None or isinstance(error, str):
+                raise ValueError(error)
+
+            # Errors may include a code
+            # https://docs.alchemy.com/reference/error-reference#json-rpc-error-codes
```
fselmo: Maybe link to the JSON-RPC 2.0 spec instead of Alchemy here? And I would change the verbiage of "may" to "must include an integer code" as per the JSON-RPC 2.0 spec. Unless you know of some cases that deviate from this and are part of Ethereum JSON-RPC specs. I know that can happen too unfortunately (see: eth subscriptions).

```suggestion
            # https://www.jsonrpc.org/specification#error_object
```

web3/manager.py
```diff
@@ -198,43 +208,66 @@ async def _coro_make_request(
...
+            code = error.get("code")
+            if not isinstance(code, int):
+                _raise_bad_response_format(response, "error['code'] must be an integer")
+            elif code == METHOD_NOT_FOUND:
+                raise MethodUnavailable(error)
+
+            if not type(error.get("message")) in (str, int, None):
```
fselmo: I'm curious where the accepted value of `int` came from here for message. I think it might be OK with being lenient here and returning the `repr()` of anything that isn't a String. But if the goal is to validate according to JSON-RPC 2.0 specs then this is always a String as per the link above.

web3/manager.py
```diff
@@ -198,43 +208,66 @@ async def _coro_make_request(
...
-            return response["result"]
-        elif response.get("result") is not None:
-            return response["result"]
+            raise ValueError(error)
+
+        # Format and validate results
+        elif "result" in response:
+            # Null values for result should apply null_result_formatters
+            # Skip when result not present in the response (fallback to False)
+            if response.get("result", False) in NULL_RESPONSES:
```
fselmo, line 283: nit: We can opt to get "result" once rather than twice

```py
if "result" in response:
    result = response.get("result", False)

    if result in NULL_RESPONSES:
        apply_null_result_formatters(null_result_formatters, response, params)

    return result
```

web3/manager.py
```diff
@@ -75,6 +75,16 @@
 
 
 NULL_RESPONSES = [None, HexBytes("0x"), "0x"]
+METHOD_NOT_FOUND = -32601
+
+
+def _raise_bad_response_format(response: RPCResponse, error: str = "") -> None:
+    error_message = f"The error is: {error}. " if error else ""
+    raise BadResponseFormat(
```
fselmo: I wonder if the raw response might be useful in a `data` field eventually, rather than clumped in with the message here. I think that's out of the scope of this PR, just thinking out loud.

Either way the message here seems to be getting pretty chunky. I think we can remove "There error is: " and still have this be a self-explanatory and useful message.

```py
    error_message = f": {error}" if error else ""
    raise BadResponseFormat(
        "The response was in an unexpected format and unable to be parsed"
        f"{error_message}. The raw response is: {response}"
    )
```

That would leave the message a bit more succinct and just gets added to if the error is known.

`The response was in an unexpected format and unable to be parsed: error['message'] must be a string, integer or null.`

Consider this very much a nit, but it might be nice to slim it down a bit.

web3/manager.py
```diff
@@ -198,43 +208,66 @@ async def _coro_make_request(
...
     def formatted_response(
         response: RPCResponse,
         params: Any,
         error_formatters: Optional[Callable[..., Any]] = None,
         null_result_formatters: Optional[Callable[..., Any]] = None,
     ) -> Any:
+        if "jsonrpc" in response and response["jsonrpc"] != "2.0":
+            _raise_bad_response_format(response, '"jsonrpc" must equal "2.0"')
+
+        if "id" in response and not isinstance(response["id"], (str, int, None)):
```
fselmo: This raises a warning for me coming from the `type` for `id` in `RPCResponse`. We define it as an `int` only. I think the `id` can be a string so maybe we need to change the type there to `Union[int, str]` to match. 

I also don't think we should be validating a value of `None` here so loosely. Per the JSON-RPC 2.0 spec:

> id
>     - This member is REQUIRED.
>     - It MUST be the same as the value of the id member in the Request Object.
>     - **If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.**

I think we grab the `id` and if it's `None` we make sure that there was an `error` as well.

```py
        if "id" in response:
            response_id = response.get("id")
            if response_id is None and "error" not in response:
                _raise_bad_response_format(
                    response,
                    '"id" was null without error in response.',
                )
            elif not isinstance(response_id, (str, int)):
                _raise_bad_response_format(
                    response,
                    '"id" must be a string or integer.',
                )
```

This won't actually fail any of the current tests, so we should add a test that includes an `"id"` key in the response but does not include an `"error"` and validate the new exception is raised.

Unfortunately subscriptions in ethereum don't include an `id` in the response so this actually deviates from JSON-RPC 2.0. Which is why we keep the `if "id" in response` rather than making sure that it is always there. And if it's in the response it can only be null when there is an error.

Thoughts?

fselmo (commented):
@reedsa I had mostly small nits (sorry, I can't help it... feel free to ignore them 😆). But I did have a couple of comments for us to think about. Let me know your thoughts. I can do another pass at this after we talk these through.

web3/manager.py
fselmo: Actually, in a bit less of a nit, we either get `result` or `error` according to spec. So if we define both and include these checks in the if / elif, it's probably even better of a validation. We should also add a test where both are present.

```py

error = response.get("error")
result = response.get("result", False)

if "error" and "result" in response:
     _raise_bad_response_format(response, 'Both "error" and "result" found in response.')

elif "error" in response:
    ...
elif "result" in response:
    ...
# Response from eth_subscribe includes response["params"]["result"]
elif (
    response.get("params") is not None
    and response["params"].get("result") is not None
):
    ...
else:
    _raise_bad_response_format(response)
```

web3/manager.py
reedsa: Oh I think I had that as part of the json schema originally, but I dont recall why I had included it. Since the string is explicitly called out here, I'll update this accordingly.

web3/manager.py
reedsa: +1 for readability
In terms of including additional arguments for the response, it would be great to include as much context as possible. Opened a new issue (#3094) to consider revamping for v7.

fselmo: Love that. If it's just new features, not sure how much of that would need to be exclusive to `v7` too. Just adding the option to include more information on our exceptions would be great though. I think we've talked about this in the past but this issue should help track that 👌🏼

web3/manager.py
reedsa: Responses with an id may contain an error or a result so no need to enforce the check for an error within the "id" block.

I also saw a [note in the "id" section of the response object](https://www.jsonrpc.org/specification#response_object) of the spec that the value must be null when the request throws a parse error or invalid request. Unfortunately a bunch of the response types appear to lack an "id" entirely and seems like we are stuck with it deviating from the spec in this case.

The specific case you called out is caught in the final `else` of `format_response`, which is covered by a few tests like https://github.com/ethereum/web3.py/blob/main/tests/core/manager/test_response_formatters.py#L31

web3/manager.py
```diff
@@ -198,43 +211,82 @@ async def _coro_make_request(
...
+            _raise_bad_response_format(response, '"id" must be a string or integer')
+
+        # Response may not include both "error" and "result"
+        if "error" in response and "result" in response:
+            _raise_bad_response_format(
+                response, 'Response cannot include both "error" and "result"'
+            )
+
+        # Format and validate errors
         if "error" in response:
```
fselmo: This can be an `elif` as a part of the `if` / `elif` / `else` block, correct?

web3/manager.py
fselmo: I think I'm missing some context. Why can't we test that if an `id` key is present and the value is in null (`EMPTY_VALUES = ("", "null", None)`?) that there should be an `error` key present? That doesn't seem to be the case you linked to in your second link.

If we decide to keep this, the `isinstance()` check needs to check for `(str, int, type(None))`. But I believe the `id` type should be `Optional[Union[int, str]]` if we sometimes don't have an `id`. I'll make a comment there as well.

web3/types.py
```diff
@@ -283,7 +283,7 @@ class GethSyncingSubscriptionResponse(SubscriptionResponse):
 
 class RPCResponse(TypedDict, total=False):
     error: Union[RPCError, str]
-    id: int
+    id: Union[int, str]
```
fselmo: I believe this should be `Optional[Union[int, str]]`. Sorry if my previous comment was misleading. We sometimes get a null (`None`) `id` value (when there is an `error` in the response), making this param `Optional`.

fselmo (commented):
Just a few remaining thoughts.

web3/manager.py
reedsa: My brain wasnt groking the error rule for some reason. It sounds good to me to enforce that case.

I read it as "if there is an error, the value of id must be NULL", therefore if id is anything but "null" it should raise an exception? Curious if we want to keep empty string and None as empty values here, especially since empty string could mean something different.

The spec for the requests also indicates if id is missing, the request is considered to be a message. I'm guessing the same is for responses and the way the logic is implemented it will be parsed the same as other errors, so no change needed for that case.

fselmo: > I read it as "if there is an error, the value of id must be NULL", therefore if id is anything but "null" it should raise an exception?

Yeah I think we're on the same page. If `id` key is present and `error` is present, `id` value must be `null`. Put in another way, if `id` is `null` and there is no `error`, we raise since it can't be `null` any other time. My only question is what do they mean by `null`. They use `NULL` and `Null` in the text and `null` (not the string `"null"`) in the examples below. My understanding is that json parsing into a Python object would yield `None` in those cases. So that's probably the only reasonable `null` value? Thoughts there?

Since the `id` can indeed be `None` when there's an `error` also present, that makes the typing an `Optional[...]`.

If we're validating that, it could just look like the example block I posted above where we check if it is indeed `None` and there is no `error`, we throw... and then we only check for `str` or `int` cases below that since it can't be `None` beyond that first bit of logic. Does that make sense?

fselmo: Some new test cases arise from this.

1. There is an `"id"` key in the response with a value of `None` and no `"error"` key -> raises exception
2. There is an `"id"` key in the response with a value other than `None` and there is an `error` key with properly formatted `error` object -> raises exception
3. There is an `"id"` key in the response witha  value of `None` and there is an `error` key with properly formatted `error` object -> does NOT raise exception

web3/manager.py
```diff
@@ -198,43 +211,90 @@ async def _coro_make_request(
...
+            # id is always None for errors
+            if response_id is None and "error" not in response:
+                _raise_bad_response_format(
+                    response, '"id" must be None when an error is present'
+                )
+            elif not isinstance(response_id, (str, int)):
+                _raise_bad_response_format(response, '"id" must be a string or integer')
+
+        # Response may not include both "error" and "result"
+        if "error" in response and "result" in response:
```
reedsa, line 250: @fselmo You had left a comment about this `if` here, it's to check the error or result after validating the id. Otherwise it wouldn't drop into this block.

web3/manager.py
```diff
@@ -198,43 +211,90 @@ async def _coro_make_request(
...
+                response, 'The "jsonrpc" field must be present with a value of "2.0"'
+            )
+
+        # id is not enforced (as per the spec) but if present, it must be a
+        # string or integer
+        # TODO: v7 - enforce id per the spec
+        if "id" in response:
+            response_id = response["id"]
+            # id is always None for errors
+            if response_id is None and "error" not in response:
```
reedsa, line 242: @fselmo here's the check to prevent None as the value for `id` for anything but error responses.

reedsa (commented):
I've wrapped up the last round of feedback, thank you! I think it is ready for 🚀 but happy to fix any remaining issues you might find.

web3/manager.py
```diff
@@ -198,43 +211,91 @@ async def _coro_make_request(
...
+                response, 'The "jsonrpc" field must be present with a value of "2.0"'
+            )
+
+        # id is not enforced (as per the spec) but if present, it must be a
+        # string or integer
+        # TODO: v7 - enforce id per the spec
+        if "id" in response:
+            response_id = response["id"]
+            # id is always None for errors
+            if (response_id is None and "error" not in response) or (
```
reedsa: @fselmo I noticed we didnt handle when an error is present but `response_id` is not None. There are integration failures so curious if this is what we want?

fselmo: Oof... I don't know why there's a deviation from the spec but we should change it accordingly. I think we should definitely check for the case where the response `id` is `None` and there is no `error`. But the case you mention, when there is an `error` and the `id` is not `None`, seems like that's fine for me. Makes sense too to show which request gave the error so I'm not against keeping that 👍🏼

reedsa: Cool, I'll take this out!

fselmo (approved):
lgtm! Thanks for scrutinizing the specs with me 😅 ... this is a great addition

Issue #3056, opened 2023-07-21
Feature: wasm32 / emscripten support


joemarshall:
So, I did some work on running web3.py on webassembly / emscripten, so as to be able to run it in a web browser / jupyterlite notebook, without having to install any software on the user's machine.

This is potentially useful for teaching purposes, to be able to run demo code in the browser, but also worth doing, as webassembly is gaining quite a lot of traction in container / cloud platforms.

In terms of how this works, I see this work as being possible in multiple stages, each of which gains some functionality:

- [ ]  **Remove / make optional dependencies based on system.platform being wasm32-emscripten/ wasm32-wasi.** Currently, aiohttp and websockets specifically don't work on webassembly. It is possible to run web3.py code by hacking in dummy modules for these two. This gets us the basic ability to run synchronous web3 calls, and to add emscripten to the project CI, so it keeps working.
- [ ] **Add emscripten provider in web3.py for async_rpc, and websockets client connections.** This would allow most functionality of web3.py, except for local ipc support. Currently aiohttp and websockets libraries don't have support for emscripten; it is possible that they won't, because there is some functionality of the libraries that will never work (server side anything), and even if a polyfill is made, some functionality may not behave exactly the same. I think it is probably worth having specific emscripten providers for this here instead..
- [ ] Potentially add support for ipc / geth on wasm32-wasi. This would be the final piece in the jigsaw in making all functionality work in webassembly container platforms, but to be honest I don't know if it is worth it - it will probably never work in browser / emscripten based webassembly.

wolovim:
i'd love to support learners with a hosted version of web3.py!

couple things on my mind as i read:
1. would this be equally supported in mobile browsers? that would be compelling.
1. maintenance impact of each stage?
1. who's the audience? if just newcomers, a pretty limited version might be just fine. 
1. are there other advanced or interesting use cases that come to mind?

ameya-deshmukh:
@wolovim would love to assist you in this! Can I take it up?

wolovim:
@ameya-deshmukh thanks for offering! just checked in with @joemarshall who's down the rabbit hole already. we talked about an MVP version that he's getting started with. keep an eye out for updates or opportunities here.

joemarshall:
@wolovim An initial PR that lets you build and run web3 on pyodide is here #3088 

[...]

PR #3088, opened 2023-08-30
Web3 on pyodide

joemarshall:
This is the basic patch for it to be possible to build and run web3.py on webassembly using pyodide.

It works for synchronous http only right now. Calls to async or websockets functions will raise exceptions,

Right now it is at the point where:

```

w3 = Web3(HTTPProvider('https://goerli.infura.io/v3/<apikey>'))
w3.is_connected()
w3.eth.get_block("latest")
```
should work 

I built a pyodide distribution with it in here:

https://joemarshall.github.io/web3_py_on_web/dist/console.html

Still todo on this is bits that are outside web3.py itself:
1) Figure out how to package this easily (I had a load of hassle running micropip.freeze to make the pyodide lock file that I use so that it loads nicely on import)
2) Build in a jupyterlite based notebook environment so that one can store private keys etc.

#3056

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,8 @@ setup(
     url="https://github.com/ethereum/web3.py",
     include_package_data=True,
     install_requires=[
-        "aiohttp>=3.7.4.post0",
+        "aiohttp>=3.7.4.post0;sys_platform!='emscripten'",
+        "micropip;sys_platform=='emscripten'",
         "eth-abi>=4.0.0",
         "eth-account>=0.8.0",
         "eth-hash[pycryptodome]>=0.5.1",
@@ -81,7 +82,7 @@ setup(
         "pywin32>=223;platform_system=='Windows'",
         "requests>=2.16.0",
         "typing-extensions>=4.0.1",
-        "websockets>=10.0.0",
+        "websockets>=10.0.0;sys_platform!='emscripten'",
         "pyunormalize>=15.0.0",
     ],
     python_requires=">=3.7.2",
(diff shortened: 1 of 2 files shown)


joemarshall:
@wolovim 

Once this PR in pyodide is in [pyodide/](https://github.com/pyodide/pyodide-lock/pull/20) we'll have an easy way to build this for webassembly and then create a pyodide environment with all of web3 available, so it is probably worth merging the changes here in now if they look okay?

wolovim:
> Once this PR in pyodide is in [pyodide/](https://github.com/pyodide/pyodide-lock/pull/20) we'll have an easy way to build this for webassembly and then create a pyodide environment with all of web3 available, so it is probably worth merging the changes here in now if they look okay?

apologies for letting this PR go without comment @joemarshall. thank you for the initiative. talked to the team and we agreed that being able to host a learning environment in a browser would be great, though likely not something we can invest much time in before end of year. there was hesitation about including the code in the main branch, so current thinking is to try maintaining this in a separate branch.

i didn't follow your comment exactly. can you clarify? by all of web3, are you including aiohttp and websockets? planned adding and removing of stubs would further support this living on a different, experimental branch.

joemarshall:
> 
i didn't follow your comment exactly. can you clarify? by all of web3, are you including aiohttp and websockets? planned adding and removing of stubs would further support this living on a different, experimental branch.

I think sync http right now, but you're right - planning for experimental branch to remove stubs as things get ported in pyodide would make sense to me.

PR #3096, opened 2023-09-13
Websockets v2 update

fselmo:
### What was wrong?

The main reason `WebsocketProviderV2` is in beta is that we have to asynchronously process a response from the original request that was made. This isn't so straightforward with the way websockets work because the socket will give back the next item it needs to, whatever that may be, when we call `recv()`. If a user has subscribed to an `eth_subscribe` subscription, say something like `"newPendingTransactions"` where it will return many transactions nearly instantly, as soon as a user has subscribed to it, one (or many) of these subscription messages may return before the original request was made for which a user may be expecting a response. Consider the following:

```py

async def wsV2_example():
    async with AsyncWeb3.persistent_websocket(
        WebsocketProviderV2(f"wss://mainnet.infura.io/ws/v3/{INFURA_PROJECT_ID}")
    ) as w3:
        # -- get pending transactions subscription -- #
        
        pending_full_txs = True
        pending_tx_sub_id = await w3.eth.subscribe("newPendingTransactions", pending_full_txs)
        print(f"pending_tx_sub_id: {pending_tx_sub_id}\n")

        # -- logs subscription -- #

        logs_sub_id = await w3.eth.subscribe("logs", {"address": ETHEREUM_WETH_CONTRACT})
        # we will likely get one of the "newPendingTransactions" subscriptions returned here 
        # before we get the response to our second `eth_subscribe` request with the `logs_sub_id` value.
        print(f"logs_sub_id: {logs_sub_id}\n")

        while True:
                # when we actually listen to the websocket for our subscriptions, we will eventually get the response
                # we needed to the second subscription request with our `logs_sub_id` value
                print(response)

asyncio.run(wsV2_example())
```

### How was it fixed?

Thanks to JSON-RPC 2.0 specs, responses are required to return the `id` of the request that asked for the response. If we keep the request `id` and pass it to the method we use to retrieve and process a response, we can wait until this particular response is received and return that. It should take no longer than a `call_timeout`. While we look for the response, we need to store every response that came in before it in a transient cache that will then be emptied on each `__anext__()` call before we pull anything from the socket again.

- Created a `RequestProcessor` class that is strapped to `PersistentConnectionProvider` classes to process two types of caches now (both transient, in the sense that they only store information until it's needed and that time window shouldn't be too long): 
   - The `RequestInformation` cache that already existed. This cache is responsible for storing a request's information so we can have that context and middleware methods to pipe the answer to the request through once we process it asynchronously.
   - A cache that stores raw responses from the open websocket connection that may come in before the response for the original request comes in. This cache gets emptied in a FIFO manner before we listen to any new messages from the websocket so that we don't lose any response that were received within that window.
- Refactored matching responses to request ids so that it exists within the `make_request()` method of the `WebsocketProviderV2` class. This allows us to actually return a response that matches the request `id` within the same method, keeping types happy throughout the code base.
- Internalize the `provider.ws` websocket connection (as `provider._ws`) and add a public API class (`WebsocketConnection`) for interacting with the websocket on the `_PersistentConnectionWeb3` class with more predictable behavior and formatting support, etc.
- Other refactoring along the way.

### Todo:
- [x] Make `provider.ws` internal (`provider._ws`) and have any "listen" to ws method that is public look in the raw response cache first and process those responses in FIFO before actually calling `ws.recv()` again. This will be important to document since a user might want to directly call `recv()` on the websocket connection without knowing there may be cached responses ahead of any new responses.
- [x] Missing a cute animal picture below 😱 
- [x] Manual testing (for now) of the new request processor transient caches and the logic therein


web3/_utils/method_formatters.py
```diff
@@ -620,30 +624,45 @@ def subscription_formatter(value: Any) -> Union[HexBytes, HexStr, Dict[str, Any]
...
+    formatted_value = {"subscription": HexStr(value.get("subscription"))}
+    result = value.get("result")
+    result_key_set = set(result.keys())
 
     # handle dict subscription responses
-    if either_set_is_a_subset(response_key_set, set(BLOCK_FORMATTERS.keys())):
+    if either_set_is_a_subset(
+        result_key_set,
+        set(BLOCK_FORMATTERS.keys()),
+        percentage=90,
```
reedsa: I'm not quite understanding this percentage and how it changes what it was doing previously. Guessing the purpose is to apply specific formatters based on the value passed to `subscription_formatter`. When the value contains most (>90%) of the keys in the formatter type, the formatter is applied? Is it often the case that the value doesnt include all of the same formatters, or is this more to future proof things as more keys are added to the `value`?

web3/main.py
```diff
@@ -538,6 +536,7 @@ def __init__(
                 "Provider must inherit from PersistentConnectionProvider class."
             )
         AsyncWeb3.__init__(self, provider, middlewares, modules, external_modules, ens)
+        self.ws = WebsocketConnection(self)
 
     # async for w3 in w3.persistent_websocket(provider)
     async def __aiter__(self) -> AsyncIterator["_PersistentConnectionWeb3"]:
```
pacrob: Since we have `Self` available in compat, it could be applied here and anywhere else a class returns an instance of itself 

```suggestion
    from web3._utils.compat import Self

    async def __aiter__(self) -> AsyncIterator[Self]:
```

web3/_utils/method_formatters.py
fselmo: > When the value contains most (>90%) of the keys in the formatter type, the formatter is applied?

Correct.

One of the messages for a `pendingTransactions` subscription uses `yParity` in the place of `v` and we lacked a formatter for `yParity`. In that case it wasn't an exact subset so the formatters for subscription transactions stopped working on those messages entirely. If we are lenient on a percentage, say 90% here, we can distinguish between the different subscription responses and apply formatters appropriately and only those fields missing a formatter line won't be formatted. Then, if / when we catch the missing formatting, all we have to do is update the formatter method for that single field and that field too will be formatted appropriately.

This is the [relevant commit](https://github.com/ethereum/web3.py/pull/3096/commits/1107480944730e366df3ccf2300e446960942d0b).

reedsa:
Nothing blocking here, I haven't yet tested though. Looks great, love the separate commits and clean messages 💯

reedsa (approved):

pacrob (approved):
lgtm!

PR #3097, opened 2023-09-14
ignore flake8 F401 qa in all __init__.py files

pacrob:
### What was wrong?

Flake8 rule F401 requires that an import be used, but `__init__.py` files are specifically for importing various functions to then be imported elsewhere

### How was it fixed?

Tell flake8 to ignore F401 qa errors in all `__init__.py` files.

### Todo:

#### Cute Animal Picture

image


fselmo (approved):
🔥 🔥

PR #3103, opened 2023-09-20
Handle edge case for initialized chain

### LINKED ISSUE #1149: Web3 crashes in time-based gas price strategies with empty chain
* Version: 4.7.1
* Python: 3.6
* OS: linux

### What was wrong?

Running our test suite web3 crashes when estimating time based gas price. Probably it is caused by not having any transactions in our test chain. However when probabilities are calculated they are empty (empty tuple) which causes then an Index Error in ` '104, in _compute_gas_price\r\n    first = probabilities[0]\r\nIndexError: tuple index out of range\`

It might be the case that there is a problem with our geth client during runtime.

```
File “/home/circleci/raiden/raiden/network/rpc/client.py”, line 346, in gas_price
return int(self.web3.eth.generateGasPrice())
File “/home/circleci/venv-3.7/lib/python3.7/site-packages/web3/eth.py”, line 385, in generateGasPrice
return self.gasPriceStrategy(self.web3, transaction_params)
File “/home/circleci/venv-3.7/lib/python3.7/site-packages/web3/gas_strategies/time_based.py”, line 168, in time_based_gas_price_strategy
gas_price = _compute_gas_price(probabilities, probability / 100)
File “/home/circleci/venv-3.7/lib/python3.7/site-packages/web3/gas_strategies/time_based.py”, line 104, in _compute_gas_price
first = probabilities[0]
IndexError: tuple index out of range
```

### How can it be fixed?

However I would suggest handling the edge case that there is an empty tuple and not unconditionally access `first = probabilities[0] `

I'm happy to give more information when needed

reedsa:
### What was wrong?

Closes #1149

The time based gas price strategy assumes there are many blocks to sample from. In cases where test chains are spun up and no blocks have been mined, this can throw exceptions for folks who are running tests.

### How was it fixed?

Check the latest block prior to attempting to sample so it can fallback to the gas price in `w3.eth.gas_price`.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1149.bugfix.rst
+++ b/newsfragments/1149.bugfix.rst
@@ -1,0 +1,1 @@
+Return `w3.eth.gas_price` when calculating time based gas price strategy for an empty chain.
--- a/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
+++ b/tests/core/gas-strategies/test_time_based_gas_price_strategy.py
@@ -171,6 +171,46 @@ def test_time_based_gas_price_strategy(strategy_params, expected):
     assert actual == expected
 
 
+def _get_initial_block(method, params):
+    return {
+        "hash": constants.HASH_ZERO,
+        "number": 0,
+        "parentHash": None,
+        "transactions": [],
+        "miner": "0x" + "Aa" * 20,
+        "timestamp": 0,
+    }
+
+
+def _get_gas_price(method, params):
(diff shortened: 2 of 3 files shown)


pacrob (approved):
lgtm!

fselmo (commented):
lgtm 👍🏼

Issue #2572, opened 2022-07-14
AttributeDict objects not always hashable


rikublock:
* Version: 5.29.2
* Python: 3.8.10
* OS: Linux

### What was wrong?
Depending on the items, `AttributeDict` might not always be hashable. 

For example, log entries returned by `eth_getLogs()` can contain list objects (in this case 'topics'), which are not hashable.

```
AttributeDict({
	'address': '0x2C846CF666f2D427ec86ca4BADe204fFD5CbC421', 
	'topics': [HexBytes('0xdccd412f0b1252819cb1fd330b93224ca42612892bb3f4f789976e6d81936496'), HexBytes('0x000000000000000000000000e54ca86531e17ef3616d22ca28b0d458b6c89106'), HexBytes('0x000000000000000000000000d2f59be407080abfd97a00a7071d67f93b526733')], 
	'data': '0x000000000000000000000000000000000000000000000000000000000018921600000000000000000000000000000000000000000000000033bd1424be462f14', 
	'blockNumber': 16955438, 
	'transactionHash': HexBytes('0x932a26038bcf3d22bd981543b7ed6aae653a8a565a2a78bd644162c7e1bff6e4'), 
	'transactionIndex': 2, 
	'blockHash': HexBytes('0xd5b1c3a7704a26b1b7b47b12d1a929ecba82bb0cbf9b97f41d227ff104612f7b'), 
	'logIndex': 16, 
	'removed': False
 })
```

### How can it be fixed?
a) Adjust the `__hash__` function. Recursively look for mutable types like lists, etc. and temporary replace them before hashing.
This would likely also require changing `__eq__` as two object can now have the same hash without being "equal". 

[source](https://github.com/ethereum/web3.py/blob/master/web3/datastructures.py#L116-L117)
```python
def __hash__(self) -> int:
    return hash(tuple(sorted(self.items())))
```

b) Recursively replace mutable types (e.g. list -> tuple) when creating the AttributeDict object.

[source](https://github.com/ethereum/web3.py/blob/master/web3/datastructures.py#L46-L51)
```python
def __init__(
    self, dictionary: Dict[TKey, TValue], *args: Any, **kwargs: Any
) -> None:
    # type ignored on 46/50 b/c dict() expects str index type not TKey
    self.__dict__ = dict(dictionary)  # type: ignore
    self.__dict__.update(dict(*args, **kwargs))
```

c) Handle it earlier, for instance when parsing response results.

[...]

PR #3104, opened 2023-09-20
Use `hash` to compare equality in `AttributeDict`s

pacrob:
### What was wrong?

Two `AttributeDicts` could have the same hash, but not compare as equal. 
Also found a bug in `tupleize_lists_nested` where it wouldn't find lists if they were inside a tuple.

Related to Issue #2572 

### How was it fixed?

Updated `AttributeDict.__eq__` to compare hashes if one is being compared to another. It still uses the previous method of converting to dicts for comparison if the `other` is another type of `Mapping`.

In `tupleize_lists_nested`, added tuples to things that need to be recursively tupleized, in case they have unhashable objects within them.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/datastructures/test_datastructures.py
+++ b/tests/core/datastructures/test_datastructures.py
@@ -1,10 +1,88 @@
 import pytest
+import random
 import re
 
 from web3.datastructures import (
     AttributeDict,
     tupleize_lists_nested,
 )
+
+
+def generate_random_value(depth=0, max_depth=3, key_type=None):
+    if depth > max_depth:
+        return None
+
+    choice = random.choice(["str", "int", "bool", "list", "tuple", "dict"])
+
+    if choice == "str":
+        return "".join(random.choices("abcdefghijklmnopqrstuvwxyz", k=5))
+    elif choice == "int":
+        return random.randint(0, 100)
+    elif choice == "bool":
+        return random.choice([True, False])
+    elif choice == "list":
+        return [
+            generate_random_value(depth + 1, key_type=key_type)
+            for _ in range(random.randint(1, 5))
+        ]
+    elif choice == "tuple":
+        return tuple(
+            generate_random_value(depth + 1, key_type=key_type)
+            for _ in range(random.randint(1, 5))
+        )
+    elif choice == "dict":
+        return generate_random_dict(depth + 1, key_type=key_type)
+
+
+def generate_random_dict(depth=0, max_depth=3, max_keys=5, key_type=None):
+    if not key_type:
+        key_type = random.choice(["str", "int", "float"])
+
+    if depth > max_depth:
+        return {}
+
+    result = {}
+    for _ in range(random.randint(1, max_keys)):
+        if key_type == "str":
+            key = "".join(random.choices("abcdefghijklmnopqrstuvwxyz", k=5))
+        elif key_type == "int":
+            key = random.randint(0, 100)
+        elif key_type == "float":
+            key = round(random.uniform(0, 100), 2)
+
+        value = generate_random_value(depth, max_depth, key_type)
+        result[key] = value
+
+    return AttributeDict.recursive(result)
+
+
+def test_attribute_dict_raises_when_mutated():
+    test_attr_dict = AttributeDict(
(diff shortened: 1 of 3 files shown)


web3/datastructures.py
```diff
@@ -114,10 +114,16 @@ def __delattr__(self, key: str) -> None:
         raise TypeError("This data is immutable -- create a copy instead of modifying")
 
     def __hash__(self) -> int:
-        return hash(tuple(sorted(tupleize_lists_nested(self).items())))
+        return hash(
+            tuple(
+                sorted(tupleize_lists_nested(self).items(), key=lambda item: str(item))
```
pacrob: ~~I don't believe the addition of the `key` arg here is sufficiently breaking. Previously, it would throw a `TypeError` if the  keys were not comparable. Every example of an `AttributeDict` I've found uses only strings as keys, but the class is explicitly typed with `TKey`, not `String`.~~

nevermind, I've read up on what the use of `TKey` actually implies.

tests/core/datastructures/test_datastructures.py
```diff
@@ -7,6 +8,67 @@
...
+        result[key] = value
+
+    return AttributeDict.recursive(result)
+
+
+def test_attribute_dict_raises_when_mutated():
+    test_attr_dict = AttributeDict(
+        {"address": "0x4CB06C43fcdABeA22541fcF1F856A6a296448B6c"}
+    )
+    with pytest.raises(TypeError):
```
fselmo, line 63: I like adding this test here. Maybe add a test for the attribute setting as well? 

```py
with pytest.raises(TypeError):
    test_attr_dict.address = "0x6C8f2A135f6ed072DE4503Bd7C4999a1a17F824B"
```

web3/datastructures.py
```diff
@@ -130,12 +133,12 @@ def tupleize_lists_nested(d: Mapping[TKey, TValue]) -> AttributeDict[TKey, TValu
...
 
-    def _to_tuple(lst: List[Any]) -> Any:
-        return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
+    def _to_tuple(value: Union[List[Any], Tuple[Any, ...]]) -> Any:
+        return tuple(_to_tuple(i) if isinstance(i, (list, tuple)) else i for i in value)
 
     ret = dict()
     for k, v in d.items():
-        if isinstance(v, List):
+        if isinstance(v, (list, tuple)):
```
fselmo, line 141: Ah, yep... we have to handle the case where there are lists inside of tuples with a recursive check. Nice catch 👍🏼

fselmo (approved):
lgtm 👍🏼.  Just had a minor suggestion.

tests/core/datastructures/test_datastructures.py
```diff
@@ -50,9 +112,18 @@
...
 )
-def test_tupleization_and_hashing(input, expected):
+def test_tupleization_and_hashing_passing_defined(input, expected):
     assert tupleize_lists_nested(input) == expected
     assert hash(AttributeDict(input)) == hash(expected)
+    assert AttributeDict(input) == expected
+
+
+def test_tupleization_and_hashing_passing_random():
+    for _ in range(1000):
```
fselmo, line 138: great test 👌🏼

tests/core/datastructures/test_datastructures.py
pacrob: good idea, done

tests/core/datastructures/test_datastructures.py
reedsa: This is super cool!

web3/datastructures.py
```diff
@@ -117,7 +118,9 @@ def __hash__(self) -> int:
         return hash(tuple(sorted(tupleize_lists_nested(self).items())))
 
     def __eq__(self, other: Any) -> bool:
-        if isinstance(other, Mapping):
+        if isinstance(other, AttributeDict):
+            return hash(self) == hash(other)
```
reedsa, line 122: 🥳

reedsa (approved):
excellent use of random! lgtm

PR #3111, opened 2023-09-28
Speed up core tests

fselmo:
### What was wrong?

Core tests take ~12 min on average to run. We should try to speed this up.

- use `pytest-xdist`
- split sync / async into separate runs
- ???

### Todo:

#### Cute Animal Picture

image


pacrob (approved):
lgtm!

PR #3116, opened 2023-10-04
WebsocketProviderV2 updates + formatting middleware bugfix

### LINKED ISSUE #3093: WebsocketProviderV2 field 'extraData'
I've encountered an issue with the web3.py. My code works fine with Ethereum contracts, but I face a ValueError about the HexBytes value being 280 bytes (expected 32) when working with contracts on the BSC network.

Name: web3
Version: 6.9.0

Error traceback:
```
Exception has occurred: ValueError
Could not format invalid value '0xd88301020a846765746888676f312e32302e34856c696e7578000000b19df4a2f8b5830cefffb860a1a315d2b3cb4db96507221c47f66e3e811db45aad4905d0950895c104cbeffa87eb4c8d2bcd1275835a08951dbeded9126c2c7985c9b5e597bde03d48c1d10325eda14b239e219baa9953384ade5f96899162cf6c52aa9d6ce2865e11617d58f84c8401e17177a03afff5ba177d1fcd859bff6a940b0e43551b38d5c85cb63e605baeeebccb0e6c8401e17178a06f53ccf029bd4c1451bbd8b956f264d13a312290307206861574b4c025a384928077cd060dd0ff0e42bb0981534e5ba1486fec61c24509d6da3e73630e2854104551e27360ac5196e18fd3665773728d98531a4cb6b9e4d628d6dcab406bceda0600' as field 'extraData'
ValueError: The value HexBytes('0xd88301020a846765746888676f312e32302e34856c696e7578000000b19df4a2f8b5830cefffb860a1a315d2b3cb4db96507221c47f66e3e811db45aad4905d0950895c104cbeffa87eb4c8d2bcd1275835a08951dbeded9126c2c7985c9b5e597bde03d48c1d10325eda14b239e219baa9953384ade5f96899162cf6c52aa9d6ce2865e11617d58f84c8401e17177a03afff5ba177d1fcd859bff6a940b0e43551b38d5c85cb63e605baeeebccb0e6c8401e17178a06f53ccf029bd4c1451bbd8b956f264d13a312290307206861574b4c025a384928077cd060dd0ff0e42bb0981534e5ba1486fec61c24509d6da3e73630e2854104551e27360ac5196e18fd3665773728d98531a4cb6b9e4d628d6dcab406bceda0600') is 280 bytes, but should be 32

The above exception was the direct cause of the following exception:

  File "test.py", line 13, in new_blocks
  File "test.py", line 22, in <module>
    asyncio.run(new_blocks())
ValueError: Could not format invalid value '0xd88301020a846765746888676f312e32302e34856c696e7578000000b19df4a2f8b5830cefffb860a1a315d2b3cb4db96507221c47f66e3e811db45aad4905d0950895c104cbeffa87eb4c8d2bcd1275835a08951dbeded9126c2c7985c9b5e597bde03d48c1d10325eda14b239e219baa9953384ade5f96899162cf6c52aa9d6ce2865e11617d58f84c8401e17177a03afff5ba177d1fcd859bff6a940b0e43551b38d5c85cb63e605baeeebccb0e6c8401e17178a06f53ccf029bd4c1451bbd8b956f264d13a312290307206861574b4c025a384928077cd060dd0ff0e42bb0981534e5ba1486fec61c24509d6da3e73630e2854104551e27360ac5196e18fd3665773728d98531a4cb6b9e4d628d6dcab406bceda0600' as field 'extraData'

```

my code:
```
import asyncio
from web3 import AsyncWeb3

async def new_blocks(): #                                           
    async with AsyncWeb3.persistent_websocket(WebsocketProviderV2("wss://bsc.getblock.io/mainnet/")) as w3:
        subscription_id = await w3.eth.subscribe("newHeads") #                  ^^^ or any BSC RPC
        try:
            while True:
                    print(response)
         
        except KeyboardInterrupt:
            await w3.eth.unsubscribe(subscription_id)

if __name__ == "__main__":
    asyncio.run(new_blocks())

```

fselmo:
### What was wrong?

- Closes #3093
- `WebsocketProviderV2` needed tests for subscriptions. While we revamp the test suite and try to get some actual integration tests for `eth_subscribe`, introduce some static tests to make sure our formatters and middleware work appropriately for subscriptions.
- Adds a timeout around trying to match the response `id` to the request `id` for a request. Also improves on the logic around matching the ids, simplifying it.
- Add unit tests for the raw response cache process. 
  - Ensure we are caching any undesired responses before returning the desired response.
  - Ensure we return the desired cached response, if cached, and that we never call recv() in that case. 
  - Test the timeout if a response with matching `id` never comes (see above bullet point for test case).
- **(breaking change for beta provider)** Refactor to a proper AsyncIterator pattern for receiving a persistent message stream from the socket.
- Update docs based on the above

### Todo:

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -277,7 +277,7 @@ asynchronous context manager, can be found in the `websockets connection`_ docs.
         ...             if some_condition:
         ...                 # unsubscribe from new block headers and break out of
         ...                 # iterator
-        ...                 unsubscribed = await w3.eth.unsubscribe(subscription_id)
+        ...                 await w3.eth.unsubscribe(subscription_id)
         ...                 break
         ...
         ...         # still an open connection, make any other requests and get
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -270,16 +270,15 @@ asynchronous context manager, can be found in the `websockets connection`_ docs.
         ...         # subscribe to new block headers
         ...         subscription_id = await w3.eth.subscribe("newHeads")
         ...
-        ...         unsubscribed = False
-        ...         while not unsubscribed:
-        ...             async for response in w3.ws.listen_to_websocket():
-        ...                 print(f"{response}\n")
-        ...                 # handle responses here
+        ...         async for response in w3.ws.listen_to_websocket():
+        ...             print(f"{response}\n")
+        ...             # handle responses here
         ...
-        ...                 if some_condition:
-        ...                     # unsubscribe from new block headers
-        ...                     unsubscribed = await w3.eth.unsubscribe(subscription_id)
-        ...                     break
+        ...             if some_condition:
+        ...                 # unsubscribe from new block headers and break out of
+        ...                 # iterator
+        ...                 unsubscribed = await w3.eth.unsubscribe(subscription_id)
+        ...                 break
         ...
         ...         # still an open connection, make any other requests and get
         ...         # responses via send / receive
--- a/web3/_utils/module_testing/persistent_connection_provider.py
+++ b/web3/_utils/module_testing/persistent_connection_provider.py
@@ -292,7 +292,6 @@ class PersistentConnectionProviderTest:
         expected_formatted_result: Any,
     ) -> None:
         sub_id = await async_w3.eth.subscribe(*subscription_params)
-        assert sub_id is not None
         assert is_hexstr(sub_id)
 
         async def _mocked_recv_coro() -> bytes:
--- a/web3/_utils/module_testing/persistent_connection_provider.py
+++ b/web3/_utils/module_testing/persistent_connection_provider.py
@@ -1,0 +1,272 @@
+import pytest
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Tuple,
+    cast,
+)
+
+from eth_utils import (
+    is_hexstr,
+    to_bytes,
+)
+from hexbytes import (
+    HexBytes,
+)
+
+from web3._utils.encoding import (
+    FriendlyJsonSerde,
+)
+from web3.datastructures import (
+    AttributeDict,
+)
+from web3.types import (
+    FormattedEthSubscriptionResponse,
+)
+
+if TYPE_CHECKING:
+    from web3.main import (
+        _PersistentConnectionWeb3,
+    )
+
+
+class PersistentConnectionProviderTest:
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "subscription_params,ws_subscription_response,expected_formatted_result",
+        (
+            (
+                ("newHeads",),
+                {
+                    "jsonrpc": "2.0",
+                    "method": "eth_subscription",
+                    "params": {
+                        "subscription": "THIS_WILL_BE_REPLACED_IN_THE_TEST",
+                        "result": {
+                            "number": "0x539",
+                            "hash": "0xb46b85928f2c2264c2bf7ad5c6d6985664f1527e744193ef990cc0d3da5afc5e",  # noqa: E501
+                            "parentHash": "0xb46b85928f2c2264c2bf7ad5c6d6985664f1527e744193ef990cc0d3da5afc5e",  # noqa: E501
+                            "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",  # noqa: E501
+                            "logsBloom": "0x00",
+                            "transactionsRoot": "0x56260fe8298aff6d360e3a68fa855693f25dcb2708d8a7e509e8519b265d3988",  # noqa: E501
+                            "stateRoot": "0x56260fe8298aff6d360e3a68fa855693f25dcb2708d8a7e509e8519b265d3988",  # noqa: E501
+                            "receiptsRoot": "0x56260fe8298aff6d360e3a68fa855693f25dcb2708d8a7e509e8519b265d3988",  # noqa: E501
+                            "miner": "0x0000000000000000000000000000000000000000",
+                            "difficulty": "0x0",
+                            "extraData": "0x496c6c756d696e61746520446d6f63726174697a6520447374726962757465",  # noqa: E501
+                            "gasLimit": "0x1c9c380",
+                            "gasUsed": "0xd1ce44",
+                            "timestamp": "0x539",
+                            "baseFeePerGas": "0x26f93fef9",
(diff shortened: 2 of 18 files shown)


docs/providers.rst
```diff
@@ -270,16 +270,15 @@ asynchronous context manager, can be found in the `websockets connection`_ docs.
...
+        ...             # handle responses here
         ...
-        ...                 if some_condition:
-        ...                     # unsubscribe from new block headers
-        ...                     unsubscribed = await w3.eth.unsubscribe(subscription_id)
-        ...                     break
+        ...             if some_condition:
+        ...                 # unsubscribe from new block headers and break out of
+        ...                 # iterator
+        ...                 unsubscribed = await w3.eth.unsubscribe(subscription_id)
```
reedsa: Any reason to set the return value here to `unsubscribed`? Now that the wile loop is gone, perhaps dont need this variable.
```suggestion
        ...                await w3.eth.unsubscribe(subscription_id)
```

web3/_utils/module_testing/persistent_connection_provider.py
```diff
@@ -0,0 +1,325 @@
...
+    )
+    async def test_async_eth_subscribe_mocked(
+        self,
+        async_w3: "_PersistentConnectionWeb3",
+        subscription_params: Tuple[Any, ...],
+        ws_subscription_response: Dict[str, Any],
+        expected_formatted_result: Any,
+    ) -> None:
+        sub_id = await async_w3.eth.subscribe(*subscription_params)
+        assert sub_id is not None
```
reedsa: Nit: should we keep this assertion or add it to the test below? Looks like `is_hexstr` returns False for None anyway.

reedsa (commented):
Remaining nits, looking good otherwise!

docs/providers.rst
fselmo: Good eye 👌🏼

web3/providers/websocket/websocket_v2.py
```diff
@@ -158,40 +157,52 @@ async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
+            # "bad" cases somewhat gracefully.
+            return await asyncio.wait_for(
+                _match_response_id_to_request_id(), self.call_timeout
+            )
+        except asyncio.TimeoutError:
+            raise TimeExhausted(
+                f"Timed out waiting for response to request id `{request_id}` after "
+                f"{self.call_timeout} seconds. This is likely due to the provider not "
+                "issuing a response with the same id that was sent in the request, "
+                "which is required by the JSON-RPC spec."
```
reedsa:
```suggestion
                f"Timed out waiting for response with request id `{request_id}` after "
                f"{self.call_timeout} seconds. This is likely due to the provider not "
                "returning a response with the same id that was sent in the request, "
                "which is required by the JSON-RPC spec."
```

web3/providers/websocket/websocket_v2.py
```diff
@@ -67,7 +66,7 @@ def __init__(
         self,
         endpoint_uri: Optional[Union[URI, str]] = None,
         websocket_kwargs: Optional[Dict[str, Any]] = None,
-        call_timeout: Optional[int] = None,
+        call_timeout: Optional[float] = DEFAULT_PERSISTENT_CONNECTION_TIMEOUT,
```
reedsa, line 69: The timeout may be `None` or could be the max float value, `wait_for` will in turn wait forever (or until the timeout of the large value provided). Since this would be intentionally set by the user, I think it is acceptable to let the connection stay open until the response id is returned, if ever. What do you think?

fselmo: Hmm. That's a good point. We use this not only for call timeouts but for the while loop timeout. What if we set the while loop timeout to `self.call_timeout or 20` so that if it is set to `None` for call timeouts, the while loop will still max out in `20` seconds. Or whatever we decide that default is. Since it differs from the timeout for each call, I think that might be a good alternative. Thoughts?

reedsa: Makes sense to have the timeout limit on the while loop. Paraphrasing what you said to make sure I have it right in my head, use the `call_timeout` unless it exceeds the threshold (20s or something). If `call_timeout` is larger than the threshold, the while loop will still  short circuit as desired.

fselmo: My suggestion would've only worked for `None` but I think yours works well.

```py
timeout = self.call_timeout if self.call_timeout and self.call_timeout <= 20 else 20
```

This also prevents it from being `0` in the `if` check. Anything from `0.1` to `20` is fair game.

pacrob (approved):
lgtm!

web3/providers/websocket/websocket_v2.py
```diff
@@ -158,40 +157,55 @@ async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
+        try:
+            # Enters a while loop, looking for a response id match to the request id.
+            # If the provider does not give responses with matching ids, this will
+            # hang forever. The JSON-RPC spec requires that providers respond with
+            # the same id that was sent in the request, but we need to handle these
+            # "bad" cases somewhat gracefully.
+            timeout = (
+                self.call_timeout
+                if self.call_timeout and self.call_timeout <= 20
+                else 20
```
reedsa, line 199: Should use the `DEFAULT_PERSISTENT_CONNECTION_TIMEOUT` constant here.
```suggestion
                if self.call_timeout and self.call_timeout <= DEFAULT_PERSISTENT_CONNECTION_TIMEOUT
                else DEFAULT_PERSISTENT_CONNECTION_TIMEOUT
```

fselmo: Isn't this an entirely different bit of logic though? What if we agree to change the default timeout? I think that was the point of hard-coding this here.

Let's say I don't want _each_ call to the provider to take more than, say, 5 seconds. I wonder if I'm subscribed to a lot of subscriptions, if the response to my call won't be bogged down by the latency behind so many subscriptions where it may actually take longer than a call timeout to receive it.

Thoughts there?

reedsa: That's fine then, I have no concerns with using the hardcoded value as is. Definitely would be overkill to create yet another constant just for this logic.

reedsa (approved):
🚀 :shipit:

web3/providers/websocket/websocket_v2.py
fselmo: Yeah in the same way I think using `self.call_timeout` feels a bit disingenuous to use here too but at least the user can tweak that to make sense if desired. Definitely open to re-thinking this a bit more before ripping off the `beta` tag.

PR #3120, opened 2023-10-09
Add custom list and backoff factor to http_retry_middleware

### LINKED ISSUE #958: HTTPRequestRetry middleware should accept a parameter for allowing retry on certain methods
* Version: 4.3.0
* Python: 3.6.5
* OS: osx

This default middleware is not part of the middleware_stack to remove or add during runtime easily. Please provide a mechanism to perform this during runtime and also some mechanism to exclude certain methods from retrying.

### LINKED ISSUE #1911: http_retry_request_middleware needs to be more graceful and sleep a bit
* Version: 5
* Python: 3.x

### What was wrong?

The current `http_retry_request_middleware` is not graceful and configurable enough.

### How can it be fixed?

The current `exception_retry_middleware`, which is used by `http_retry_request_middleware`, looks like:

```python
def exception_retry_middleware(
    make_request: Callable[[RPCEndpoint, Any], RPCResponse],
    web3: "Web3",
    errors: Collection[Type[BaseException]],
    retries: int = 5,
    """
    Creates middleware that retries failed HTTP requests. Is a default
    middleware for HTTPProvider.
    """
        if check_if_retry_on_failure(method):
            for i in range(retries):
                try:
                # https://github.com/python/mypy/issues/5349
                except errors:  # type: ignore
                    if i < retries - 1:
                        continue
                    else:
                        raise
            return None
        else:
    return middleware
```

Possible usual retryable error conditions include 

* Temporary TCP/IP routing errors

* Network card reset

* Firewall reset

* JSON-RPC server reset

Under normal conditions, any of these error conditions may last from fractions of seconds up to a minute. However the current `exception_retry_middleware` has too tight loop without a sleep. It simply bombs the network interface as fast as it can, without backing off.

I recommend making `exception_retry_middleware`, and `http_retry_request_middleware`, configurable so that the library user can pass parameters for timeouts and retry counts.

fselmo:
### What was wrong?

closes #958 
closes #1911 

### How was it fixed?

- Add kwarg for importing custom list of RPC methods to retry on
- Add kwarg, as in the async version, for a backoff factor to prevent timeout issues and provide flexibility

### Todo:

### DIFF
--- a/web3/middleware/exception_retry_request.py
+++ b/web3/middleware/exception_retry_request.py
@@ -1,9 +1,11 @@
 import asyncio
+import time
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Collection,
+    List,
     Optional,
     Type,
 )
@@ -28,7 +30,7 @@ if TYPE_CHECKING:
         Web3,
     )
 
-whitelist = [
+DEFAULT_ALLOWLIST = [
     "admin",
     "miner",
     "net",
@@ -87,11 +89,16 @@ whitelist = [
 ]
 
 
-def check_if_retry_on_failure(method: RPCEndpoint) -> bool:
+def check_if_retry_on_failure(
+    method: str, allow_list: Optional[List[str]] = None
+) -> bool:
+    if allow_list is None:
+        allow_list = DEFAULT_ALLOWLIST
+
     root = method.split("_")[0]
-    if root in whitelist:
+    if root in allow_list:
         return True
-    elif method in whitelist:
+    elif method in allow_list:
         return True
     else:
         return False
@@ -102,6 +109,8 @@ def exception_retry_middleware(
     _w3: "Web3",
     errors: Collection[Type[BaseException]],
     retries: int = 5,
+    backoff_factor: float = 0.3,
+    allow_list: Optional[List[str]] = None,
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
     """
     Creates middleware that retries failed HTTP requests. Is a default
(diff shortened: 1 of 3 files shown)


web3/middleware/exception_retry_request.py
```diff
@@ -87,11 +89,16 @@
...
-def check_if_retry_on_failure(method: RPCEndpoint) -> bool:
+def check_if_retry_on_failure(
+    method: str, allow_list: Optional[List[str]] = None
+) -> bool:
+    if allow_list is None:
+        allow_list = DEFAULT_ALLOWLIST
+
     root = method.split("_")[0]
-    if root in whitelist:
+    if root in allow_list:
```
reedsa, line 99: 💯

web3/middleware/exception_retry_request.py
```diff
@@ -87,11 +89,16 @@
 ]
 
 
-def check_if_retry_on_failure(method: RPCEndpoint) -> bool:
+def check_if_retry_on_failure(
+    method: str, allow_list: Optional[List[str]] = None
+) -> bool:
+    if allow_list is None:
```
reedsa, line 95: What do you think about using the default if `allow_list` has zero items as well as `None`?

reedsa (approved):
lgtm, curious to hear about further validating the allow_list param.

web3/middleware/exception_retry_request.py
fselmo: I think I could go either way on that but it seems a bit disingenuous. I don't think it's ideal to submit an empty list for a method like this but also in python I'd always expect if I pass in a kwarg, that is what's used / overrides the default kwarg value. What are your thoughts there? 👀

reedsa: Logically I see your point, if looking for an item in an empty list, I would expect it to return false. I'm good leaving it as is!

PR #3122, opened 2023-10-11
bump and pin mypy to latest version that supports py37

pacrob:
### What was wrong?

`mypy` was mistakenly given only a lower-bound, should have been pinned, and its new 1.6.0 release introduced a change that breaks CI.

### How was it fixed?
Pinned to latest version that still supports py37, 1.4.1

### Todo:

#### Cute Animal Picture

image


fselmo (approved):

reedsa (approved):
Lgtm

PR #3125, opened 2023-10-12
WebsocketProviderV2 `recv()` lock + other multi-tasking support

fselmo:
### What was wrong?

Some use cases for `WebsocketProviderV2` don't allow `asyncio.gather(tasks...)`, for example, as it will halt when two coroutines try to call `recv()` on the websocket connection at once. This PR attempts to ameliorate these scenarios.

- Add an `asyncio.Lock` around the websocket `recv()` method.
- Change `listen_to_websocket()` so that it only listens to subscription (or one-to-many request / response) messages. This is because every `make_request()` expects a 1-to-1 response-to-request and expects it to be returned on the same line e.g. `variable_assignment = await w3.eth.block_number`. If we are persistently listening to the socket, we clearly expect many messages. So, we only listen to messages from one-to-many calls, triggered by a `ws.send()`.
- Add an API for retrieving active subscriptions: `w3.ws.subscriptions` (WebsocketConnection.subscriptions)
- Because of the separation of 1-to-1 request / response and one-to-many send() to recv() messages, separate the `RequestProcessor` caches into two types of "caches". The subscription cache is really a deque that always uses `popleft()`, so make it one. The response cache is looking for a particular cache key so this still makes sense to keep as an `OrderedDict`.

- Defines an ``__await__()`` on the ``_PersistentConnectionWeb3`` class so the ``AsyncWeb3.persistent_websocket()`` instantiation can be awaited 
(i.e. ``w3 = await AsyncWeb3.persistent_websocket(WebsocketProviderV2(...))``). New integration test suite added with this pattern of instantiation.

### Todo:
- [x] Add tests for new additions

#### Cute Animal Picture

snake socks

```
    Y
  .-^-.
 /     \      .- ~ ~ -.
()     ()    /   _ _   `.                     _ _ _
 \_   _/    /  /     \   \                . ~  _ _  ~ .
   | |     /  /       \   \             .' .~       ~-. `.
   | |    /  /         )   )           /  /             `.`.
   \ \_ _/  /         /   /           /  /                `'
    \_ _ _.'         /   /           (  (               |=====|
                    /   /             \  \`             |=====|
                   /   /               \  \             |     |
                  /   /                 )  )            |      \___
                 (   (                 /  /             (        (_)
                  `.  `.             .'  /               ``````````
                    `.   ~ - - - - ~   .'
                       ~ . _ _ _ _ . ~

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -201,13 +201,13 @@ message and receive a response to that particular request,
 response *id* values coming back from the websocket connection. Any provider that does
 not adhere to the `JSON-RPC 2.0 specification <https://www.jsonrpc.org/specification>`_
 in this way will not work with ``PersistentConnectionProvider`` instances. The specifics
-of how the request processor handles this is outlined below.
+of how the request processor handles this are outlined below.
 
 One-To-One Requests
 ~~~~~~~~~~~~~~~~~~~
 
-One-to-one requests can be summarized as any request that expects one response back.
-An example is using the ``eth`` module API to request the latest block number.
+One-to-one requests can be summarized as any request that expects only one response
+back. An example is using the ``eth`` module API to request the latest block number.
 
 .. code-block:: python
 
@@ -220,13 +220,13 @@ An example is using the ``eth`` module API to request the latest block number.
 
     >>> asyncio.run(wsV2_one_to_one_example())
 
-With websockets we have to call ``ws_send()`` and asynchronously receive responses via
-``ws.recv()``. In order to make the one-to-one request-to-response call work, we
-have to save the request information somewhere so that, when the response is received,
-we can match it to the original request that was made (the request with a matching *id*
-to the response that was received), and use that request information to process the
-response. Processing the response, in this case, means running it through the
-formatters and middlewares internal to the *web3.py* library.
+With websockets we have to call ``send()`` and asynchronously receive responses via
+``recv()``. In order to make the one-to-one request-to-response call work, we
+have to save the request information somewhere so that when the response is received
+we can match it to the original request that was made i.e. the request with a matching
+*id* to the response that was received). The stored request information is then used to
+process the response when it is received, piping it through the response formatters and
+middlewares internal to the *web3.py* library.
 
 In order to store the request information, the ``RequestProcessor`` class has an
 internal ``RequestInformation`` cache. The ``RequestInformation`` class saves important
@@ -279,7 +279,7 @@ successful. For this reason, the original request is considered a one-to-one req
 so that a subscription *id* can be returned to the user on the same line, but the
 ``listen_to_websocket()`` method on the
 :class:`~web3.providers.websocket.WebsocketConnection` class, the public API for
-interacting with the active websocket connection, is set up to receive many-to-one
+interacting with the active websocket connection, is set up to receive
 ``eth_subscription`` responses over an asynchronous interator pattern.
 
 .. code-block:: python
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -185,3 +185,138 @@ Managers
 The Manager acts as a gatekeeper for the request/response lifecycle.  It is
 unlikely that you will need to change the Manager as most functionality can be
 implemented in the Middleware layer.
+
+
+Request Processing for Persistent Connection Providers
+------------------------------------------------------
+
+.. py:class:: web3.providers.websocket.request_processor.RequestProcessor
+
+The ``RequestProcessor`` class is responsible for the storing and syncing up of
+asynchronous requests to responses for a ``PersistentConnectionProvider``. The best
+example of one such provider is the
+:class:`~web3.providers.websocket.WebsocketProviderV2`. In order to send a websocket
+message and receive a response to that particular request,
+``PersistentConnectionProvider`` instances have to match request *id* values to
+response *id* values coming back from the websocket connection. Any provider that does
+not adhere to the `JSON-RPC 2.0 specification <https://www.jsonrpc.org/specification>`_
+in this way will not work with ``PersistentConnectionProvider`` instances. The specifics
+of how the request processor handles this is outlined below.
+
+One-To-One Requests
+~~~~~~~~~~~~~~~~~~~
+
+One-to-one requests can be summarized as any request that expects one response back.
+An example is using the ``eth`` module API to request the latest block number.
+
+.. code-block:: python
+
+    >>> async def wsV2_one_to_one_example():
+    ...     async with AsyncWeb3.persistent_websocket(
+    ...         WebsocketProviderV2(f"ws://127.0.0.1:8546")
+    ...     ) as w3:
+    ...         # make a request and expect a single response returned on the same line
+    ...         latest_block_num = await w3.eth.block_number
+
+    >>> asyncio.run(wsV2_one_to_one_example())
+
+With websockets we have to call ``ws_send()`` and asynchronously receive responses via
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: docs/internals.rst [hunk 1]
```diff
@@ -185,3 +185,138 @@ Managers
...
+    ...         latest_block_num = await w3.eth.block_number
+
+    >>> asyncio.run(wsV2_one_to_one_example())
+
+With websockets we have to call ``ws_send()`` and asynchronously receive responses via
+``ws.recv()``. In order to make the one-to-one request-to-response call work, we
+have to save the request information somewhere so that, when the response is received,
+we can match it to the original request that was made (the request with a matching *id*
+to the response that was received), and use that request information to process the
+response. Processing the response, in this case, means running it through the
```
reedsa: Long sentence, tried to clean it up a little but no hard feelings about it.
```suggestion
``ws.recv()``. A one-to-one request-to-response call works by matching the request 
and response to ensure the result is processed appropriately. When a request is made,
a request ID is stored so that it can be used when a response is returned to determine 
if it came from that same request. When the original request is found it processes the 
result by running it through the
```

web3/main.py
```diff
@@ -524,6 +527,8 @@ def persistent_websocket(
 class _PersistentConnectionWeb3(AsyncWeb3):
     provider: PersistentConnectionProvider
 
+    # w3 = AsyncWeb3.persistent_websocket(provider)
```
reedsa, line 530: The comments that follow kind of look like code that was commented out. Recommend creating comment blocks that call these examples out more explicitly.

web3/manager.py
```diff
@@ -364,22 +368,32 @@ async def _ws_recv_stream(self) -> AsyncGenerator[RPCResponse, None]:
...
-            for (
-                cache_key,
-                cached_response,
-            ) in self._request_processor._raw_response_cache.items():
-                await self._request_processor.pop_raw_response(cache_key)
-                yield await self._process_ws_response(cached_response)
-        else:
-            response = await self._provider._ws_recv()
-            yield await self._process_ws_response(response)
+        while True:
```
reedsa, line 371: Could be worth adding a comment for this block. At least one that references the docs about this implementation.

reedsa (commented):
Nothing major, looks good overall!

docs/internals.rst
```diff
@@ -185,3 +185,138 @@ Managers
...
+The ``RequestProcessor`` class is responsible for the storing and syncing up of
+asynchronous requests to responses for a ``PersistentConnectionProvider``. The best
+example of one such provider is the
+:class:`~web3.providers.websocket.WebsocketProviderV2`. In order to send a websocket
+message and receive a response to that particular request,
+``PersistentConnectionProvider`` instances have to match request *id* values to
+response *id* values coming back from the websocket connection. Any provider that does
+not adhere to the `JSON-RPC 2.0 specification <https://www.jsonrpc.org/specification>`_
+in this way will not work with ``PersistentConnectionProvider`` instances. The specifics
+of how the request processor handles this is outlined below.
```
pacrob: "are outlined below", so verb matches noun (specifics).

docs/internals.rst
```diff
@@ -185,3 +185,138 @@ Managers
...
+
+One-To-One Requests
+~~~~~~~~~~~~~~~~~~~
+
+One-to-one requests can be summarized as any request that expects one response back.
```
pacrob: nit, take or leave: "expects a single response", or "expects one and only one response" to clarify that it won't be more than one.

docs/internals.rst
```diff
@@ -185,3 +185,138 @@ Managers
...
+    >>> async def wsV2_one_to_one_example():
+    ...     async with AsyncWeb3.persistent_websocket(
+    ...         WebsocketProviderV2(f"ws://127.0.0.1:8546")
+    ...     ) as w3:
+    ...         # make a request and expect a single response returned on the same line
+    ...         latest_block_num = await w3.eth.block_number
+
+    >>> asyncio.run(wsV2_one_to_one_example())
+
+With websockets we have to call ``ws_send()`` and asynchronously receive responses via
```
pacrob: It looks like these should match - `ws.send` and `ws.recv` - but I don't know for certain. Is it correct as is?

docs/internals.rst
```diff
@@ -185,3 +185,138 @@ Managers
...
+
+One-to-many requests can be summarized by any request that expects many responses as a
+result of the initial request. An example is the ``eth_subscribe`` request. The initial
+``eth_subscribe`` request expects only one response, the subscription *id* value, but
+it also expects to receive many ``eth_subscription`` messages if and when the request is
+successful. For this reason, the original request is considered a one-to-one request
+so that a subscription *id* can be returned to the user on the same line, but the
+``listen_to_websocket()`` method on the
+:class:`~web3.providers.websocket.WebsocketConnection` class, the public API for
+interacting with the active websocket connection, is set up to receive many-to-one
```
pacrob: "set up to receive many eth_subscription responses"  - the "-to-one" here feels unnecessary, no?

docs/providers.rst
```diff
@@ -327,43 +367,61 @@ When an ``AsyncWeb3`` class is connected to a persistent websocket connection, v
...
 
-        This class handles interactions with a websocket connection. It is available
-        via the ``ws`` attribute of the ``_PersistentConnectionWeb3`` class. The
-        ``WebsocketConnection`` class has the following methods:
+    .. py:method:: listen_to_websocket()
 
-        .. py:method:: listen_to_websocket()
+        This method is available for listening to websocket responses indefinitely.
+        It is an asynchronous iterator that yields strictly one-to-many
+        (e.g. eth_subscription responses) request-to-response responses from the
```
pacrob: `eth_subscription` should be code, like `eth_subscribe` below.

docs/providers.rst
```diff
@@ -327,43 +367,61 @@ When an ``AsyncWeb3`` class is connected to a persistent websocket connection, v
...
+        websocket connection. To receive responses for 1-to-1 request-to-response
```
pacrob: This is "one-to-one" everywhere else, recommend change just for consistency.

web3/main.py
pacrob, line 530: I like these examples of usage. Would a label, something like `# Example usage:` or `# Access this method with pattern:` make it more clear for users?

web3/providers/websocket/websocket_v2.py
```diff
@@ -138,76 +138,99 @@ async def connect(self) -> None:
...
+                            # cache all responses that are not the desired response
+                            self.logger.debug("Undesired response received, caching.")
+                            is_subscription = (
+                                response.get("method") == "eth_subscription"
+                            )
+                            await self._request_processor.cache_raw_response(
+                                response, subscription=is_subscription
+                            )
+
+                    # this is important to let asyncio run other tasksx
```
pacrob: `tasksx` needs less x, please

pacrob (approved):
Nothing came up in messing with it. Once existing notes are handled, lgtm!

web3/main.py
fselmo: I'd like to leave that up to the documentation. This is more so we separate the logic for each of those design patterns into code blocks and so contributors know where to look for it.

fselmo: I really only meant these to be for organization / reference. The websockets library does a similar thing with their `connect()` method [here](https://github.com/python-websockets/websockets/blob/be551424ecf17b92ce5cc24f5eb73f2796eb1daa/src/websockets/legacy/client.py#L587-L676). I feel like we could let the documentation do the explaining. Thoughts?

PR #3127, opened 2023-10-16
add upper pin to hexbytes dep due to incoming breaking update in 0.4.0

pacrob:
### What was wrong?

The hexbytes lib will be changing the way `Hexbytes.hex()` works - https://github.com/ethereum/hexbytes/issues/14

### How was it fixed?

Add upper pin to hexbytes dep now, then web3 v7 will use hexbytes 0.4.0 as a lower bound.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setup(
         "eth-hash[pycryptodome]>=0.5.1",
         "eth-typing>=3.0.0",
         "eth-utils>=2.1.0",
-        "hexbytes>=0.1.0",
+        "hexbytes>=0.1.0,<0.4.0",
         "jsonschema>=4.0.0",
         "lru-dict>=1.1.6",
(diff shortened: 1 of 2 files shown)


reedsa (approved):

setup.py
```diff
@@ -74,7 +74,7 @@
         "eth-hash[pycryptodome]>=0.5.1",
         "eth-typing>=3.0.0",
         "eth-utils>=2.1.0",
-        "hexbytes>=0.1.0",
+        "hexbytes>=0.1.0,<=0.3.1",
```
fselmo: See comment [here](https://github.com/ethereum/eth-account/pull/240). I think we should use `<0.4.0` as the pin if that is the planned version with the breaking changes.

fselmo (approved):

PR #3129, opened 2023-10-18
Add formatter for ``data`` field for transaction results

fselmo:
### What was wrong?

- Nethermind returns both ``data`` and ``input`` with the same value for each. Add a formatter for ``data`` that will format the field if it is present.
- Add a test for transaction result formatters via ``eth.get_transaction()`` test.

### Todo:

#### Cute Animal Picture

```
   00
 8====--
   00
```


reedsa (approved):
Lgtm

PR #3135, opened 2023-10-25
[WS_V2] Improve upon yielding control to the event loop

fselmo:
### What was wrong?

- In order to guarantee that the event loop can run other tasks more efficiently, asyncio.sleep(0) seems to be the most efficient way to yield control back to the event loop in a way that many tasks can still run concurrently without quickly timing out.
- Increase the default timeout to look for a response to a request from 20 seconds to 50 seconds.
- Make the caching methods in the request processor synchronous since they don't need to be async.

### Todo:


reedsa (approved):
Nice work, lgtm!

PR #3137, opened 2023-11-01
Improve docs for abi required-ness

### LINKED ISSUE #2539: web3.eth.contract with no abi
* Version: 5.29.1
* Python: 3.10.4
* OS: linux
* `pip freeze` output

```
eth-abi==2.1.1
eth-account==0.5.7
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.3.4
eth-rlp==0.2.1
eth-typing==2.3.0
eth-utils==1.10.0
web3==5.29.1

```

### What was wrong?

testContract = w3.eth.contract(address=add)
testContract.get_function_by_selector("0x370158ea")
Gives an error: 
```
  File "/home/dean/.local/lib/python3.10/site-packages/web3/_utils/abi.py", line 89, in filter_by_type
    return [abi for abi in contract_abi if abi['type'] == _type]
TypeError: 'NoneType' object is not iterable
```
running the same commands but with abi included gives the correct value:
testContract = web3.eth.contract(address=add,abi=abi)
testContract.get_function_by_selector("0x370158ea")
<Function info()>

The documentation [documentation ](https://web3py.readthedocs.io/en/stable/web3.eth.html#contracts)
Does not show abi as a required parameter when instantiating the Contract class. But all the examples show it being used. 
I can instantiate a contract by passing it an address without an abi parameter. But later calls fail. Is this expected behavior? 
I should still be able to interact with the bytecode at a specific address as long as I can either generate a function signature or call the selector directly. 

**Is it expected behavior to require abi for all contract interaction, even those that don't require it?** 

I used a remote infura node. I noticed this issue when working with brownie and verified that it was happening in using web3 directly. 

### How can it be fixed?

Updating the documentation if this is expected behavior. If this is not expected behavior then adding a check on for set abi on the contract function calls. 

---

reedsa:
### What was wrong?

Lacking clarity on whether abi must be provided.

Closes #2539

### How was it fixed?

Explicit docs for ABI being required for contract interactions to work.

### Todo:

#### Cute Animal Picture


docs/web3.contract.rst
```diff
@@ -35,7 +35,13 @@ After ``py-solc-x`` is installed, you will need to install a version of ``solc``
     >>> from solcx import install_solc
     >>> install_solc(version='latest')
 
-You should now be set up to run the contract deployment example below:
+You should now be set up to compile and deploy a contract.
+
+The following example runs through these steps:
+1. Compile Solidity contract into bytecode and an ABI
```
pacrob: Formatting in the docs isn't coming out as intended: https://web3py--3137.org.readthedocs.build/en/3137/web3.contract.html#contract-deployment-example

Looks like you want `#.` instead of the actual numbers: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html

docs/web3.eth.rst
```diff
@@ -1238,6 +1238,11 @@ Contracts
 
 .. py:method:: Eth.contract(address=None, contract_name=None, ContractFactoryClass=Contract, **contract_factory_kwargs)
 
+    An ``abi`` specifies how a contract can be interacted with. The ``abi``
```
pacrob: I feel like this clarification should go down below, as in the "following arguments are accepted" section. This first section discusses the actual named arguments associated with the function, in order. `abi`, while necessary for making contract calls, isn't a named arg, just an optional kwarg (whether this is a good design is perhaps another question :thinking: ) Maybe adding it as a description to the optional kwargs list is a better place? That then opens it as a place for describing usage of the others in that list. 

Just my feels, open for discussion!

reedsa: Would be ideal to call something out here in the contract docs. I wonder if it would help to include abi as an argument in the method signature above, or some way of indicating it is necessary for the interface to work.

Agree this paragraph doesnt seem quite right in the contract API docs. I think it might clutter the list down below, but there is a link to the Contracts guide from there. I think the paragraph might be a good replacement for the contract docs under `Contract.abi`. What do you think?

docs/web3.contract.rst
reedsa: Oops, thanks. Fixed!

docs/web3.eth.rst
pacrob: Agree with putting more info under `Contract.abi`. Could even explain what ABI stands for there. I also agree with some indication here that abi is required if you want to interact with the contract, maybe just not first thing?

docs/web3.eth.rst
```diff
@@ -1282,22 +1282,28 @@ Contracts
...
-    - ``decode_tuples``
-    - ``interface``
-    - ``metadata``
-    - ``opcodes``
-    - ``src_map``
-    - ``src_map_runtime``
-    - ``user_doc``
-
-    See :doc:`web3.contract` for more information about how to use contracts.
+    :param abi: Application Binary Interface. Usually provided since an ``abi`` is required to interact with any contract.
```
reedsa, line 1285: @fselmo @pacrob I like the sphinx `param` block! I'm not sure about all of these params, but I took my best guess at as many as I could. Are there any words we can add or change here?

Thanks!

docs/web3.eth.rst
```diff
@@ -1282,22 +1282,28 @@ Contracts
...
+    :param asm: Asssembly code generated by the compiler
+    :param ast: Abstract Syntax Tree of the contract generated by the compiler
+    :param bytecode: Bytecode of the contract generated by the compiler
+    :param bytecode_runtime: Bytecode stored at the contract address, excludes the constructor and initialization code
+    :param clone_bin:
+    :param dev_doc:
+    :param decode_tuples: Optionally convert tuples/structs to named tuples
+    :param interface:
+    :param metadata: Contract Metadata generated by the compiler
+    :param opcodes: Opcodes of the contracts generated by the compiler
```
fselmo:
```suggestion
    :param opcodes: Opcodes for the contract generated by the compiler
```

Is this meant to be singular here?

docs/web3.eth.rst
```diff
@@ -1282,22 +1282,28 @@ Contracts
...
+    :param src_map:
+    :param src_map_runtime:
+    :param user_doc:
+    :return: Instance of the contract
+    :rtype: Contract
+    :raises TypeError: if the address is not provided
```
fselmo:
```suggestion
    :raises TypeError: If the address is not provided
```
 Everything else seems to be capitalized here.

docs/web3.eth.rst
```diff
@@ -1282,22 +1282,28 @@ Contracts
...
+    :raises AttributeError: if the contract class is not initialized
```
fselmo:
```suggestion
    :raises AttributeError: If the contract class is not initialized
```

newsfragments/2539.internal.rst
```diff
@@ -0,0 +1 @@
+Additional contract ``abi`` documentation to make it a clear requirement for contract instances.
```
fselmo, line 1:
```suggestion
Additional contract ``abi`` documentation to make it a clear requirement for interacting with contract instances.
```

Perhaps that distinction? Are we allowed to create a contract instance without an ABI? I'm not at my computer at the moment to verify but, if so, is there a particular reason we even allow that and don't validate it as an input to `__init__()`?

Just curious on improvements that could come in `v7` but also maybe we should clarify the exceptions in the docs as well, if they exist.

fselmo (commented):
Nothing holding this back really. Good updates. Just some food for thought.

newsfragments/2539.internal.rst
reedsa: I recall an example that used a contract without an abi, I'll dig it up in the docs in a bit. It is possible to create, though very limited in what it can do.

I have no clear idea why it isn't enforced but maybe it is so that really basic operations can be done.maybe comparing contract instances or something?

PR #3138, opened 2023-11-02
Add types for constants

Uxio0:
### What was wrong?

No typing in web3 constants

### How was it fixed?

Adding types

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/constants.py
+++ b/web3/constants.py
@@ -1,7 +1,14 @@
+from eth_typing import (
+    ChecksumAddress,
+    HexAddress,
+    HexStr,
+)
+
 # Constants as Strings
-ADDRESS_ZERO = "0x0000000000000000000000000000000000000000"
-MAX_INT = "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
-HASH_ZERO = "0x0000000000000000000000000000000000000000000000000000000000000000"
+ADDRESS_ZERO = HexAddress(HexStr("0x0000000000000000000000000000000000000000"))
+CHECKSUM_ADDRESSS_ZERO = ChecksumAddress(ADDRESS_ZERO)
+MAX_INT = HexStr("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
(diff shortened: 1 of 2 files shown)


web3/constants.py
```diff
@@ -1,7 +1,15 @@
...
+    ChecksumAddress,
+    HexAddress,
+    HexStr,
+)
+
 # Constants as Strings
-ADDRESS_ZERO = "0x0000000000000000000000000000000000000000"
-MAX_INT = "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
-HASH_ZERO = "0x0000000000000000000000000000000000000000000000000000000000000000"
+ADDRESS_ZERO = ChecksumAddress(
```
reedsa: @fselmo Might be an opportunity to clean this up. I noticed this constant is defined within the ens module as well, but it does not type it with ChecksumAddress: https://github.com/ethereum/web3.py/blob/d1842bc0a05657b703bb3ce8bbba6fbf0a21d426/ens/constants.py#L16

Does the ens module usage require a separate constant or could these be combined? Other option might be to just keep as is, `ADDRESS_ZERO` for the checksummed address type and `EMPTY_ADDR_HEX` is the non-checksummed type.

fselmo: It would be nice to combine them actually! Good thought.

Uxio0: I totally got that from ens constants indeed, I thought it might be like that for a reason

reedsa: Great! I'll help get this merged in. Will push a refactor of the constants along with release notes in a bit.

Thanks for your contribution @Uxio0 !

web3/middleware/cache.py
```diff
@@ -195,7 +195,7 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
 
 
 _time_based_cache_middleware = construct_time_based_cache_middleware(
-    cache_class=functools.partial(lru.LRU, 256),
+    cache_class=functools.partial(lru.LRU, 256)(),
```
reedsa: @fselmo started seeing CI lint failures for this incorrect default argument. Should be callable but I'm not typing it correctly.

PR #3140, opened 2023-11-02
Add better validation around matching response with request information

fselmo:
### What was wrong?

- Add better validation that the cached request information really matches the response we are trying to process. This is done by passing the actual `response_id` for the matching request once we go back through the middleware onion and the response is present. This area of the code hadn't been iterated on since some of the earlier commits in the building out of the provider. This is a very important and necessary improvement on `id` matching.

This resolves an issue reported on Discord where ``await asyncio.gather(one_request, second_request)`` was guessing the wrong request id for each request since they are basically being triggered / yielding to each other in a shared context.

### Steps to reproduce

```py
async def reproducible_example():
    async with AsyncWeb3.persistent_websocket(WebsocketProviderV2(PROVIDER_URI)) as w3:
        (
            latest,
            chain_id,
            block_num,
            pending,
            chain_id2,
            chain_id3,
            finalized,
            balance,
        ) = await asyncio.gather(
            w3.eth.get_block("latest"),
            w3.eth.chain_id,
            w3.eth.block_number,
            w3.eth.get_block("pending"),
            w3.eth.chain_id,
            w3.eth.chain_id,
            w3.eth.get_block("finalized"),
            w3.eth.get_balance("shaq.eth")
        )
        print(f"block num from latest block: {latest['number']}\n")
        print(f"chain_id: {chain_id}\n")
        print(f"block_num: {block_num}\n")
        print(f"block num from pending block: {pending['number']}\n")
        print(f"chain_id2: {chain_id2}\n")
        print(f"chain_id3: {chain_id3}\n")
        print(f"block num from finalized block: {finalized['number']}\n")
        print(f"balance: {balance}\n")

asyncio.run(reproducible_example())
```

### Todo:

### DIFF
--- a/web3/providers/websocket/request_processor.py
+++ b/web3/providers/websocket/request_processor.py
@@ -177,15 +177,15 @@ class RequestProcessor:
 
     def append_middleware_response_processor(
         self,
+        response_id: int,
         middleware_response_processor: Callable[..., Any],
     ) -> None:
-        request_id = next(copy(self._provider.request_counter)) - 1
-        cache_key = generate_cache_key(request_id)
-        current_request_cached_info: RequestInformation = (
+        cache_key = generate_cache_key(response_id)
+        cached_request_info_for_id: RequestInformation = (
             self._request_information_cache.get_cache_entry(cache_key)
         )
-        if current_request_cached_info:
-            current_request_cached_info.middleware_response_processors.append(
+        if cached_request_info_for_id is not None:
+            cached_request_info_for_id.middleware_response_processors.append(
                 middleware_response_processor
             )
 
--- a/web3/providers/websocket/request_processor.py
+++ b/web3/providers/websocket/request_processor.py
@@ -177,16 +177,29 @@ class RequestProcessor:
 
     def append_middleware_response_processor(
         self,
-        response_id: int,
+        response: RPCResponse,
         middleware_response_processor: Callable[..., Any],
     ) -> None:
-        cache_key = generate_cache_key(response_id)
-        cached_request_info_for_id: RequestInformation = (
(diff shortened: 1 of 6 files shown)


fselmo:
On second thought, I think this is an area where allowing for too much flexibility can be a bad thing. If we can't determine the request id with certainty, we should be guessing too much because methods with the same params can and will have different responses based on when they are called. If 10 `w3.eth.get_block("latest")` requests are gathered, for example, then matching them all to the appropriate one would be impossible by guessing based on params and methods and they could return different responses if they were requested around the transition to a new block.

web3/providers/websocket/request_processor.py
```diff
@@ -177,15 +177,15 @@ def get_request_information_for_response(
 
     def append_middleware_response_processor(
         self,
+        response_id: int,
         middleware_response_processor: Callable[..., Any],
     ) -> None:
-        request_id = next(copy(self._provider.request_counter)) - 1
-        cache_key = generate_cache_key(request_id)
-        current_request_cached_info: RequestInformation = (
+        cache_key = generate_cache_key(response_id)
```
reedsa: In the examples we were walking through earlier, I think there was a sticky point that some responses didnt appear to include an `id`. If that happens, `generate_cache_key` will raise. Should we swallow that exception and just skip the processors here instead?

fselmo: That's a good point. We could still return the raw response and try to not halt things here. I think passing in the `response` to this method would allow us to retrieve the `id` if it's there or debug log the response that doesn't have an `id` and keep moving.

reedsa: Yeah that sounds like it would work great!

reedsa (approved):
Lgtm

Issue #3152, opened 2023-11-28
where can i download the document pdf on web3.py version 6.x


shifenhutu:
https://readthedocs.org/projects/web3py/downloads/

there is no  version 6.x   document  

thank you very much

pacrob:
I've updated to build the Downloadable docs going forward. Unfortunately there doesn't appear to be a way to make them available for past versions. The `latest` is the current state of our docs, and the next numbered version will be available when a new release goes out (within a week).

PR #3153, opened 2023-11-28
add formats key to rtd to enable downloadable docs

pacrob:
### What was wrong?

Related to Issue #3152

### How was it fixed?

Added `formats` key to `.readthedocs.yaml` to enable downloadable docs

It doesn't look like I can actually verify that they are built and available for download until this update is merged, but this is how to do it per the RTD [docs](https://docs.readthedocs.io/en/stable/guides/enable-offline-formats.html#how-to-enable-offline-formats).

Added the `fail_on_warning` key (from the template) while updating RTD settings.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,7 +6,8 @@ build:
     python: "3.8"
 
 sphinx:
-   configuration: docs/conf.py
+  configuration: docs/conf.py
+  fail_on_warning: true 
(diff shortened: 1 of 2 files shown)


fselmo (approved):
Lgtm with a newsfragment 👍🏼

PR #3160, opened 2023-12-04
Pin pytest-asyncio

kclowes:
### What was wrong?
`pytest-asyncio v0.23` changed something that caused all of our async tests to fail. Putting this up so we can move on, but I'll circle back soon.

### How was it fixed?
Pinned to `<0.23`

### Todo:

#### Cute Animal Picture


pacrob (approved):
lgtm!

PR #3168, opened 2023-12-14
Estimate Gas with State Override

### LINKED ISSUE #3164: Add state_override argument to estimate_gas
### What was wrong?

As of geth 1.12.1, `estimate_gas` allows a `state_override` argument. 

### How can it be fixed?

Allow `eth.estimate_gas` to take the `state_override` param. Can follow the existing `eth.call` implementation.

reedsa:
### What was wrong?

Closes #3164

### How was it fixed?

Adds `state_override` param to `estimate_gas` method.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -1062,9 +1062,9 @@ The following methods are available on the ``web3.eth`` namespace.
     The ``transaction`` and ``block_identifier`` parameters are handled in the
     same manner as the :meth:`~web3.eth.Eth.send_transaction()` method.
 
-    The ``state_override`` parameter allows gas estimation for cases when there
-    are multiple calls that require a specific state as a side effect of the
-    earlier transaction, such as an `approve` followed by a `transfer` call.
+    The ``state_override`` is useful when there is a chain of transaction calls.
+    It overrides state so that the gas estimate of a transaction is accurate in
+    cases where prior calls produce side effects.
 
     .. code-block:: python
 
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -1051,7 +1051,7 @@ The following methods are available on the ``web3.eth`` namespace.
         })
 
 
-.. py:method:: Eth.estimate_gas(transaction, block_identifier=None)
+.. py:method:: Eth.estimate_gas(transaction, block_identifier=None, state_override=None)
 
     * Delegates to ``eth_estimateGas`` RPC Method
 
@@ -1061,6 +1061,10 @@ The following methods are available on the ``web3.eth`` namespace.
 
     The ``transaction`` and ``block_identifier`` parameters are handled in the
     same manner as the :meth:`~web3.eth.Eth.send_transaction()` method.
+
+    The ``state_override`` parameter allows gas estimation for cases when there
+    are multiple calls that require a specific state as a side effect of the
+    earlier transaction, such as an `approve` followed by a `transfer` call.
 
     .. code-block:: python
 
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -455,12 +455,31 @@ call_with_override: Callable[
 estimate_gas_without_block_id: Callable[[Dict[str, Any]], Dict[str, Any]]
 estimate_gas_without_block_id = apply_formatter_at_index(transaction_param_formatter, 0)
 estimate_gas_with_block_id: Callable[
-    [Tuple[Dict[str, Any], Union[str, int]]], Tuple[Dict[str, Any], int]
+    [Tuple[Dict[str, Any], BlockIdentifier]], Tuple[Dict[str, Any], int]
 ]
 estimate_gas_with_block_id = apply_formatters_to_sequence(
     [
         transaction_param_formatter,
         to_hex_if_integer,
+    ]
+)
+ESTIMATE_GAS_OVERRIDE_FORMATTERS = {
+    "balance": to_hex_if_integer,
+    "nonce": to_hex_if_integer,
+    "code": to_hex_if_bytes,
+}
+estimate_gas_with_override: Callable[
+    [Tuple[Dict[str, Any], BlockIdentifier, CallOverrideParams]],
+    Tuple[Dict[str, Any], int, Dict[str, Any]],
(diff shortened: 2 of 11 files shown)


web3/_utils/method_formatters.py
```diff
@@ -454,14 +454,33 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
 estimate_gas_without_block_id: Callable[[Dict[str, Any]], Dict[str, Any]]
 estimate_gas_without_block_id = apply_formatter_at_index(transaction_param_formatter, 0)
 estimate_gas_with_block_id: Callable[
-    [Tuple[Dict[str, Any], Union[str, int]]], Tuple[Dict[str, Any], int]
+    [Tuple[Dict[str, Any], BlockIdentifier]], Tuple[Dict[str, Any], int]
```
fselmo, line 458: 💯

docs/web3.eth.rst
```diff
@@ -1076,6 +1076,10 @@ The following methods are available on the ``web3.eth`` namespace.
     The ``transaction`` and ``block_identifier`` parameters are handled in the
     same manner as the :meth:`~web3.eth.Eth.send_transaction()` method.
 
+    The ``state_override`` parameter allows gas estimation for cases when there
```
fselmo: Just curious, where did this description come from? It seems really specific, which may indeed be useful for that specific case but it seems like it may be more broadly used and might be better to describe it as simply overriding parts of the state before estimating the gas. Thoughts?

web3/_utils/module_testing/eth_module.py
```diff
@@ -4149,6 +4149,36 @@ def test_eth_estimate_gas_with_block(
...
+            {
+                "nonce": HexStr("0x1"),  # HexStr
+                "balance": HexStr("0x1"),  # HexStr
+                "code": b"\x00",  # bytes
+                # with stateDiff
+                "stateDiff": {HexStr(f"0x{'00' * 32}"): HexStr(f"0x{'00' * 32}")},
+            },
+        ),
+    )
+    def test_eth_estimate_gas_with_override_param_type_check(
```
fselmo, line 4145: We added the async functionality in this PR, let's add an analogous async test for this too. I noticed the `eth_call` equivalent also doesn't have an async test... it would be nice to add that as well and it could be in this PR if not too much of a burden.

web3/eth/base_eth.py
```diff
@@ -95,17 +95,28 @@ def set_gas_price_strategy(
...
-            params: Sequence[Union[TxParams, BlockIdentifier]] = [transaction]
-        else:
-            params = [transaction, block_identifier]
+            return [transaction]
 
-        return params
+        if state_override is None:
+            return [
+                transaction,
+                block_identifier,
```
fselmo: This is a nit but we could condense these lines into one. If you remove the comma after the last item in a list, or dict, and it fits into the line without triggering a line-length violation, it will look like this after `black` runs:

```py
return [transaction, block_identifier]
```

The same can be done by removing the trailing comma after `state_override` below. It makes things more concise / easier to read.

fselmo (commented):
This looks good. It would be good to add a test for the async functionality that is in here too. Also, do you know if we're testing the contract functions somewhere, related to `contract.functions.my_func().estimate_gas()`, etc... with all the params? Would be very nice to know that those are working as expected as well.

web3/eth/base_eth.py
```diff
@@ -95,17 +95,28 @@ def set_gas_price_strategy(
...
-    ) -> Sequence[Union[TxParams, BlockIdentifier]]:
+        self,
+        transaction: TxParams,
+        block_identifier: Optional[BlockIdentifier] = None,
+        state_override: Optional[CallOverride] = None,
+    ) -> Sequence[Union[TxParams, BlockIdentifier, CallOverride]]:
         if "from" not in transaction and is_checksum_address(self.default_account):
             transaction = assoc(transaction, "from", self.default_account)
 
         if block_identifier is None:
```
fselmo, line 109: I don't think this logic is doing what we want. The `eth_call` munger will set the block to the default block but geth also will use the "pending" block if the block identifier is missing... but even if it is `None` (or set to default), we want to be able to send a `state_override`. It doesn't look like we allow for that with this `return` here.

We should probably abstract the logic of the `call_munger` out and use it here as well. Instead of renaming it let's keep `estimate_gas_munger()` and `call_munger()` separated and let's abstract only one logic block into a `_eth_call_and_estimate_gas_munger()` and use that in each. This way we are not trying to recreate the same logic in both places and possibly behaving one way in one and another way in the other (as it currently appears to be happening).

```py
def _eth_call_and_estimate_gas_munger(...):
    # eth_call munger logic here since that's already written out
    
def call_munger(...):
   return  _eth_call_and_estimate_gas_munger(...)
   
def estimate_gas_munger(...):
   return  _eth_call_and_estimate_gas_munger(...)
```

That way if they ever don't have parity we can still tweak, etc... but for now they share the same logic so let's reduce the surface for friction / errors to occur. Thoughts on that?

docs/web3.eth.rst
reedsa: You're right, I pulled this from the example they provided in the [original issue in geth](https://github.com/ethereum/go-ethereum/issues/27800). Happy to change it so that it doesnt include the specific transaction call examples.

Would this be better? Feel free to call out your own idea or modify my words.

> The ``state_override`` parameter is useful when there is a chain of transaction calls. It overrides state so that the gas estimate of a transaction is accurate in cases where prior calls produce side effects.

fselmo: That makes more sense to me in a broader scope 👍🏼. Ship it.

web3/eth/base_eth.py
reedsa: Oh yep, good catch there. Not sure why I thought returning just the transaction would work, will fix!

pacrob (approved):
lgtm!

kclowes (approved):
This is looking good! Like @fselmo mentioned, it would be nice to add some tests for the contract `estimate_gas` `state_override`. Those can be found in `tests/core/contracts/test_contract_estimate_gas.py`

PR #3169, opened 2023-12-15
Refactor the middleware setup [starts v7 breaking changes]

### LINKED ISSUE #899: rename geth_poa_middleware to something more generic
### What was wrong?

It seems like many other chains have set higher length for `extraData`.  see #898

### How can it be fixed?

Should `geth_poa_middleware` be renamed to something like `long_extraData_middleware`?  Or could the extraData length be configurable?

### LINKED ISSUE #2990: Improve user experience around provider-specific middlewares
### What was wrong?

- Currently there is a middleware onion on the `RequestManager` and there are provider-specific middlewares. This leads to confusion like in #2770. We should re-think the way that request retry is handled in a more user-friendly way.

### How can it be fixed?

One possible solution is to remove the idea of middlewares on providers and leave it to the middleware onion only. If we want to add some provider-specific features, these might be nice features as flags on the provider instead or as provider-specific middleware that may be added to the onion itself in the `RequestManager` class.

Suggestions / opinions encouraged!

### LINKED ISSUE #2995: Remove ABI Middleware
Default middlewares should not include the ABI middleware. 

Dependencies on the middleware should be identified and removed.

Validation errors caused when removing ABI from the defaults. 

Tests failing in this build without `abi_middleware`: https://app.circleci.com/pipelines/github/ethereum/web3.py/4559/workflows/8c5fe87f-fe5f-4a45-97ee-8391158c502f/jobs/308436

_Originally posted by @reedsa in https://github.com/ethereum/web3.py/pull/2982#discussion_r1223350822_

### LINKED ISSUE #3021: Remove reliance on `web3` instance for `name_to_address_middleware`
When we have a `v7` branch going, the sync ENS name-to-address middleware can be re-written as:

```py
def name_to_address_middleware(
    make_request: Callable[[RPCEndpoint, Any], Any],
    w3: "Web3",
) -> Middleware:
    normalizers = [
        abi_ens_resolver(w3),
    ]
    middleware = construct_formatting_middleware(
        request_formatters=abi_request_formatters(normalizers, RPC_ABIS)
    )
    return middleware(make_request, w3)
 ```
 
 Removing the need to pass in a web3 instance to `default_middlewares()` in `manager.py`. It can then be passed in just as the other default middlewares, as the method reference `name_to_address_middleware`, rather than the method instance `name_to_address_middleware(w3)`.

### PR DESCRIPTION
**Pull request by fselmo**
### What was wrong?

In order to process requests and responses separately and to eventually support _batch requests_, we need to uncouple the `make_request()` method from being sandwiched along with all the middleware in each request. If we want APIs that can process request params separately from making a request, we need to refactor quite a bit of the code to massage this sort of logic in.

### How was it fixed?

In this PR I am refactoring the middleware from the functional programming paradigm of a sandwiched request into a class-based structure with request processors and response processors. The ultimate end result for most requests is these classes still wrap the `make_request`, as it is often convenient to be able to replace the `make_request` function in order for a specific request to not be made under certain conditions some middleware necessitate (currently this is only true for the `local_filter_middleware`).

Imo this class-based structure makes the middleware a bit easier to read and create. If a middleware only processes request, it only needs to define a request processor; if the response is the only thing the middleware touches, it can override the response processor on the class, etc. In case the `make_request` method itself needs to be altered, the `_wrap_make_request()` method can be overridden and that would behave similarly to the way the middlewares currently work.

Some things to note:

- This refactor removes the request retry middleware from the provider and removes the concept of provider middleware altogether. Instead, exception retry logic is done as a configuration on the HTTP providers.
- This refactor removes the result generator middlewares and opts for a `request_mocker` fixture for dealing with mocked requests. This has proved to be quite nice to work with, at least for the current tests, and it removes the need to create, inject, and clean up middleware for each test. Instead, the context manager takes care of mocking while inside the manager and cleaning up when exiting the context manager for both sync and async.
- The `simple_cache_middleware` is moved to a decorator on the `make_request` method for the providers. This can be turned on and configured as configurations on the appropriate provider classes.

---

- closes #3021 - Removes the need to pass `w3` into the default middlewares 
- closes #2995 - Remove the need for `abi_middleware` by fixing where the `name_to_address` middleware would override request formatters
- closes #2990 - Removes provider-specific middleware except for a hardcoded specific set of middleware for the in-house `EthereumTesterProvider`
- closes #899 - The current version of the middleware takes `extraData` field and changes it to `proofOfAuthority`. The name was changed from `geth_poa_middleware` to `extradata_to_poa_middleware`

---

### Todo:
- [x] Revise some of the type hinting. In particular the type hinting around the Middleware within manager and provider classes
- Update documentation to reflect above changes ( will be in a separate PR)

#### Cute Animal Picture

IMG-20240103-WA0005

### CODE REVIEW

File: tests/core/contracts/test_contract_example.py [hunk 1]
```diff
@@ -121,10 +119,9 @@ def async_eth_tester():
 
 @pytest_asyncio.fixture()
 async def async_w3():
-    provider = AsyncEthereumTesterProvider()
-    w3 = Web3(provider, modules={"eth": [AsyncEth]}, middlewares=provider.middlewares)
-    w3.eth.default_account = await w3.eth.coinbase
-    return w3
+    async_w3 = AsyncWeb3(AsyncEthereumTesterProvider())
```
reedsa, line 122: Very nice 😻

tests/core/manager/conftest.py
```diff
@@ -15,15 +19,19 @@ class Wrapper:
...
-                    method = "|".join((method, key))
-                    response = make_request(method, params)
-                    response["result"]["middlewares"].append(key)
-                    return response
+            def __call__(self, web3):
+                class Middleware(Web3Middleware):
+                    def _wrap_make_request(self, make_request):
+                        def middleware_fn(method, params):
+                            params.append(key)
+                            method = "|".join((method, key))
```
reedsa, line 27: Is this pipe a delimiter for the request so we have a way to determine each unique method call?

fselmo: Yeah, so we have a way of keeping track of unique middleware "keys" and asserting the order they wrap things in. I think it's really only used in [this test](https://github.com/ethereum/web3.py/blob/0fc0e064fbba6eca7a78b044dc8bc5888d5a0c45/tests/core/manager/test_provider_request_wrapping.py#L29-L29) if I'm searching correctly. I didn't change the logic here at all though, this code is from 2017 it looks like.

tests/core/caching-utils/test_request_caching.py
```diff
@@ -0,0 +1,238 @@
...
+        with pytest.raises(ValueError) as err_a:
+            w3.manager.request_blocking("fake_endpoint", [])
+        with pytest.raises(ValueError) as err_b:
+            w3.manager.request_blocking("fake_endpoint", [])
+
+        assert str(err_a) != str(err_b)
+        assert err_a.value.args != err_b.value.args
+
+
+def test_request_caching_does_not_cache_endpoints_not_in_whitelist(w3):
```
pacrob:
```suggestion
def test_request_caching_does_not_cache_endpoints_not_in_allowlist(w3):
```

tests/core/caching-utils/test_request_caching.py
```diff
@@ -0,0 +1,238 @@
...
+        with pytest.raises(ValueError) as err_a:
+            await async_w3.manager.coro_request("fake_endpoint", [])
+        with pytest.raises(ValueError) as err_b:
+            await async_w3.manager.coro_request("fake_endpoint", [])
+
+    assert str(err_a) != str(err_b)
+
+
+@pytest.mark.asyncio
+async def test_async_request_caching_does_not_cache_non_whitelist_endpoints(
```
pacrob:
```suggestion
async def test_async_request_caching_does_not_cache_non_allowlist_endpoints(
```

docs/middleware.rst
```diff
@@ -446,16 +446,15 @@ Time-based Cache Middleware
...
 ~~~~~~~~~~~~~~~~~~
 
-.. py:method:: web3.middleware.geth_poa_middleware
-               web3.middleware.async_geth_poa_middleware
+.. py:class:: web3.middleware.extradata_to_poa_middleware
 
 .. note::
     It's important to inject the middleware at the 0th layer of the middleware onion:
-    ``w3.middleware_onion.inject(geth_poa_middleware, layer=0)``
+    ``w3.middleware_onion.inject(extradata_to_poa_middleware, layer=0)``
```
kclowes, line 441: this is a nit you can definitely leave, but maybe call this: `poa_extradata_middleware`?

web3/_utils/module_testing/utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def __enter__(self) -> "Self":
+        setattr(self.w3.provider, "make_request", self._mock_request_handler)
+        # reset request func cache to re-build request_func with mocked make_request
+        self.w3.provider._request_func_cache = (None, None)
+
+        return self
+
+    # define __exit__ with typing information
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
```
kclowes, line 75: nit: Can we tighten up any of these types?

tests/core/caching-utils/test_request_caching.py
```diff
@@ -0,0 +1,238 @@
...
+    mock_results_c = {RPCEndpoint("eth_chainId"): 33333}
+
+    with request_mocker(w3_a, mock_results=mock_results_a):
+        with request_mocker(w3_b, mock_results=mock_results_b):
+            with request_mocker(w3_c, mock_results=mock_results_c):
+                result_a = w3_a.manager.request_blocking("eth_chainId", [])
+                result_b = w3_b.manager.request_blocking("eth_chainId", [])
+                result_c = w3_c.manager.request_blocking("eth_chainId", [])
+
+    assert result_a != result_b != result_c
```
kclowes: nit: this is redundant if we're already checking what the actual values are in the next couple lines 🤷

web3/middleware/exception_retry_request.py
pacrob, line 1: It looks like this file was emptied but not deleted. Any reason it shouldn't be?

web3/middleware/filter.py
```diff
@@ -411,9 +373,9 @@ async def async_iter_latest_block(
     is_bounded_range = to_block is not None and to_block != "latest"
 
     while True:
-        latest_block = await w3.eth.block_number  # type: ignore
+        latest_block = await w3.eth.block_number
         # type ignored b/c is_bounded_range prevents unsupported comparison
```
pacrob, line 377: can remove `type ignored b/c` because `type: ignore` removed

web3/middleware/cache.py
pacrob, line 1: empty file?

web3/main.py
```diff
@@ -472,8 +469,8 @@ async def is_connected(self, show_traceback: bool = False) -> bool:
         return await self.provider.is_connected(show_traceback)
 
     @property
-    def middleware_onion(self) -> AsyncMiddlewareOnion:
-        return cast(AsyncMiddlewareOnion, self.manager.middleware_onion)
+    def middleware_onion(self) -> MiddlewareOnion:
```
kclowes: nit: we could probably put this on `BaseWeb3` now?

tests/core/providers/test_http_request_retry.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+@patch("web3.providers.rpc.rpc.make_post_request", side_effect=ConnectionError)
+def test_valid_method_retried(make_post_request_mock, w3):
+    with pytest.raises(ConnectionError):
+        w3.provider.make_request(RPCEndpoint("eth_getBalance"), [f"0x{'00' * 20}"])
+    assert make_post_request_mock.call_count == TEST_RETRY_COUNT
+
+
+def test_exception_retry_config_is_strictly_on_http_provider():
```
pacrob, line 82: Could do WS, and AsyncWeb3 versions too, for completeness. But a take/leave as you like.

web3/middleware/base.py
```diff
@@ -0,0 +1,129 @@
...
+        w3.middleware_onion.inject(my_middleware, layer=0)
+        ```
+        """
+        raise NotImplementedError("Must be implemented by subclasses")
+
+
+# --- type definitions --- #
+
+Middleware = Type[Web3Middleware]
+MiddlewareOnion = NamedElementOnion[str, Middleware]
```
kclowes, line 129: nit: is this `str` accurate any more?

web3/middleware/filter.py
fselmo: This is separate. It's talking about the still-type-ignored `is_bounded_range` line below the comment.

tests/core/providers/test_http_request_retry.py
fselmo: Yeah I just ported the test as it was over here. Refraining from adding too many additional things at the moment but this seems simple enough. I'll put it in 👍🏼

pacrob (approved):
Nicely done! :weight_lifting:
GTG for me.

web3/providers/async_base.py
```diff
@@ -40,53 +48,61 @@
...
+    def __init__(self) -> None:
+        self._request_cache = SimpleCache(1000)
 
     async def request_func(
-        self, async_w3: "AsyncWeb3", outer_middlewares: AsyncMiddlewareOnion
+        self, async_w3: "AsyncWeb3", middlewares: MiddlewareOnion
     ) -> Callable[..., Coroutine[Any, Any, RPCResponse]]:
         # type ignored b/c tuple(MiddlewareOnion) converts to tuple of middlewares
-        all_middlewares: Tuple[AsyncMiddleware] = tuple(outer_middlewares) + tuple(self.middlewares)  # type: ignore  # noqa: E501
+        middlewares: Tuple[Middleware, ...] = tuple(middlewares)  # type: ignore
```
kclowes: nit: does this `type: ignore` still need to be here?

web3/providers/base.py
```diff
@@ -47,44 +72,42 @@ class BaseProvider:
...
 
         cache_key = self._request_func_cache[0]
-        if cache_key is None or cache_key != all_middlewares:
+        if cache_key != middlewares:  # type: ignore
             self._request_func_cache = (
-                all_middlewares,
-                self._generate_request_func(w3, all_middlewares),
+                middlewares,
+                combine_middlewares(
+                    middlewares=middlewares,  # type: ignore
```
kclowes: why do we need to ignore the type here? might be worth a comment?

kclowes (approved):
Left a few nits to take or leave, but looks good!

web3/_utils/module_testing/utils.py
fselmo: That comment was supposed to be a `# TODO`, thanks for catching 👍🏼

Nov1kov:
Sorry for an almost off-topic. But why the package [eth_tester](https://github.com/ethereum/web3.py/tree/main/web3/providers/eth_tester) couldn't be fully in the [eth-tester pypi](https://github.com/ethereum/eth-tester)?

For adding implementation for one method in the eth-tester, I have to implement formatter also in web3.py

web3/middleware/base.py
fselmo: All keys (`TKey` names) in the onion should still be strings, yeah. That logic hasn't changed other than to use the `repr()` of the element if the `TValue` isn't hashable.

web3/providers/async_base.py
fselmo: Unfortunately yeah. I played around with this a bit but didn't get too far. It comes fromt he fact that the `__iter__()` method defined before my time will turn the onion into a `Tuple[Middleware, ...]` but to type it as such isn't straightforward. The type here _is_ correct, we just don't have a very elegant solution for this (yet) and maybe we can tackle that as a separate issue altogether. Could be worth looking into. I think it has to do with satisfying the `Mapping` inheritance for the `NamedElementOnion`.

fselmo: Just kidding. The moment I said a definitive "yeah we have to keep it" something clicked. It was easily solved by creating a method to tupelize the onion into a tuple of middlewares with the correct type hinting. I pushed that up as a separate commit in case that isn't ideal but I prefer than to the `type: ignore`s myself. Let me know what you think on that.

web3/providers/base.py
fselmo: This comment actually made me solve the comment above. All these `type: ignores` go away with the `as_tuple_of_middlewares()` method. I could use help naming it though 😆. I like how explicit that name is and I felt like `tupleize` just doesn't explain what it does for a reader trying to move on past the code. Any thoughts there would be appreciated.

docs/middleware.rst
fselmo: My thought with `extradata_to_poa_middleware` is that it tells you exactly what the middleware does. I may leave as is but I'm not tied to it if you have a strong opinion here. Let me know.

fselmo:
@kclowes I think I addressed all your comments. Let me know if there's anything else outstanding and how you feel about the [tupleization commit](a548a4e420b304d548270657b2868cb0be8ee6d6).

docs/middleware.rst
kclowes: Nope, no strong opinion!

web3/providers/async_base.py
kclowes: Nice! Yeah, I think I prefer that too

web3/providers/base.py
kclowes: I can't think of anything better off the top of my head but will let you know if I do!

kclowes:
LGTM! 🚀

web3/middleware/base.py
fselmo: Actually, I changed that due to [this comment](https://github.com/ethereum/web3.py/blob/45db990ec8240930a11a38e467bfcf200e46aa50/web3/datastructures.py#L32-L39). If `TKey` is bound to a hashable we should define a `__hash__()` that would reflect any changes to the middleware class.

fselmo: False alarm, this was mostly because we borrow the defined ``TKey`` for ``AttributeDict`` data structure. We can revisit this later but we already break these rules in the middleware onion when we set a ``TKey`` as a "name" string. I took out that last commit, adding a newsfragment and merging!

Issue #3174, opened 2023-12-19
Excessive waits when listening to events through a websocket


poolpitako:
* Version: 6.12.0
* Python: 3.12
* OS: osx/linux/win
* `pip freeze` output

```
aiohttp==3.9.1
aiosignal==1.3.1
annotated-types==0.6.0
anyio==3.7.1
asyncpg==0.29.0
attrs==23.1.0
beautifulsoup4==4.12.2
bitarray==2.8.5
cachetools==5.3.2
certifi==2023.11.17
charset-normalizer==3.3.2
click==8.1.7
cytoolz==0.12.2
eth-abi==4.2.1
eth-account==0.10.0
eth-hash==0.5.2
eth-keyfile==0.7.0
eth-keys==0.4.0
eth-rlp==1.0.0
eth-typing==3.5.2
eth-utils==2.3.1
fastapi==0.104.1
frozenlist==1.4.0
h11==0.14.0
hexbytes==0.3.1
idna==3.6
jsonschema==4.20.0
jsonschema-specifications==2023.11.2
lru-dict==1.2.0
multidict==6.0.4
parsimonious==0.9.0
pipdeptree==2.13.1
proto-plus==1.23.0
protobuf==4.25.1
pyasn1==0.5.1
pyasn1-modules==0.3.0
pycryptodome==3.19.0
pydantic==2.5.2
pydantic-settings==2.0.3
pydantic_core==2.14.5
python-dotenv==1.0.0
pyunormalize==15.1.0
PyYAML==6.0.1
referencing==0.32.0
regex==2023.10.3
requests==2.31.0
rlp==4.0.0
rpds-py==0.13.2
rsa==4.9
setuptools==69.0.2
six==1.16.0
sniffio==1.3.0
soupsieve==2.5
starlette==0.27.0
toolz==0.12.0
typing_extensions==4.9.0
urllib3==2.1.0
uvicorn==0.23.2
web3==6.12.0
websockets==12.0
yarl==1.9.4
```

### What was wrong?

When user does `listen_to_websocket()`, web3.py ends up doing:
```
response = await self._provider._ws_recv(timeout=0.5)
```

This means that web3.py will wait 0.5 secs for a possible event since websockets' `recv()` will block.
We noticed the issue since we are running a web server with several ws connections to different RPCs. Each ws connection have different subscriptions.

In our tests, we were seeing events arriving several seconds after the websockets library was getting the frame.

### How can it be fixed?

As mentioned in
[...]

PR #3176, opened 2023-12-20
Improve efficiency checking cache and ws.recv() for WebsocketProviderV2 responses

fselmo:
### What was wrong?

Instead of re-creating some complex while loop to look for request id to match a response id and another to look for subscription responses, both in the cache and in the websocket, use `asyncio.wait(tasks, FIRST_COMPLETED)` with some more-simply-defined tasks to check for both at the same time. Previously we had to check the websocket for some x amount of time (0.5 seconds) and then check back in with the cache and then try the socket again. This way, we check both tasks at the same time and return when the winner finds the response we want.

Related to #3174

Bonus:

- Adds some `INFO` level logs to websocket provider V2 around request caching to help visualize what responses are coming from the cache and what responses are coming directly from checking the websocket.

### Todo:

### DIFF
--- a/tests/core/providers/test_wsv2_provider.py
+++ b/tests/core/providers/test_wsv2_provider.py
@@ -40,6 +40,7 @@ async def test_async_make_request_caches_all_undesired_responses_and_returns_des
     method_under_test = provider.make_request
 
     _mock_ws(provider)
+
     undesired_responses_count = 10
     ws_recv_responses = [
         to_bytes(
@@ -53,6 +54,7 @@ async def test_async_make_request_caches_all_undesired_responses_and_returns_des
         )
         for i in range(0, undesired_responses_count)
     ]
+
     # The first request we make should have an id of `0`, expect the response to match
     # that id. Append it as the last response in the list.
     ws_recv_responses.append(b'{"jsonrpc": "2.0", "id":0, "result": "0x1337"}')
@@ -90,11 +92,12 @@ async def test_async_make_request_returns_cached_response_with_no_recv_if_cached
     desired_response = {"jsonrpc": "2.0", "id": 0, "result": "0x1337"}
     provider._request_processor.cache_raw_response(desired_response)
 
+    # request with id `0` should return the cached response
     response = await method_under_test(RPCEndpoint("some_method"), ["desired_params"])
     assert response == desired_response
 
     assert len(provider._request_processor._request_response_cache) == 0
-    assert not provider._ws.recv.called  # type: ignore
+    assert provider._ws.recv.not_called  # type: ignore
 
 
 @pytest.mark.asyncio
--- a/web3/providers/websocket/request_processor.py
+++ b/web3/providers/websocket/request_processor.py
@@ -1,3 +1,4 @@
+import asyncio
 from collections import (
     deque,
 )
@@ -20,6 +21,7 @@ from web3._utils.caching import (
 )
 from web3.types import (
     RPCEndpoint,
+    RPCId,
     RPCResponse,
 )
 from web3.utils import (
@@ -204,7 +206,9 @@ class RequestProcessor:
 
     # raw response cache
 
-    def cache_raw_response(self, raw_response: Any, subscription: bool = False) -> None:
+    def cache_raw_response(self, raw_response: Any) -> None:
+        subscription = raw_response.get("method") == "eth_subscription"
+
         if subscription:
             self._provider.logger.debug(
                 f"Caching subscription response:\n    response={raw_response}"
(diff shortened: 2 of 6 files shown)


web3/providers/websocket/request_processor.py
```diff
@@ -246,12 +250,53 @@ def pop_raw_response(
...
+
+        - If subscription is True, wait for the next cached subscription response.
+        - If subscription is False, wait for the next cached response that matches the
+            ``request_id``.
+        """
+        if request_id is None and not subscription:
+            raise ValueError("Must provide a request_id when subscription is False.")
+        elif subscription and request_id is not None:
+            raise ValueError(
+                "``request_id`` must be ``None`` when ``subscription`` is True."
```
reedsa, line 276: Nit: Could include the value of `request_id` in the error message.

web3/providers/websocket/websocket_v2.py
```diff
@@ -160,71 +160,46 @@ async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
 
+        # if not in cache, kick off two competing tasks to wait for the desired response
         try:
-            # Add the request timeout around the while loop that checks the request
-            # cache and tried to recv(). If the request is neither in the cache, nor
-            # received within the request_timeout, raise ``TimeExhausted``.
-            return await asyncio.wait_for(
-                _match_response_id_to_request_id(), self.request_timeout
+            completed, pending = await asyncio.wait_for(
+                asyncio.wait(
```
reedsa, line 177: Any chance this could be extracted to a utility for DRYness? Similar `asyncio.wait` happening in [web3/manager.py](https://github.com/ethereum/web3.py/pull/3176/files#diff-c902d776bb4f676fa5cca4ee81273cd66d1113db8109d4f607167681b0b8bdeaR376).
Maybe could pass an argument for the `request_id` vs `subscription`?

web3/providers/websocket/websocket_v2.py
```diff
@@ -234,5 +209,54 @@ async def _match_response_id_to_request_id() -> RPCResponse:
...
+
+        - If subscription is True, wait for the next subscription response.
+        - If subscription is False, wait for the next response that matches the
+            ``request_id``.
+        """
+        if request_id is None and not subscription:
+            raise ValueError("Must provide a request_id when subscription is False.")
+        elif subscription and request_id is not None:
+            raise ValueError(
+                "``request_id`` must be ``None`` when ``subscription`` is True."
```
reedsa, line 227: Same Nit as above, might include `request_id` value in the error message.

reedsa (commented):
Minor nits, but glad this update works well!

web3/providers/websocket/request_processor.py
fselmo: This is an internal method to us and I validated it just so we know we shouldn't use it in our code under certain conditions. If it isn't `None`, it shouldn't be _anything_, do you think it would still be nice to have it? 👀

web3/providers/websocket/websocket_v2.py
fselmo: Good point. Not sure where that could reside 🤔. Any ideas?

fselmo: It may make sense in a `web3/providers/websocket/utils.py` or something similar. I'll toy with that idea tomorrow.

web3/providers/websocket/request_processor.py
reedsa: That's good to know. The logic implies the value is only relevant in the case that `subscription` is false. Thought about including a comment in the block at the top of this method but even that seems redundant. Should be ok to leave as is!

web3/providers/websocket/websocket_v2.py
reedsa: I noticed references to `web3._utils` as well if we want to consider creating `websockets.py` within that. Can't tell if using `web3/providers/websocket/utils.py` would be a problem if using it within other modules. Keeping it private is probably better though.

tests/core/providers/test_wsv2_provider.py
```diff
@@ -90,11 +92,12 @@ async def test_async_make_request_returns_cached_response_with_no_recv_if_cached
     desired_response = {"jsonrpc": "2.0", "id": 0, "result": "0x1337"}
     provider._request_processor.cache_raw_response(desired_response)
 
+    # request with id `0` should return the cached response
     response = await method_under_test(RPCEndpoint("some_method"), ["desired_params"])
     assert response == desired_response
 
     assert len(provider._request_processor._request_response_cache) == 0
-    assert not provider._ws.recv.called  # type: ignore
+    assert provider._ws.recv.not_called  # type: ignore
```
pacrob, line 100: I tried a few lint runs without the `type: ignore` and it seems to be unnecessary. Does it work for you if you remove it?

pacrob (approved):
minor question on typing, otherwise LGTM!

tests/core/providers/test_wsv2_provider.py
fselmo: Oh good to know. This may have just been a PyCharm warning then. I'll pull it, thanks 👍🏼

fselmo:
Closing in favor of #3179

fselmo:
Though this improved efficiency when checking the cache, the timeout on ``recv()`` creates an issue when multiple providers are instantiated side by side, subscribing to different subscriptions. Since instantiating an instance for each subscription makes more sense overall, we should prioritize support for that design pattern.

The main issue I was seeing is that though we added ``asyncio.sleep(0)`` to yield to the event loop to possibly multiple tasks, the timeout was being triggered in that scenario when a message was retrieved in one instance and it yielded to another, etc.

#3179 was created as an attempt to address the ``recv()`` as a separate call in general. With each instance having a listener task that sorts messages to internal message stores, we are only ever listening on the connection in one place and iterating on the messages. Since the message iterator pattern is triggered by events, this ends up being more efficient than checking the ``recv()`` over and over again in a `while True` loop.

Moving the discussion to that PR.

Issue #3032, opened 2023-07-10
Drop Python 3.7


kclowes:
* Version: v7

### What was wrong?

As of 6/2023, Python 3.7 is no longer supported, so we shouldn't support either.

### How can it be fixed?

Remove it from CI, and search the code base to update anywhere we're checking for a Python version of <3.8

akash-kumar-saw:
I would love to contribute... 
/assign me please 🙏

kclowes:
Thanks for your willingness to contribute! It'll be a while before we're ready to start the v7 branch, but we'll let you know when we are.

DanielFrantes:
@kclowes can I contribu on this.

DanielFrantes:
Also on what version we want to update it

fselmo:
closed by #3198

Issue #2351, opened 2022-02-18
[proposal] add support for poetry.


wiseaidev:
IMHO, using requirements.txt along with setup.py is quite outdated, and it is arguably insecure. What you can do instead is to use [Poetry](https://github.com/python-poetry/poetry) for a proper and modern way of package management. Apparently, Poetry is well maintained and has a good performance. However, I guess that web3 is not currently compatible with the poetry architecture.

What do you think about the project's likelihood of adopting this architecture? Do you think that this project could support pyproject.toml and the [tool.poetry.dependencies] section as used by poetry?

Are you guys down for using poetry as a dependency manager? If so, I would like to give it a shot and start working on it.

Fun fact: "Writing poetry is very powerful for maintaining your well-being, and your mental health." - TrustMeBro.

Consider RTFM of [poetry](https://python-poetry.org/docs/).

kclowes:
We are not going to use Poetry in the near future.

PR #3178, opened 2023-12-23
Modern Tooling for Modern Groove Dealers

wiseaidev:
### What was wrong?

Related to Issue #3032 & #2351 & #2355

### How was it fixed?
I have dropped support for python 3.7, migrated to poetry by following the docs:
- https://python-poetry.org/docs/
- https://docs.python.org/3/library/typing.html

P.S. Just realized I dropped support for python 3.8 as well. Probably merge this PR once 3.8 is deprecated (9 months from now):

Screenshot from 2023-12-23 16-01-47

### Todo:

#### Cute Animal Picture

dope cat groove dealer


pacrob:
Hey @wiseaidev ! Thanks for the attention, but this isn't something we can accept right now. 

Regarding the build tooling, `setuptools` is still meeting all our needs. That may change in the future, but it's not a decision taken lightly.

Regarding the python versions, dropping python 3.7 will be in our next major, version 7.  Dropping 3.8 would then be done with another major, likely sometime 2025.

PR #3179, opened 2023-12-27
Use a message listener background task over ws.recv()

### LINKED ISSUE #3174: Excessive waits when listening to events through a websocket

fselmo:
### What was wrong?

- Get rid of ``websockets.recv()`` calls in favor of a background task that listens for messages and puts them into appropriate queues / caches.

- This rids us of the need to have an ``asyncio.Lock()`` around the ``recv()`` calls.

- Each provider instance has a listener task that error logs unless it is configured to raise exceptions, in which case exceptions will be raised and halt the listener task.

- In the (ideally) rare case that the subscription queue fills up to its max size, this will trigger an ``asycio.Event`` that will cause the listener task to wait until messages are consumed from the queue to resume listening.

closes #3174 

### Todo:
- [x] Add tests around the subscription cache being full and triggering the `asyncio.Event` until messages are being consumed again
- [x] Add tests around exception suppression in the listener task
- [x] Update documentation to reflect above changes
- [x] Do not yield subscription messages that are no longer active. It isn't very easy to remove items from an `asyncio.Queue` so we should probably skip over the messages that are no longer active (ones that no longer have a `RequestInformation` object mapping to the subscription `id`). This effectively loses the subscriptions without having to re-create the queue every time we unsubscribe.

### DIFF
--- a/tests/core/providers/test_wsv2_provider.py
+++ b/tests/core/providers/test_wsv2_provider.py
@@ -3,12 +3,16 @@ import json
 import pytest
 from unittest.mock import (
     AsyncMock,
+    patch,
 )
 
 from eth_utils import (
     to_bytes,
 )
 
+from web3._utils.module_testing.module_testing_utils import (
+    WebsocketMessageStreamMock,
+)
 from web3.exceptions import (
     TimeExhausted,
 )
@@ -24,13 +28,22 @@ def _mock_ws(provider):
     provider._ws = AsyncMock()
 
 
+async def _coro():
+    return None
+
+
 @pytest.mark.asyncio
 async def test_async_make_request_caches_all_undesired_responses_and_returns_desired():
     provider = WebsocketProviderV2("ws://mocked")
 
-    method_under_test = provider.make_request
+    with patch(
+        "web3.providers.websocket.websocket_v2.connect", new=lambda *_1, **_2: _coro()
+    ):
+        await provider.connect()
 
     _mock_ws(provider)
+    method_under_test = provider.make_request
+
     undesired_responses_count = 10
     ws_recv_responses = [
         to_bytes(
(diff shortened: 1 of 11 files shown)


web3/providers/websocket/websocket_v2.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         endpoint_uri: Optional[Union[URI, str]] = None,
         websocket_kwargs: Optional[Dict[str, Any]] = None,
         request_timeout: Optional[float] = DEFAULT_PERSISTENT_CONNECTION_TIMEOUT,
```
pacrob: If I want to define a different `subscription_response_queue_size` to be passed on to the `PersistentConnectionProvider`, should it not be available as an optional arg here, to then be passed on via `super().__init__`? Or is there a different place I'm missing?

fselmo: Good catch, We can probably leave default as `None` here and let the superclass set the default value too. I'll put that config in 👍🏼

pacrob (approved):
lgtm!

tests/core/providers/test_wsv2_provider.py
```diff
@@ -27,21 +47,25 @@ def _mock_ws(provider):
...
+
 @pytest.mark.asyncio
-@pytest.mark.skipif(
-    # TODO: remove when python 3.7 is no longer supported in web3.py
-    #  python 3.7 is already sunset so this feels like a reasonable tradeoff
-    sys.version_info < (3, 8),
-    reason="Uses AsyncMock, not supported by python 3.7",
-)
-async def test_async_make_request_caches_all_undesired_responses_and_returns_desired():
+@skip_if_py37
```
reedsa: Nice refactor!

reedsa (approved):
Testing on this looks great. Lgtm

fselmo:
note: I just rebased due to significant changes on main since this was opened. No code should have changed 👀

web3/manager.py
```diff
@@ -324,47 +326,35 @@ async def ws_send(self, method: RPCEndpoint, params: Any) -> RPCResponse:
...
-
-                    if response.get("method") == "eth_subscription":
-                        break
-                    else:
-                        self._provider._request_processor.cache_raw_response(response)
-
-        yield await self._process_ws_response(response)
+            if (
+                response is not None
+                and response.get("params").get("subscription")
```
kclowes: `response.get("params")` returns `None` if params isn't found, which will make the second `.get('subscription')` throw an: `AttributeError: 'NoneType' object has no attribute 'get'`. Maybe change to something like: 
```suggestion
                and response.get("params", {}).get("subscription")
```

kclowes (approved):
I like the change from `Recv` -> `Message`, and like that this will be more extensible! I just left one minor comment.

PR #3180, opened 2023-12-29
Mark tests for async test suite more accurately

fselmo:
### What was wrong?

- All async tests need the ``@pytest.mark.asyncio`` decorator to be picked up by pytest-asyncio. This already marks them as asyncio tests, so we should use the ``-m`` option to call them. Looking for "async" in the test name with the``-k`` option caused us to miss some tests that we didn't name with "async" in them in the async test suite.

- Note to any reviewers: Be sure to check that the selected number in one test run equals the deselected number in the other.

### Todo:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,8 +20,8 @@ allowlist_externals=/usr/bin/make
 install_command=python -m pip install {opts} {packages}
 usedevelop=True
 commands=
-    core: pytest {posargs:tests/core -m "not asyncio" -k "not async"}
-    core_async: pytest {posargs:tests/core -k "asyncio or async"}
+    core: pytest {posargs:tests/core -m "not asyncio"}
+    core_async: pytest {posargs:tests/core -m asyncio}
     ens: pytest {posargs:tests/ens --ignore=tests/ens/normalization/test_normalize_name_ensip15.py}
     ensip15: pytest {posargs:tests/ens/normalization/test_normalize_name_ensip15.py -q}
     ethpm: pytest {posargs:tests/ethpm}
--- a/tox.ini
+++ b/tox.ini
@@ -20,8 +20,8 @@ allowlist_externals=/usr/bin/make
 install_command=python -m pip install {opts} {packages}
 usedevelop=True
 commands=
-    core: pytest {posargs:tests/core -k "not async"}
-    core_async: pytest {posargs:tests/core -k async}
+    core: pytest {posargs:tests/core -m "not asyncio" -k "not async"}
+    core_async: pytest {posargs:tests/core -k "asyncio or async"}
     ens: pytest {posargs:tests/ens --ignore=tests/ens/normalization/test_normalize_name_ensip15.py}
     ensip15: pytest {posargs:tests/ens/normalization/test_normalize_name_ensip15.py -q}
     ethpm: pytest {posargs:tests/ethpm}
--- a/newsfragments/3180.internal.rst
+++ b/newsfragments/3180.internal.rst
@@ -1,0 +1,1 @@
+Re-define how async vs sync core test suites are ran.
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -20,8 +20,8 @@ allowlist_externals=/usr/bin/make
 install_command=python -m pip install {opts} {packages}
 usedevelop=True
 commands=
-    core: pytest {posargs:tests/core -k "not async"}
-    core_async: pytest {posargs:tests/core -k async}
+    core: pytest {posargs:tests/core -m "not asyncio" -k "not async"}
+    core_async: pytest {posargs:tests/core -k "asyncio or async"}
```
reedsa: Should `asyncio` be part of the mark expression flag? I think this will find any test names containing `asyncio` or `async` (which in turn finds anything that would include `asyncio` I believe).

fselmo: Yep, good catch. I think it should be the opposite of the previous. Will change.

fselmo: I think this approach needs to be revised actually... I don't believe the `-m` and `-k` work well enough together. We may just have to run `asyncio` marked tests and the opposite as: 

- `pytest tests/core -m asyncio`: run tests that are actually asynchronous in nature
- `pytest tests/core -m "not asyncio"`: run all other tests, including those that may involve async classes but aren't asynchronous in nature (e.g. testing some static, non-asynchronous, property of some asynchronous class)

reedsa (approved):
Good to see the marker flag, way more robust than relying on the test name itself.

kclowes:
> Note to any reviewers: Be sure to check that the selected number in one test run equals the deselected number in the other

I looked at the `py310-core` and `py310-core_async` tests, and it doesn't look like the numbers are quite adding up? I could be wrong though 

  | passed | skipped | deselected | xfailed | warnings
-- | -- | -- | -- | -- | --
sync | 2035 | 14 | 676 | 5 | 7
async | 667 | 2 | 2054 | 7 | 3


</body>
</html>

fselmo:
> I looked at the `py310-core` and `py310-core_async` tests, and it doesn't look like the numbers are quite adding up? I could be wrong though

Oh weird. Where is that? I'm seeing the opposite numbers from each other ([sync](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379099?invite=true#step-105-4841_79) vs [async](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379082/steps?invite=true#step-105-4895_79)).

kclowes:
Oh, I see what you're talking about. I think we're good. The numbers I was looking at are at the bottom in the output after the run. [sync](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379099?invite=true#step-105-285202_69), and [async](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379082/steps?invite=true#step-105-112681_69)

kclowes (approved):
LGTM! Much cleaner! 🧹

fselmo:
> Oh, I see what you're talking about. I think we're good. The numbers I was looking at are at the bottom in the output after the run. [sync](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379099?invite=true#step-105-285202_69), and [async](https://app.circleci.com/pipelines/github/ethereum/web3.py/5452/workflows/1db6f4cf-924d-4e8c-973e-cc7c25734286/jobs/379082/steps?invite=true#step-105-112681_69)

ohh I see... yeah the skipped and xfailed ones, etc, skew the numbers a bit or at least make it a bit more confusing to compare 👍🏼

Issue #3184, opened 2024-01-04
Optional parameter in `fee_history` causes problems when omitted


antazoey:
* Version: 6.13.0
* Python: 3.10.13
* OS: macos
* `pip freeze` output

```
<put the output from running pip freeze here>
```

### What was wrong?

We have been doing this in our code:

```python
fee_history = self.web3.eth.fee_history(1, BlockNumber(latest_block_number))
```

However, it now failed with the following error:

```
ValueError: {'code': -32602, 'message': 'Errors encountered in param 2: Invalid value null supplied to : Array<Float number> | undefined', 'data': {'message': 'Errors encountered in param 2: Invalid value null supplied to : Array<Float number> | undefined'}}
```

To fix it on my end, I had to change the line to:

```python
fee_history = self.web3.eth.fee_history(1, BlockNumber(latest_block_number), reward_percentiles=[])
```

### How can it be fixed?

If given `None` , change to `[]` in the code implementation.

---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
please refer to our [documentation](https://web3py.readthedocs.io/en/latest/) and/or join the conversation 
on [discord](https://discord.gg/GHryRvPB84).

antazoey:
Ok, I am realizing this does not happen on normal geth nodes but does happen on Anvil and Hardhat.

antazoey:
maybe the fix is to actually remove the fee `reward_percentiles` parameter from the arguments list if it is null

reedsa:
Closed in https://github.com/ethereum/web3.py/pull/3185

PR #3185, opened 2024-01-04
change fee_history arg reward_percentiles default to pass empty list instead of None

pacrob:
### What was wrong?

`fee_history`s optional `reward_percentiles` errors when omitted for Anvil and Hardhat providers.

Related to Issue #3184 

### How was it fixed?

Changed the default value that is passed to the provider from `None` to an empty list.

Manually tested with Anvil to confirm it works. The other option is to remove the `reward_percentiles` arg completely if it's `None` and only call `fee_history` with `block_count` and `newest_block`. That also works locally, but requires more code to be changed.

In looking through the geth and anvil implementations (with some chatgpt assistance), it looks like it's a difference in how they handle the `None` default. Geth accepts it and moves on, while Anvil will only apply its default empty vector if the value is missing. So defaulting to an empty list seems appropriate on our end.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/eth/async_eth.py
+++ b/web3/eth/async_eth.py
@@ -234,6 +234,7 @@ class AsyncEth(BaseEth):
         newest_block: Union[BlockParams, BlockNumber],
         reward_percentiles: Optional[List[float]] = None,
     ) -> FeeHistory:
+        reward_percentiles = reward_percentiles or []
         return await self._fee_history(block_count, newest_block, reward_percentiles)
 
     # eth_call
--- a/web3/eth/eth.py
+++ b/web3/eth/eth.py
@@ -226,6 +226,7 @@ class Eth(BaseEth):
         newest_block: Union[BlockParams, BlockNumber],
         reward_percentiles: Optional[List[float]] = None,
     ) -> FeeHistory:
+        reward_percentiles = reward_percentiles or []
         return self._fee_history(block_count, newest_block, reward_percentiles)
 
     # eth_call
--- a/newsfragments/3185.bugfix.rst
+++ b/newsfragments/3185.bugfix.rst
@@ -1,0 +1,1 @@
+Change ``fee_history`` default behavior. If ``reward_percentiles`` arg not included, pass it to the provider as an empty list instead of ``None``.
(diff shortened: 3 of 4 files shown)


reedsa (commented):
The change looks good here, I would be curious if testing for the default and non-default values would be possible. Given the disparity between geth and other implementations maybe not, but curious if you have ideas there.

web3/eth/async_eth.py
```diff
@@ -232,7 +232,7 @@ async def fee_history(
         self,
         block_count: int,
         newest_block: Union[BlockParams, BlockNumber],
-        reward_percentiles: Optional[List[float]] = None,
+        reward_percentiles: Optional[List[float]] = [],
```
antazoey: you may want to be careful defaulting to a mutable variable in a function signature!

https://softwareengineering.stackexchange.com/questions/157373/python-mutable-default-argument-why

web3/eth/eth.py
```diff
@@ -224,7 +224,7 @@ def fee_history(
         self,
         block_count: int,
         newest_block: Union[BlockParams, BlockNumber],
-        reward_percentiles: Optional[List[float]] = None,
+        reward_percentiles: Optional[List[float]] = [],
```
antazoey: instead, you can do `reward_percentiles = reward_percentiles or []` as the first line in the function's implementation

pacrob: Thanks for the heads up! Updated.

pacrob:
> The change looks good here, I would be curious if testing for the default and non-default values would be possible. 

There are tests for `fee_history` both with and without the `reward_percentiles` argument already. I can't think of any way to test it further. At least not without adding testing against an Anvil, etc. instance.

antazoey:
> There are tests for `fee_history` both with and without the `reward_percentiles` argument already. I can't think of any way to test it further. At least not without adding testing against an Anvil, etc. instance.

Don't be afraid to use mocks in unit tests: https://github.com/ApeWorX/ape/blob/main/tests/functional/geth/test_provider.py#L347-L360

You don't need to spin up a whole node just to have the code hit a certain condition!

antazoey (approved):

reedsa (approved):
Nice mocks and tests, thanks for the effort there! Lgtm

PR #3189, opened 2024-01-10
Remove unneeded type ignores

fselmo:
### What was wrong?

- Some type ignores trickled through as unnecessary, likely related to a recent `eth-abi` update.

### How was it fixed?

- Removed type ignores

#### Cute Animal Picture

\ /
O ====
 /\        /\ /\ /\


pacrob (approved):
gtg!

PR #3192, opened 2024-01-18
Fix types - AsyncContractConstructor

Nov1kov:
### What was wrong?

IDE shows warning 
image

### How was it fixed?

There is wrong return type.

### Todo:

#### Cute Animal Picture

image


pacrob:
Thanks for catching this, @Nov1kov!

PR #3193, opened 2024-01-18
Validate geth version in test fixture against binary used for tests

fselmo:
### What was wrong?

We've had instances where we forget to update the `geth` version being used in the circleCI runs and this goes unnoticed since two versions can be compatible with each other. This means we create the fixture with one version, but run it with another - and, as long as no noticeable changes have happened, this is silently unintended.

### How was it fixed?

- Validate the geth binary version being used to run the tests against the version in the generated zip file for the test fixture. This version is automatically named in the zip file to be the geth version used to create the zip so it should be a decent enough check.

### Todo:

### DIFF
--- a/newsfragments/3193.internal.rst
+++ b/newsfragments/3193.internal.rst
@@ -1,0 +1,1 @@
+Validate geth version used to generate the integration test fixture against the version in the binary that is used to run the tests.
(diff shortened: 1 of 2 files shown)


reedsa (approved):
Looks good.

As a follow up, we might consider updating the version as part of the fixture generation script so we don't have to remember in the future.

pacrob (approved):
lgtm!

Issue #1002, opened 2018-08-13
Clearer error on contract execution error during call()


carver:
### What was wrong?

When a contract error happens during a call, usually we let the eth-abi exception raise through. This makes users think that web3.py is broken. eg~ https://github.com/ethereum/web3.py/issues/185

### How can it be fixed?

After checking that there is code at the contract address, raise a `ContractError` indicating that something went wrong in the execution of the contract, so it didn't return any value at all.

Ashhar-24:
Hey @carver could you please help me know how should this issue be resolved. Like which file in the code base should be changed and it's directive. 
Actually I'm new to open source and currently I'm facing issues understanding the code base.
Thanks for your cooperation :)

pacrob:
Hey @Ashhar-24 , thanks for taking a look! The first step would be to read through the referenced issue #185 to understand the background behind this issue. Then create some different test cases based on that discussion and see what error messages you get. From there, you should be able to search the codebase for where the new logic would belong. `web3/contract/utils.py -> call_contract_function` might be it, but I'd need to look deeper to be sure.

MRLab12:
@carver @pacrob I was working on this and have a test to trigger the errors discussed in [#185](https://github.com/ethereum/web3.py/issues/185) to handle cases where the contract call returns unexpected or empty data. If no one else is working with this issue, I can take it on.

MRLab12:
Quick update here - I was able to replicate the `eth_abi.exceptions.InsufficientDataBytes` and `web3.exceptions.BadFunctionCallOutput` errors as described in the reference ticket. Working on handling the `ContractError` error. I can start a PR since the contributing guide recommends starting a PR early on development to start a discussion.

kclowes:
Thanks @MRLab12! Opening a PR would be great!

Mridul-Dev123:
I would like to work on this issue

albin-george-kurian:
Hi @fselmo,
Is this still available? If so, can I work on it.

PR #3194, opened 2024-01-19
#1002 Handle error when interacting with a bad contract

MRLab12:
### What was wrong?
When a contract error happens during a call, usually we let the eth-abi exception raise through. This makes users think that web3.py is broken.

Related to Issue [#1002](https://github.com/ethereum/web3.py/issues/1002)

### How was it fixed?
A `ContractLogicError` is now raised in `call_contract_function` when a contract has no code.

### Todo:


MRLab12:
Created this draft PR to initiate a conversation on this fix. 

I was able to reproduce the error when connecting to a Base Sepolia test wallet and setting up an invalid contract address. In order to not have the test actually connect to the network, I have mocked the functionalities. 

This is my first contribution to this repo so any suggestions on contributing and code reviewing are welcome!

MRLab12:
The `test_call_undeployed_contract` was failing and I noticed this was returning the `eth_abi.exceptions.InsufficientDataBytes` and `web3.exceptions.BadFunctionCallOutput` errors, so maybe we just keep this one instead of the one I added.

kclowes:
@MRLab12 Thanks for the PR! I haven't had a chance to look at this yet, but I hope to be able to check it out in the next few days.

Issue #961, opened 2018-07-16
Add API to iterate through all events in a contract


carver:
### What was wrong?

No easy way to get all events for a contract (while still parsing the results). See [this StackExchange question](https://ethereum.stackexchange.com/questions/54473/how-to-read-allevents-using-python-web3-theres-capability-in-web3-js). One option is to iterate over all the events, but it's a bit awkward right now. I think the easiest way is:

```py
    from web3.contract import ContractEvent
    
    filters = [
      event.createFilter(fromBlock='latest')
      for event in myContract.events
      if isinstance(event, ContractEvent)
    ]
```

### How can it be fixed?

Some options: 
- Implement `__iter__` on `Contract.events` to iterate through all events in the ABI (my favorite option, except that it's inconsistent with `contract.functions`, which is doing the wrong thing IMO)
- Add a new `Contract.all_events()` equivalent to `Contract.all_functions()`

Then the example changes to:

```py
    filters = [
      event.createFilter(fromBlock='latest')
      for event in myContract.events
    ]
```

---

Of course, we could also implement `contract.create_filter()` like web3.js's `contract.events.allEvents`. I kind of like that the filters are event specific right now, though. I don't think it's too big a deal to require callers to write a filter loop on events.

dylanjw:
+1 for implementing `__iter__`.  I need to think about `all_events`.  While I see the use, it would be a simple filter with just the contract address, and not allow one to filter on arguments or topics.  It might be better to just steer users towards instantiating filter object with just the contract address rather than providing a helper function to do the same.

dylanjw:
Maybe we could publish a recipe on combining filters into a single object that can be polled.

carver:
>  It might be better to just steer users towards instantiating filter object with just the contract address rather than providing a helper function to do the same.

The key difference is that they want the automatic argument decoding, which they don't get with the generic filter on an address. I like the recipe idea.

carver:
> Is there something specific about that implementation that you would like to see changed?

Yeah, it's only returning the name (`str`) of the function. I think it should return the full `ContractFunction` object.

HarryR:
Wow, this ticket is still open! Here is a snippet that emits all decode-able events for a specific contract.

```python

from web3.contract import Contract
from web3._utils.events import get_event_data
from eth_utils import event_abi_to_log_topic


def fetch_events(contract: Contract, fromBlock, toBlock=None):
    """
[...]

PR #3200, opened 2024-01-23
Return function in contract function iterator

reedsa:
### What was wrong?

Related to Issue #961

### How was it fixed?

`contract.functions` yields `ContractFunction` rather than just the `str` name.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_contract_call_interface.py
+++ b/tests/core/contracts/test_contract_call_interface.py
@@ -30,6 +30,12 @@ from web3._utils.contract_sources.contract_data.bytes_contracts import (
 from web3._utils.ens import (
     contract_ens_addresses,
 )
+from web3.contract.async_contract import (
+    AsyncContractFunction,
+)
+from web3.contract.contract import (
+    ContractFunction,
+)
 from web3.exceptions import (
     BadFunctionCallOutput,
     BlockNumberOutofRange,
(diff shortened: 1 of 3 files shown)


newsfragments/3200.feature.rst
```diff
@@ -0,0 +1 @@
+Return iterable of `ABIFunction`s from the `BaseContractFunctions` iterator.
```
fselmo: We should label this as `3200.breaking.rst` since it breaks the current behavior.

fselmo (commented):
I like this change. Adding a test for it in this PR would go a long way to assert the functionality doesn't break in the future. Are you up for that? 👀

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -2313,3 +2325,12 @@ async def test_async_changing_default_block_identifier(async_w3, async_math_cont
     assert (
         await async_math_contract.functions.counter().call(block_identifier=None) == 7
     )
+
+
+def test_async_functions_iterator(async_w3, async_math_contract):
```
reedsa, line 2333: No async stuff happening in here, is it worth having this separate test?

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -1149,6 +1152,15 @@ def test_changing_default_block_identifier(w3, math_contract):
     assert math_contract.functions.counter().call(block_identifier=None) == 7
 
 
+def test_functions_iterator(w3, math_contract):
+    all_functions = math_contract.all_functions()
+    functions_iter = math_contract.functions
+
+    for fn, expected_fn in zip(iter(functions_iter), all_functions):
+        assert isinstance(fn, BaseContractFunction)
```
reedsa: Want to check this type is correct, or is there a better one I may have missed?

pacrob: Since the tests you have are for sync and async providers, not just an instance of BaseProvider, you could `assert isinstance(fn, ContractFunction)` here and `AsyncContractFunction` below and keep both tests. Not sure it's necessary, but checks that the class inherits properly from base :shrug:

pacrob (approved):
lgtm!

tests/core/contracts/test_contract_call_interface.py
pacrob: It has the Async type now. Resolving.

kclowes (approved):
🚢 !

fselmo (approved):
lgtm 👍🏼

PR #3202, opened 2024-01-25
Raise exceptions by default; opt to silence in message listener task

fselmo:
### What was wrong?

It is often best to fail hard and fast and be able to silence exceptions rather than let exceptions possibly go unnoticed. We were defaulting to silencing exceptions in the message listener task for `WebsocketProviderV2`. This hasn't been released yet and so this change just aims to tweak the configuration and opt to silence exceptions that are raised by default.

### How was it fixed?

- `raise_listener_task_exceptions` -> `silence_listener_task_exceptions`
- Modify tests accordingly
- Close all current tasks in the loop before raising the exception

### Todo:

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -186,6 +186,7 @@ The Manager acts as a gatekeeper for the request/response lifecycle.  It is
 unlikely that you will need to change the Manager as most functionality can be
 implemented in the Middleware layer.
 
+.. _internals__persistent_connection_providers:
 
 Request Processing for Persistent Connection Providers
 ------------------------------------------------------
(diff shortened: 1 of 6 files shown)


pacrob:
Update looks good, but I don't see any reference to this option in docs. Is that coming separately, or should be added here?

pacrob (approved):
lgtm!

fselmo:
@pacrob I added a more extensive update of the docs here since it was stale. Thanks for the poke. Take a look again and lmk if anything looks off. I also took the `endpoint_uri` out of the base ``PersistentConnectionProvider`` class, opting to only keep it in the inheriting classes which are the actual provider classes.

pacrob (approved):
lgtm!

Issue #3212, opened 2024-01-29
Waiting for a transaction fails on Geth 1.13.11


antazoey:
* Version: 6.15.0
* Python: 3.10
* OS: macos
* `pip freeze` output

```
<put the output from running pip freeze here>
```

### What was wrong?

The line:

```python
tx = web3.eth.wait_for_transaction_receipt(hex_hash)
```

has issues on my local testing Geth node (inspired from geth --dev).
So I run into this issue if I check too fast:

```
ValueError {"message": "(-32000) transaction indexing is in progress"}
```

### How can it be fixed?

I think raising + catching a custom error for this situation.
Like it should still wait for the timeout if we are indexing.
Right now, it is only looking for "tx not found" error.
---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
please refer to our [documentation](https://web3py.readthedocs.io/en/latest/) and/or join the conversation 
on [discord](https://discord.gg/GHryRvPB84).

antazoey:
Realizing this is likely related to a recent Geth change: https://github.com/ethereum/go-ethereum/issues/28877

rjl493456442:
The question is: how does web3.py handle receipt waiting?

My hunch is that there is an internal loop that queries the receipts repeatedly until they are available.

Originally, eth_getTransactionReceipt would return null if the receipt was not available yet. After the implementation of https://github.com/ethereum/go-ethereum/pull/28703, eth_getTransactionReceipt returns an error if the receipt is not available, and transaction indexing is not finished yet (note that transaction indexing only occurs during the initial sync stage).

I suspect that the new behavior breaks the logic in web3.py, which immediately returns when the error is returned by the GETH RPC.

fselmo:
closed by #3216 and #3217

PR #3217, opened 2024-02-01
[v6] Catch newly raised geth errors when tx indexing in progress

fselmo:
### What was wrong?

- Geth ``1.13.11`` started to return an error when transaction indexing is in progress and a ``eth_getTransactionReceipt`` call is made.

Related to Issue #3212 

### How was it fixed?

- Handle this in the ``wait_for_transaction_receipt`` method by catching the error in order to continue waiting.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/tests/core/eth-module/test_transactions.py
+++ b/tests/core/eth-module/test_transactions.py
@@ -24,6 +24,9 @@ from web3.exceptions import (
     Web3ValidationError,
 )
 from web3.middleware import (
+    async_construct_error_generator_middleware,
+    async_construct_result_generator_middleware,
+    construct_error_generator_middleware,
     construct_result_generator_middleware,
 )
 from web3.middleware.simulate_unmined_transaction import (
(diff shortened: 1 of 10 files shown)


tests/core/eth-module/test_transactions.py
```diff
@@ -193,6 +206,28 @@ def test_unmined_transaction_wait_for_receipt(w3):
...
 
 
+def test_eth_wait_for_transaction_receipt_transaction_indexing_in_progress(w3):
+    result_middleware = construct_result_generator_middleware(
+        {RPC.eth_getTransactionReceipt: lambda *_: {"status": 1}}
+    )
+    error_middleware = construct_error_generator_middleware(
+        {
+            RPC.eth_getTransactionReceipt: lambda *_: (
+                _swap_error_middleware_for_result_middleware_and_return_error(
```
reedsa, line 216: Super funky but this seems like a good way to test the error/response.

tests/core/eth-module/test_transactions.py
```diff
@@ -193,6 +206,28 @@ def test_unmined_transaction_wait_for_receipt(w3):
...
+                )
+            )
+        }
+    )
+    w3.middleware_onion.inject(error_middleware, "error_middleware", layer=0)
+
+    receipt = w3.eth.wait_for_transaction_receipt(f"0x{'00' * 32}")
+    assert receipt == {"status": 1}
+
+    assert "error_middleware" not in w3.middleware_onion.middlewares
```
reedsa, line 227: Nit: This seems like it would always be true given the expected receipt came back. Curious if you have thoughts for this assertion?

reedsa (approved):
Minor nit, not blocking. lgtm!

tests/core/eth-module/test_transactions.py
fselmo: This is just to add an extra layer of assertion for better code readability. Sanity check that the `error_middleware` is getting cleaned up when we add the `result_middleware` which is why we don't remove it from the onion below.

kclowes (approved):
LGTM! :shipit:

PR #3225, opened 2024-02-06
Replace legacy websocket provider with V2

fselmo:
### What was wrong?

- Replace the legacy WebSocket provider with V2.

We wanted to get these changes in before the first release of `v7` beta.

### How was it fixed?

- ``WebsocketProvider`` -> ``LegacyWebSocketProvider``
- ``WebsocketProviderV2`` -> ``WebSocketProvider``
- Update documentation to reflect the above changes

### Todo:


docs/quickstart.rst
```diff
@@ -102,7 +104,7 @@ You can connect to a remote node by specifying the endpoint, just like the previ
 
    >>> w3 = AsyncWeb3(AsyncWeb3.AsyncHTTPProvider('https://<your-provider-url>'))
 
-   >>> w3 = Web3(Web3.WebsocketProvider('wss://<your-provider-url>'))
+   >>> w3 = await AsyncWeb3(AsyncWeb3.WebsocketProvider('wss://<your-provider-url>'))
```
kclowes: There is a very good chance this is just me not knowing how to use the API, but I always get an error: `TypeError: object AsyncWeb3 can't be used in 'await' expression` when I run this. Heres the code for reference: 
```
>>> from web3 import AsyncWeb3
>>> w3 = await AsyncWeb3(
>>>     AsyncWeb3.WebsocketProvider('ws://localhost:8546')
>>> )
```

kclowes (approved):
Just putting this here so we don't forget... we need to add a deprecation warning (or some type of warning) in v6 on the `WebsocketProvider` to let people know about the change from `WebsocketProvider` -> `LegacyWebsocketProvider`, and provide the legacy alias. I'm also happy to add an issue, just let me know!

Looks good! I think there are a couple spots in the Quickstart docs where the new WebsocketProvider API doesn't work quite like the docs say it does, but that may just be user error :)

web3/providers/websocket/websocket.py
```diff
@@ -2,134 +2,243 @@
...
+from toolz import (
+    merge,
+)
 from websockets.client import (
     connect,
 )
-from websockets.legacy.client import (
-    WebSocketClientProtocol,
+from websockets.exceptions import (
+    WebSocketException,
```
pacrob: consistency nit - All the Providers are `Websocket...`, but the Exception (and a couple other places I found in search) use `WebSocket...` Probably good to make them all `Websocket`, yah?

pacrob:
The organization of providers feels unintuitive:

image

Maybe org the HTTP Providers together? Do we have future `PersistentConnectionProvider`s planned that would merit it staying as an umbrella term? If not, I feel like it's good to have the names of the actual providers users will be implementing show up first in docs. Can then include deets about `PersistentConnectionProvider` in more detailed description.

Just thinking out loud. What'chu think?

kclowes:
> The organization of providers feels unintuitive: ...

Yeah, we don't neatly split between Sync and Async providers any more. We really have Sync Providers, Async Providers, and Persistent Connection Providers, IMHO. Might be good to reflect that in the docs

docs/quickstart.rst
fselmo: Nope, you're right. This sparked a whole discussion but to note here:

We can completely remove the need for `persistent_websocket()` by implementing the magic methods in `AsyncWeb3` itself. Those changes will come in here soon.

ethpm/backends/registry.py
```diff
@@ -50,10 +50,10 @@ class RegistryURIBackend(BaseURIBackend):
     def __init__(self) -> None:
         from web3 import (
             Web3,
-            WebsocketProvider,
+            WebsocketProviderLegacy,
```
reedsa: What is required to make use of `WebsocketProvider` in all examples?

reedsa (approved):

ethpm/backends/registry.py
fselmo: Good thought. `ethpm` is deprecated though and being removed in `v7` so I don't think it will be a good time investment here. If you see this somewhere else in the tests that would be good to note.

reedsa (approved):
lgtm!

pacrob (approved):
More of a nit, but it's still bugging me that everything we get from the `websockets` lib uses capital S `WebSocket` and we use mostly `Websocket`. But as long as ya'll are good with it, lgtm :)

fselmo:
> More of a nit, but it's still bugging me that everything we get from the `websockets` lib uses capital S `WebSocket` and we use mostly `Websocket`. But as long as ya'll are good with it, lgtm :)

Being that the protocol is WebSocket, let's make it happen since we're already breaking everything. Updated in the last commit.

PR #3227, opened 2024-02-06
``CallOverride`` -> ``StateOverride``

### LINKED ISSUE #3175: Type name change: ``CallOverride*`` -> ``StateOverride*``
### What was wrong?

We currently will be using the type `CallOverride` and `CallOverrideParams` for *eth_estimateGas*, since this was added to the API recently. Now that this param applies to more methods, these types can arguably be better named `StateOverride` and `StateOverrideParams`, respectively.

### How can it be fixed?

- `CallOverrideParams` -> `StateOverrideParams`
- `CallOverride` -> `StateOverride`

This is a breaking change and so should consider this for `v7` branch.

fselmo:
### What was wrong?

closes #3175 

### How was it fixed?

State override is used in more requests than just ``eth_call`` (see recent updated to `eth_estimateGas`). We should use a more generic name that suits this type more. ``StateOverride`` seems the most appropriate here.

### Todo:


web3/eth/eth.py
```diff
@@ -619,15 +619,15 @@ def modify_transaction(
...
-    filter: Method[Callable[[Optional[Union[str, FilterParams, HexStr]]], Filter]] = (
-        Method(
-            method_choice_depends_on_args=select_filter_method(
-                if_new_block_filter=RPC.eth_newBlockFilter,
-                if_new_pending_transaction_filter=RPC.eth_newPendingTransactionFilter,
-                if_new_filter=RPC.eth_newFilter,
-            ),
-            mungers=[BaseEth.filter_munger],
-        )
+    filter: Method[
```
fselmo: My `black` has been reformatting some things lately locally. Anyone else having this issue? Not seeing why this needed to change.

fselmo: nm fixed... I must have conflicting versions

kclowes (approved):
Looks good to me! Maybe add a snippet to the v7 migration guide? I think this is technically breaking

pacrob (approved):
lgtm!

fselmo:
> Looks good to me! Maybe add a snippet to the v7 migration guide? I think this is technically breaking

Yep, good remembering. Definitely breaking. I added the right newsfragment for that just didn't update the migration guide. Will do that and merge 👍🏼

PR #3228, opened 2024-02-08
Contract Events Sorted by `ContractEvent.logIndex` and `ContractEvent.blockNumber`

### LINKED ISSUE #961: Add API to iterate through all events in a contract

reedsa:
### What was wrong?

Closes #961

### How was it fixed?

Pass log entries list through `sorted` to return a list sorted by all event's `blockNumber` and `logIndex`es

### Todo:

#### Cute Animal Picture

Screen Shot 2024-02-08 at 3 09 27 PM

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -944,6 +944,8 @@ For example:
 
    Fetches all logs for a given event within the specified block range or block hash.
 
+   Returns a list of decoded event logs sorted by ``logIndex``.
+
     ``argument_filters`` is an optional dictionary argument that can be used to filter
     for logs where the event's argument values match the values provided in the
     dictionary. The keys must match the event argument names as they exist in the ABI.
--- a/newsfragments/3228.feature.rst
+++ b/newsfragments/3228.feature.rst
@@ -1,0 +1,1 @@
+Contract event ``get_logs`` results sorted by each ``ContractEvent`` ``logIndex``.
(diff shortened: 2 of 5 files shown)


kclowes (approved):
LGTM! Do we need to update any docs?

reedsa:
Committed a change to also sort the events by `blockNumber`, I think that makes more sense than just sorting by `logIndex`.

kclowes (approved):
sweet!

PR #3229, opened 2024-02-12
Move newsfragment to newsfragment directory

kclowes:
### What was wrong?
There was a newsfragment sitting in the root. I don't think it's been released yet, so it can just get moved into the newsfragment file.

### How was it fixed?
Moved it!

#### Cute Animal Picture


reedsa (approved):
Lgtm

PR #3238, opened 2024-02-21
[V6] Deprecate geth miner namespace

### LINKED ISSUE #2856: Deprecate `miner` namespace support and tests

reedsa:
### What was wrong?

Closes #2856

Geth removed support for their [miner API](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-miner).

### How was it fixed?

Change `GethMiner` methods to `DeprecatedMethod`.

### Todo:

#### Cute Animal Picture


pacrob (approved):
lgtm!

tests/integration/generate_fixtures/common.py
```diff
@@ -210,7 +211,8 @@ def get_process(run_command):
 
 
 def mine_block(w3):
-    origin_block_number = w3.eth.block_number
+    with pytest.warns(DeprecationWarning):
```
kclowes: Do we need this here? It seems like this is just used for the fixture generation so you'd only see it if you were generating the fixture?

tests/integration/generate_fixtures/go_ethereum.py
```diff
@@ -189,7 +190,8 @@ def verify_chain_state(w3, chain_data):
 
 
 def mine_block(w3):
-    origin_block_number = w3.eth.block_number
+    with pytest.warns(DeprecationWarning):
```
kclowes: Same question here

kclowes (approved):
Looks good to me! I left a question around the `DeprecationWarning` in the fixture generation, but looks good otherwise.

kclowes:
I also had to add `docs/web3.providers.rpc.rst` to the `exclude_patterns` in `docs/conf.py` but that is likely just me since the doc build is passing 🤷

web3/method.py
```diff
@@ -214,17 +214,25 @@ def process_params(self, module: "Module", *args: Any, **kwargs: Any) -> Tuple[
...
 
     def __get__(
         self, obj: Optional["Module"] = None, obj_type: Optional[Type["Module"]] = None
     ) -> Any:
-        warnings.warn(
-            f"{self.old_name} is deprecated in favor of {self.new_name}",
-            category=DeprecationWarning,
-        )
+        if self.msg is not None:
+            warnings.warn(self.msg)
```
reedsa: @kclowes should this include the `category=DeprecationWarning` argument as well?

Issue #2911, opened 2023-04-11
`web3.providers.ipc.get_default_ipc_path()` returned `None` instead of `str`


antazoey:
* Version: 6.0.0
* Python: 3.10
* OS: macky whacky
* `pip freeze` output

```
will provide if needed
```

### What was wrong?

The method `get_deafult_ipc_path()` located in `web3.providers.ipc` returned `None` even though its return type is documented as `str`.

I am not sure why it returned `None` in my case, probably forgot some flag in `geth`.

Edit: it returned None in my case because i was checking it before starting geth, so it didnt exist yet.

### How can it be fixed?

Either change type to `Optional[str]` or raise an error instead of returning `None` ?

---
**Note:** We prefer to use issues to track our work. If you think you've encountered a bug in web3py or 
have a feature request, you're in the right place. If you have implementation or usage questions, 
please refer to our [documentation](https://web3py.readthedocs.io/en/latest/) and/or join the conversation 
on [discord](https://discord.gg/GHryRvPB84).

unparalleled-js:
For us, it is most useful to have this path even if it does not currently exist yet, since it may not exist at config time but may exist before starting geth

kclowes:
> OS: macky whacky

This is cracking me up 😆 

It would be a small breaking change to return the path  even if it doesn't exist, but I changed the type and made the `return None` explicit in #2917. I don't see a problem with returning the path whether or not it exists, so I'll add this issue to our [v7 megathread](https://github.com/ethereum/web3.py/issues/2854)

Issue #2854, opened 2023-02-27
``v7`` breaking changes megathread


fselmo:
Main thread for tracking breaking changes slated for the next major version release of web3.py - `v7`. This will be an evolving issue much like #1416 was for `v6`.

Candidates:

```[tasklist]
### Tasks
- [ ] https://github.com/ethereum/web3.py/issues/3188
- [ ] https://github.com/ethereum/web3.py/pull/2857
- [ ] #2685
- [ ] #2816
- [ ] #2849
- [ ] https://github.com/ethereum/web3.py/issues/3085
- [ ] https://github.com/ethereum/web3.py/pull/2964
- [ ] https://github.com/ethereum/web3.py/issues/3108
- [ ] https://github.com/ethereum/web3.py/issues/715
- [ ] https://github.com/ethereum/web3.py/issues/3115
- [ ] https://github.com/ethereum/web3.py/issues/901
- [ ] https://github.com/ethereum/web3.py/issues/3119
- [ ] https://github.com/ethereum/web3.py/issues/3175
- [ ] https://github.com/ethereum/web3.py/issues/3164
- [ ] https://github.com/ethereum/web3.py/issues/2844
- [ ] https://github.com/ethereum/web3.py/issues/3032
- [ ] #2884
- [ ] https://github.com/ethereum/web3.py/issues/2995
- [ ] https://github.com/ethereum/web3.py/issues/3021
```

kclowes:
Return ipc path even if it doesn't yet exist. See #2911

pacrob:
Remove the `ethPM` module, deprecated in #2953

kclowes:
#2964 - Change provider User-Agent header to be more readable.

kclowes:
Better UX for provider-specific middlewares. See: https://github.com/ethereum/web3.py/issues/2770

edit: issue created (#2990)

reedsa:
Remove the miner API, deprecated in #2856.

pacrob:
Here, then upstream libs:
 - Remove python 3.7 support
 - Branch name change: master -> main

pacrob:
`AttributeDict` updates #3105

pacrob:
web3 v7 should set `hexbytes>=1.0.0`.

Bottom pins for our other libs should move up to at least the release that dropped py37 support

kclowes:
Where possible, have all errors thrown inherit from `Web3Exception`, or `ENSException`. See #1864.

[...]

PR #2917, opened 2023-04-13
Fix typing for get_ipc_path and get_dev_ipc_path

kclowes:
### What was wrong?
The typing was incorrect on `get_ipc_path` and `get_dev_ipc_path`. 

Closes #2911 

### How was it fixed?
Added the correct typing and explicit `None` returns where needed. 

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3245, opened 2024-02-22
return default ipc path whether it exists or not

pacrob:
### What was wrong?

Related to Issue #2911 
Related to Issue #2854

### How was it fixed?

Remove check to see if the `.ipc` file exists and just return the default value. Add tests.
Also, the default path for `get_default_ipc_path` was fixed in #3060, but doesn't look like it was for `get_dev_ipc_path`, so updated that too.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -14,7 +14,6 @@ from types import (
 )
 from typing import (
     Any,
-    Optional,
     Type,
     Union,
 )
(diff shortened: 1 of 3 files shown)


web3/providers/ipc.py
```diff
@@ -91,21 +91,15 @@ def get_default_ipc_path() -> Optional[str]:
         ipc_path = os.path.expanduser(
             os.path.join("~", "Library", "Ethereum", "geth.ipc")
         )
-        if os.path.exists(ipc_path):
-            return ipc_path
-        return None
```
reedsa, line 96: I noticed in https://github.com/ethereum/web3.py/pull/2917/files the `return None` was added to make typing work. I think we can remove the `Optional` from the return type now that `None` is being removed.

web3/providers/ipc.py
```diff
@@ -91,21 +91,15 @@ def get_default_ipc_path() -> Optional[str]:
         ipc_path = os.path.expanduser(
```
reedsa: Nit: Since these `ipc_path` vars are not used anywhere else, just return in a single line.
```suggestion
        return os.path.expanduser(
```

reedsa (commented):
What do you think about adding a parameterized test to check the paths are correctly returned from `get_default_ipc_path()` and `get_dev_ipc_path()`? I think `sys.platform` can be mocked to make that possible.

reedsa (approved):
lgtm!

reedsa (changes requested):
My apologies, I did not mean to approve yet. Let me know what you think about testing those functions.

reedsa (approved):
This looks good to me, nice additions! 👍

PR #3248, opened 2024-02-23
Update `eth-utils` to v4.0.0

### LINKED ISSUE #715: Web3.isAddress doesn't work for non prefixed checksumed values
* Version: 4.0.0b11
* Python: 3.6
* OS: linux

### What was wrong?

As stated in the docs http://web3py.readthedocs.io/en/latest/overview.html#Web3.isAddress the function Web3.isAddress(value) should **allow both 0x prefixed and non prefixed values**.

If the address is not checksumed, it's ok not to have the **0x**:

```
>>> Web3.isAddress('d3cda913deb6f67967b99d67acdfa1712c293601')
>>> True
```

But if it's checksumed

```
>>> Web3.isAddress('d3CdA913deB6f67967B99D67aCDFa1712C293601')
>>> False
```

No problem if we add the **0x**:

```
>>> Web3.isAddress('0xd3CdA913deB6f67967B99D67aCDFa1712C293601')
>>> True
```

### How can it be fixed?

Changing the documentation to state that checksumed addresses must have 0x or changing the function to accept checksumed addresses with 0x. I would just remove 0x at the beginning of the function (if found) and work with the address as that.

reedsa:
### What was wrong?

Closes #715

### How was it fixed?

Major version bump of `eth-utils` to make `Web3.is_address` return True for non-checksummed addresses.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-02-23 at 9 08 36 AM

### DIFF
--- a/newsfragments/3248.misc.rst
+++ b/newsfragments/3248.misc.rst
@@ -1,0 +1,1 @@
+Upgrade ``eth-utils`` to ``v4.0.0``.
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import (
 
 extras_require = {
     "tester": [
-        "eth-tester[py-evm]==v0.10.0-b.1",
+        "eth-tester[py-evm]==v0.10.0-b.3",
         "py-geth>=4.1.0",
     ],
     "linter": [
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setup(
         "eth-account>=0.8.0",
         "eth-hash[pycryptodome]>=0.5.1",
         "eth-typing>=3.0.0",
-        "eth-utils>=2.1.0",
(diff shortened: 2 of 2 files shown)


pacrob:
I think your newsfragments got crossed :grin:

pacrob (approved):
lgtm, w/ an updated newsfragment!

fselmo (approved):

newsfragments/3248.misc.rst
```diff
@@ -0,0 +1 @@
+Upgrade ``eth-utils`` to ``v4.0.0``.
```
reedsa, line 1: @pacrob was this more what you were thinking?

pacrob: It's fine. It previously had something about checksum addresses, but no more!

Issue #1441, opened 2019-08-30
processReceipt() returns  "Tried to read 96 bytes.  Only got 68 bytes"


natonamco:
* Version: x.x.x
* Python: 2.7/3.4/3.5
* OS: osx/linux/win
* `pip freeze` output
```
alabaster==0.7.12
anaconda-client==1.7.2
anaconda-navigator==1.9.7
anaconda-project==0.8.2
appnope==0.1.0
appscript==1.0.1
argh==0.26.2
asn1crypto==0.24.0
astroid==2.2.5
astropy==3.1.2
atomicwrites==1.3.0
attrdict==2.0.1
attrs==19.1.0
Babel==2.6.0
backcall==0.1.0
backports.os==0.1.1
backports.shutil-get-terminal-size==1.0.0
beautifulsoup4==4.7.1
bitarray==0.8.3
bitcoin==1.1.42
bkcharts==0.2
bleach==3.1.0
bokeh==1.0.4
boto==2.49.0
Bottleneck==1.2.1
certifi==2019.3.9
cffi==1.12.2
chardet==3.0.4
Click==7.0
cloudpickle==0.8.0
clyent==1.2.2
colorama==0.4.1
conda==4.6.11
conda-build==3.17.8
conda-verify==3.1.1
contextlib2==0.5.5
cryptography==2.6.1
cycler==0.10.0
Cython==0.29.6
cytoolz==0.9.0.1
dask==1.1.4
dataclasses-json==0.2.14
decorator==4.4.0
defusedxml==0.5.0
distributed==1.26.0
docutils==0.14
entrypoints==0.3
et-xmlfile==1.0.1
eth-abi==1.3.0
eth-account==0.3.0
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.1.2
eth-tester==0.1.0b39
eth-testrpc==1.3.0
eth-typing==2.1.0
eth-utils==1.6.4
ethereum==1.6.1
fastcache==1.0.2
filelock==3.0.10
Flask==1.0.2
future==0.17.1
gevent==1.4.0
glob2==0.6
gmpy2==2.0.8
gql==0.1.0
gql-next==0.1.1
graphql-core==2.1
GraphQL-core-next==1.0.5
greenlet==0.4.15
h5py==2.9.0
heapdict==1.0.0
hexbytes==0.2.0
html5lib==1.0.1
idna==2.6
imageio==2.5.0
[...]

Issue #2489, opened 2022-05-27
Allow `get_event_data` to handle invalid padding bytes


cwelton:
* Version: 4.7.2
* Python: 3.9.11
* OS: osx/linux
* `pip freeze` output

```
attrdict==2.0.1
certifi==2022.5.18.1
charset-normalizer==2.0.12
cytoolz==0.11.2
eth-abi==1.3.0
eth-account==0.3.0
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.2.1
eth-typing==2.3.0
eth-utils==1.10.0
hexbytes==0.2.2
idna==3.3
lru-dict==1.1.7
parsimonious==0.8.1
pycryptodome==3.14.1
requests==2.27.1
rlp==2.0.1
six==1.16.0
toolz==0.11.2
urllib3==1.26.9
web3==4.7.2
websockets==6.0
```

### What was wrong?

Trying to decode [an event with bad padding](https://etherscan.io/tx/0x93f2a3e74d2d6d636c14d5f63348dbd03e256f4e35f222814554881ce7b2d26a#eventlog) causes a NonEmptyPaddingBytes exception

Note the above linked event is a `Transfer` event with a  `to_address` of `0x0000000000000000000000010000000000000000000000000000000000000006`, which has invalid data in the padding bytes, but is a real event in the blockchain.

* The code which produced the error

```
from web3.utils.events import get_event_data
from hexbytes import HexBytes


TRANSFER_ABI = {"name":"Transfer","type":"event","anonymous":False,"inputs":[{"indexed":True,"name":"from","type":"address"},{"indexed":True,"name":"to","type":"address"},{"indexed":False,"name":"raw_value","type":"uint256"}]}

myevent = {
  'log_index': 0,
  'transaction_hash': HexBytes('0x93f2a3e74d2d6d636c14d5f63348dbd03e256f4e35f222814554881ce7b2d26a'),
  'transaction_index': 7,
  'block_hash': HexBytes('0xe87449740310af0b5b819b4978694a72bf1a959ac88e387749b8f73c0d7481fb'),
  'block_number': 4086410,
  'address': '0x48c80f1f4d53d5951e5d5438b54cba84f29f32a5',
  'data': HexBytes('0x0000000000000000000000000000000000000000000000000de0b6b3a7640000'),
  'topics': [
[...]

PR #3256, opened 2024-02-26
add abi.decode strictness flag and catch InsufficientDataBytes error …

pacrob:
…when parsing logs

This should use the existing global strictness flag.

### What was wrong?

Related to Issue #1441 , if a user is decoding logs with length not a multiple of 32 bytes, it throws an `eth-abi` `InsufficientBytesLength` error which is not properly caught.

### How was it fixed?
Add catching the error type `InsuffiecientDataBytes` and allow the passing of `strict` flag through to `eth_abi.decode` in `process_receipt`.

Since the flag is then passed to `get_event_data`, I considered making the flag available everywhere `get_event_data` is called, but that would greatly expand the scope and doesn't seem necessary, given the original issue.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -1040,7 +1040,7 @@ For example:
 
    In the case of an ``InsufficientDataBytes`` error, it may be possible to decode the
    log by setting the ``strict_bytes_type_checking`` flag to ``False`` on the Web3
-   instance.
+   instance before decoding.
 
     .. code-block:: python
 
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -969,7 +969,7 @@ For example:
 
 .. _process_receipt:
 
-.. py:method:: ContractEvents.myEvent(*args, **kwargs).process_receipt(transaction_receipt, errors=WARN, abi_decode_strict=True)
+.. py:method:: ContractEvents.myEvent(*args, **kwargs).process_receipt(transaction_receipt, errors=WARN)
    :noindex:
 
    Extracts the pertinent logs from a transaction receipt.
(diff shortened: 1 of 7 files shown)


fselmo:
Instead of creating a new flag and passing that flag around, we should let this be determined by the [Web3.strict_bytes_type_checking](https://github.com/fselmo/web3.py/blob/main/web3/main.py#L262-L273) flag. We should use this flag in the decoder for web3 to mark the `strict` flag value for the abi methods. Thoughts there?

fselmo:
> Instead of creating a new flag and passing that flag around, we should let this be determined by the [Web3.strict_bytes_type_checking](https://github.com/fselmo/web3.py/blob/main/web3/main.py#L262-L273) flag. We should use this flag in the decoder for web3 to mark the `strict` flag value for the abi methods. Thoughts there?

Building on this, it looks like the previous version of the global strict bytes checking only re-builds encoders with `is_strict` set to `False`. We could, and maybe should, build decoders with these values as well and override certain areas where the strictness is checked. I'm not sure how much more effort this would take nor that it is the right avenue since it might just re-write a lot of the `eth-abi` code.

If we can find a good balance to set the strictness at the decoder level, then we can maybe even solve issues like #2489 when we pass in the `w3.codec` if `w3.strict_bytes_type_checking` is set to `False`. This would ensure the decoder itself is already essentially using `strict=False` when it decodes.

I think we should take a step back and re-evaluate how we want to add this globally since it doesn't seem like the global flag is handling decoding strictness very gracefully.

---
edit: Looking at the *eth-abi* code again, I think we just need to override the `strict` property on each decoder when we build the non-strict registry which should be pretty simple. But don't want to speak too soon.

Curious on your thoughts there.

pacrob:
As far as I can tell, I don't think this will work without updates to `eth-abi`. The `strict` flag only has any effect on decoding `bytes` and `string` types (not `bytes[]`, not `bytes2`, etc.) Since logs can be anything, it wouldn't work for most issues, the example in #2489 being one.

From `eth_abi/registry.py`:
```python
    def _get_decoder_uncached(self, type_str, strict=True):
        decoder = self._get_registration(self._decoders, type_str)

        if hasattr(decoder, "is_dynamic") and decoder.is_dynamic:
            # Set a transient flag each time a call is made to ``get_decoder()``.
            # Only dynamic decoders should be allowed these looser constraints. All
            # other decoders should keep the default value of ``True``.
            decoder.strict = strict

        return decoder
```

pacrob:
Let me know if I'm missing something. Otherwise, I think scaling this back to just catching the `InsufficientDataBytes` error is the way to go for now.

PR #3257, opened 2024-02-26
`ContractCaller` initializes without validating the `block_identifier`

### LINKED ISSUE #2816: Call contract function using block hash makes two requests inside
Hi!

All call to contract functions with block identifier as block hash makes two requests:
1. Fetch block by block hash.
2. Send call request to contract with block number from previous response.

https://github.com/ethereum/web3.py/blob/master/web3/contract/contract.py#L406

### How can it be fixed?

I propose to replace this part
https://github.com/ethereum/web3.py/blob/a35351d8bb3305a3b31bcbcac9f76a00c83c6d23/web3/_utils/contracts.py#L419-L421
with
```python
elif is_hex_encoded_block_hash(block_identifier):
    return block_identifier
elif isinstance(block_identifier, bytes):
    retrurn '0x' + block_identifier.hex()
```

reedsa:
### What was wrong?

Closes #2816 

### How was it fixed?

Remove call to `parse_block_identifier` when initializing `ContractCaller` functions. The validation occurs when calling the function itself.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-02-26 at 4 17 26 PM

### DIFF
--- a/newsfragments/3257.performance.rst
+++ b/newsfragments/3257.performance.rst
@@ -1,0 +1,1 @@
+Remove call to ``parse_block_identifier`` when initializing ``ContractCaller`` functions.
--- a/tests/core/contracts/test_contract_caller_interface.py
+++ b/tests/core/contracts/test_contract_caller_interface.py
@@ -149,24 +149,38 @@ def test_caller_with_invalid_block_identifier(w3, math_contract):
 
 
 def test_caller_with_block_identifier(w3, math_contract):
-    start_num = w3.eth.get_block("latest").number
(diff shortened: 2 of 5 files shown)


pacrob (approved):
lgtm!

fselmo (approved):
lgtm. Do we have a test that tests a block `hash` as the block id? I think that was the case in question right?

reedsa:
@fselmo Good catch, added those in and cleaned up those tests a bit more.

fselmo:
Sounds good. Ship it!

PR #3262, opened 2024-03-04
Set all call timeout defaults on providers to ``30``:

### LINKED ISSUE #1939: Set default timeouts to mirror Geth
* Version: all
* Python: all
* OS: all

### What was wrong?
Geth specifies default timeouts of 30s [1] per request. Clients that use this library should expect the default behavior to match that of the node. This is particularly true for writes (sendRawTransaction) because of the data inconsistencies that can arise from a successful transaction that returns a client-side timeout.

[1] https://github.com/ethereum/go-ethereum/blob/053ed9cc847647a9b3ef707d0efe7104c4ab2a4c/rpc/http.go#L99

### How can it be fixed?

Fill this section in if you know how this could or should be fixed.

fselmo:
### What was wrong?

- In the HTTP cases, this actually increases the timeout which matches the timeouts in some clients (like Geth); closes #1939.
- This commit also sets the persistent connection timeout to be the same as the HTTP timeout for a request which seems reasonable and keeps consistency.

### Todo:

### DIFF
--- a/newsfragments/3262.misc.rst
+++ b/newsfragments/3262.misc.rst
@@ -1,0 +1,1 @@
+Set all provider default timeouts to ``30``, effectively increasing HTTP timeouts to better match those of certain clients.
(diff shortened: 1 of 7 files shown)


reedsa (approved):
Lgtm

Issue #3094, opened 2023-09-12
include more context with exceptions


reedsa:
RPC errors are handled in various ways depending on the format of the response. When a response is malformed in a way that raises a BadResponseFormat, the raw response is appended to the error message, making it difficult for humans to parse.

Other exceptions like ValueError could also provide better messages and information.

For this effort, consider splitting up the message from the response as separate arguments to the excpetions. Also provide more context about the errors if possible.

It may be useful to audit other exceptions in web3.py to see if they can be improved as well.

fselmo:
related to #2798

Wenzhi-Ding:
Yes. I have a related issue. Currently, when there is a timeout error from RPC provider, current `manager.py` returns `ValueError`, which is difficult to catch and handle. There are too many possibilities for `ValueError`. Handling a timeout is different from handling ordinary `ValueError`. I suggest to seperate this error to someting like `RPCTimeoutError`.

```
Traceback (most recent call last):
  File "/home/ubuntu/collection_2401/app/submit_txn.py", line 160, in <module>
    pick_txns()
  File "/home/ubuntu/collection_2401/app/submit_txn.py", line 73, in pick_txns
    submit_txn(txn, nonce)
  File "/home/ubuntu/collection_2401/app/submit_txn.py", line 100, in submit_txn
    txn_hash = _mint(
               ^^^^^^
  File "/home/ubuntu/collection_2401/utils/web3_write.py", line 44, in _mint
    "gasPrice": w3.eth.generate_gas_price() * 2
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/lib/python3.11/site-packages/web3/eth/base_eth.py", line 89, in generate_gas_price
    return self._gas_price_strategy(self.w3, transaction_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/lib/python3.11/site-packages/web3/gas_strategies/time_based.py", line 232, in time_based_gas_price_strategy
    probabilities = _compute_probabilities(
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/lib/python3.11/site-packages/eth_utils/functional.py", line 45, in inner
    return callback(fn(*args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2796, opened 2023-01-26
Raise new MethodUnavailable error for methods that are unavailable

kclowes:
### What was wrong?
When methods are not supported by the node, we were raising a generic ValueError. It would be good to raise a custom error and maybe add more details to the error message.

Closes #2448 

### How was it fixed?
Now if a method is unsupported, we raise `MethodUnavailable`, which inherits from `Web3Exception`.

@fselmo / @pacrob - This is one of the error messages that we get asked about most in the Discord channel. Does anyone have thoughts on adding a custom "hint" here or something? Maybe something along the lines of "The provider you're using does not support this method. Check the provider docs for more info."?

The response comes back like:
```py
response = { 
    "jsonrpc": "2.0",
    "error": {
        "code": -32601,
        "message": "the method eth_getTransactionByHash does not exist/is not "
        "available",
    },
}
```
and we return `response["error"]`  to the user. But maybe we should add something like: 

```py
 {
    "code": -32601,
    "message": "the method eth_getTransactionByHash does not exist/is not available",
    "hint": "Check your node provider's API docs to see what methods are supported",
}
```
Open to other ideas too or just leaving it as-is!

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3263, opened 2024-03-04
Add `user_message` to `Web3Exception`

### LINKED ISSUE #2798: Add optional user message to Web3Exception
### What was wrong?

It would be nice to add an optional user message on the `Web3Exception` class that would allow us to communicate info about the exception to users. It came up in the context of #2796. In that case, we only return the info from the node, but we get a lot of questions about what the node error info means. It would be nice if we could include the error message from the node along with a hint to the users.

reedsa:
### What was wrong?

Closes #2798

Related to #3094 

In #2796 there was a recommendation to add `user_message` to the `Web3Exception` class to support hint messages and other information for users. 

### How was it fixed?

Implemented an optional `user_message` named argument in `Web3Exception`. This adds the ability to provide a more human friendly message.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-03-04 at 2 26 21 PM


pacrob:
I like it! Could probably use a test or 2 though, no?

reedsa:
@pacrob Yeah I can add tests that throw this and assert on the `user_message`. Good call!

fselmo (approved):
nice change, lgtm 👍🏼

web3/exceptions.py
```diff
@@ -30,6 +30,16 @@ class Web3Exception(Exception):
...
 
+    def __init__(
+        self,
+        *args: Any,
+        user_message: Optional[str] = None,
+        **kwargs: Any,
+    ):
+        super().__init__(*args)
+        # Assign properties of Web3Exception
+        self.__dict__.update(locals())
```
fselmo: Sorry I missed how we were updating this the first time around. Is there a reason we are using `locals()` here instead of assigning `self.user_message = user_message`?

reedsa: I did not use user_message explicitly, though I could. `locals()` here also saves the other args and kwargs on the `Web3Exception`. `kwargs` are not supported by `Exception` as far as I could tell so this felt like a good place to keep that data.

Also thought keeping it open to anything the user would want to store might be useful. If we want to enforce it strictly to `user_message` I am fine doing so. Not sure if there are risks to saving any named property on the `Web3Exception`.

What do you think?

fselmo: I'll always lean toward easier reading of the code / debugging as long as there's not a big performance hit. Since explicit property assignment is slightly more performant I'd just lean that way. To me this feels like I don't know what the class is inheriting when I look at it and that would take up more dev time to fix a bug, etc.

That said, if you feel strongly about it I can be convinced :)

reedsa: Totally agree with having that more explicit. I guess any class properties could be overwritten, so protecting against that might be best.

I do not feel strongly about supporting `kwargs` in general. It might be convenient to store some data in the exception but maybe we wait until a use case comes about. Do you think there's significant value to storing custom `kwargs` somehow? What about `args`, since they are on the `Exception` it may or may not be useful to have them accessible on both.

I'm on board with setting the `user_message` explicitly. Let me know what you think about `kwargs` and `args` for `Web3Exception`.

pacrob (approved):
lgtm!

PR #3264, opened 2024-03-05
Replace `ipc_path` variable with `endpoint_uri`

ehdgks0627:
### What was wrong?

The `AsyncIPCProvider` class was using `ipc_path` instead of `endpoint_uri` from its super class, `PersistentConnectionProvider`. 

### How was it fixed?

Remove the usage of `ipc_path` and replaced it with `endpoint_uri` in the `PersistentConnectionProvider` class.

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -254,19 +254,19 @@ Persistent Connection Base Class
 AsyncIPCProvider
 ++++++++++++++++
 
-.. py:class:: web3.providers.persistent.AsyncIPCProvider(ipc_path=None, max_connection_retries=5)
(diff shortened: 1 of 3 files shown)


fselmo:
I don't think this is ideal as `endpoint_uri` implies there exists an endpoint URI. I think we should remove the typing from `PersistentConnectionProvider`, [L33](https://github.com/ethereum/web3.py/blob/main/web3/providers/persistent/persistent.py#L33-L33), and let `WebSocketProvider` define and use `endpoint_uri` and let `AsyncIPCProvider` define and use `ipc_path`.

Does that make sense?

PR #3267, opened 2024-03-05
catch eth-abi DecodeError on failed tester provider make_request v6

pacrob:
### What was wrong?

`eth-abi v5.0.1` introduced a new `InvalidPointer` exception type. In `EthereumTesterProvider->_make_request`, we catch the `InsufficientDataBytes` exception, but no other types of decoding errors. Specifically, this caused 2 vyper tests to start failing.

This is the v6 version of pr #3271 

### How was it fixed?

Instead of catching just the `InsufficientDataBytes` exception, catch its parent `DecodingError`

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/3267.bugfix.rst
+++ b/newsfragments/3267.bugfix.rst
@@ -1,0 +1,1 @@
+Catch all types of ``eth-abi`` ``DecodingError`` in ``EthereumTesterProvider->_make_request()``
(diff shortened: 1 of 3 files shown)


fselmo (approved):
Ideally we add a good test here. See my comment on #3271. Otherwise, lgtm 👍🏼

Issue #3036, opened 2023-07-11
Port useful abi utility methods to public utils directory


fselmo:
We have many handy abi utility methods that live under the `web3._utils.abi` directory that could be made stable / public via our public utilities directory in `web3.utils.abi`. Something that comes to mind is a method like `encode_method_call(fn_signature, params)`, for example.

Some of these _could_ make more sense if they lived in ~~*eth-abi*~~ (maybe *eth-utils* makes more sense as mentioned below) and could be exposed through `web3.py` if desired, sort of like how the *eth-utils* methods `to_hex()`, `to_bytes()`, etc, are passed through via `Web3.to_bytes()`...

This seems like a low-hanging fruit since that logic exists internally in the *web3.py* library. Making it public and massaging it into methods that are easy to understand and use could go a long way.

reedsa:
I'm curious if we also want to consider moving abi functions that reside in `eth-utils`. They seem to be closely related to other abi functions, so keeping everything in one place (probably `eth-abi`) would be ideal.

Another option, which may not make sense, is to move all things to `eth-utils` and just deprecate `eth-abi` entirely. I'm not sure what value one has over the other to be honest. My guess is everything "utility" would be easiest to find under the `eth-utils` package. Just food for thought.

fselmo:
> deprecate eth-abi entirely.

*eth-abi* houses the implementation for ABI encoding & decoding. I think it's a reasonable abstraction to leave that to its own library / repository.

I wouldn't be against using *eth_utils* a bit more for "public" utils though if it makes sense to - even if we end up exposing it via *web3.py* where / if necessary. The distinction of what should live in *eth-utils* vs *web3.utils* is one that would be good to think about. If the answer is "everything" then we can just port useful fxns over to *eth_utils* and call _that_ the public utilities.

reedsa:
In #1596 we decided most functions which deal with a contract's ABI belong closer to its context within Web3. We started releasing event and function helpers publicly via `web3.contract.utils`.

When other functions are exposed, consider whether it is a generalized util and might fit best in `eth-utils` or if an ABI is involved it may belong in `web3.contract.utils`.

PR #3268, opened 2024-03-05
Expose `get_event_data`, `get_function_info` utilities

### LINKED ISSUE #1596: Move `get_event_data` and `get_function_info` from web3._utils modules to Public API
### What was wrong?
Breaking changes were made to utility functions depended upon for client application functionality, making version updates more difficult.

### How can it be fixed?
Python clients are using utility functions such as `get_event_data` and `get_function_info` extensively to interact with the blockchain. Given their utility to clients, these functions should be moved out of the `_utils` module and made public to ensure API stability across non-major releases.

reedsa:
### What was wrong?

Closes #1596
Relates to #3036

`get_event_data`, `get_function_info` utilities were inaccessible for public use even though many people have the need to use these functions without instantiating a `Contract`.

Further refactoring to be covered in #3279 

### How was it fixed?

Expose these functions via `web3.contract.utils`.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-03-05 at 2 50 20 PM

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -1178,6 +1178,87 @@ Utils
           '_debatingPeriod': 604800,
           '_newCurator': True})
 
+.. py:method:: web3.contract.utils.get_function_info(fn_name, abi_codec, contract_abi, fn_abi, args, kwargs)
+
+    Get function signature and information from a contract ABI given a name, contract 
+    ABI, and arguments.
+
+    If the function exists and its signature matches the provided arguments, the 
+    function ABI, selector and arguments will be returned.
+
+    Inspect a contract function from an ABI as follows:
+
+    .. code-block:: python
+
+        # Contract ABI with a single "counter" function
+        contract_abi = [
+            {
+                "inputs": [],
+                "name": "counter",
+                "outputs": [{"internalType": "uint256", "name": "", "type": "uint256"}],
+                "stateMutability": "view",
+                "type": "function",
+            },
+        ]
+
+        >>> web3.contract.utils.get_function_info("counter", w3.codec, contract_abi)
+        ({'inputs': [], 'name': 'counter', 'outputs': [{'internalType': 'uint256', 'name': '', 'type': 'uint256'}], 'stateMutability': 'view', 'type': 'function'}, '0x61bc221a', ())
+
+.. py:method:: web3.contract.utils.get_event_data(abi_codec, event_abi, log_entry)
+
+    Get decoded event data from a log entry using an event ABI.
+
+    For example, to find all WETH "Transfer" events in the latest block:
+
+    .. code-block:: python
+
+        WETH_ADDRESS = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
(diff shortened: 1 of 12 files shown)


pacrob:
Overall great, my only question is how to obtain an `event_abi`, as this is the first I've heard that term. Looks like it's obtained using `find_matching_event_abi`, so maybe we need to expose that too? Not sure how deep the rabbit hole will go :rabbit:

reedsa:
There are now two additional utilities I made public to assist with the `event_abi`.  A similar utility exists to extract a `function_abi` with it's name and argument signature, but those can be made public in a subsequent refactor (https://github.com/ethereum/web3.py/issues/3279).

`find_matching_event_abi` - Looks in the contract ABI for the event that correlates with the provided name.
`construct_event_filter_params` - Builds the keccak signatures from the provided filters and event ABI

fselmo:
Not to throw a wrench in the spokes, but before we make this public via an API, do you think this is worth a re-consideration: https://github.com/ethereum/web3.py/issues/2489 ?

fselmo:
> Not to throw a wrench in the spokes, but before we make this public via an API, do you think this is worth a re-consideration: #2489 ?

On second thought, if we can handle this strictness in the codec itself this should remain unchanged. See my comment [here](https://github.com/ethereum/web3.py/pull/3256#issuecomment-1992619139). That might be a more elegant solution than passing `strict` into every method that uses `eth_abi`. Although if it makes sense to here, since it is a utility method, then we can consider adding a `strict` kwarg to `get_event_data` too.

Thoughts?

reedsa:
I like the idea of putting this within the codec itself as it seems much more flexible. I think if we decide to add more contract utility functions the flag could be required for other functions that deal with events.

It looks like the decode issue could happen for any kind of log, so if that is true it would be easier to maintain in the codec.

docs/web3.contract.rst
```diff
@@ -1178,6 +1178,87 @@ Utils
...
+
+    Get function signature and information from a contract ABI given a name, contract 
+    ABI, and arguments.
+
+    If the function exists and its signature matches the provided arguments, the 
+    function ABI, selector and arguments will be returned.
+
+    Inspect a contract function from an ABI as follows:
+
+    .. code-block:: python
```
kclowes, line 1191: These code blocks would be really good candidates for doctests.  I think there are some examples of doctests in this file, but let me know if you need more direction than that!

kclowes (commented):
I think this is okay as a first step to expose these methods, but I think we could really improve on this API. For example, it might be nice to return an object for `get_function_info` instead of a tuple. It also feels clunky to be passing around the codec. Feel free to take or leave, but I think if we put these methods on the web3 instance instead, we don't need to pass the codec through. I'm curious too if you have any other ideas about where we might make the API better.

reedsa:
@kclowes Great ideas, I also think we can do better with these. I see several functions that deal with `contract_abi`'s which might be useful for public consumption.

In addition to `get_event_data`, `get_function_info` and subsequent related functions I've exposed in this PR, many of the ABI helpers for functions and events seem useful too, namely those in `_utils/abi.py`, `_utils/events.py`.

I think the `eth_utils.abi` helpers are a little out of place too but I dont think it's worth moving those at this point.

I think we can expose the same functionality of the function and event abi helpers but with descriptive names. After an initial think through those candidates, I came up with the following:
```python
# get all functions in a contract
fns = w3.get_contract_functions(contract_abi: ABI)

# ABIFunction
fn = fns[0]
fn_name = fn["name"]

fn_abi = w3.get_contract_function_abi(contract_abi, fn_name, fn_args, fn_kwargs)
fn_arguments = w3.get_contract_function_arguments(fn_abi, fn_args, fn_kwargs)
fn_selector = w3.get_contract_function_selector(fn_abi)
fn_signature = w3.get_contract_function_signature(fn_abi)

# get all events in a contract
events = w3.get_contract_events(contract_abi)

# ABIEvent
event = events[0]
event_name = event["event"]

event_abi = w3.get_contract_event_abi(contract_abi, event_name, event_arg_names)
event_argument_names = w3.get_contract_event_argument_names(event_abi)
event_inputs = w3.get_contract_event_inputs(event_abi, indexed=True)
event = w3.get_contract_event_log(event_abi, log_entry)
```

kclowes:
Good ideas! I like those! [This](https://docs.ethers.org/v5/api/utils/abi/interface/#Interface) is what Ethers exposes too, for reference.

reedsa:
Closing this, proceeding in https://github.com/ethereum/web3.py/pull/3401

PR #3269, opened 2024-03-05
clear new abi lint errors

pacrob:
### What was wrong?

Some lint errors have appeared in `web3._utils.abi.py`. Possibly related to recent `eth-abi` releases, but same lint errors after installing older `eth-abi` versions. :shrug:

### How was it fixed?
Three main mypy complaints:
 - The `def is_dynamic` attempts to overwrite the class attribute `is_dynamic` with the `@property` function `is_dynamic`. I changed it to be a regular class attribute, as is standard across `eth-abi`
 - `validate_value`'s signature was being altered by the `combomethod` decorator so it didn't match the parent ABC function sig. I determined that because the function uses `subencoder`, which only exists if the class has been instantiated, it cannot be a `classmethod` or a `combomethod`. Removing the decorator removes the mypy error. Removed the `return` as well - the function is supposed to either not return or throw an error.
 - `from_type_str` returns an instance of an encoder class, with the `cls(...)` calling the `__init__` function of the parent class. But (I think) because the `@parse_type_str` decorator turns `from_type_str` into a `classmethod`, mypy thinks `cls(...)` is calling the `__call__` method, so the args are wrong (as is the return type of `bytes`, so I updated that here too.)

I'll make a note to come back and see if `@parse_type_str` can be removed, but for now I think ignoring the last error is the best (most timely) option.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/3269.bugfix.rst
+++ b/newsfragments/3269.bugfix.rst
@@ -1,0 +1,1 @@
+Fix/update methods and decorators in ``web3/_utils/abi.py`` to address issues raised by ``mypy``
--- a/web3/_utils/abi.py
+++ b/web3/_utils/abi.py
@@ -51,7 +51,6 @@ from eth_typing import (
     TypeStr,
 )
 from eth_utils import (
-    combomethod,
     decode_hex,
     is_bytes,
     is_list_like,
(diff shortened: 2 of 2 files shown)


kclowes (approved):
I think this is okay for now, but I'm still confused as to why this suddenly started failing. It looks like using eth-abi 5.0.0 passes the mypy lint, but both 5.0.1 and 4.2.1 cause it to fail. Super weird.

EDIT: Ha never mind. Removing the typing files (as happened in v5.0.0) would definitely make this pass. 😆

PR #3261, opened 2024-03-01
Remove ethpm

pacrob:
### What was wrong?

`ethpm` was deprecated, now being fully removed.

Related to Issue #2854 

### Todo:
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

PR #3270, opened 2024-03-06
remove repetitive words

teslaedison:
remove repetitive words


pacrob:
Thanks for your help, @teslaedison , but the entire `ethpm` module is being removed per #3261.

PR #3271, opened 2024-03-06
catch eth-abi DecodeError on failed tester provider make_request

pacrob:
### What was wrong?

`eth-abi v5.0.1` introduced a new `InvalidPointer` exception type. In `EthereumTesterProvider->_make_request`, we catch the `InsufficientDataBytes` exception, but no other types of decoding errors. Specifically, this caused 2 vyper tests to start failing.

### How was it fixed?

Instead of catching just the `InsufficientDataBytes` exception, catch its parent `DecodingError`

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/3271.bugfix.rst
+++ b/newsfragments/3271.bugfix.rst
@@ -1,0 +1,2 @@
+Catch all types of ``eth-abi`` ``DecodingError`` in ``EthereumTesterProvider->_make_request()``
+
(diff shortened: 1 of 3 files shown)


fselmo (approved):
It would be nice to add a test that fails before this change. Maybe a parametrized test that raises `InvalidPointer` `InsufficientDataBytes` and `DecodingError` and makes sure they're all caught and that the message is successfully parsed. If we ever need to catch more exceptions, they can be added to that list in the test?

Not a blocker though, the change lgtm.

Issue #3119, opened 2023-10-09
Terminology updates across the library


fselmo:
### What was wrong?

- Some middleware use the term `whitelist`. Let's change these over to `allowlist`.
- Consider scraping the whole of the library for terminology that can be changed before the next major version (`v7`).

Consider other terms as well, here is a good list: https://www.aswf.io/blog/inclusive-language/

fselmo:
Can we mark this closed by #3275?

PR #3275, opened 2024-03-08
Lint non-inclusive language with `blocklint`

reedsa:
### What was wrong?

Related to Issue #3119

Our codebase should eliminate language that can be offensive.

### How was it fixed?

`blocklint` is a tool which checks for words that are configured as a `blocklist`. If found, the linter will fail and output the errors.

### Todo:

#### Cute Animal Picture
Screen Shot 2024-03-08 at 4 10 41 PM

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,20 +52,34 @@ basepython =
     py310: python3.10
     py311: python3.11
 
+[blocklint]
+max_issue_threshold=1
+
+[common-lint-args]
+files={toxinidir}/web3 {toxinidir}/ens {toxinidir}/tests
+blocklint_files={toxinidir}/web3/* {toxinidir}/ens/* {toxinidir}/tests/*
+blocklint_skip_files="docs/Makefile,docs/releases.rst,tox.ini"
+
 [common-lint]
 extras=linter
+allowlist_externals=
+    ; bash passes args to blocklint with *
+    /bin/bash
 commands=
-    flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/tests
-    black {toxinidir}/ens {toxinidir}/web3 {toxinidir}/tests {toxinidir}/setup.py --check
-    isort --check-only --diff {toxinidir}/web3/ {toxinidir}/ens/ {toxinidir}/tests/
+    flake8 {[common-lint-args]files}
+    black {[common-lint-args]files} {toxinidir}/setup.py --check
+    isort --check-only --diff {[common-lint-args]files}
     mypy -p web3 -p ens --config-file {toxinidir}/mypy.ini
+    /bin/bash -c "blocklint {[common-lint-args]blocklint_files} --skip-files {[common-lint-args]blocklint_skip_files}"
 
 [testenv:lint]
 basepython: python
+allowlist_externals: {[common-lint]allowlist_externals}
 extras: {[common-lint]extras}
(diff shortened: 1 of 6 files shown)


reedsa:
This configuration will be changed to use `pre-commit` once the latest template is applied to this repo.

pacrob (approved):
lgtm!

docs/releases.rst
```diff
@@ -331,7 +331,7 @@ Features
 
 - Improved error messaging for exceptions from malformed JSON-RPC responses. (`#3053 <https://github.com/ethereum/web3.py/issues/3053>`__)
 - Enable filtering by non-indexed arguments for contract event ``get_logs()``. (`#3078 <https://github.com/ethereum/web3.py/issues/3078>`__)
-- Add ``eth_maxPriorityFeePerGas`` to ``exception_retry_middleware`` whitelist (`#3090 <https://github.com/ethereum/web3.py/issues/3090>`__)
+- Add ``eth_maxPriorityFeePerGas`` to ``exception_retry_middleware`` allowlist (`#3090 <https://github.com/ethereum/web3.py/issues/3090>`__)
```
kclowes: feel free to take or leave, but I think since this was called the `whitelist` at this point in time, it makes sense to leave it?

tox.ini
```diff
@@ -52,20 +52,30 @@ basepython =
...
 
+[blocklint]
+max_issue_threshold=1
+blocklist=blacklist,whitelist,slave
+exactlist=master ; allow URLs to contain some terms
+
 [common-lint]
 extras=linter
+allowlist_externals=
+    /bin/bash
```
kclowes, line 67: I don't see that we're using bash anywhere within the lint tasks so I think this can go.

tox.ini
```diff
@@ -52,20 +52,30 @@ basepython =
...
 commands=
     flake8 {toxinidir}/web3 {toxinidir}/ens {toxinidir}/tests
     black {toxinidir}/ens {toxinidir}/web3 {toxinidir}/tests {toxinidir}/setup.py --check
     isort --check-only --diff {toxinidir}/web3/ {toxinidir}/ens/ {toxinidir}/tests/
     mypy -p web3 -p ens --config-file {toxinidir}/mypy.ini
+    blocklint docs ens tests web3
 
 [testenv:lint]
 basepython: python
+allowlist_externals: {[common-lint]allowlist_externals}
```
kclowes, line 77: This can go too, I believe

tox.ini
```diff
@@ -52,20 +52,30 @@ basepython =
...
 extras: {[common-lint]extras}
 commands: {[common-lint]commands}
 
 [testenv:py{38,39,310,311}-lint]
+allowlist_externals: {[common-lint]allowlist_externals}
```
kclowes, line 82: This can go too, I believe

docs/releases.rst
reedsa: Agree with you here, I am looking for ways to ignore this case but still check `releases.rst`. There are inline ignores for code but not docs unfortunately.

I think since all of these come from commit messages and those are unchangeable we wouldn't want to change these here. I think I'll just ignore this file entirely.

tox.ini
reedsa: Ended up needing bash 😭 

It will be scorched from existence with pre-commit.

PR #3288, opened 2024-03-14
[v6] EventABI Topics with Indexed Arguments

### LINKED ISSUE #3286: web3py cannot match Event ABIs when indexed arguments are not placed first
* Version: 6.15.1
* Python: 3.10
* OS: macOS
* `pip freeze` output

```
eth_abi                   5.0.1
eth-account               0.11.0
eth-bloom                 3.0.0
eth-hash                  0.5.2
eth-keyfile               0.7.0
eth-keys                  0.5.0
eth-retry                 0.1.19
eth-rlp                   1.0.1
eth-typing                3.5.2
eth-utils                 2.3.1
web3                      6.15.1
```

### What was wrong?

Somewhere in the library stack, `.process_log()` on Contract instances assume that ABI Event definitions always begin with their indexed arguments. 

Circle's [TokenMessenger Contract](https://github.com/circlefin/evm-cctp-contracts/blob/master/src/TokenMessenger.sol#L45) contains an event called `DepositForBurn()` where an indexed argument appears after an unindexed one.  (See also [the ABI](https://raw.githubusercontent.com/circlefin/evm-cctp-contracts/master/docs/abis/cctp/TokenMessenger.json) for the contract)

The correct hashing method of this Event would be `DepositForBurn(uint64,address,uint256,address,bytes32,uint32,bytes32,bytes32)`, or `0x2fa9ca894982930190727e75500a97d8dc500233a5065e0f3126c48fbe0343c0` (evidenced by [any txn](https://arbiscan.io/tx/0x51d5c2d87cdf8b7dc560c807521869f47ca8d7a1ad879c0237747ec45f5d1bfb#eventlog) that emits the event)

However, when trying to process any log using `contract.events.DepositForBurn().process_log(event)`, web3py throws a `MismatchedABI` exception. 
Debugging that call, it has somehow resolved topic0 to `0xfe7cd5fd79e991f2a72bb5f9aa66d76845cd7d1b8f205b6a264148cf0d1f8876`, which happens to be the keccak of the event signature if indexed params are assumed to be first, like:
`DepositForBurn(uint64,address,address,uint256,bytes32,uint32,bytes32,bytes32)`

### How can it be fixed?

Unfortunately, I was unable to properly trace whether web3py, eth-abi, or eth-utils is altering the evaluated ABI in this manner.  However, retaining exact order as published in the ABI will likely correct this issue.

reedsa:
### What was wrong?

Closes #3286

### How was it fixed?

When event ABI topics are converted to a "log topic" in `event_abi_to_log_topic`, it returns a value in bytes. Comparing this with the topic value often fails when the value is just a hex string and not `HexBytes`. So this is to ensure everything is converted to hex before comparing event signatures from logs with that from the event ABI.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-03-14 at 4 30 58 PM

### DIFF
--- a/newsfragments/3288.bugfix.rst
+++ b/newsfragments/3288.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``process_log()`` when parsing logs for events with indexed and non-indexed inputs. ``get_event_data()`` now compares log topics and event ABIs as hex values.
--- a/tests/core/contracts/test_extracting_event_data.py
+++ b/tests/core/contracts/test_extracting_event_data.py
@@ -143,6 +143,115 @@ def test_event_data_extraction(
     assert event_data["event"] == event_name
 
 
+def test_event_data_with_ordered_indexed_inputs(w3):
+    event_abi = {
+        "anonymous": False,
+        "inputs": [
+            {
+                "indexed": True,
+                "internalType": "uint64",
+                "name": "nonce",
+                "type": "uint64",
+            },
+            {
+                "indexed": True,
+                "internalType": "address",
+                "name": "burnToken",
+                "type": "address",
+            },
+            {
+                "indexed": False,
+                "internalType": "uint256",
+                "name": "amount",
+                "type": "uint256",
+            },
+            {
+                "indexed": True,
+                "internalType": "address",
+                "name": "depositor",
+                "type": "address",
+            },
+            {
+                "indexed": False,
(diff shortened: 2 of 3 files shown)


pacrob (approved):
lgtm!

DefiDebauchery:
These changes work on my end, as well. Thanks!

PR #3291, opened 2024-03-16
feat: gasless dnssec ccip support

stanleyjzheng:
### What was wrong?

Gasless DNSSEC was unsupported; eg.

```py
w3 = Web3(HTTPProvider("https://cloudflare-eth.com/"))

w3.ens.address("gregskril.com")
```

resulted in `web3.exceptions.MultipleFailedRequests: Offchain lookup failed for supplied urls.`

---
@fselmo:

- Re-compiled test contract with url that does not include ``{sender}`` and fixed relevant tests to reflect this change so we can make sure that logic is working as expected.
- Added newsfragment.
---

### How was it fixed?
Copied the CCIP handler from ENS's Rust implementation [here](https://github.com/ensdomains/ethers-ccip-read/blob/8838667e0e7768d14011cc3bb6b519b7e849864c/src/ccip.rs#L42-L48)

Apologies, I couldn't figure out how to correctly mock tests since I couldn't find a suitable sequences of tx's for `async_ens_setup`. But the above snippet works.

### Todo:
- [x] Review tests and make sure we have POST accounted for (I'm almost certain we do, just another set of eyes on this)

#### Cute Animal Picture
_MG_0866


fselmo:
Thanks for submitting this @stanleyjzheng and for the ping @mdtanrikulu. Can you confirm this is the expected behavior? I can merge once I have your 👍🏼.

PR #3292, opened 2024-03-18
[V6] Fix ``process_log`` for ``HexStr`` inputs

reedsa:
### What was wrong?

Related to Issue #3286 

Realized the topics and data values had not been explicitly converted to bytes. The initial fix I made was to convert the log topics to a `HexStr` but since bytes are needed later on, it made sense to just ensure everything is coerced to bytes before handling them.

### How was it fixed?

Explicit type coercion of entry ``topics`` and ``data`` values.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-03-18 at 12 19 02 PM


kclowes (approved):
Looks good to me! 🚀

PR #3293, opened 2024-03-18
Fix ``process_log`` for ``HexStr`` inputs

reedsa:
### What was wrong?

Related to Issue #3286 

Realized the topics and data values had not been explicitly converted to bytes. The initial fix I made was to convert the log topics to a `HexStr` but since bytes are needed later on, it made sense to just ensure everything is coerced to bytes before handling them.

### How was it fixed?

Explicit type coercion of entry ``topics`` and ``data`` values.

### Todo:

#### Cute Animal Picture

Screen Shot 2024-03-18 at 12 24 54 PM

### DIFF
--- a/web3/_utils/events.py
+++ b/web3/_utils/events.py
@@ -29,6 +29,7 @@ from eth_abi.codec import (
 from eth_typing import (
     ChecksumAddress,
     HexStr,
+    Primitives,
     TypeStr,
 )
 from eth_utils import (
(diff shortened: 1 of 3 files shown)


web3/_utils/events.py
```diff
@@ -102,6 +99,13 @@
     )
 
 
+@curry
```
fselmo: nit: There's probably no need to curry this, right? Doesn't seem like this method is being used in that context.

fselmo (approved):
Just one nit which adds a bit of unnecessary overhead, wrapping with `@curry` where we don't need to. Looks good otherwise 👍🏼

pacrob (approved):
Agree with the `@curry` nit, otherwise lgtm!

pacrob (approved):
lgtm!

PR #3295, opened 2024-03-19
bump to hexbytes>=1.2.0

pacrob:
### What was wrong?

Hexbytes had breaking changes at v1.0.0, and small improvements since. web3 v6 is pinned to <0.4.0.

### How was it fixed?

Move lower hexbytes pin to >=1.2.0

### Todo:

#### Cute Animal Picture

image


pacrob:
will handle in #3298

PR #3297, opened 2024-03-20
change linting to use pre-commit 

pacrob:
### What was wrong?

`pre-commit` is totally rad and we should use it for web3 v7!

### How was it fixed?
Added pre-commit, removed previous linting setup, moved config to `pyproject.toml` where possible.

To aid in review, I've made one commit for each new lint tool. It's still a big pr but should be more manageable that way.

### Todo:

#### Cute Animal Picture

image


reedsa (approved):
Nice improvements! lgtm

web3/providers/rpc/async_rpc.py
```diff
@@ -56,7 +56,9 @@ def __init__(
         request_kwargs: Optional[Any] = None,
         exception_retry_configuration: Optional[
             ExceptionRetryConfiguration
-        ] = ExceptionRetryConfiguration(errors=(ClientError, TimeoutError)),
+        ] = ExceptionRetryConfiguration(  # noqa: B008
```
fselmo: Instead of this I think we should set it to `None` here and below we should use:

```py
        self.exception_retry_configuration = (
            exception_retry_configuration
            or ExceptionRetryConfiguration(errors=(ClientError, TimeoutError))
        )
```

web3/providers/rpc/rpc.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         request_kwargs: Optional[Any] = None,
         session: Optional[Any] = None,
         exception_retry_configuration: Optional[ExceptionRetryConfiguration] = (
-            ExceptionRetryConfiguration(
+            ExceptionRetryConfiguration(  # noqa: B008
```
fselmo: Same thing as the `async_rpc.py` case here. This feels like a valid `B008` case. I think we should set it to `None` here and below:

```py
        self.exception_retry_configuration = (
            exception_retry_configuration
            or ExceptionRetryConfiguration(
                errors=(
                    ConnectionError,
                    requests.HTTPError,
                    requests.Timeout,
                )
            )
        )
```

ens/async_ens.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa: B008
```
fselmo: I don't think we should ignore whole files like this unless we have good reason to. We can resolve the three specific issues by setting arguments to `None` and addressing them in the method with `if x is None: {{set the default value}}`.

This guarantees it's not just set once. Even if that's intended behavior, we should add `noqa: B008` to single lines and explain why they're there with a comment instead of whole files. That would make it easy to mask future errors since it's sneakily up at the top of the file.

Specifically, in this file, we can address:

- `provider` in `__init__()`
- `address` in `setup_address()`
- `new_owner` in `setup_owner()`.

ens/ens.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa: B008
```
fselmo: See my comment on `async_ens.py` we should address the issues instead, set to `None`, etc...

ens/utils.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa: B008
```
fselmo: We should address the issues here instead.

web3/main.py
```diff
@@ -198,14 +198,14 @@ def middleware_onion(self) -> MiddlewareOnion:
     def to_bytes(
         primitive: Primitives = None, hexstr: HexStr = None, text: str = None
     ) -> bytes:
-        return to_bytes(primitive, hexstr, text)
+        return cast(bytes, to_bytes(primitive, hexstr, text))
```
fselmo: These don't seem, and shouldn't be, necessary locally. If there's any issue here we should address it some other way. Casting everything that `to_bytes`, `to_int`, etc, returns will not be an ideal way to resolve things across the library.

However, taking these out does not give me lint issues locally... so not sure if they are needed in the first place?

ens/specs/nf.json
```diff
@@ -1 +1 @@
...
...0,194931,194932,194933,194934,194935,194936,194937,194938,194939,194940,194941,194942,194943,194944,194945,194946,194947,194948,194949,194950,194951,194952,194953,194954,194955,194956,194957,194958,194959,194960,194961,194962,194963,194964,194965,194966,194967,194968,194969,194970,194971,194972,194973,194974,194975,194976,194977,194978,194979,194980,194981,194982,194983,194984,194985,194986,194987,194988,194989,194990,194991,194992,194993,194994,194995,194996,194997,194998,194999,195000,195001,195002,195003,195004,195005,195006,195007,195008,195009,195010,195011,195012,195013,195014,195015,195016,195017,195018,195019,195020,195021,195022,195023,195024,195025,195026,195027,195028,195029,195030,195031,195032,195033,195034,195035,195036,195037,195038,195039,195040,195041,195042,195043,195044,195045,195046,195047,195048,195049,195050,195051,195052,195053,195054,195055,195056,195057,195058,195059,195060,195061,195062,195063,195064,195065,195066,195067,195068,195069,195070,195071,195072,195073,195074,195075,195076,195077,195078,195079,195080,195081,195082,195083,195084,195085,195086,195087,195088,195089,195090,195091,195092,195093,195094,195095,195096,195097,195098,195099,195100,195101]}
```
fselmo, line 1: Should we ignore spec files?

fselmo (commented):
Had a few comments. I don't think turning off `noqa B008` on whole files is ideal. Also, should we ignore large json spec files for ENS? I don't see a benefit in formatting those but could go either way. I don't see a specific json formatter in pre-commit, just yaml and toml.

According to the github merge check there's a conflict too that likely needs to be resolved before merging.

ens/specs/nf.json
pacrob: I looked at them. All it did was add a newline at the end, which is best practice, at least for Unix-based systems.

web3/providers/rpc/async_rpc.py
pacrob: I tried using that pattern and `mypy` did not approve. Or maybe tests started failing?  Maybe I got something out of order? Will take another look.

ens/async_ens.py
pacrob: I put those there instead of inline because I made a note "talk to Felipe about the B008 ignores". And then forgot about it.

web3/providers/rpc/async_rpc.py
fselmo: Ok lmk if you run into issues. It worked for me locally.

ens/specs/nf.json
fselmo: Sweet. That makes sense then.

web3/main.py
pacrob: I was attempting to fix `no-any-return` `mypy` errors and started by `cast`ing them before I realized that it was a much bigger problem. If you remove the cast and set `mypy` `warn_return_any = true`, it'll come up. Removing them now.

web3/providers/rpc/async_rpc.py
pacrob: Yeah, it passes lint, but for some reason doesn't configure correctly. Setting this and the sync one to your example gives me:

```
FAILED tests/core/providers/test_http_request_retry.py::test_check_without_retry_config - AssertionError: assert 5 == 1
FAILED tests/core/providers/test_http_request_retry.py::test_async_check_without_retry_config - AssertionError: assert 5 == 1
```

pacrob:
I've cleaned up per comments. All added `cast`s have been removed as unneeded due to setting `mypy`'s `warn_return_any = false`.

There are 4 remaining `B008` rule ignores. 2 are related to ENS. Fails these tests:
`tests/ens/test_setup*.py`

The other 2 are related to HTTP provider retry config. Removing them fails:
`tests/core/providers/test_http_request_retry.py`

I tried a few things to fix them but no dice. Let me know if you'd like to try to fix them here or open a new issue to deal with them separately.

web3/_utils/request.py
```diff
@@ -109,9 +109,9 @@ def get_response_from_post_request(
 
 
 def make_post_request(
-    endpoint_uri: URI, data: Union[bytes, Dict[str, Any]], *args: Any, **kwargs: Any
```
kclowes, line 112: Curious about this change, and what library fixed it. I don't think many people are using it, but is breaking if they are. We should highlight in v7 migration docs maybe?

web3/tools/benchmark/main.py
```diff
@@ -144,18 +144,20 @@ def main(logger: logging.Logger, num_calls: int) -> None:
...
-                    "async_exec": lambda: async_w3_http.eth.get_block(1),
+                    "exec": lambda w3_http=w3_http: w3_http.eth.get_block(1),
+                    "async_exec": lambda async_w3_http=async_w3_http: async_w3_http.eth.get_block(  # noqa: E501
+                        1
+                    ),
                 },
             ]
 
-            def benchmark(method: Dict[str, Any]) -> None:
+            def benchmark(method: Dict[str, Any], loop=loop) -> None:
```
kclowes: This seems like a strange addition. Do you know why we need it?

pyproject.toml
```diff
@@ -20,15 +20,15 @@ use_parentheses = true
 check_untyped_defs = true
 disallow_any_generics = true
 disallow_incomplete_defs = true
-disallow_subclassing_any = true
+disallow_subclassing_any = false
```
kclowes: Should/do we have an issue to track these couple changes from `true` to `false`?

CONTRIBUTING.md
```diff
@@ -1,7 +1,7 @@
 # Contributing
 
 Thanks for the interest! See the [contributor documentation][contribute]
-to get started, and reach out on [Discord][discord] if you get stuck.
+to get started, and reach out on [Discord] if you get stuck.
```
kclowes, line 4: I think we still need this. It's linking to the discord link below. Alternatively, we could inline the links.

docs/README-windows.md
kclowes, line 1: I think we can probably delete this file? I think it's maybe from the template or py-evm?

kclowes (commented):
That is a terrifying cute animal picture, lol. I think the typing changes will be breaking, so I _think_ that this can only go into v7. I may be thinking about that wrong though, so let me know if I am. Thanks for getting this done, it will be so nice to have precommit in web3 finally! The commits made it super easy(/manageable :) ) to review, thank you!

web3/_utils/request.py
fselmo: Is it breaking if it's inside `_utils`?

web3/tools/benchmark/main.py
pacrob: The error is:

`B023 Function definition does not bind loop variable 'loop'.`

It's because `loop` and the function are defined within a `for` loop, the `loop=loop` fastens the value of `loop` to the function each time it's created, rather than just the value of `loop` at the end of the `for`. Or something like that.

web3/_utils/request.py
pacrob: The error was having `*args` after a keyword argument - places it was called were passing `data` as a kwarg. And yeah, as it was in `_utils` and we had no usage of the `*args`, I removed it.

web3/providers/rpc/async_rpc.py
fselmo: Ah yeah, I'm glad we have that test in there. We want to turn this off by setting it to `None` so we probably want to use the `Empty()` class as the default here instead. You can check out how `ENS` is configured for `Web3` in `main.py`

web3/tools/benchmark/main.py
kclowes: Got it. Thanks for the clarification!

web3/_utils/request.py
kclowes: Ah, that makes sense. Yeah, not breaking if it's inside `_utils`

CONTRIBUTING.md
pacrob: It was removed by the `mdformat` tool - it looks like it creates the link if spelling is the same:

https://github.com/pacrob/web3.py/blob/add-pre-commit/CONTRIBUTING.md

kclowes: ooo, fancy!

docs/README-windows.md
pacrob: It's not from the template in any form I've seen it, nor in `py-evm`. Should I delete all 4 OS READMEs then?

kclowes: weird. TIL that we have os-specific READMEs. I guess we can leave it for consistency 🤷

pyproject.toml
pacrob: The `true` to `false` is from the template default. web3.py has never had an entry for that rule, and `mypy` defaults to `false`, so the errors appeared for the first time when merging the template. I figured to leave it so at least we know we are explicitly ignoring them. It's not too bad, though - only 8 errors. I could put `ignore` lines in for them and then turn it back on so we don't ignore any more in the future.

kclowes: I'm good leaving it as `false` then

kclowes (approved):
Looks good to me. Thanks for clearing up those comments! 🚀

pyproject.toml
pacrob: Meh, I tried it. Several errors were resolved by adding pydantic types to mypy checking. Several more will be resolved when `eth-abi` gets fully typed and we'll get an `unused ignore` warning. Seems good to have it on to avoid adding more.

fselmo (approved):
Looks good! This will be amazing to get in, ty.

Issue #1864, opened 2021-01-31
Unhandled ValueError when calling contract functions


ml31415:
Version: master

### What was wrong?

Whenever a contract function call fails, for whatever reason, it throws an unspecified `ValueError`, which pops up through all the call stack. For libraries on top it's not possible to properly reason and deal with such an unspecific error. This problem shows up in a lot of issue reports, e.g. https://github.com/ethereum/web3.py/issues/1725 https://github.com/ethereum/web3.py/issues/1717 https://github.com/ethereum/web3.py/issues/651 and https://github.com/nosofa/uniswap-v2-py/issues/10 .

### How can it be fixed?

To my humble understanding, this should best be handled somewhere in `call_contract_function()` https://github.com/ethereum/web3.py/blob/d8996bd4b5c6c69415ef95ccac335471d1075d4b/web3/contract.py#L1470. E.g. in https://github.com/ethereum/web3.py/blob/d8996bd4b5c6c69415ef95ccac335471d1075d4b/web3/contract.py#L1528 the function nicely throws a perfectly understandable `BadFunctionCallOutput`. Imho the two function calls in https://github.com/ethereum/web3.py/blob/d8996bd4b5c6c69415ef95ccac335471d1075d4b/web3/contract.py#L1497 and below should probably also be wrapped with a try-except-clause and throw `BadFunctionCallOutput` or another more specific error.

Another and maybe even the better option might be, to replace all the `ValueError` and `KeyError` in https://github.com/ethereum/web3.py/blob/master/web3/manager.py with some kind of server side response error right away.

ml31415:
I'd be happy to put a PR together, but I'd like to hear the developers view on that first.

kclowes:
Sorry, @ml31415 I lost track of this. I'm for changing ValueError to be something more specific to web3. I think the best option would be to raise something other than `ValueError` in `manager.py`. In order to keep this backwards compatible, however, we'd need to inherit from ValueError in v5. Once we have a v6 branch going, we can remove the inheritance. I'm not sure what we'd call the error class... Maybe `Web3ClientError`, `ClientResponseError`, `Web3ResponseError`? /cc master namer @marcgarreau

ml31415:
[...]

PR #3300, opened 2024-03-25
Use in-house Web3 exceptions for all of the codebase

### LINKED ISSUE #3241: Use `Web3Exception` for all exceptions within the library
All exceptions for the library should inherit from `Web3Exception` for UX improvements on user-side exception handling

closes #1864

fselmo:
### What was wrong?

- closes #3241 

### How was it fixed?

- Instead of `ValueError`, raise `Web3ValueError` and extend from `ValueError` and `Web3Exception`
- Instead of `TypeError`, raise `Web3TypeError` and extend from `TypeError` and `Web3Exception`
- Instead of `AssertionError`, raise `Web3AssertionError` and extend from `AssertionError` and `Web3Exception`
- Instead of `AttributeError`, raise `Web3AttributeError` and extend from `AttributeError` and `Web3Exception`
- Add `ENSValueError` and `ENSTypeError` as well. Seems like that could only help with better control over exception handling.

### Todo:

#### Cute Animal Picture

Screenshot 2024-03-26 at 18 44 18


tests/library_wide_test.py
```diff
@@ -0,0 +1,32 @@
...
+    for root, dirs, files in os.walk(WEB3_PATH):
+        if "module_testing" in dirs:
+            dirs.remove("module_testing")
+            continue
+        for file in files:
+            if file.endswith(".py"):
+                file_path = os.path.join(root, file)
+                with open(file_path, "r", encoding="utf-8") as f:
+                    line_number = 0
+                    for line in f:
```
reedsa: Nit: Could use `enumerate` to keep track of the line number instead.

```suggestion
                    for line_number, line in enumerate(f):
```

reedsa (approved):
No major concerns, lgtm!

Noticed two empty files, [tests/core/pm-module/test_ens_integration.py](https://github.com/ethereum/web3.py/pull/3300/files#diff-7cdd106902cec475ae2f95918c3e33b9fc4b984bfd07a2345958073fee748702) and [web3/tools/pytest_ethereum/deployer.py](https://github.com/ethereum/web3.py/pull/3300/files#diff-6a9aada34ef357017b2a273df3d6835c9f52845cf668bfaf9fac5eada34d35e2)

fselmo:
My ultimate question here is whether or not it answers the call to have a more specific error thrown? I suppose we can take care of that part after as well. This helps with user exception handling, ideally catching all (or close to all) `web3.py` errors by catching `Web3Exception` or any combination of exceptions that inherit from it for fine-tuning. But if we want to throw an error that isn't so generic as _any_ value error is bound to be I think we should probably end up tracking that in a separate ticket maybe?

@kclowes I'll wait for your review / input on this since you seem to be in this conversation quite a while longer than I.

cc: @LefterisJP for input via [this comment](https://github.com/ethereum/web3.py/issues/1864#issuecomment-1902072719).

LefterisJP:
> This helps with user exception handling, ideally catching all (or close to all) `web3.py` errors by catching `Web3Exception` or any combination of exceptions that inherit from it for fine-tuning

Exactly this is what prompted my original comment. To be able to catch all the library errors easily and not have to add extra `except` clauses whenever a user does something that raises a `ValueError`, or `TypeError` which we did not expect.

> But if we want to throw an error that isn't so generic as any value error is bound to be I think we should probably end up tracking that in a separate ticket maybe?

I am not sure I understand what you mean here. What kind of error would not be generic? Can you maybe give an example?

cc @yabirgb from the rotki team here just an fyi for the upcoming changes and for potential input.

fselmo:
> I am not sure I understand what you mean here. What kind of error would not be generic? Can you maybe give an example?

I think your answer covered it. I just mean that in some cases a more specific class name may make more sense. Something like `Web3RPCError` but I get the sense that the important part for you was fine tuning what is coming from the library and not how generic the exception scope was.

reedsa (approved):
Lgtm

tests/core/library_wide_test.py
kclowes, line 1: nit: I think most test files start with a `test_`, maybe something like `tests/core/test_library_for_generic_exceptions.py`?

kclowes (approved):
This looks great! I left one nit on a file name, but that was all I saw. It'll be great to get this in!

PR #3302, opened 2024-03-26
add "python 3.12" to setup.py classifiers

agodbug:


fselmo:
Thanks @agodbug but we don't yet test against python 3.12 in our CI and don't want to advertise that we support 3.12 until that is in and passing. This is coming very soon and this change will be included in that PR.

PR #3304, opened 2024-03-26
validate percentile param & simplify percentile function @math.py

agodbug:


fselmo:
This seems harmless enough to bring in while the `percentile` method is there. I'm not sure how much use that part of the code base gets but this adds validation so 👍🏼

PR #3309, opened 2024-03-27
simplify check_if_retry_on_failure function @web3/providers/rpc/utils.py

agodbug:


fselmo:
These small changes as separate PRs do not seem like necessary changes

PR #3310, opened 2024-03-28
add wsv2 upgrade path + rearrange migration guide

wolovim:
### What was wrong?

nits: 
- i thought the provider updates made sense as the first items highlighted in the migration guide
- opportunity to include smol note about the upgrade path for `WebsocketProviderV2` users
- additional smol language tweaks

### Todo:

#### Cute Animal Picture

### DIFF
--- a/docs/v7_migration.rst
+++ b/docs/v7_migration.rst
@@ -3,11 +3,70 @@
 Migrating your code from v6 to v7
 =================================
 
-web3.py follows `Semantic Versioning <http://semver.org>`_, which means
-that version 7 introduced backwards-incompatible changes. If your
-project depends on web3.py ``v7``, you'll probably need to make some changes.
+web3.py follows `Semantic Versioning <http://semver.org>`_, which means that
+version 7 introduced backwards-incompatible changes. If you're upgrading from
+web3.py ``v6`` or earlier, you can expect to need to make some changes. Refer
+to this guide for a summary of breaking changes when updating from ``v6`` to
+``v7``. If you are more than one major version behind, you should also review
+the migration guides for the versions in between.
 
-Breaking Changes:
+
+Provider Updates
+~~~~~~~~~~~~~~~~
+
+
+WebSocketProvider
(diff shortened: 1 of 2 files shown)


docs/v7_migration.rst
```diff
@@ -3,11 +3,69 @@
...
+  .. code-block:: python
+
+      # WebsocketsProviderV2:
+      AsyncWeb3.persistent_websocket(WebsocketProviderV2('...'))
+
+      # WebSocketProvider:
+      AsyncWeb3(WebSocketProvider('...'))
+
+- Handling incoming subscription messages has a more flexible namespace and method name,
+  given that it now supports the AsyncIPCProvider in addition to the WebSocketProvider:
```
fselmo: nit: I'd try to clarify what "it" is here a bit more. Maybe something along the lines of:

> Handling incoming subscription messages has a more flexible namespace and method name,
  given that the new ``AsyncIPCProvider`` was implemented as a ``PersistentConnectionProvider`` alongside the ``WebSocketProvider``:

docs/v7_migration.rst
```diff
@@ -3,11 +3,69 @@
...
+ 
+  .. code-block:: python
+
+    # WebsocketsProviderV2:
+    async for message in w3.ws.listen_to_websocket():
```
fselmo: This isn't the case in `V2` anymore. It is also `process_subscriptions()`. I think the verbiage above was changed but the code was not?

fselmo: Ah, "namespace and method name" should be changed to "namespace" only in the description above to reflect: `ws` -> `socket`.

wolovim: changed in [v6.15.0](https://web3py.readthedocs.io/en/stable/releases.html#web3-py-v6-15-0-2024-01-25), so if you kept up with latest versions, then you're right

fselmo (approved):
lgtm 👍🏼

PR #3319, opened 2024-03-29
Retrieve endpoint_uri or ipc_path from base persistent provider class

fselmo:
### What was wrong?

- closes #3264

Since we defined `endpoint_uri` at the `PersistentConnectionProvider` level, we used it in the `manager.py` since it wouldn't raise any attribute errors. Rather than rename the `ipc_path` in `AsyncIPCProvider`, I think it makes sense to have this separation of paths but be able to retrieve either path from the base class as well since we may want to implement logic for the base class (as is the case in `manager.py`).

### How was it fixed?

- Implement `endpoint_uri_or_ipc_path` property in the base class to be able to return the proper path depending on the type of persistent connection provider.

### Todo:

#### Cute Animal Picture

### DIFF
--- a/web3/providers/persistent/persistent.py
+++ b/web3/providers/persistent/persistent.py
@@ -49,8 +49,7 @@ class PersistentConnectionProvider(AsyncJSONBaseProvider, ABC):
         self.request_timeout = request_timeout
         self.silence_listener_task_exceptions = silence_listener_task_exceptions
 
-    @property
-    def endpoint_uri_or_ipc_path(self) -> str:
+    def get_endpoint_uri_or_ipc_path(self) -> str:
         if hasattr(self, "endpoint_uri"):
             return str(self.endpoint_uri)
         elif hasattr(self, "ipc_path"):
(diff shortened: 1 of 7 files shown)


pacrob (approved):
lgtm

web3/providers/persistent/persistent.py
```diff
@@ -50,6 +49,18 @@ def __init__(
...
         self.silence_listener_task_exceptions = silence_listener_task_exceptions
 
+    @property
+    def endpoint_uri_or_ipc_path(self) -> str:
+        if hasattr(self, "endpoint_uri"):
+            return str(self.endpoint_uri)
+        elif hasattr(self, "ipc_path"):
+            return str(self.ipc_path)
+        else:
+            raise AttributeError(
```
kclowes, line 58: depending on when this gets merged, may need to swap it for a `Web3AttributeError`

kclowes (approved):
Nice! I think this is a good solution. Edit - Nit: maybe call the method `get_endpoint_uri_or_ipc_path`? Feel free to take or leave

PR #3322, opened 2024-04-01
Format max fee per blob gas for requests

fselmo:
### What was wrong?

We were missing some field formatters for Cancun

### How was it fixed?

- Add request formatter for `maxFeePerBlobGas`
- Add result formatters for ``blobGasPrice`` and ``blobGasUsed`` for *eth_getTransactionReceipt*.

### Todo:

- [x] Add entry to the [release notes](https://github.com/ethereum/<REPO_NAME>/blob/main/newsfragments/README.md)


pacrob (approved):
lgtm!

PR #3353, opened 2024-04-17
Snake case the remaining camels

### LINKED ISSUE #3085: Variables in code should follow standard conventions
### What was wrong?

Most of the python codebase is using the standard [pep8 variable naming convention](https://peps.python.org/pep-0008/#method-names-and-instance-variables) with a few exceptions. We'll need to look at and decide whether it makes sense to conform to this for all variables throughout the codebase.

In this recent [PR](https://github.com/ethereum/web3.py/pull/3078/files), a couple variables used in filtering (`fromBlock` and `toBlock`) were considered for refactoring into snake_case names. As the request is constructed, the filter arguments can be adapted to the appropriate naming convention so that the code itself has no concern with how the requests look.

### How can it be fixed?

As a result of this thread, we will agree on the conventions we would like to proceed with. Refactoring can proceed as necessary for the v7 release.

fselmo:
### What was wrong?

Closes #3085

### How was it fixed?

- snake-case remaining camel-cased arguments: `fromBlock`, `toBlock`, and `blockHash`

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2024-04-17 at 17 02 17

### DIFF
--- a/docs/v7_migration.rst
+++ b/docs/v7_migration.rst
@@ -174,6 +174,21 @@ been functional since around October 2022. It was deprecated in ``v6`` and has b
 completely removed in ``v7``.
 
 
+Remaining cameCase -> snake_case Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following arguments have been renamed across the library from camelCase to
+snake_case in all methods where they are passed in as a kwarg.
+
+- ``fromBlock`` -> ``from_block``
+- ``toBlock`` -> ``to_block``
+- ``blockHash`` -> ``block_hash``
+
+Note that if a dictionary is used instead, say to a call such as `eth_getLogs`, the
+keys in the dictionary should be camelCase. This is because the dictionary is passed
+directly to the JSON-RPC request, where the keys are expected to be in camelCase.
+
+
 Miscellaneous Changes
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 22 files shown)


pacrob (approved):
:sheep:

reedsa (approved):
lgtm!

docs/v7_migration.rst
```diff
@@ -174,6 +174,21 @@ been functional since around October 2022. It was deprecated in ``v6`` and has b
...
+The following arguments have been renamed across the library from camelCase to
+snake_case in all methods where they are passed in as a kwarg.
+
+- ``fromBlock`` -> ``from_block``
+- ``toBlock`` -> ``to_block``
+- ``blockHash`` -> ``block_hash``
+
+Note that if a dictionary is used instead, say to a call such as `eth_getLogs`, the
+keys in the dictionary should be camelCase. This is because the dictionary is passed
+directly to the JSON-RPC request, where the keys are expected to be in camelCase.
```
kclowes, line 189: What do you think about adding a formatter for the case of `getLogs` and friends? Basically it would look at the TxDict and flip any snake_case to camelCase to be passed on to the node. Then on the way back out it would do the same.

fselmo: I'd be good with that if the user experience would benefit from it. The way it is now, users are aware the transactions need to have camelCase arguments as well i.e. `{"maxFeePerGas": 10**9, ...}`. I think this is a similar case to that but I do think it's a touch more confusing because of the way it can be used by passing in kwargs which are snake-cased. I could go either way.

PR #3357, opened 2024-04-22
Fix web3.eth.wait_for_transaction_receipt example output

### LINKED ISSUE #3336: Incorrect documentation of `send_raw_transaction` and `wait_for_transaction_receipt`
### What happened?

The value returned from function `send_raw_transaction` is a `HexBytes`, not a plain hexadecimal string.

The `transactionHash` attribute returned from function `wait_for_transaction_receipt` is a `HexBytes`, not a plain hexadecimal string. The input to this function, can apparently be any one of them with no issues.

The problem that I've personally encountered is that when you assert `input_tx_hash == output['transactionHash']` it may fail only because the LHS is a plain hexadecimal string, and the RHS is a `HexBytes`.

But in either case, the documentation should just state the expected return-value type correctly, in order to remove all doubts.

In the case of function `send_raw_transaction`, the documentation is confusing, with one place stating the return-value type being a `HexBytes` (see [here](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.send_raw_transaction)), and another place stating the return-value type being a plain hexadecimal string (scroll to the bottom of the example [here](https://web3py.readthedocs.io/en/stable/web3.eth.account.html#sign-a-contract-transaction)).

In the case of function `wait_for_transaction_receipt`, the documentation is just wrong, stating that the returned `transactionHash` attribute is a plain string (see [here](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.wait_for_transaction_receipt)).

_No response_

### Full error output

_No response_

Fix the docs

### web3 Version

6.16.0

### Python Version

3.9.13

### Operating System

MacOS

### Output from `pip freeze`

_No response_

tiendq:
### What was wrong?

Related to Issue #3336 
Closes #3336

### How was it fixed?

Updated `wait_for_transaction_receipt` example.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -618,17 +618,17 @@ The following methods are available on the ``web3.eth`` namespace.
         # ...
         # Then when the transaction is added to a block, its receipt is returned:
         AttributeDict({
-            'blockHash': '0x4e3a3754410177e6937ef1f84bba68ea139e8d1a2258c5f85db9f1cd715a1bdd',
+            'blockHash': HexBytes('0x4e3a3754410177e6937ef1f84bba68ea139e8d1a2258c5f85db9f1cd715a1bdd'),
             'blockNumber': 46147,
             'contractAddress': None,
             'cumulativeGasUsed': 21000,
             'from': '0xA1E4380A3B1f749673E270229993eE55F35663b4',
             'gasUsed': 21000,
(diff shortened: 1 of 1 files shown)


fselmo:
Looks good, thanks 👍🏼

Issue #3210, opened 2024-01-29
Exponential backoff factor for HTTP request retries


fselmo:
Make the ``backoff_factor`` for the ``ExceptionRetryConfiguration`` class work as an actual exponential "backoff factor" rather than just defining how long we sleep between retries. See how we implemented this in ``WebsocketProviderV2`` and do something similar to that.

fselmo:
closed by #3358

PR #3358, opened 2024-04-23
feat: Exponential backoff factor for HTTP request retries

hr98w:
### What was wrong?

Related to Issue https://github.com/ethereum/web3.py/issues/3210

### How was it fixed?
Make the backoff_factor for the ExceptionRetryConfiguration class work as an actual exponential "backoff factor" And Update the doc.

Reuse `backoff_factor` as the initial back off time

Introduce a new field `exponential_factor` 
### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2024-04-24 at 10 40 03

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -146,8 +146,10 @@ Retry Requests for HTTP Providers
 
 ``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
 on exceptions. This can be configured via configuration options on the provider
-instance. Below is an example showing the default options for the retry configuration
-and how to override them.
+instance. The retry mechanism employs an exponential backoff strategy, starting from
+the initial value determined by the ``backoff_factor``, and doubling the delay with
+each attempt, up to the ``retries`` value. Below is an example showing the default
+options for the retry configuration and how to override them.
 
 .. code-block:: python
 
@@ -165,8 +167,8 @@ and how to override them.
             # number of retries to attempt
             retries=5,
 
-            # how long to wait between retries
-            backoff_factor=0.5,
+            # initial delay multiplier, doubles with each retry attempt
+            backoff_factor=0.125,
 
             # an in-house default list of retryable methods
             method_allowlist=REQUEST_RETRY_ALLOWLIST,
--- a/newsfragments/3358.breaking.rst
+++ b/newsfragments/3358.breaking.rst
@@ -1,0 +1,1 @@
+Employ an exponential backoff strategy using the ``backoff_factor`` in ``ExceptionRetryConfiguration`` for ``HTTPProvider`` and ``AsyncHTTPProvider``. Reduce the default initial delay to ``0.125`` seconds.
--- a/web3/providers/rpc/async_rpc.py
+++ b/web3/providers/rpc/async_rpc.py
@@ -123,7 +123,7 @@ class AsyncHTTPProvider(AsyncJSONBaseProvider):
                 except tuple(self.exception_retry_configuration.errors):
                     if i < self.exception_retry_configuration.retries - 1:
                         await asyncio.sleep(
-                            self.exception_retry_configuration.backoff_factor
+                            self.exception_retry_configuration.backoff_factor * 2**i
                         )
                         continue
                     else:
--- a/web3/providers/rpc/rpc.py
+++ b/web3/providers/rpc/rpc.py
@@ -130,7 +130,9 @@ class HTTPProvider(JSONBaseProvider):
                     )
                 except tuple(self.exception_retry_configuration.errors) as e:
                     if i < self.exception_retry_configuration.retries - 1:
-                        time.sleep(self.exception_retry_configuration.backoff_factor)
+                        time.sleep(
+                            self.exception_retry_configuration.backoff_factor * 2**i
+                        )
                         continue
                     else:
                         raise e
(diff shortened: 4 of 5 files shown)


hr98w:
@fselmo as per this issue https://github.com/ethereum/web3.py/issues/3210
I was considering add a new field `initial_backoff_time` and re-use `backoff_factor` as the exponential factor initially. But this will affect existing usage of field `backoff_factor` and might cause backward compatibility issue. Please help review the PR and share your thoughts.

fselmo:
Hey @hr98w. Thanks for getting this started. I'm not sure that we need to add yet another configuration option here. If we make the sleep time itself exponential, the backoff factor can apply directly to that. Something like:

```py
for i in range(self.exception_retry_configuration.retries):
    try:
        return make_post_request(
            self.endpoint_uri, request_data, **self.get_request_kwargs()
        )
    except tuple(self.exception_retry_configuration.errors) as e:
        if i < self.exception_retry_configuration.retries - 1:
            time.sleep(
                # the `backoff_factor` works as a factor on the inherent
               # exponential setup of 2**i
                self.exception_retry_configuration.backoff_factor * 2**i
            )
            continue
        else:
            raise e
return None
```

Thoughts on that?

hr98w:
@fselmo sure, this also works.

I introduced a new variable to let users determine the exponential factor. Keeping that factor as 1 means no exponenial backoff. We can keep it simple here, will follow your advice.

fselmo:
> I introduced a new variable to let users determine the exponential factor. Keeping that factor as 1 means no exponenial backoff. We can keep it simple here, will follow your advice.

I hear you. That would allow for more configuration for sure. Exponential is pretty standard and keeping that factor low still allows for pretty quick succession if desired. If you're good with my simpler approach then let's go for it 👍🏼

fselmo:
We should maybe decrease the default value here to `0.125`? That would yield `.125` -> `.25` -> `.5` -> `1` -> `2`. That feels pretty good for JSON-RPC calls. Thoughts there?

hr98w:
Starting from 0.5 seems too large for me. But i'm not sure for JSON-RPC calls, any latency stats you can share?

hr98w:
@fselmo Updated the PR, please help review

docs/internals.rst
```diff
@@ -146,8 +146,8 @@ Retry Requests for HTTP Providers
 
 ``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
 on exceptions. This can be configured via configuration options on the provider
-instance. Below is an example showing the default options for the retry configuration
-and how to override them.
+instance. Retry backoff time would be ``backoff_factor * 2**retry_times``. Below is an
```
fselmo: I think we can be a bit less analytical here, use a touch more words, and explain that it doubles each time. This should help give a bit clearer picture on what sort of intervals the user would want to configure to and perhaps why we went with a number like `0.125` as the default.

```suggestion
instance. The retry mechanism employs an exponential backoff strategy, starting from the initial value determined by the ``backoff_factor``, and doubling the delay with each attempt, up to the ``retries`` value. Below is an
```

docs/internals.rst
```diff
@@ -165,7 +165,7 @@ and how to override them.
             # number of retries to attempt
             retries=5,
 
-            # how long to wait between retries
+            # time interval between retries would be backoff_factor * 2**retry_times
```
fselmo: I don't think the technical aspect is super important here either. A curious user can look at the code to see the implementation.

```suggestion
            # initial delay multiplier, doubles with each retry attempt
```

docs/internals.rst
```diff
@@ -165,7 +165,7 @@ and how to override them.
...
             backoff_factor=0.5,
```
fselmo: This should also be `0.125`

```suggestion
            backoff_factor=0.125,
```

newsfragments/3358.breaking.rst
```diff
@@ -0,0 +1 @@
+Update to exponential backoff for ``HTTPProvider`` and ``AsyncHTTPProvider``. Decrease default initial backoff time to 0.125 second.
```
fselmo:
```suggestion
Employ an exponential backoff strategy using the ``backoff_factor`` in ``ExceptionRetryConfiguration`` for ``HTTPProvider`` and ``AsyncHTTPProvider``. Reduce the default initial delay to ``0.125`` seconds.
```

fselmo (commented):
With a few tweaks to the wording this LGTM. I gave some suggestions to the wording. The important thing here is clarity and I don't think we need to be super technical in the documentation description since we can conceptually describe the doubling strategy. The technical details can be sniffed out by a curious user if they'd like to see the implementation. I'd love your thoughts on it though. Thanks again for getting the ball rolling on this... this is one of the final pieces for getting ``v7`` wrapped up 👍🏼

hr98w:
@fselmo Thanks for your comment! Got your point to make it easy to understand. Resolved them.

docs/internals.rst
```diff
@@ -146,8 +146,10 @@ Retry Requests for HTTP Providers
 
 ``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
 on exceptions. This can be configured via configuration options on the provider
-instance. Below is an example showing the default options for the retry configuration
-and how to override them.
+instance. The retry mechanism employs an exponential backoff strategy, starting from
+the initial value determined by the ``backoff_factor``, and doubling the delay with
+each attempt, up to the ``retries`` value.Below is an example showing the default
```
fselmo: Just one tiny nit to wrap this up. Thanks for the quick replies 👌🏼 

```suggestion
each attempt, up to the ``retries`` value. Below is an example showing the default
```

fselmo (approved):

fselmo:
Thanks again 👍🏼

PR #3359, opened 2024-04-23
Stricter validation for JSON-RPC responses

### LINKED ISSUE #3095: Validate JSON-RPC response `id` field more strictly
### What was wrong?

The new `WebsocketProviderV2` depends on the JSON-RPC 2.0 specification to be correct for providers. This means it depends on a request `id` matching with a response `id`. Because of the asynchronous nature of a websocket connection, and sometimes many-to-one response-to-request relationship (e.g. eth_subscribe), this is the only way to know how to format a response based on its request and to know which response from the socket to return for a particular request.

We don't currently validate that a response has an `id` very strictly. Starting with `v7` of web3.py, we should check if an `id` is present and if it isn't we should make sure that the response matches that of a subscription. If it doesn't, it should be an invalid response.

### How can it be fixed?

Validate whether an `RPCResponse` contains an `id` and if it doesn't, it is required to match all fields of an `eth_subscribe` subscription. There should be no reason for our library to make any assumptions based on malformed responses.

fselmo:
### What was wrong?

Closes #3095 by more tightly validating the JSON-RPC responses coming from providers (not just the `id` as the issue states) while providing good messaging on errors.

This PR also includes the following changes:

- Use `Web3RPCError` instead of a generic `Web3ValueError` for JSON-RPC errors returned by the provider.
- Error log the `user_message`, debug log the `response` for RPC errors for better clarity / debugging.
- In the spirit of wrapping up breaking changes for `v7` and snake-casing some final remnants of camelCase args, capitalize "Of": `BlockNumberOutofRange` -> `BlockNumberOutOfRange`
- Distinguish between `MethodUnavailable` (a JSON-RPC error from the node) and `MethodNotSupported` (*web3.py* does not support the method, internally, under some conditions).
- Re-organize the formatted_response tests to be more readable. Separate test cases into tests where they make sense together: test valid responses, test invalid fields, test valid error responses, test invalid error object, test error formatters and null formatters together.

### Todo:

- [x] Clean up commit history
- [x] Add or update documentation related to these changes (including the `v7` migration guide)

### DIFF
--- a/tests/core/manager/test_response_formatters.py
+++ b/tests/core/manager/test_response_formatters.py
@@ -4,6 +4,9 @@ import re
 from eth_utils.toolz import (
     identity,
 )
+from toolz import (
+    merge,
+)
 
 from web3._utils.method_formatters import (
     raise_block_not_found,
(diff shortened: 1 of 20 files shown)


web3/exceptions.py
```diff
@@ -40,6 +43,12 @@ def __init__(
         # Assign properties of Web3Exception
         self.user_message = user_message
 
+    def __str__(self) -> str:
+        # append a clarifying user message if one is provided
+        return super().__str__() + (
```
reedsa: I like this approach, though the attribute allows more flexibility in how users use it when they catch the exception. Having both seems fine with documentation.

It looks like we could do something similar with exceptions that inherit this, like for the `ContractLogicError` data attribute. Might be useful to append `data` in a `__str__` method there too.

web3/manager.py
```diff
@@ -117,6 +118,100 @@ def apply_null_result_formatters(
         return response
 
 
+def _validate_subscription_fields(response: RPCResponse) -> None:
+    params = response["params"]
+    subscription = params["subscription"]
+    if not isinstance(subscription, str) and not len(subscription) == 34:
+        _raise_bad_response_format(
```
reedsa, line 125: Similar to the `user_message`, Had a thought here that we could add a `response_property` and/or `response_property_value` or something to `BadResponseFormat`. The attribute(s) would save the specific field and/or value that caused the error (invalid value or missing field)

web3/types.py
```diff
@@ -295,7 +295,7 @@ class GethSyncingSubscriptionResponse(SubscriptionResponse):
 
 
 class RPCResponse(TypedDict, total=False):
-    error: Union[RPCError, str]
```
fselmo, line 298: Note: This change is also breaking but this is how it should be defined. The only reason we allowed `str` is bc of how `EthereumTesterProvider` returned errors but that should not be allowed imo.

pacrob (approved):
lgtm!

web3/manager.py
```diff
@@ -117,6 +118,111 @@ def apply_null_result_formatters(
...
+
+def _validate_response(
+    response: RPCResponse,
+    error_formatters: Optional[Callable[..., Any]],
+    is_subscription_response: bool = False,
+    logger: Optional[logging.Logger] = None,
+) -> None:
+    if "jsonrpc" not in response or response["jsonrpc"] != "2.0":
+        _raise_bad_response_format(
+            response, 'The "jsonrpc" field must be present with a value of "2.0".'
```
kclowes, line 138: I know that this is part of the JSON-RPC spec, but if you haven't already, it may be worth a quick provider check to make sure that this actually does get sent back with most major providers. I can imagine some test providers or something not sending this (or something else we're validating) back.

tests/core/manager/test_response_formatters.py
```diff
@@ -4,6 +4,9 @@
 from eth_utils.toolz import (
     identity,
 )
+from toolz import (
+    merge,
```
kclowes, line 6: this should be imported via `eth_utils` like identity above. There's some logic in eth-utils around using toolz or cytoolz

kclowes (approved):
I like that `BlockNumberOutOfRange` change, nice! I left a comment around making sure we've tested against a range of providers to make sure it makes sense to validate the rpc response. (I have scarring from [this chunk of code](https://github.com/ethereum/web3.py/blob/main/web3/_utils/error_formatters_utils.py#L57-L170), lol) I also left a nit about importing `merge` from eth-utils instead of from `toolz`. But once those are addressed, this LGTM!

tests/core/manager/test_response_formatters.py
fselmo: ah, yep!

web3/manager.py
fselmo: Tested against infura mainnet + sepolia, alchemy mainnet + sepolia, my local nethermind sepolia node, and polygon 👍🏼

PR #3370, opened 2024-04-29
V7 support for request batching

### LINKED ISSUE #832: Perform batch RPC requests to ethereum node
* Version: 4.2.0
* Python: 3.6
* OS: osx/linux/win

### What was wrong?

There's currently not support for batch requests, when using `HTTPProvider`. Even using IPC/WS connectors I think a speed up could be possible using batch requests.

### How can it be fixed?

Implementing batch requests

fselmo:
### What was wrong?

Closes #832

### How was it fixed?

Add batch request support to the `RequestManager` via a context manager. Expose this as a high-level API on the `Web3` / `AsyncWeb3` class itself (`w3.batch_requests()`).

### Todo:

- [x] Handle / skip `EthereumTesterProvider` test case
- [x] Handle IPC & WS provider batch request cases
- [x] Handle persistent connection provider cases
- [x] Clean up commit history
- [ ] Add or update documentation related to these changes -- separate PR ?

#### Cute Animal Picture

Screenshot 2024-04-30 at 15 38 09

### DIFF
--- a/tests/integration/go_ethereum/test_goethereum_http.py
+++ b/tests/integration/go_ethereum/test_goethereum_http.py
@@ -21,10 +21,11 @@ from .common import (
     GoEthereumAsyncEthModuleTest,
     GoEthereumAsyncNetModuleTest,
     GoEthereumAsyncTxPoolModuleTest,
+    GoEthereumAsyncWeb3ModuleTest,
     GoEthereumEthModuleTest,
     GoEthereumNetModuleTest,
-    GoEthereumTest,
     GoEthereumTxPoolModuleTest,
+    GoEthereumWeb3ModuleTest,
 )
 from .utils import (
     wait_for_aiohttp,
@@ -70,7 +71,7 @@ def w3(geth_process, endpoint_uri):
     return Web3(Web3.HTTPProvider(endpoint_uri))
 
 
-class TestGoEthereumTest(GoEthereumTest):
+class TestGoEthereumWeb3ModuleTest(GoEthereumWeb3ModuleTest):
     pass
 
 
@@ -114,6 +115,10 @@ async def async_w3(geth_process, endpoint_uri):
     await wait_for_aiohttp(endpoint_uri)
     _w3 = AsyncWeb3(AsyncHTTPProvider(endpoint_uri))
     return _w3
+
+
+class TestGoEthereumAsyncWeb3ModuleTest(GoEthereumAsyncWeb3ModuleTest):
+    pass
 
 
 class TestGoEthereumAsyncAdminModuleTest(GoEthereumAsyncAdminModuleTest):
--- a/web3/_utils/batching.py
+++ b/web3/_utils/batching.py
@@ -1,0 +1,86 @@
+from types import (
+    TracebackType,
+)
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Generic,
+    List,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
+
+from web3._utils.compat import (
+    Self,
+)
+from web3.method import (
+    Method,
+)
+from web3.types import (
+    TFunc,
+    TReturn,
+)
+
+if TYPE_CHECKING:
+    from web3 import (  # noqa: F401
+        AsyncWeb3,
+        Web3,
+    )
+    from web3.types import (  # noqa: F401
+        RPCEndpoint,
+        RPCResponse,
+    )
+
+BatchRequestInformation = Tuple[Tuple["RPCEndpoint", Any], Sequence[Any]]
+
+
+class BatchRequestContextManager(Generic[TFunc]):
+    def __init__(self, web3: Union["AsyncWeb3", "Web3"]) -> None:
+        self.web3 = web3
+        self._requests_info: List[BatchRequestInformation] = []
+
+    def add(self, batch_payload: Union[TReturn, Sequence[TReturn]]) -> None:
+        # When batching, we don't make a request. Instead, we will get the request
+        # information and store it in the _requests_info list.
+        if isinstance(batch_payload, Sequence):
+            for request_information in batch_payload:
+                self._requests_info.append(
+                    cast(BatchRequestInformation, request_information)
+                )
+        else:
+            self._requests_info.append(cast(BatchRequestInformation, batch_payload))
+
+    def __enter__(self) -> Self:
+        self.web3._is_batching = True
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_val: BaseException,
+        exc_tb: TracebackType,
+    ) -> None:
+        self.web3._is_batching = False
+
+    def execute(self) -> List["RPCResponse"]:
+        return self.web3.manager.make_batch_request(self._requests_info)
+
+    # -- async -- #
+
+    async def __aenter__(self) -> Self:
+        self.web3._is_batching = True
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_val: BaseException,
+        exc_tb: TracebackType,
+    ) -> None:
+        self.web3._is_batching = False
+
+    async def async_execute(self) -> List["RPCResponse"]:
+        return await self.web3.manager.async_make_batch_request(self._requests_info)
--- a/web3/_utils/batching.py
+++ b/web3/_utils/batching.py
@@ -50,6 +50,9 @@ if TYPE_CHECKING:
     from web3.providers.async_base import (  # noqa: F401
         AsyncJSONBaseProvider,
     )
+    from web3.providers.base import (  # noqa: F401
+        JSONBaseProvider,
+    )
     from web3.types import (  # noqa: F401
         RPCEndpoint,
         RPCResponse,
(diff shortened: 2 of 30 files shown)


fselmo:
This is good for an initial review. It's still missing documentation but that may be best in a separate PR? Thoughts there? Usage examples can be seen in the tests for now.

tests/integration/go_ethereum/test_goethereum_http.py
```diff
@@ -70,7 +71,7 @@ def w3(geth_process, endpoint_uri):
     return Web3(Web3.HTTPProvider(endpoint_uri))
 
 
-class TestGoEthereumTest(GoEthereumTest):
+class TestGoEthereumWeb3ModuleTest(GoEthereumWeb3ModuleTest):
```
pacrob, line 74: There's a lot of this renaming in the integration tests. Is it just cleaning up for clarity, or more relevant in a way I'm not seeing?

fselmo: The biggest change is making a proper `AsyncWeb3ModuleTest` and then yeah the renaming is for clarity. The old `Web3ModuleTest` was checking all the static utility methods for conversions, etc. Since we have some actual async methods on the `AsyncWeb3` class directly now, this PR adds a separation of logic between sync and async and gets rid of the `@pytest.mark.parametrize` that cycled between `Web3` and `AsyncWeb3`, opting for two separate test classes instead.

web3/_utils/batching.py
```diff
@@ -0,0 +1,171 @@
...
+    "eth_unsubscribe",
+    "eth_sendRawTransaction",
+    "eth_sendTransaction",
+    "eth_signTransaction",
+    "eth_sign",
+    "eth_signTypedData",
+}
+
+
+class BatchRequestContextManager(Generic[TFunc]):
```
fselmo: Now that it's no longer just a context manager, e.g.:

```py
batch = w3.batch_requests()
batch.add(...)
batch.add(...)
result1, result2 = batch.execute()
```

... this class should probably be renamed to a more general name like `RequestBatcher`. We already have `RequestManager` and `RequestProcessor` classes. `RequestBatcher` seems like a fitting enough name.

---
edit: done in latest commit

web3/contract/utils.py
```diff
@@ -60,10 +66,50 @@
...
     )
+    from web3.providers.persistent import (  # noqa: F401
+        PersistentConnectionProvider,
+    )
 
 ACCEPTABLE_EMPTY_STRINGS = ["0x", b"0x", "", b""]
 
 
+@curry
+def format_contract_call_return_data_curried(
```
kclowes, line 77: I didn't think about this too hard, but could we repurpose `call_contract_function` to have a `curry` decorator? It looks like the logic and inputs are fairly similar?

web3/contract/utils.py
```diff
@@ -108,6 +154,23 @@ def call_contract_function(
...
+        resp_formatters = compose(
+            format_contract_call_return_data_curried(
+                w3,
+                decode_tuples,
+                fn_abi,
+                function_identifier,
+                normalizers,
+                output_types,
+            ),
+            return_data[1][0],
```
fselmo: make this more readable

web3/contract/utils.py
```diff
@@ -319,6 +382,33 @@ async def async_call_contract_function(
...
+    if async_w3.provider._is_batching:
+        formatter = format_contract_call_return_data_curried(
+            async_w3,
+            decode_tuples,
+            fn_abi,
+            function_identifier,
+            normalizers,
+            output_types,
+        )
+        if async_w3.provider.has_persistent_connection:
```
fselmo, line 405: make this more readable pls

web3/providers/ipc.py
```diff
@@ -189,6 +192,47 @@ def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
                         timeout.sleep(0)
                         continue
 
+    def make_batch_request(
```
fselmo, line 200: refactor similar logic to `make_request()` (on all providers where this isn't yet done)

web3/method.py
```diff
@@ -149,15 +153,29 @@ def __init__(
...
+            if self.json_rpc_method in RPC_METHODS_UNSUPPORTED_DURING_BATCH:
+                raise MethodNotSupported(
+                    f"Method `{self.json_rpc_method}` is not supported within a batch "
+                    "request."
+                )
+            return module.retrieve_request_information(self)
+        else:
+            return module.retrieve_caller_fn(self)
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
```
fselmo, line 179: ~~Think about removing this?~~ [see below]

fselmo (commented):
From team code review 5/6/24

web3/method.py
fselmo: For clarity, this explicitly makes `Method` class callable... otherwise we get mypy conflicts, among other issues, such as:

```
web3/_utils/batching.py:128: error: "Method[Callable[..., Any]]" not callable  [operator]
```

The `batching.py` line in question:

```py
       for method, params in batch_payload.items():
            for param in params:
                self.add(method(param))  # <--- conflicting line
```

web3/providers/persistent/request_processor.py
```diff
@@ -217,6 +232,28 @@ def get_request_information_for_response(
...
+    ) -> None:
+        cache_key = generate_cache_key(request_id)
+        cached_request_info_for_id: RequestInformation = (
+            self._request_information_cache.get_cache_entry(cache_key)
+        )
+        if cached_request_info_for_id is not None:
+            cached_request_info_for_id.response_formatters = (
+                compose(
+                    result_formatter,
+                    cached_request_info_for_id.response_formatters[0],
```
kclowes: What do you think about making these `response_formatters` their own attributes instead of a tuple? Then we could reference like:
```suggestion
                    cached_request_info_for_id.result_formatters,
                    cached_request_info_for_id.error_formatters,
                    ...
```

pacrob (approved):
I like it!

kclowes (approved):
This is awesome! Feel free to take or leave any comments I left. I wonder if we make the RequestsInformation object a little more robust we could avoid all of the indexing - [here](https://github.com/ethereum/web3.py/pull/3370/files#diff-c902d776bb4f676fa5cca4ee81273cd66d1113db8109d4f607167681b0b8bdeaR467) for example. I was thinking maybe have `RequestsInfo.method`, `RequestsInfo.params`, `RequestInfo.result_formatters`, `RequestInfo.error_formatters`, etc. We could either pass around the object, or just whatever attributes each function needs. If that doesn't make sense, feel free to push back :)

web3/contract/utils.py
fselmo: Hmm I'm not sure. It's really only the `return_data` that we want to be able to pass in at a later time, everything else is preconfigured in the curried method. Since `call_contract_function` doesn't take the `return_data` argument I don't see how it can replace this. Is that kind of what you meant?

web3/providers/persistent/request_processor.py
fselmo: Ah, yeah that could make this more readable for sure. I like this. I'll play around with that idea.

web3/contract/utils.py
kclowes: oh, gotcha. don't mind me :)

web3/providers/persistent/request_processor.py
fselmo: Since I think it's only in this part of the code base that unpacks them like this, what do you think about [unpacking the values](https://github.com/ethereum/web3.py/pull/3370/commits/56adab130945a29fd5f4d62990f54557c4792889) for better readability?

web3/providers/persistent/request_processor.py
```diff
@@ -95,7 +102,11 @@ def cache_request_information(
         self,
         method: RPCEndpoint,
         params: Any,
-        response_formatters: Tuple[Callable[..., Any], ...],
+        response_formatters: Tuple[
```
reedsa, line 105: Might be good to create a new type for this, something like `ResponseFormatters`.

web3/providers/persistent/request_processor.py
kclowes: That works for me!

web3/module.py
```diff
@@ -55,9 +57,43 @@ def apply_result_formatters(
...
 
+@curry
+def retrieve_request_information_for_batching(
+    w3: Union["AsyncWeb3", "Web3"],
+    module: "Module",
+    method: Method[Callable[..., Any]],
+) -> Union[
+    Callable[..., Tuple[Tuple[RPCEndpoint, Any], Sequence[Any]]],
+    Callable[..., Coroutine[Any, Any, Tuple[Tuple[RPCEndpoint, Any], Sequence[Any]]]],
+]:
```
reedsa, line 68: Maybe worth creating a type for `RequestInformation` and use that inside the `Callable[...]`s. There are a few places these are used. It could end up something like:
```suggestion
) -> Union[
    Callable[..., RequestInformation],
    Callable[..., Coroutine[Any, Any, RequestInformation],
]:
```

reedsa (approved):
Minor suggestions for typing but lgtm!

web3/providers/persistent/request_processor.py
fselmo: Yeah... these exist elsewhere in the code base and could honestly use some love throughout all of it. Separate issue to track maybe?

PR #3378, opened 2024-05-03
Propagate listener task exceptions to the main loop.

### LINKED ISSUE #3375: Async iteration of AsyncWeb3 doesn't work as expected
### What happened?

I am using the latest stable release - 6.18.0

I'm trying out the AsyncWeb3 class following the usage as described in the docs (https://web3py.readthedocs.io/en/stable/providers.html#usage). I would like to maintain an indefinite subscription, automatically reconnecting where necessary due to connection drops. From my understanding this should be possible via the second example using the `async for w3 in AsyncWeb3.persistent_websocket(...` pattern, however this does not work as expected for me.

Using that example, I am able to connect and retrieve some values, however after a short while when receiving a ConnectionClosedError the connection is not automatically retried. Instead, the main asyncio task is cancelled meaning that I am unable to retry by catching `websockets.ConnectionClosed`.

```python
import asyncio
import logging

from websockets.exceptions import ConnectionClosed
from web3 import AsyncWeb3


logging.basicConfig(level=logging.INFO)


async def main():
    async for w3 in AsyncWeb3.persistent_websocket(
        WebsocketProviderV2(...)
    ):
        try:
            subscription_id = await w3.eth.subscribe("newHeads")

            async for response in w3.ws.process_subscriptions():
                block_number = response["result"]["number"]
                logging.info("block number: %d", block_number)
        except ConnectionClosed:
            logging.info("connection closed, retrying")


if __name__ == "__main__":
    asyncio.run(main())
```

### Full error output

```shell
File "/Users/ryancollingham/code/block-scholes/defiDataGrabber/main.py", line 24, in <module>
    asyncio.run(main())
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/code/block-scholes/defiDataGrabber/main.py", line 19, in main
    except ConnectionClosed:
           ^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/code/block-scholes/defiDataGrabber/repro.py", line 21, in main
    async for response in w3.ws.process_subscriptions():
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/manager.py", line 454, in __anext__
    return await self.manager._get_next_ws_message()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/manager.py", line 367, in _get_next_ws_message
    return await self._ws_message_stream().__anext__()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/manager.py", line 380, in _ws_message_stream
    response = await self._request_processor.pop_raw_response(subscription=True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/providers/websocket/request_processor.py", line 230, in pop_raw_response
    raw_response = await self._subscription_response_queue.get()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/queues.py", line 158, in get
    await getter
asyncio.exceptions.CancelledError
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-5' coro=<WebsocketProviderV2._ws_message_listener() done, defined at /Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/providers/websocket/websocket_v2.py:212> exception=ConnectionClosedError(None, Close(code=<CloseCode.INTERNAL_ERROR: 1011>, reason='keepalive ping timeout'), None)>
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 1301, in close_connection
    await self.transfer_data_task
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 963, in transfer_data
    message = await self.read_message()
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 1033, in read_message
    frame = await self.read_data_frame(max_size=self.max_size)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 1108, in read_data_frame
    frame = await self.read_frame(max_size)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 1165, in read_frame
    frame = await Frame.read(
            ^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/framing.py", line 68, in read
    data = await reader(2)
           ^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/streams.py", line 750, in readexactly
    await self._wait_for_data('readexactly')
  File "/Users/ryancollingham/.pyenv/versions/3.11.8/lib/python3.11/asyncio/streams.py", line 543, in _wait_for_data
    await self._waiter
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/providers/websocket/websocket_v2.py", line 236, in _ws_message_listener
    raise e
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/web3/providers/websocket/websocket_v2.py", line 223, in _ws_message_listener
    async for raw_message in self._ws:
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 498, in __aiter__
    yield await self.recv()
          ^^^^^^^^^^^^^^^^^
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 568, in recv
    await self.ensure_open()
  File "/Users/ryancollingham/.pyenv/versions/defiDataGrabber/lib/python3.11/site-packages/websockets/legacy/protocol.py", line 948, in ensure_open
    raise self.connection_closed_exc()
websockets.exceptions.ConnectionClosedError: sent 1011 (internal error) keepalive ping timeout; no close frame received
```

It seems like the error handling here is problematic: https://github.com/ethereum/web3.py/blob/08665fcd54e0e593ba6d19a884b4ba8c9b5753d0/web3/providers/persistent/websocket.py#L198-L209

Since the listener is run in a separate background task, when it cancels all tasks this prevents the ConnectionClosed error from being caught and handled by reconnecting.

### web3 Version

6.18.0

### Python Version

3.11.8

### Operating System

osx

### Output from `pip freeze`

```shell
aiohttp==3.9.5
aiosignal==1.3.1
attrs==23.2.0
bitarray==2.9.2
black==24.4.2
certifi==2024.2.2
charset-normalizer==3.3.2
ckzg==1.0.1
click==8.1.7
cytoolz==0.12.3
eth-account==0.11.2
eth-hash==0.7.0
eth-keyfile==0.8.1
eth-keys==0.5.1
eth-rlp==1.0.1
eth-typing==4.2.2
eth-utils==4.1.0
eth_abi==5.1.0
frozenlist==1.4.1
greenlet==3.0.3
hexbytes==0.3.1
idna==3.7
iniconfig==2.0.0
jsonschema==4.22.0
jsonschema-specifications==2023.12.1
lru-dict==1.2.0
msgpack==1.0.8
multidict==6.0.5
mypy==1.10.0
mypy-extensions==1.0.0
neovim==0.3.1
packaging==24.0
parsimonious==0.10.0
pathspec==0.12.1
platformdirs==4.2.1
pluggy==1.5.0
protobuf==5.26.1
pycryptodome==3.20.0
pynvim==0.5.0
pytest==8.2.0
pyunormalize==15.1.0
referencing==0.35.1
regex==2024.4.28
requests==2.31.0
rlp==4.0.1
rpds-py==0.18.0
ruff==0.4.2
toolz==0.12.1
typing_extensions==4.11.0
urllib3==2.2.1
web3==6.18.0
websockets==12.0
yarl==1.9.4
```

fselmo:
### What was wrong?

Closes #3375 (along with #3387)

### How was it fixed?

As I understand it, the way asyncio works is each task operates under its own context. Directly communicating with another context and trying to immediately propagate an exception to the main loop, for example, does not seem to be a very straightforward thing to do surprisingly. For better or worse, I think when we poll for messages we should check if the listener task is done and, if it is, if an exception was set. If an exception was set then we can raise it in the main loop where we are polling for messages.

- When we poll for messages, poll to see if the background task is finished. If it is, handle any exceptions in the main loop. On subscription streams, since we hang awaiting a message to appear in the subscription queue, we need to signal when the background task is `done` since no messages will be coming in and we will hang indefinitely. Push a new exception `TaskNotRunning` to the queue. Revamp the queue to raise when this is pushed to it. Handle this exception in the subscription polling.
- Refactor some of the logic. When disconnecting persistent connection providers, shut down the listener task before closing the connection to avoid any sort of errors when the task is still trying to read from the closing connection.

---

I added some nice logging that really shouldn't be too noisy unless the connection is reconnecting all the time (which I forced for the example here).

```
INFO:web3.providers.WebSocketProvider:Connecting to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:Successfully connected to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:WebSocketProvider listener background task started. Storing all messages in appropriate request processor queues / caches to be processed.
newHeads sub_id: 0xad65c6e2aefa417dafda6999d0bd774b
INFO:web3.providers.WebSocketProvider:Subscription response queue synced with websocket message stream.
blocknum: 19829151
blocknum: 19829152
ERROR:web3.providers.WebSocketProvider:Connection interrupted, attempting to reconnect...
INFO:web3.providers.WebSocketProvider:Message listener background task successfully shut down.
INFO:web3.providers.WebSocketProvider:Successfully disconnected from: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:Connecting to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:Successfully connected to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:WebSocketProvider listener background task started. Storing all messages in appropriate request processor queues / caches to be processed.
newHeads sub_id: 0x718852a668ed437fbdc86243db929405
blocknum: 19829153
blocknum: 19829154
ERROR:web3.providers.WebSocketProvider:Connection interrupted, attempting to reconnect...
INFO:web3.providers.WebSocketProvider:Message listener background task successfully shut down.
INFO:web3.providers.WebSocketProvider:Successfully disconnected from: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:Connecting to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:Successfully connected to: wss://mainnet.infura.io/ws/v3/c364...
INFO:web3.providers.WebSocketProvider:WebSocketProvider listener background task started. Storing all messages in appropriate request processor queues / caches to be processed.
newHeads sub_id: 0xf46690d2853c4afba527ac71dd1915e0
blocknum: 19829155
...
```
---
Bonus:

I've been meaning to do this for some time and I think this is an appropriate time. Refactor the commonality of the listener task down to the base `PersistentConnectionProvider` class and define provider-specific methods that either need to be implemented or can be overridden to fine-tune logic specific to each persistent connection provider. This keeps things quite a bit DRYer.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/web3/_utils/module_testing/module_testing_utils.py
+++ b/web3/_utils/module_testing/module_testing_utils.py
@@ -177,6 +177,12 @@ class WebSocketMessageStreamMock:
         self.messages = deque(messages) if messages else deque()
         self.raise_exception = raise_exception
 
+    def __await__(self) -> Generator[Any, Any, "Self"]:
+        async def __async_init__() -> "Self":
+            return self
+
+        return __async_init__().__await__()
+
     def __aiter__(self) -> "Self":
         return self
 
@@ -188,6 +194,13 @@ class WebSocketMessageStreamMock:
             raise StopAsyncIteration
 
         return self.messages.popleft()
+
+    @staticmethod
+    async def pong() -> bool:
+        return False
+
+    async def connect(self) -> None:
+        pass
 
     async def send(self, data: bytes) -> None:
         pass
--- a/web3/_utils/module_testing/module_testing_utils.py
+++ b/web3/_utils/module_testing/module_testing_utils.py
@@ -196,7 +196,7 @@ class WebSocketMessageStreamMock:
         return self.messages.popleft()
 
     @staticmethod
-    async def pong() -> bool:
+    async def pong() -> Literal[False]:
         return False
 
     async def connect(self) -> None:
--- a/web3/main.py
+++ b/web3/main.py
@@ -85,6 +85,7 @@ from web3.eth import (
     Eth,
 )
 from web3.exceptions import (
+    ProviderConnectionError,
     Web3TypeError,
     Web3ValueError,
 )
@@ -511,12 +512,18 @@ class AsyncWeb3(BaseWeb3):
         "when instantiating via ``async for``."
     )
     async def __aiter__(self) -> AsyncIterator[Self]:
-        if not await self.provider.is_connected():
-            await self.provider.connect()
-
         while True:
             try:
+                await self.provider.connect()
                 yield self
+            except ProviderConnectionError:
+                # We use an exponential retry in `connect()` so if we raise a
+                # `ProviderConnectionError`, there is likely an issue with the provider
+                # that will likely not be resolved by continuing to iterate.
+                raise
+            except KeyboardInterrupt:
+                # If the user interrupts the iteration, stop iterating.
+                break
             except Exception:
-                # provider should handle connection / reconnection
+                # If we encounter any other exception, continue iterating.
                 continue
--- a/web3/main.py
+++ b/web3/main.py
@@ -85,7 +85,6 @@ from web3.eth import (
     Eth,
 )
 from web3.exceptions import (
-    ProviderConnectionError,
     Web3TypeError,
     Web3ValueError,
 )
@@ -513,17 +512,6 @@ class AsyncWeb3(BaseWeb3):
     )
     async def __aiter__(self) -> AsyncIterator[Self]:
         while True:
-            try:
-                await self.provider.connect()
-                yield self
-            except ProviderConnectionError:
-                # We use an exponential retry in `connect()` so if we raise a
-                # `ProviderConnectionError`, there is likely an issue with the provider
-                # that will likely not be resolved by continuing to iterate.
-                raise
-            except KeyboardInterrupt:
-                # If the user interrupts the iteration, stop iterating.
-                break
-            except Exception:
-                # If we encounter any other exception, continue iterating.
-                continue
+            await self.provider.connect()
+            yield self
+            await self.provider.disconnect()
--- a/web3/main.py
+++ b/web3/main.py
@@ -144,6 +144,7 @@ from web3.types import (
 
 if TYPE_CHECKING:
     from web3._utils.empty import Empty  # noqa: F401
+    from web3.providers.persistent import PersistentConnectionProvider  # noqa: F401
 
 
 def get_async_default_modules() -> Dict[str, Union[Type[Module], Sequence[Any]]]:
@@ -511,7 +512,11 @@ class AsyncWeb3(BaseWeb3):
         "when instantiating via ``async for``."
     )
     async def __aiter__(self) -> AsyncIterator[Self]:
+        provider = self.provider
         while True:
-            await self.provider.connect()
+            await provider.connect()
             yield self
-            await self.provider.disconnect()
+            cast("PersistentConnectionProvider", provider).logger.error(
+                "Connection interrupted, attempting to reconnect..."
+            )
+            await provider.disconnect()
--- a/web3/providers/persistent/websocket.py
+++ b/web3/providers/persistent/websocket.py
@@ -147,9 +147,10 @@ class WebSocketProvider(PersistentConnectionProvider):
         if self._message_listener_task is not None:
             try:
                 self._message_listener_task.cancel()
-                await self._message_listener_task
             except (asyncio.CancelledError, StopAsyncIteration):
                 pass
+            finally:
+                self._message_listener_task = None
 
         if self._ws is not None and not self._ws.closed:
             await self._ws.close()
@@ -178,7 +179,7 @@ class WebSocketProvider(PersistentConnectionProvider):
 
         return response
 
-    async def _message_listener_provider_specific_logic(self) -> None:
+    async def _provider_specific_message_listener(self) -> None:
         async for raw_message in self._ws:
             await asyncio.sleep(0)
 
--- a/web3/providers/persistent/websocket.py
+++ b/web3/providers/persistent/websocket.py
@@ -144,6 +144,13 @@ class WebSocketProvider(PersistentConnectionProvider):
                 _backoff_time *= _backoff_rate_change
 
     async def disconnect(self) -> None:
+        if self._message_listener_task is not None:
+            try:
+                self._message_listener_task.cancel()
+                await self._message_listener_task
+            except (asyncio.CancelledError, StopAsyncIteration):
+                pass
+
         if self._ws is not None and not self._ws.closed:
             await self._ws.close()
             self._ws = None
@@ -151,11 +158,6 @@ class WebSocketProvider(PersistentConnectionProvider):
                 f'Successfully disconnected from endpoint: "{self.endpoint_uri}'
             )
 
-        try:
-            self._message_listener_task.cancel()
-            await self._message_listener_task
-        except (asyncio.CancelledError, StopAsyncIteration):
-            pass
         self._request_processor.clear_caches()
 
     @async_handle_request_caching
@@ -197,9 +199,6 @@ class WebSocketProvider(PersistentConnectionProvider):
                     )
             except Exception as e:
                 if not self.silence_listener_task_exceptions:
-                    loop = asyncio.get_event_loop()
-                    for task in asyncio.all_tasks(loop=loop):
-                        task.cancel()
                     raise e
 
                 self.logger.error(
--- a/web3/providers/persistent/websocket.py
+++ b/web3/providers/persistent/websocket.py
@@ -61,7 +61,6 @@ class WebSocketProvider(PersistentConnectionProvider):
     logger = logging.getLogger("web3.providers.WebSocketProvider")
     is_async: bool = True
 
-    _max_connection_retries: int = 5
     _ws: Optional[WebSocketClientProtocol] = None
 
     def __init__(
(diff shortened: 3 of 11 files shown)


fselmo:
I'm actually going to re-structure this a bit. Will ping for review when ready again.

fselmo:
@kclowes @pacrob @reedsa This is ready for review again

fselmo:
Just kidding this is still in progress 😔

fselmo:
@kclowes @pacrob @reedsa Hey it's me again. This is ready now 😄. We were missing a good test for subscription polling with the iterator pattern. This would've failed or hanged forever. That should be addressed now.

web3/_utils/module_testing/module_testing_utils.py
```diff
@@ -189,6 +195,13 @@ async def __anext__(self) -> bytes:
 
         return self.messages.popleft()
 
+    @staticmethod
+    async def pong() -> bool:
```
kclowes: nit: 
```suggestion
    async def pong() -> False:
```

web3/providers/persistent/websocket.py
```diff
@@ -116,47 +115,13 @@ async def is_connected(self, show_traceback: bool = False) -> bool:
...
-            self.logger.debug(
-                f'Successfully disconnected from endpoint: "{self.endpoint_uri}'
-            )
-
-        try:
-            self._message_listener_task.cancel()
-            await self._message_listener_task
-        except (asyncio.CancelledError, StopAsyncIteration):
-            pass
-        self._request_processor.clear_caches()
```
kclowes, line 159: I think we lost this line in the persistent provider's `disconnect`. Was that on purpose?

web3/main.py
```diff
@@ -511,12 +512,11 @@ async def __aexit__(
         "when instantiating via ``async for``."
     )
     async def __aiter__(self) -> AsyncIterator[Self]:
-        if not await self.provider.is_connected():
```
kclowes, line 514: Does this check matter? I can't decide, ha

kclowes (approved):
Nice tests and nice refactor! This looks good to me overall. Pushing `None` to the queue is the only thing that stands out as something that might be a little foot-gun-y. In my head, it seems like it would be more descriptive and maybe less error prone if we could raise a custom exception there, but I can't quite see the path to how that could work. So I'm good leaving as-is, but wanted to flag in case you saw a nice solution off the bat.

web3/main.py
fselmo: I think we should try to keep `connect` where it re-creates the state each time and force a re-connect on each iteration to make sure we are creating a clean slate. Thoughts there?

web3/providers/persistent/websocket.py
fselmo: Nope, this should be there! Thanks for catching. I'll see if I can fit this into a test too.

fselmo:
> Pushing None to the queue is the only thing that stands out as something that might be a little foot-gun-y.

I think you're absolutely right. I don't see `None` being pushed to the cache in another way but I think we can isolate this case to be more unique. If I can't, I think we can still use `None` to un-brick the awaiting of a message and we can make sure we make the proper checks of whether or not the task really did finish and if it didn't we just move on to the next iteration of the subscription messages - essentially causing a no-op in that scenario.

web3/main.py
fselmo: Also, since we're disconnecting after yielding now, I think this makes more sense.

kclowes: okay, cool. That makes sense to me!

fselmo:
@kclowes I came up with a more elegant solution for the queue to operate correctly as long as the listener task is running

> In my head, it seems like it would be more descriptive and maybe less error prone if we could raise a custom exception there

taking this as inspiration, lmk what you think: a4790d6

...other comments addressed in 8ed6df9

kclowes:
> I came up with a more elegant solution for the queue to operate correctly as long as the listener task is running

super elegant 🤩  🚀

kclowes (approved):

PR #3385, opened 2024-05-09
move v6 windows wheel check to py312

pacrob:
### What was wrong?

Circle updated their default image to have py312 instead of py311, causing CI to fail.

### How was it fixed?

Updated existing windows wheel job to use py312 instead (rather than importing the new windows wheel structure that exists in v7 (which won't have this problem))

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -786,11 +786,6 @@ jobs:
         environment:
           TOXENV: py310-wheel-cli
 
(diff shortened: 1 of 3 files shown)


fselmo (approved):
lgtm 👍🏼

reedsa (approved):
Nice work, lgtm!

PR #3395, opened 2024-05-14
Allow provider init with cache configuration & add tests

### LINKED ISSUE #3394: Request Caching can't be configured via providers classes constructors
### What happened?

Following [v7 docs](https://web3py.readthedocs.io/en/latest/internals.html#request-caching), request caching should be enabled using `cache_allowed_requests` and RPC cacheable methods using `cacheable_requests`, but as of today the params are not available in the init of the RPC providers classes.

```python

w3 = Web3(HTTPProvider(
    endpoint_uri="...",

    # optional flag to turn on cached requests, defaults to False
    cache_allowed_requests=True,

    # optional, defaults to an internal list of deemed-safe-to-cache endpoints
    cacheable_requests={"eth_chainId", "eth_getBlockByNumber"},
))
```

### Full error output

```shell
  File "/mnt/c/Users/guill/AppData/Roaming/JetBrains/PyCharm2024.1/scratches/scratch_7.py", line 3, in <module>
    w3 = Web3(HTTPProvider(
              ^^^^^^^^^^^^^
TypeError: HTTPProvider.__init__() got an unexpected keyword argument 'cache_allowed_requests'
```

Add `cache_allowed_requests` and `cacheable_requests` arguments to `__init__()` of `BaseProvider`. The properties of the class already exist but it's not configurable via parameters.

### web3 Version

7.0.0-b2f5574

### Python Version

3.12

### Operating System

linux

### Output from `pip freeze`

_No response_

fselmo:
### What was wrong?

Closes #3394 

### How was it fixed?

- Allow provider init with request cache configuration
- Add test to test that all providers init without caching by default and can be instantiated with caching configs

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -141,7 +141,6 @@ class IPCProvider(JSONBaseProvider):
         self,
         ipc_path: Union[str, Path] = None,
         timeout: int = 30,
-        *args: Any,
         **kwargs: Any,
     ) -> None:
         if ipc_path is None:
@@ -154,7 +153,7 @@ class IPCProvider(JSONBaseProvider):
         self.timeout = timeout
         self._lock = threading.Lock()
         self._socket = PersistantSocket(self.ipc_path)
-        super().__init__()
+        super().__init__(**kwargs)
 
     def __str__(self) -> str:
         return f"<{self.__class__.__name__} {self.ipc_path}>"
--- a/docs/gas_price.rst
+++ b/docs/gas_price.rst
@@ -110,10 +110,10 @@ Available gas price strategies
 
     .. code-block:: python
 
-        from web3 import Web3, middleware
+        from web3 import Web3
         from web3.gas_strategies.time_based import medium_gas_price_strategy
 
-        w3 = Web3()
+        w3 = Web3(...)
         w3.eth.set_gas_price_strategy(medium_gas_price_strategy)
 
         w3.provider.cache_allowed_requests = True
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -184,7 +184,7 @@ The following properties are available on the ``web3.eth`` namespace.
 
        .. code-block:: python
 
-          >>> w3.provider.cache_allowed_requests
+          >>> w3.provider.cache_allowed_requests = True
 
 
 .. _web3-eth-methods:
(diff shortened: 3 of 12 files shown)


clover-es:
Maybe tests [`test_caching_requests_does_not_share_state_between_providers`](https://github.com/fselmo/web3.py/blob/5ddca0ec5e415942f1c961f7d2370476b6b4d406/tests/core/caching-utils/test_request_caching.py#L118) and [`test_async_request_caching_does_not_share_state_between_providers`](https://github.com/fselmo/web3.py/blob/5ddca0ec5e415942f1c961f7d2370476b6b4d406/tests/core/caching-utils/test_request_caching.py#L258) should be adapted as well?

I know previously wasn't tested either (probably that was an error), but I think that if the purpose is to test whether cache is shared, they should be initialized with `cache_allowed_requests=True`.

reedsa (approved):
nice work, lgtm!

fselmo:
> I know previously wasn't tested either (probably that was an error), but I think that if the purpose is to test whether cache is shared, they should be initialized with `cache_allowed_requests=True`.

That is definitely the case, good catch. Will add that in here.

---
edit: I also added to those tests the case where a cache _is_ shared, as a sanity check and for quicker code readability.

web3/providers/ipc.py
```diff
@@ -154,7 +154,7 @@ def __init__(
         self.timeout = timeout
         self._lock = threading.Lock()
         self._socket = PersistantSocket(self.ipc_path)
-        super().__init__()
+        super().__init__(**kwargs)
```
kclowes, line 156: Should we pass `*args` here too? Or maybe remove it from the signature if not?

kclowes (approved):
LGTM! Just left a small comment on the IPCProvider ⛵

web3/providers/ipc.py
fselmo: I think removing from signature is probably the best route. I don't think we should define any args in the base classes since it would be hard to envision arg positioning to pass along as a user. Best to more explicitly set certain properties on the class, via kwargs, imo. Thoughts on that?

kclowes: Yeah, removing it makes sense to me!

PR #3408, opened 2024-05-23
Function ABI Utilities

reedsa:
### What was wrong?

Function ABI utilties are needed for some use cases, so exposing them as public methods will allow people to work with ABIs and encode/decode function arguments.

Related to Issue #3036, #1596

Dependent upon https://github.com/ethereum/eth-utils/pull/271.

### How was it fixed?

Implement functions for parsing and validating contract data using the ABI spec.

Public ABI methods in `web3.utils.abi`:
    `check_if_arguments_can_be_encoded`,
    `get_abi_element_info`,
    `get_abi_element`,
    `get_event_abi`,
    `get_event_log_topics`,
    `log_topic_to_bytes`,

Additionally, all `eth_utils.abi` functions are included as part of the `web3.utils` module.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


web3/utils/abi.py
```diff
@@ -1,21 +1,242 @@
...
+    "`{0}` is deprecated and will be removed in a future release. Use " "`{1}` instead."
+)
+
 
+@deprecated(
+    DEPRECATION_MESSAGE.format(
+        "get_abi_input_names", "eth_utils.abi.get_abi_output_names"
+    )
+)
+@deprecated("Use eth_utils.abi.get_abi_input_names")
```
fselmo: `@deprecated` is on here twice.

But also, do we need this? I thought we would re-expose it here and do something like:

```py
from eth_utils import get_abi_input_names
```

and leave it at that. Since this is for `v7`, and we're in beta, I'd don't think we need to deprecate it and we can just re-expose it here? What are your thoughts on that?

newsfragments/3408.deprecation.rst
```diff
@@ -0,0 +1,4 @@
+Deprecated ``web3.utils.abi`` utils and types.
```
fselmo: This newsfragment pattern doesn't display as I _think_ was intended, based on the format that's here. We should either revise it or accept that it will be one sentence followed by a text block or one sentence after another.

web3/utils/abi.py
```diff
@@ -1,21 +1,242 @@
...
     if "inputs" not in abi and abi["type"] == "fallback":
         return []
     return [arg["name"] for arg in abi["inputs"]]
 
 
+@deprecated(
+    DEPRECATION_MESSAGE.format(
+        "get_abi_output_names", "eth_utils.abi.get_abi_output_names"
+    )
+)
```
fselmo, line 84: Same here. Get rid of this implementation entirely for `v7` and just import it from *eth_utils* to re-expose? Either way, deprecation should not be necessary for `v7` beta unless I'm missing context.

web3/_utils/abi.py
```diff
@@ -385,6 +386,7 @@ def check_if_arguments_can_be_encoded(
     )
 
 
+@deprecated_for("eth_utils.abi.get_normalized_abi_inputs")
```
fselmo: Again, do we need to deprecate in `v7`? Now might be the time to break it and deprecate in `v6`. I would argue anything living under `_utils` does _not_ need a deprecation message in either case, so this is probably fine to just break in `v7` and not even deprecate in `v6`.

web3/utils/abi.py
```diff
@@ -1,21 +1,242 @@
...
 def get_abi_output_names(abi: Union[ABIFunction]) -> List[str]:
     if "outputs" not in abi and abi["type"] == "fallback":
         return []
     return [arg["name"] for arg in abi["outputs"]]
+
+
+def get_function_info(
```
fselmo: By adding new methods here, it doesn't seem like we deprecated `web3.utils.abi.py` as the newsfragment states so I'm confused by that in the newsfragment. My lingering question remains, are we re-exposing `eth_utils` or asking users to import different utils from different places? I lean towards the former for a better UX of centralized utils but I'm not sure I'm caught up on the conversation around it.

fselmo (commented):
At first pass the code looks good. I just have a big lingering question about the organization of utils methods. I think my inline comments will be clear enough but I was under the impression we were moving things to `eth_utils` and re-exposing them in a centralized spot where all abi utils can be located, even the in-house web3 ones.

I'll take a second pass once I'm caught up on that.

web3/_utils/abi.py
reedsa: Agreed, deprecations should be made in `v6`. Trying to keep track of all the things, deprecating this private method for now but I do intend to remove it in this PR.

web3/utils/abi.py
reedsa: Bad merge here but you're right, the method itself should just be removed.

web3/utils/abi.py
reedsa: The newsfragment is outdated, sorry about that.

tests/core/contracts/test_contract_method_to_argument_matching.py
```diff
@@ -1,15 +1,15 @@
...
-)
 from web3._utils.function_identifiers import (
     FallbackFn,
     ReceiveFn,
 )
 from web3.exceptions import (
-    Web3ValidationError,
+    MismatchedABI,
+)
+from web3.utils.abi import (
```
fselmo: nit: import this from the source `from eth_utils.abi`, take or leave

web3/utils/abi.py
```diff
@@ -1,21 +1,505 @@
...
+
+def get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
+    fallback_abis = filter_by_type("fallback", contract_abi)
+    if fallback_abis and fallback_abis[0]["type"] == "fallback":
+        return cast(ABIFallback, fallback_abis[0])
+    else:
+        raise FallbackNotFound("No fallback function was found in the contract ABI.")
+
+
+def filter_abi_by_name(
```
fselmo: Should this get moved to `eth_utils.abi` as well? Seems like we have `filter_by_type` there.

web3/contract/utils.py
```diff
@@ -118,7 +124,7 @@ def call_contract_function(
     transaction: TxParams,
     block_id: Optional[BlockIdentifier] = None,
     contract_abi: Optional[ABI] = None,
-    fn_abi: Optional[ABIFunction] = None,
+    fn_abi: Optional[Union[ABIFunction]] = None,
```
fselmo: I don't think this `Union` is doing anything

web3/types.py
```diff
@@ -74,33 +77,53 @@ class AccessListEntry(TypedDict):
 
 AccessList = NewType("AccessList", Sequence[AccessListEntry])
 
+ABI_COMPONENT_DEPRECATION_MSG = "`{0}` is deprecated, use `{1}` instead."
+
 
-# todo: move these to eth_typing once web3 is type hinted
+@deprecated(
```
fselmo: Since this is going into `main`, I'd vote to remove these altogether, document in the `v7` migration notes, and deprecate in `v6`. That applies to all `@deprecated` here.

fselmo (commented):
I think this looks great, just left some minor comments. Since we still have some minor stuff to address and we have to wait until the versions are released and updated here, I'm going to refrain from approving until then.

pacrob (changes requested):
`eth-utils` dep needs updating before gtg, but no other concerns.

web3/_utils/validation.py
```diff
@@ -21,6 +24,11 @@
     is_list_like,
     is_string,
 )
+from eth_utils.abi import (
+    abi_to_signature,
+    filter_abi_by_type,
+    function_abi_to_4byte_selector,
```
kclowes: This `function_abi_to_4byte_selector` makes me wonder if we should expose all of these at the top level?

docs/migration.rst
```diff
@@ -252,6 +252,26 @@ The ``personal`` namespace was used for managing accounts and keys and was depre
 in geth in ``v1.11.0``. Geth has moved to using ``clef`` for account and key management.
 
 
+ABI Types Removed
+`````````````````
+
+The type definitions for ABIs, deprecated in ``v6``, have been removed in ``v7``. New
+types have been introduced in the ``eth_typing`` package for ABIs. Improvements have
```
fselmo: nit: We could mention the *eth-typing* version that they were introduced in here.

tests/core/contracts/test_contract_call_interface.py
```diff
@@ -532,7 +532,7 @@ def test_call_receive_fallback_function(
 
 
 def test_call_nonexistent_receive_function(fallback_function_contract):
-    with pytest.raises(FallbackNotFound, match="No receive function was found"):
+    with pytest.raises(ABIReceiveNotFound, match="No receive function was found"):
```
fselmo, line 535: 👌🏼

tests/core/utilities/test_abi_named_tree.py
```diff
@@ -89,15 +81,6 @@ def test_named_arguments_decode_with_misshapen_inputs(values):
         named_tree(short_abi_inputs_with_disallowed_names, values)
 
 
-def test_namedtuples_encodable():
```
fselmo: Just curious why we moved this test? This feels like a more specific place for it, no?

web3/types.py
```diff
@@ -75,46 +75,6 @@ class AccessListEntry(TypedDict):
 AccessList = NewType("AccessList", Sequence[AccessListEntry])
 
 
-# todo: move these to eth_typing once web3 is type hinted
```
fselmo, line 78: 🎉

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    return [
+        cast(ABICallable, function_abi)
+        for function_abi in contract_abi
+        if check_if_arguments_can_be_encoded(
+            cast(ABICallable, function_abi), args, kwargs, abi_codec
+        )
+    ]
+
+
+def _mismatched_abi_error_diagnosis(
```
fselmo, line 161: 🔥 🔥

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+
+    Returns the ABI which matches the provided identifier, named arguments (``args``)
+    and keyword args (``kwargs``).
+
+    :param abi: Contract ABI.
+    :type abi: `ABI`
+    :param function_identifier: Find a function ABI with matching identifier.
+    :type function_identifier: `str`
+    :param args: Find a function ABI with matching args.
+    :type args: `list[Any]`
```
fselmo:
```suggestion
    :type args: `Optional[Sequence[Any]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param kwargs: Find a function ABI with matching kwargs.
+    :type kwargs: `Any`
```
fselmo:
```suggestion
    :type kwargs: `Optional[Dict[str, Any]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    The `abi_codec` may be overridden if custom encoding and decoding is required. The
+    default is used if no codec is provided. More details about customizations are in
+    the `eth-abi Codecs Doc <https://eth-abi.readthedocs.io/en/latest/codecs.html>`__.
+
+    :param abi: Contract ABI.
+    :type abi: `ABI`
+    :param function_identifier: Find a function ABI with matching name.
+    :type function_identifier: `FunctionIdentifier`
+    :param args: Find a function ABI with matching args.
+    :type args: `list[Any]`
```
fselmo:
```suggestion
    :type args: `Optional[Sequence[Any]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param kwargs: Find a function ABI with matching kwargs.
+    :type kwargs: `Any`
```
fselmo:
```suggestion
    :type kwargs: `Optional[Dict[str, Any]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param abi_codec: Codec used for encoding and decoding. Default with \
+    `strict_bytes_type_checking` enabled.
+    :type abi_codec: `Any`
```
fselmo:
```suggestion
    :type abi_codec: `Optional[Any]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :return: ABI for the function interface.
+    :rtype: `ABIFunction`
```
fselmo:
```suggestion
    :rtype: `ABIElement`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    function_abi: ABIElement,
+    args: Sequence[Any],
+    kwargs: Dict[str, Any],
+    abi_codec: Optional[Any] = None,
+) -> bool:
+    """
+    Check if the provided arguments can be encoded with the function ABI.
+
+    :param function_abi: Function ABI.
+    :type function_abi: `ABIFunction`
```
fselmo: See comment above but I think this is the right change to make out of the two?

```suggestion
    :type abi_element: `ABIElement`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param args: Positional arguments.
+    :type args: `list[Any]`
```
fselmo:
```suggestion
    :type args: `Sequence[Any]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param kwargs: Keyword arguments.
+    :type kwargs: `Any`
```
fselmo:
```suggestion
    :type kwargs: `Dict[str, Any]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+        return ABIElementInfo(
+            abi=fn_abi, selector=fn_selector, arguments=aligned_fn_inputs
+        )
+
+
+def get_abi_element(
+    abi: ABI,
+    function_identifier: FunctionIdentifier,
+    args: Optional[Sequence[Any]] = None,
+    kwargs: Optional[Any] = None,
```
fselmo:
```suggestion
    kwargs: Optional[Dict[str, Any]] = None,
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    ]
+
+
+def _mismatched_abi_error_diagnosis(
+    function_identifier: FunctionIdentifier,
+    matching_function_signatures: Sequence[str],
+    arg_count_matches: int,
+    encoding_matches: int,
+    args: Optional[Sequence[Any]] = None,
+    kwargs: Optional[Any] = None,
```
fselmo:
```suggestion
    kwargs: Optional[Dict[str, Any]] = None,
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
     ABIEvent,
+    ABIFallback,
     ABIFunction,
+    ABIReceive,
+)
+from eth_typing.abi import (
+    ABI,
+    ABIElementInfo,
+)
+from eth_utils.abi import (
```
fselmo: This is re-imported below, this can go.

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    :param function_identifier: Find a function ABI with matching name.
+    :type function_identifier: `FunctionIdentifier`
+    :param args: Find a function ABI with matching args.
+    :type args: `list[Any]`
+    :param kwargs: Find a function ABI with matching kwargs.
+    :type kwargs: `Any`
+    :param abi_codec: Codec used for encoding and decoding. Default with \
+    `strict_bytes_type_checking` enabled.
+    :type abi_codec: `Any`
+    :return: ABI for the function interface.
```
fselmo: Not sure here but it's not only function, correct?
```suggestion
    :return: ABI for the specific ABI element.
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    """
+    fallback_abis = filter_abi_by_type("fallback", contract_abi)
+    if fallback_abis and fallback_abis[0]["type"] == "fallback":
+        return cast(ABIFallback, fallback_abis[0])
+    else:
+        raise ABIFallbackNotFound("No fallback function was found in the contract ABI.")
+
+
+def check_if_arguments_can_be_encoded(
+    function_abi: ABIElement,
```
fselmo:
```suggestion
    abi_element: ABIElement,
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+    args: Sequence[Any],
+    kwargs: Dict[str, Any],
+    abi_codec: Optional[Any] = None,
+) -> bool:
+    """
+    Check if the provided arguments can be encoded with the function ABI.
```
fselmo:
```suggestion
    Check if the provided arguments can be encoded for the ABI element.
```

web3/utils/abi.py
```diff
@@ -1,21 +1,512 @@
...
+
+    :param function_abi: Function ABI.
```
fselmo:
```suggestion
    :param abi_element: The ABI element.
```

fselmo (commented):
I did a first pass. Let me know if I misunderstood any of the type changes.

web3/utils/__init__.py
```diff
@@ -3,9 +3,28 @@
...
+    get_abi_output_types,
+    get_aligned_abi_inputs,
+    get_all_event_abis,
+    get_all_function_abis,
+    get_normalized_abi_inputs,
+)
+from .abi import (
+    get_abi_element_info,
+    get_abi_element,
+    check_if_arguments_can_be_encoded,
```
wolovim: should `get_fallback_function_abi` and `get_receive_function_abi` be included here and below?

tests/core/utilities/test_abi_named_tree.py
reedsa: I moved it thinking that ``check_if_arguments_can_be_encoded`` could belong with the other ABI utils tests. I see your point, this test is more beneficial as part of these named tree tests. No reason to move it, I'll put this back.

ens/utils.py
kclowes, line 1: Yay! Nice cleanup 🧹

tests/core/contracts/test_args_and_kwargs_merger.py
kclowes, line 1: Lazy review comment: I assume that these tests got moved to `eth_utils` where `merge_args_and_kwargs` is defined now?

web3/_utils/abi.py
```diff
@@ -515,30 +359,12 @@ def _align_abi_input(arg_abi: ABIFunctionParams, arg: Any) -> Tuple[Any, ...]:
...
-    return (
-        tuple(collapse_if_tuple(abi) for abi in input_abis),
-        type(args)(_align_abi_input(abi, arg) for abi, arg in zip(input_abis, args)),
-    )
-
-
-def get_constructor_abi(contract_abi: ABI) -> ABIFunction:
-    candidates = [abi for abi in contract_abi if abi["type"] == "constructor"]
+def get_constructor_abi(contract_abi: ABI) -> ABIConstructor:
+    candidates = [
```
kclowes: We could use `filter_abi_by_type` here if we want to?

web3/_utils/contracts.py
```diff
@@ -124,11 +114,11 @@ def find_matching_event_abi(
     argument_names: Optional[Sequence[str]] = None,
 ) -> ABIEvent:
     filters = [
-        functools.partial(filter_by_type, "event"),
+        functools.partial(filter_abi_by_type, "event"),
     ]
 
     if event_name is not None:
-        filters.append(functools.partial(filter_by_name, event_name))
+        filters.append(functools.partial(_filter_event_abi_by_name, event_name))
```
kclowes: Can we just call `filter_abi_by_name` here? It looks like `_filter_event_abi_by_name` is doing the same thing, but I might be mistaken

web3/_utils/contracts.py
```diff
@@ -143,82 +133,25 @@ def find_matching_event_abi(
...
+    except ValueError:
+        # Use the default argument_types if the abi doesn't have inputs
+        pass
+
+    if not check_if_arguments_can_be_encoded(
+        abi,
+        arguments,
+        {},
+        w3.codec,
+    ):
```
kclowes: Moving around the order of the args is a breaking change. It's an internal method it looks like, so that's fine, but we should make sure that this doesn't break anywhere else that calls `check_if_arguments_can_be_encoded`

web3/contract/async_contract.py
```diff
@@ -577,7 +580,10 @@ def __init__(
             if transaction is None:
                 transaction = {}
 
-            self._functions = filter_by_type("function", self.abi)
+            self._functions = [
+                cast(ABIFunction, fn_abi)
```
kclowes: I don't really have any ideas other than making them separate functions - for example: calling this something like `def get_all_functions(self.abi) -> ABIFunction:`- but I wonder if there is a way that we could generally stop casting to `ABIFunction`/`ABIEvent` when we call `filter_abi_by_type`. It feels a little smelly, but I'm not hugely concerned about it either.

kclowes (commented):
This is looking good! I left a few comments inline and then here are a few general comments:
- We should make a note in either the migration guide and/or the newsfragments about error type changes. For example: `Web3ValidationError` -> `MismatchedABIError`
- I think it would be okay to export the `eth_utils.abi` helpers you're using on the top-level `eth_utils`, but don't feel too strongly about that. I wonder what @fselmo or @pacrob think?
- It looks like the order of some arguments (specifically `w3.codec`) got switched around. Hopefully something would have popped up by now if it was a problem, but it's worth a check if you want to keep that change.
- I put this inline somewhere, but it feels clunky to have to cast all the `filter_abi_by_type` to `ABIFunction`/`ABIEvent`. I wonder if there is a way to clean that up. It looks like(?) the only reason the iteration is happening is to cast each element?

tests/core/contracts/test_args_and_kwargs_merger.py
reedsa: Yep! `get_normalized_abi_inputs` is the new function and those tests are now in [`eth-utils`](https://github.com/reedsa/eth-utils/blob/e1047974c9941f62ee38ac86182ef28fe8ddd8ed/tests/core/abi-utils/test_abi_utils.py#L1249)

reedsa:
@kclowes @fselmo @pacrob 

I haven't come across a way to return specific ABI types from those functions, so casting to the types makes mypy happy. The problem is the type is based on the value of the TypedDict at runtime. There might be a way to pass a separate parameter and create overrides that return the expected type, then it would require logic to branch each of those wherever those functions are used.

Maybe something like this would work?
```python
@override
def get_abi_element(
    abi: ABI,
    function_identifier: FunctionIdentifier,
    *args: Optional[Sequence[Any]],
    abi_codec: Optional[Any] = None,
    **kwargs: Optional[Dict[str, Any]],
    abi_type: Literal['function']
): ABIFunction

def get_abi_element(
    abi: ABI,
    function_identifier: FunctionIdentifier,
    *args: Optional[Sequence[Any]],
    abi_codec: Optional[Any] = None,
    **kwargs: Optional[Dict[str, Any]],
    abi_type: str
) -> ABIElement:
  ...

def some_func():
  if(abi["type"] == "function"):
    abi = get_abi_element(abi, ..., "function")
  if(abi["type"] == "constructor"):
    abi = get_abi_element(abi, ..., "constructor")
```
I think the types of ``abi`` would then follow the overrides to make mypy happy. Havent tried that yet though. Let me know if there are other ideas or feedback on the changes I've made thus far!

web3/contract/async_contract.py
reedsa: 🤦 Oh you're right! I can simply change this to the new `get_all_function_abis` and update the type of `_functions`. There are other places which can use logic branching to make the typings happy without having to cast.

web3/_utils/abi.py
reedsa: Looks like this wasnt being used anyway. I removed it.

web3/_utils/validation.py
reedsa: I think I understand what you're saying, the functions are exposed on the top level `eth-utils`, so I can `from eth_utils import function_abi_to_4byte_selector`. Should I update these imports to pull from the top level? Is that to just keep everything together in the imports, or is there another benefit?

kclowes: Yeah, exactly. No benefit other than consistency. I can also see leaving them in `eth_utils.abi` if you think that makes more sense

reedsa: This is definitely something that is inconsistent in a lot of places. My only reasoning for using the submodule import style is for functions which have ambiguous names. When I import using vscode it automatically uses the most specific module path but I dont have a strong preference either way.

 The naming for ABI features has been intentionally specific and for that reason I slightly prefer the top level style.

kclowes: Sounds good to me!

reedsa:
@pacrob @fselmo @kclowes I've merged changes that include functions for event ABIs and added some polish. I am ready for another pass on this now that it's functioning with the latest version of `eth-utils`.

web3/contract/base_contract.py
```diff
@@ -656,20 +672,21 @@ def __init__(
...
 
         if self.abi:
-            self._functions = filter_by_type("function", self.abi)
+            self._functions = filter_abi_by_type("function", self.abi)
             for func in self._functions:
-                setattr(
-                    self,
-                    func["name"],
-                    contract_function_class.factory(
+                if func["type"] == "function":
```
kclowes: nit: looks like this is redundant since we're filtering on L675?

web3/contract/base_contract.py
```diff
@@ -865,43 +888,45 @@ def _prepare_transaction(
...
     def _find_matching_fn_abi(
         cls,
-        fn_identifier: Optional[str] = None,
-        args: Optional[Any] = None,
-        kwargs: Optional[Any] = None,
-    ) -> ABIFunction:
-        return find_matching_fn_abi(
-            cls.abi, cls.w3.codec, fn_identifier=fn_identifier, args=args, kwargs=kwargs
+        fn_identifier: Optional[FunctionIdentifier] = None,
+        *args: Optional[Sequence[Any]],
```
kclowes: If these are `Optional`ly typed, should they default to None?

web3/contract/utils.py
```diff
@@ -330,7 +344,7 @@ def find_functions_by_identifier(
             abi=fn_abi,
         )
         for fn_abi in fns_abi
-        if callable_check(fn_abi)
+        if callable_check(fn_abi) and fn_abi["type"] == "function"
```
kclowes: redundant 'function' check here?

web3/contract/base_contract.py
```diff
@@ -865,43 +888,45 @@ def _prepare_transaction(
...
     ) -> ABIEvent:
-        return find_matching_event_abi(
+        return get_event_abi(
             abi=cls.abi, event_name=event_name, argument_names=argument_names
         )
 
     @combomethod
     def _encode_constructor_data(
-        cls, args: Optional[Any] = None, kwargs: Optional[Any] = None
+        cls, *args: Optional[Sequence[Any]], **kwargs: Optional[Dict[str, Any]]
```
kclowes: Same here: should this `Optional` type default to None?

web3/contract/base_contract.py
```diff
@@ -415,19 +421,20 @@ def __init__(
...
         if abi:
             self.abi = abi
-            self._events = filter_by_type("event", self.abi)
+            self._events = filter_abi_by_type("event", self.abi)
             for event in self._events:
-                setattr(
-                    self,
-                    event["name"],
-                    contract_event_type.factory(
+                if event["type"] == "event":
```
kclowes: redundant check?

kclowes (commented):
This looks great! I didn't see anything substantial really. The only think is that we're doing the check to see `if fn_abi['type'] == 'function'` a lot. I think I commented on most of them, but not all. I think some are redundant, and I wonder if it could/should be abstracted out? Feel free to take or leave!

web3/contract/base_contract.py
reedsa: The defaults cause an error: `Parameter with "*" or "**" cannot have default value`. I haven't found an example where this would be possible as an `Optional`, so these `Optional` types don't make sense here.

I think the function itself will always initialize these with values internally, so it probably means we should just get rid of the `Optional`.

web3/contract/base_contract.py
reedsa: Ah yes, the types work now with the overloads on `get_abi_by_type`. Fixed!

newsfragments/3408.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Upgrades to use latest ``ABI`` utilities and typings from ``eth-utils`` and ``eth-typing``.
+* New ``ABI`` functions from ``eth-utils`` are now included in the ``web3.utils.abi`` module.
```
fselmo: I think this is covered in the `feature` newsfragment? New functions shouldn't be breaking.

web3/contract/base_contract.py
```diff
@@ -747,18 +762,21 @@ def encode_abi(
 
         :param data: defaults to function selector
         """
-        fn_abi, fn_selector, fn_arguments = get_function_info(
+        args = args or tuple()
+        kwargs = kwargs or {}
+
+        fn_info = get_abi_element_info(
```
fselmo: How is this filtering by "function" type? It's not immediately clear to me. This used to be `get_function_info` and now it seems like it returns all abi element info as one. Do we need to use `get_all_function_abis(cls.abi)` instead of `cls.abi`?

web3/_utils/contracts.py
```diff
@@ -252,20 +174,26 @@ def prepare_transaction(
     w3: Union["AsyncWeb3", "Web3"],
     fn_identifier: Union[str, Type[FallbackFn], Type[ReceiveFn]],
     contract_abi: Optional[ABI] = None,
-    fn_abi: Optional[ABIFunction] = None,
+    fn_abi: Optional[ABIElement] = None,
```
fselmo: Is there a reason we opened this type up? Is that desirable / is there no way to keep it within the function scope as intended? Should this be `ABICallable`?

web3/_utils/contracts.py
```diff
@@ -294,120 +222,102 @@ def prepare_transaction(
...
-
-    return fn_abi, fn_selector, aligned_fn_arguments
+        fn_abi = get_receive_function_abi(contract_abi)
+    fn_selector = encode_hex(b"")
+    fn_arguments: Tuple[Any, ...] = tuple()
+    return fn_abi, fn_selector, fn_arguments
 
 
-def validate_payable(transaction: TxParams, abi: ABIFunction) -> None:
+def validate_payable(transaction: TxParams, abi_element: ABICallable) -> None:
```
fselmo: nit, distinguish the input name as the type to avoid confusion with `ABIElement` / `abi_element` methods.


```suggestion
def validate_payable(transaction: TxParams, abi_callable: ABICallable) -> None:
```

kclowes (approved):
🚀 LGTM! Nice work!

web3/_utils/contracts.py
```diff
@@ -252,20 +174,26 @@ def prepare_transaction(
     w3: Union["AsyncWeb3", "Web3"],
     fn_identifier: Union[str, Type[FallbackFn], Type[ReceiveFn]],
```
fselmo:
```suggestion
    fn_identifier: FunctionIdentifier,
```

web3/_utils/contracts.py
```diff
@@ -294,120 +222,102 @@ def prepare_transaction(
...
     fn_selector = encode_hex(b"")
     fn_arguments: Tuple[Any, ...] = tuple()
     return fn_abi, fn_selector, fn_arguments
 
 
-def get_receive_function_info(
-    contract_abi: Optional[ABI] = None, fn_abi: Optional[ABIFunction] = None
-) -> Tuple[ABIFunction, HexStr, Tuple[Any, ...]]:
+def get_fallback_function_info(
+    contract_abi: Optional[ABI] = None, fn_abi: Optional[ABIFallback] = None
```
fselmo: nit

```suggestion
    contract_abi: Optional[ABI] = None, fallback_abi: Optional[ABIFallback] = None
```

web3/_utils/contracts.py
```diff
@@ -294,120 +222,102 @@ def prepare_transaction(
...
-    args: Optional[Sequence[Any]] = None,
-    kwargs: Optional[Any] = None,
-) -> Tuple[ABIFunction, HexStr, Tuple[Any, ...]]:
-    if args is None:
-        args = tuple()
-    if kwargs is None:
-        kwargs = {}
-
+def get_receive_function_info(
+    contract_abi: Optional[ABI] = None, fn_abi: Optional[ABIReceive] = None
```
fselmo: nit

```suggestion
    contract_abi: Optional[ABI] = None, receive_abi: Optional[ABIReceive] = None
```

web3/_utils/contracts.py
```diff
@@ -294,120 +222,102 @@ def prepare_transaction(
...
     if normalizers:
         decoded = map_abi_data(normalizers, types, decoded)
     return named_tree(fn_abi["inputs"], decoded)
 
 
-def get_fallback_function_info(
-    contract_abi: Optional[ABI] = None, fn_abi: Optional[ABIFunction] = None
-) -> Tuple[ABIFunction, HexStr, Tuple[Any, ...]]:
+def get_constructor_function_info(
+    contract_abi: Optional[ABI] = None, fn_abi: Optional[ABIConstructor] = None
```
fselmo:
```suggestion
    contract_abi: Optional[ABI] = None, constructor_abi: Optional[ABIConstructor] = None
```

web3/utils/abi.py
```diff
@@ -1,21 +1,686 @@
...
+) -> ABIEvent:
+    """
+    Find the event interface with the given name and/or arguments.
+
+    :param abi: Contract ABI.
+    :type abi: `ABI`
+    :param event_name: Find an event abi with matching event name.
+    :type event_name: `str`
+    :param argument_names: Find an event abi with matching arguments.
+    :type argument_names: `list[str]`
```
fselmo:
```suggestion
    :type argument_names: `Optional[Sequence[str]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,686 @@
...
+    The `abi_codec` may be overridden if custom encoding and decoding is required. The
+    default is used if no codec is provided. More details about customizations are in
+    the `eth-abi Codecs Doc <https://eth-abi.readthedocs.io/en/latest/codecs.html>`__.
+
+    :param abi: Contract ABI.
+    :type abi: `ABI`
+    :param function_identifier: Find a function ABI with matching name.
+    :type function_identifier: `FunctionIdentifier`
+    :param args: Find a function ABI with matching args.
+    :type args: `Optional[list[Any]]`
```
fselmo:
```suggestion
    :type args: `Optional[Sequence[Any]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,686 @@
...
+    """
+    Returns the class name of the argument, or `address` if the argument is an address.
+    """
+    if is_checksum_address(arg) or is_binary_address(arg):
+        return "address"
+
+    return arg.__class__.__name__
+
+
+def _element_abi_to_type(
```
fselmo: Does this need overloads to help with typing on the consumer side?

web3/utils/abi.py
```diff
@@ -1,21 +1,686 @@
...
+    event_abi: ABIEvent,
+    topics: Optional[Sequence[HexBytes]] = None,
+) -> Sequence[HexBytes]:
+    r"""
+    Return topics for an event ABI.
+
+    :param event_abi: Event ABI.
+    :type event_abi: `ABIEvent`
+    :param topics: Transaction topics from a `LogReceipt`.
+    :type topics: `list[HexBytes]`
```
fselmo: Should this be optional? Is there a reason that it is? imo this should be a required arg and not an optional kwarg unless there's utility in keeping this optional. If the latter, then this doctstring should change to:

```suggestion
    :type topics: `Optional[Sequence[HexBytes]]`
```

web3/utils/abi.py
```diff
@@ -1,21 +1,686 @@
...
+) -> bytes:
+    r"""
+    Return topic signature as bytes.
 
-def get_abi_input_names(abi: Union[ABIFunction, ABIEvent]) -> List[str]:
-    if "inputs" not in abi and abi["type"] == "fallback":
-        return []
-    return [arg["name"] for arg in abi["inputs"]]
+    :param log_topic: Event topic from a `LogReceipt`.
+    :type log_topic: `Primitive`, `HexStr` or `str`
```
fselmo: nit, since we've been giving the literal type everywhere else:

```suggestion
    :type log_topic: `Union[Primitives, HexStr, str]`
```

fselmo (commented):
lgtm overall. Some of these are things to think about and discuss before merging, others are minor tweaks / catches.

newsfragments/3408.breaking.rst
reedsa: It was included here because we decided the `eth-utils` upgrade itself was breaking. I can leave it out of this though since we have it covered in the other newsfragment.

fselmo: We can either remove or rephrase it as something that is actually breaking like:

```suggestion
* New versions of existing ABI functions were added to ``eth-utils`` and are now exposed in `web3.py` via ``web3.utils.abi``.
```

or something along those lines?

web3/contract/base_contract.py
reedsa: `get_abi_element_info` calls `get_abi_element` internally with the arguments and `cls.abi` to filter the correct function and put it in `fn_info`.

This information is needed for the encode call and only uses the ABI of the function being called, as well as the data being encoded using `selector`, `arguments`, and `data`. The return type of `get_function_info` was originally a `Tuple[ABIFunction, HexStr, Tuple[Any, ...]]`. There was a comment in an earlier thread somewhere to adapt the Tuple into a new type, which became a TypedDict called `ABIElementInfo`.

Maybe I could add a comment or something to clear that up, what do you think?

web3/_utils/contracts.py
reedsa: I think I recall changing this to encompass `constructor`, `fallback` and `receive` but you are right this should be `ABICallable`. That will affect a couple parts of this function so I'll get those fixed!

web3/contract/base_contract.py
fselmo: > `get_abi_element_info` calls `get_abi_element` internally with the arguments and `cls.abi` to filter the correct function and put it in `fn_info`.

My biggest concern is that we're using `cls.abi` (the whole abi object) and not `get_all_function_abis(cls.abi)` (only function abis) as the argument here. I think that may be redundant to filter for only function abis, but it just feels like a decent gut check and makes reading the code a bit easier. Does that make sense?

web3/utils/abi.py
reedsa: This was a failed attempt at typing the return value, but it isn't actually needed inside of `get_event_abi`/`get_event_abi` so I'll remove this.

web3/utils/abi.py
reedsa: That makes sense, I'll update!

web3/_utils/contracts.py
reedsa: I dont think these are all that useful since `get_abi_element` achieves this.

I could see these being convenient for users by returning the appropriate typed ABI element though. Do you think it's worth refactoring these functions to use `get_abi_element`?

fselmo: This still lives under `_utils`. Let's keep the scope of this PR more narrow and focus on breaking *eth-utils* related changes for now. If we want to add convenient methods for users, we can do so as features later on. Whatever makes sense for internal use, let's make those changes here.

reedsa: Oh being private makes things easier then, these are not used anywhere and can be removed. 👍

web3/contract/base_contract.py
reedsa: That's reasonable, I will add that filter!

reedsa: Diving deeper here, `encode_abi` should work for any ABI element, so it's unfortunate the function argument is named `fn_name`. I'm dont think a filter would make it easier to tell what's going on here, ideally we would change the variable names but probably not a good time to touch `encode_abi` just yet. Do you think we should still do the filter, a comment, or something else?

web3/utils/abi.py
```diff
@@ -1,21 +1,530 @@
...
+    """
+    if is_checksum_address(arg) or is_binary_address(arg):
+        return "address"
+
+    return arg.__class__.__name__
+
+
+def get_abi_element_info(
+    abi: ABI,
+    function_identifier: FunctionIdentifier,
```
fselmo: Somewhat related to [this thread](https://github.com/ethereum/web3.py/pull/3408/files#r1695774197), should this type reflect an abi element, rather than just a function identifier? Otherwise the name should be something closer to `get_abi_callable_info()`.

web3/utils/abi.py
```diff
@@ -1,21 +1,530 @@
...
+        _, aligned_fn_inputs = get_aligned_abi_inputs(fn_abi, fn_inputs)
+
+        return ABIElementInfo(
+            abi=fn_abi, selector=fn_selector, arguments=aligned_fn_inputs
+        )
+
+
+def get_abi_element(
+    abi: ABI,
+    function_identifier: FunctionIdentifier,
```
fselmo: Also here... is this just for functions or other elements (e.g. ABIEvent). It looks like it returns `ABIElement` so maybe this arg should reflect this.

web3/contract/base_contract.py
fselmo: It looks like the old logic here didn't actually restrict things to functions, even though the arg names would lead you to believe it. I wouldn't be against changing `fn_name` -> `abi_element_name` and leaving the function filtering out of it. Thoughts on that?

reedsa: I like that plan and I will document that change in the migration guide!

web3/utils/abi.py
reedsa: The identifier has been historically used as a way to distinguish `fallback` and `receive` types, but the others are all used with just a `str` name. I think this should be named `element_identifier`, and the type could potentially change too. Should I update both?

fselmo: > I think this should be named element_identifier, and the type could potentially change too. Should I update both?

I think it might make sense to, yeah. It seems to support more than just a "function identifier".

web3/contract/base_contract.py
```diff
@@ -766,7 +772,7 @@ def encode_abi(
 
         fn_info = get_abi_element_info(
```
fselmo: I think this `fn_info` could be `element_info` for clarity, since we left the function filtering logic out of this method.

fselmo (approved):
lgtm! Just left 1 rename nit.

PR #3409, opened 2024-05-29
Update provider explicit exports

### LINKED ISSUE #3396: Explicitly export symbols via __all__

fselmo:
### What was wrong?

closes #3396 together with #3410

### How was it fixed?

- Adds `AsyncEthereumTesterProvider` to _web3/\_\_init\_\_.py_.
- Adds `EthereumTesterProvider` and `AsyncEthereumTesterProvider` to _web3/providers/\_\_init\_\_.py_.
- Update `__all__` in _web3/\_\_init\_\_.py_ to include all provider classes, including base classes.
- Add `__all__` for _web3/providers/\_\_init\_\_.py_ including all explicit exports.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -4,7 +4,7 @@ Events and Logs
 ===============
 
 If you're on this page, you're likely looking for an answer to this question:
-**How do I know when a specific contract is used?** You have at least three options:
+**How do I know when a specific contract is used?** You have several options:
 
 1. Query blocks for transactions that include the contract address in the ``"to"`` field.
    This contrived example is searching the latest block for any transactions sent to the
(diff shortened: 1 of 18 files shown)


kclowes (approved):
LGTM! Do we need to export `__all__` in other `__init__` too? like in #3402?

fselmo:
> LGTM! Do we need to export `__all__` in other `__init__` too? like in #3402?

Ah, I missed that PR. Yeah, I think that's nice to have. I will rebase off of it and add the missing exports after the fact.

fselmo:
@kclowes rebased off #3402. My only remaining questions are whether things like `combine_middleware`, re-exporting `Web3` even though it's only used internally in places like _web3/middleware/\_\_init\_\_.py_, etc are desired. I lean towards a no on that but can also leave it as it was in #3402.

kclowes:
No, I don't think we want to re-export `Web3` or export `combine_middleware`

PR #3410, opened 2024-05-29
[v6] Update provider explicit exports for `__all__`

### LINKED ISSUE #3396: Explicitly export symbols via __all__

fselmo:
### What was wrong?

closes #3396 together with #3409

### How was it fixed?

- Adds `AsyncEthereumTesterProvider` to _web3/\_\_init\_\_.py_.
- Adds `EthereumTesterProvider` and `AsyncEthereumTesterProvider` to _web3/providers/\_\_init\_\_.py_.
- Update `__all__` in _web3/\_\_init\_\_.py_ to include all provider classes, including base classes.
- Add `__all__` for _web3/providers/\_\_init\_\_.py_ including all explicit exports.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/newsfragments/3410.feature.rst
+++ b/newsfragments/3410.feature.rst
@@ -1,0 +1,1 @@
+Provide explicit ``__all__`` exports for providers in `web3/providers/__init__.py`; update `web3/__init__.py` to include all provider classes including base classes.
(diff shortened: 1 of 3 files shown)


kclowes (approved):
🚢

PR #3412, opened 2024-05-30
Strap a `RequestSessionManager` to each instance of http providers

### LINKED ISSUE #3265: response is ERROR when using 'multiprocessing'
* Version: 6.15.1
* Python: 3.8.10
* OS: linux/
* `pip freeze` output

```
aiohttp==3.9.3
aiosignal==1.3.1
async-timeout==4.0.3
attrs==23.2.0
bitarray==2.9.2
brokenaxes==0.6.0
certifi==2024.2.2
charset-normalizer==3.3.2
contourpy==1.1.1
cycler==0.12.1
cytoolz==0.12.3
eth-account==0.11.0
eth-hash==0.6.0
eth-keyfile==0.7.0
eth-keys==0.5.0
eth-rlp==1.0.1
eth-typing==4.0.0
eth-utils==4.0.0
eth_abi==5.0.0
fonttools==4.49.0
frozenlist==1.4.1
hexbytes==0.3.1
idna==3.6
importlib-resources==6.1.1
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
kiwisolver==1.4.5
lru-dict==1.2.0
matplotlib==3.7.5
multidict==6.0.5
numpy==1.24.4
packaging==23.2
parsimonious==0.9.0
pillow==10.2.0
pkgutil_resolve_name==1.3.10
protobuf==4.25.3
py-solc==3.2.0
pycryptodome==3.20.0
pyparsing==3.1.1
python-dateutil==2.8.2
pyunormalize==15.1.0
referencing==0.33.0
regex==2023.12.25
requests==2.31.0
rlp==4.0.0
rpds-py==0.18.0
semantic-version==2.10.0
six==1.16.0
toolz==0.12.1
typing_extensions==4.9.0
urllib3==2.2.1
web3==6.15.1
websockets==12.0
yarl==1.9.4
zipp==3.17.0
```

### What was wrong?

```python

import multiprocessing

url = "http://ip:port"


def conn():
    w3 = Web3(HTTPProvider(endpoint_uri=url, request_kwargs={"timeout": 1000}))
    genesis_account = lambda: (
        w3.to_checksum_address(w3.eth.accounts[0]) if w3.eth.accounts else None
    )
    flag = w3.geth.personal.unlock_account(genesis_account(), "", 0)
    # execute one more time, then the exception occurs !!!
    genesis_account()


if __name__ == "__main__":
    conn()

    proc1 = multiprocessing.Process(target=conn)

    proc2 = multiprocessing.Process(target=conn)

    proc1.start()
    proc2.start()

    proc1.join()
    proc2.join()
```
```bash
MainProcess, session: 140271829895392
MainProcess, session: 140271829895392
MainProcess, session: 140271829895392
MainProcess, session: 140271829895392
MainProcess, session: 140271829895392
Process-1, session: 140271829895392
Process-2, session: 140271829895392
Process-2, session: 140271829895392
Process-1, session: 140271829895392
Process-2, session: 140271829895392
Process-1, session: 140271829895392
Process-2, session: 140271829895392
Process Process-2:
  File "/home/miniconda3/envs/py38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/miniconda3/envs/py38/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "temp/para_web3.py", line 15, in conn
    genesis_account()
  File "temp/para_web3.py", line 11, in <lambda>
    w3.to_checksum_address(w3.eth.accounts[0]) if w3.eth.accounts else None
  File "/home/.local/share/virtualenvs/tor-experiment-L97qUE1y/lib/python3.8/site-packages/web3/eth/eth.py", line 114, in accounts
    return self._accounts()
  File "/home/.local/share/virtualenvs/tor-experiment-L97qUE1y/lib/python3.8/site-packages/web3/module.py", line 78, in caller
    return apply_result_formatters(result_formatters, result)
  File "cytoolz/functoolz.pyx", line 267, in cytoolz.functoolz.curry.__call__
  File "cytoolz/functoolz.pyx", line 263, in cytoolz.functoolz.curry.__call__
  File "/home/.local/share/virtualenvs/tor-experiment-L97qUE1y/lib/python3.8/site-packages/web3/module.py", line 49, in apply_result_formatters
    formatted_result = pipe(result, result_formatters)
  File "cytoolz/functoolz.pyx", line 680, in cytoolz.functoolz.pipe
  File "cytoolz/functoolz.pyx", line 655, in cytoolz.functoolz.c_pipe
  File "/home/.local/share/virtualenvs/tor-experiment-L97qUE1y/lib/python3.8/site-packages/eth_utils/functional.py", line 47, in inner
    return callback(fn(*args, **kwargs))
TypeError: 'bool' object is not iterable
Process-1, session: 140271829895392
Process-1, session: 140271829895392
```

```txt
Private ethereum node. Geth version is 1.10.0-stable.
```

### How can it be fixed?

**explain the output**

I print session id in the format `"{current_process().name}, session: {id(session)}"`.

The exception happens when querying `eth.accounts`. The response shoule be `type: List[Account]`, while it is `type: bool` which should be the response type of `unlock_account`.

The detailed exception line is the [code](https://github.com/ethereum/web3.py/blob/main/web3/module.py#L78). It has obtained the response but faces the error type.

**reason**

When sending requests, web3 uses the [cached session](https://github.com/ethereum/web3.py/blob/main/web3/_utils/request.py#L49) with `SimpleCache()`. 'SimpleCache()' is [instantiated](https://github.com/ethereum/web3.py/blob/main/web3/_utils/request.py#L45) upon importing the web3 and won't be re-instantiated  whenever creating a new `Web3()` .

In the main process, after executing `conn()` the first session is cached. When forking the children process, all objects including the cached session in the main process are also forked. However, the underlying connections of this session will be shared across all processes, which could cause the responses mixed up in different processes.

***It is an absolutely dangerous BUG !!!***

**solution**

- solution 1.

The easiest solution is to change the [cache key](https://github.com/ethereum/web3.py/blob/main/web3/_utils/request.py#L53). Now the key is just tied to the thread identifier which is identical in the above codes. Just add the process identifier into the key.

- solution 2.

Solution 1 is not the best because all process are still using the same one `SimpleCache()`. I think, it should check whether it is in a new forked process or not. If it is, then a new SimpleCache()` should be instantiated.

fselmo:
### What was wrong?

Though #3265 is not reproducible, this hopefully closes #3265 as there should be no more conflicts across separate instances due to unique instantiation of session managing per provider.

### How was it fixed?

- Use a unique session manager for each instance of HTTPProvider, and AsyncHTTPProvider, by moving all the shared session cache logic into a class and strapping it to each provider instance.
- Update tests related to session caching
- Update docs and update the typing for `request_retry_configuration` along the way (`Optional`)

### Todo:

- [x] Clean up commit history

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -144,6 +144,29 @@ options on the provider instance:
 Retry Requests for HTTP Providers
 `````````````````````````````````
 
+.. py:class:: web3.providers.rpc.utils.ExceptionRetryConfiguration
+
+    .. py:attribute:: errors
+
+        A tuple of exceptions that the provider should retry on. The default is
+        ``HTTPProvider``: ``(ConnectionError, requests.HTTPError, requests.Timeout)``
+        and ``AsyncHTTPProvider``: ``(aiohttp.ClientError, asyncio.TimeoutError)``.
+
+    .. py:attribute:: retries
+
+        The number of retries to attempt. The default is 5.
+
+    .. py:attribute:: backoff_factor
+
+        The initial delay multiplier, which doubles with each retry attempt. The default
+        is 0.125.
+
+    .. py:attribute:: method_allowlist
+
+        A list of retryable methods. The default is an in-house list of deemed-safe-to-
+        retry methods.
+
+
 ``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
 on exceptions. This can be configured via configuration options on the provider
 instance. The retry mechanism employs an exponential backoff strategy, starting from
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -144,6 +144,16 @@ options on the provider instance:
 Retry Requests for HTTP Providers
 `````````````````````````````````
 
+``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
+on exceptions. This can be configured via the ``exception_retry_configuration``
+property on the provider instance, which takes a
+:class:`~web3.providers.rpc.utils.ExceptionRetryConfiguration` class as its value. The
+retry mechanism employs an exponential backoff strategy, starting from the initial
+value determined by the ``backoff_factor``, and doubling the delay with each attempt,
+up to the ``retries`` value. Below is an example showing the default options for the
+retry configuration and how to override them.
+
+
 .. py:class:: web3.providers.rpc.utils.ExceptionRetryConfiguration
 
     .. py:attribute:: errors
@@ -165,14 +175,6 @@ Retry Requests for HTTP Providers
 
         A list of retryable methods. The default is an in-house list of deemed-safe-to-
         retry methods.
-
-
-``HTTPProvider`` and ``AsyncHTTPProvider`` instances retry certain requests by default
-on exceptions. This can be configured via configuration options on the provider
-instance. The retry mechanism employs an exponential backoff strategy, starting from
-the initial value determined by the ``backoff_factor``, and doubling the delay with
-each attempt, up to the ``retries`` value. Below is an example showing the default
-options for the retry configuration and how to override them.
 
 .. code-block:: python
 
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -107,7 +107,7 @@ local and remote JSON-RPC servers.
 HTTPProvider
 ~~~~~~~~~~~~
 
-.. py:class:: web3.providers.rpc.HTTPProvider(endpoint_uri[, request_kwargs, session])
+.. py:class:: web3.providers.rpc.HTTPProvider(endpoint_uri, request_kwargs={}, session=None, exception_retry_configuration=ExceptionRetryConfiguration())
 
     This provider handles interactions with an HTTP or HTTPS based JSON-RPC server.
 
@@ -119,6 +119,11 @@ HTTPProvider
       will be passed onto each http/https POST request made to your node.
     * ``session`` allows you to pass a ``requests.Session`` object initialized
       as desired.
+    * ``exception_retry_configuration`` is an instance of the
+      :class:`~web3.providers.rpc.utils.ExceptionRetryConfiguration`
+      class which allows you to configure how the provider should handle exceptions
+      when making certain requests. Setting this to ``None`` will disable
+      exception retries.
 
     .. code-block:: python
 
@@ -180,7 +185,7 @@ IPCProvider
 AsyncHTTPProvider
 ~~~~~~~~~~~~~~~~~
 
-.. py:class:: web3.providers.async_rpc.AsyncHTTPProvider(endpoint_uri[, request_kwargs])
+.. py:class:: web3.providers.rpc.AsyncHTTPProvider(endpoint_uri, request_kwargs={}, exception_retry_configuration=ExceptionRetryConfiguration())
 
     This provider handles interactions with an HTTP or HTTPS based JSON-RPC server asynchronously.
 
@@ -190,7 +195,14 @@ AsyncHTTPProvider
       be omitted from the URI.
     * ``request_kwargs`` should be a dictionary of keyword arguments which
       will be passed onto each http/https POST request made to your node.
-    * the ``cache_async_session()`` method allows you to use your own ``aiohttp.ClientSession`` object. This is an async method and not part of the constructor
+    * ``exception_retry_configuration`` is an instance of the
+      :class:`~web3.providers.rpc.utils.ExceptionRetryConfiguration`
+      class which allows you to configure how the provider should handle exceptions
+      when making certain requests. Setting this to ``None`` will disable
+      exception retries.
+
+    The ``cache_async_session()`` method allows you to use your own
+    ``aiohttp.ClientSession`` object.
 
     .. code-block:: python
 
--- a/web3/_utils/http_session_manager.py
+++ b/web3/_utils/http_session_manager.py
@@ -29,18 +29,23 @@ from web3._utils.async_caching import (
 from web3._utils.caching import (
     generate_cache_key,
 )
+from web3._utils.http import (
+    DEFAULT_HTTP_TIMEOUT,
+)
 from web3.utils.caching import (
     SimpleCache,
 )
 
-DEFAULT_TIMEOUT = 30
 
-
-class RequestSessionManager:
-    logger = logging.getLogger("web3._utils.request.RequestSessionManager")
+class HTTPSessionManager:
+    logger = logging.getLogger("web3._utils.http_session_manager.HTTPSessionManager")
     _lock: threading.Lock = threading.Lock()
 
-    def __init__(self, cache_size: int = 100, session_pool_max_workers: int = 5):
+    def __init__(
+        self,
+        cache_size: int = 100,
+        session_pool_max_workers: int = 5,
+    ) -> None:
         self.session_cache = SimpleCache(cache_size)
         self.session_pool = ThreadPoolExecutor(max_workers=session_pool_max_workers)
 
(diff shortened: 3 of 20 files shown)


docs/providers.rst
```diff
@@ -107,7 +107,7 @@ local and remote JSON-RPC servers.
 HTTPProvider
 ~~~~~~~~~~~~
 
-.. py:class:: web3.providers.rpc.HTTPProvider(endpoint_uri[, request_kwargs, session])
+.. py:class:: web3.providers.rpc.HTTPProvider(endpoint_uri, request_kwargs={}, session=None, exception_retry_configuration=ExceptionRetryConfiguration())
```
kclowes, line 110: I believe the default of `request_kwargs` is `None`
```suggestion
.. py:class:: web3.providers.rpc.HTTPProvider(endpoint_uri, request_kwargs=None, session=None, exception_retry_configuration=ExceptionRetryConfiguration())
```

docs/providers.rst
```diff
@@ -180,7 +185,7 @@ IPCProvider
 AsyncHTTPProvider
 ~~~~~~~~~~~~~~~~~
 
-.. py:class:: web3.providers.async_rpc.AsyncHTTPProvider(endpoint_uri[, request_kwargs])
+.. py:class:: web3.providers.rpc.AsyncHTTPProvider(endpoint_uri, request_kwargs={}, exception_retry_configuration=ExceptionRetryConfiguration())
```
kclowes, line 188: same here:
```suggestion
.. py:class:: web3.providers.rpc.AsyncHTTPProvider(endpoint_uri, request_kwargs=None, exception_retry_configuration=ExceptionRetryConfiguration())
```

web3/_utils/module_testing/module_testing_utils.py
```diff
@@ -152,13 +146,12 @@ async def _mock_specific_request(
 
         # mock response only to specified url while validating appropriate fields
         if url_from_args == mocked_request_url:
-            assert kwargs["timeout"] == ClientTimeout(30)
```
kclowes, line 155: I'm curious about this removal and the one on line 25: `assert kwargs["timeout"] == 30`. Can we still pass a timeout here?

kclowes (approved):
The idea of strapping this manager to a provider is cracking me up for some reason. I'm imagining ratchet straps 😆 Aaaanyway, this looks good to me! There wasn't really a good place to put this in the PR, but what do you think about moving the name of `web3/_utils/request.py` to something like `web3/_utils/request_session_manager.py` since the `RequestSessionManager` class is basically all the file has now? Otherwise I left some small nits, but nothing big.

docs/internals.rst
```diff
@@ -144,6 +144,29 @@ options on the provider instance:
 Retry Requests for HTTP Providers
 `````````````````````````````````
 
+.. py:class:: web3.providers.rpc.utils.ExceptionRetryConfiguration
```
reedsa, line 157: In the text following this addition, it says configuration is done on the provider. I thought that it was talking about configuration for how retries are implemented and got confused.

Edit: I reread it and maybe I was confused without the additional context. Might add an intro blurb or move this below the text I mentioned.

web3/_utils/module_testing/module_testing_utils.py
fselmo: Yeah I think that's reasonable to centralize the `DEFAULT_TIMEOUT` and use it for all http requests. I will add that back for the ccip read requests

docs/providers.rst
fselmo: If it's none we set it to `{}` in the `__init__()` though... do we think it's better to hint at the type here by using `{}` or to be technically correct and use `None` even though that it doesn't give a good type hint?

fselmo: We can use `None` and clarify below that it sets it to an empty dict? I'm ok with either approach.

kclowes: Eh, I guess since it's just in the docs, we can leave it as a dict. I'll leave it up to you :)

fselmo:
> The idea of strapping this manager to a provider is cracking me up for some reason. I'm imagining ratchet straps 😆

😆 

> Aaaanyway, this looks good to me! There wasn't really a good place to put this in the PR, but what do you think about moving the name of `web3/_utils/request.py` to something like `web3/_utils/request_session_manager.py` since the `RequestSessionManager` class is basically all the file has now? Otherwise I left some small nits, but nothing big.

I think I'll still put the `DEFAULT_REQUEST_TIMEOUT` here and use it for the CCIP read http requests, which it doesn't matter what provider we are using, it still makes a http GET or POST as part of the spec. For that reason it might make sense to still keep this as request utils? Or should I move the class definition into `request_session_manager.py` and keep the minimal `request.py` as a space for adding more than just the `DEFAULT_REQUEST_TIMEOUT`?

docs/providers.rst
fselmo: I'll leave my reasoning for keeping it here. We have to do a dance for `ExceptionRetryConfiguration` as well where we use `Empty` since we want the ability to set it to `None` to turn it off. So I think if we go the technically correct route we'd have to put `Empty` as the default value for `ExceptionRetryConfiguration` and then explain why and I think that complicates everything more than any user cares about 😆 .

I'm going to opt to use values that if you don't touch the configuration at all, those are what we will hard code in the `__init__()`, which feels more valuable to a user than what the signature represents.

kclowes:
> I think I'll still put the DEFAULT_REQUEST_TIMEOUT here and use it for the CCIP read http requests, which it doesn't matter what provider we are using, it still makes a http GET or POST as part of the spec. For that reason it might make sense to still keep this as request utils? Or should I move the class definition into request_session_manager.py and keep the minimal request.py as a space for adding more than just the DEFAULT_REQUEST_TIMEOUT?

Keeping DEFAULT_REQUEST_TIMEOUT in there sounds good to me, and therefore keeping this as request utils also probably makes the most sense? I wouldn't be opposed to splitting it out either though. Whatever you think!

fselmo:
> Keeping DEFAULT_REQUEST_TIMEOUT in there sounds good to me, and therefore keeping this as request utils also probably makes the most sense? I wouldn't be opposed to splitting it out either though. Whatever you think!

Actually, we have a *_utils/http.py* already... I'm thinking that since this is really only managing HTTP sessions, we rename the class to `HTTPSessionManager`, create a *_utils/http_session_manager.py* and put it there, and then move the default timeout to *_utils/http.py* as `HTTP_REQUEST_TIMEOUT`. Thoughts there?

kclowes:
Yeah, I like that. More specific!

fselmo:
> Yeah, I like that. More specific!

Sweet. The changes were significant enough. Can I ask for one last pass starting at [this commit](https://github.com/ethereum/web3.py/pull/3412/commits/f9352e0a42c53772677d82af82d3dc93cd60f635) before we call it done? I also added passing the ``timeout`` through to the respective sync and async `cache_and_return_session()` methods so we can use a more accurate value for triggering the session closing tasks in that method.

---
edit

Just kidding I had to update the commit hash, should be good now. I added the timeout check to the integrations tests as well.

web3/_utils/http_session_manager.py
```diff
@@ -80,7 +86,7 @@ def cache_and_return_session(
                     f"{evicted_session}",
                 )
             threading.Timer(
-                DEFAULT_TIMEOUT + 0.1,
+                request_timeout + 0.1,
```
kclowes: If this is None like the type allows, I think this line will throw?  Maybe add a `if request_timeout is None` check? But I'm also surprised mypy didn't catch that so maybe I missed something.

kclowes (approved):
I left one comment about a None check, but other than that, looks good!

web3/_utils/http_session_manager.py
fselmo: Good eye... I'll make sure to address it.

PR #3414, opened 2024-06-06
Run mypy with all deps installed

pacrob:
### What was wrong?

Running `mypy` via the pre-commit `mirrors-mypy` hook misses the context of other libs in our dependencies.

### How was it fixed?

 - Update the pre-commit `mypy` hook to run locally.
 - Bump `requests` dep to >=2.23.0, where they first support py38
 - Add `types-requests>=2.0.0` to `install_requires` deps
 - bump `eth-account` to >=0.13.1 to use `TransactionDictType` in `web3/middleware/signing.py`

fix all the new mypy errors!

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture
image


web3/_utils/filters.py
```diff
@@ -78,9 +79,9 @@ def construct_event_filter_params(
     address: Optional[ChecksumAddress] = None,
 ) -> Tuple[List[List[Optional[HexStr]]], FilterParams]:
     filter_params: FilterParams = {}
-    topic_set: Sequence[HexStr] = construct_event_topic_set(
-        event_abi, abi_codec, argument_filters
-    )
+    topic_set: Union[
```
reedsa: It's not clear to me where the nested list would come from. Is that potentially from the `topics` argument? Or should the return type on `construct_event_topic_set` be updated?

In either case I think the type would also be `Union[Sequence[HexStr], Sequence[Sequence[HexStr]]]`.

reedsa (commented):
One clarifying question, otherwise looks good!

web3/_utils/filters.py
pacrob: I did not dig into `construct_event_topic_set` output. I just took the check on line 94 to mean it sometimes comes back as a nested list. But I suppose yes, if the return type of `List[HexStr]` is to be taken seriously, I should match it here.

pacrob: Decided to take `construct_event_topic_set` at its word and leave `topic_set` as `Sequence[HexStr`. If there are cases where it comes back as a nested list, we should address that as a bug and not just have a :shrug:  check here.

reedsa (approved):
lgtm!

fselmo (approved):
lgtm 👍🏼

web3/_utils/filters.py
```diff
@@ -90,10 +90,7 @@ def construct_event_filter_params(
             )
         topic_set = topics
 
-    if len(topic_set) == 1 and is_list_like(topic_set[0]):
```
kclowes, line 93: I assume this was a flake8 check that flagged? do we have it tested somewhere?

web3/_utils/module_testing/web3_module.py
```diff
@@ -712,8 +712,8 @@ async def test_batch_requests_cancel(self, async_w3: AsyncWeb3) -> None:
         assert isinstance(block_num, int)
 
     @pytest.mark.asyncio
-    async def test_batch_requests_raises_for_common_unsupported_methods(
-        self, async_w3: AsyncWeb3, async_math_contract: "AsyncContract"
+    async def test_batch_requests_raises_for_common_unsupported_methods(  # type: ignore[override]  # noqa: E501
```
kclowes, line 715: I wonder if we should change the method names here so we're not ignoring all the types?

web3/middleware/filter.py
```diff
@@ -255,7 +255,7 @@ def __init__(
         if from_block is None or from_block == "latest":
             self._from_block = BlockNumber(w3.eth.block_number + 1)
         elif is_string(from_block) and is_hex(from_block):
-            self._from_block = BlockNumber(hex_to_integer(from_block))
+            self._from_block = BlockNumber(hex_to_integer(cast(HexStr, from_block)))
```
kclowes, line 258: seems like this type should be an `int` instead of casting to `HexStr`?

web3/middleware/filter.py
```diff
@@ -272,7 +272,7 @@ def to_block(self) -> BlockNumber:
         elif self._to_block == "latest":
             to_block = self.w3.eth.block_number
         elif is_string(self._to_block) and is_hex(self._to_block):
-            to_block = BlockNumber(hex_to_integer(self._to_block))
+            to_block = BlockNumber(hex_to_integer(cast(HexStr, self._to_block)))
```
kclowes, line 275: same here - should be an `int`?

kclowes (approved):
I left a couple comments but nothing big. Hooray!!

web3/_utils/filters.py
pacrob: No, this is what Stu's original comment was related to. The type of `topic_set` in the function definition is `Sequence[HexStr]`, but here we had this "but maybe it's a nested list sometimes?" check. The return type of `construct_event_topic_set` is `List[HexStr]`, so correct behavior seems to be to assume it will be a `Sequence[HexStr]`, then if  some edge case crops up that returns a nested list, fix that as a bug and add a test there.

kclowes: Okay, sounds good!

web3/middleware/filter.py
pacrob: I'm not sure I understand. In the line above, we confirm that it's a hex string, so we then want to convert it to an `int`. `hex_to_integer` is typed to accept only `HexStr`, so needs that casting. Maybe the answer is to open up typing on `hex_to_integer` to avoid these castings?

kclowes: oops, I misread. You're right. I thought we were casting the result of `hex_to_integer` to a `HexStr`. Don't mind me :)

web3/_utils/module_testing/web3_module.py
pacrob: Comments on lines 521-522 say they need to stay the same. I remember reviewing this when Felipe set batching up, but I don't remember why they needed to be the same.

Issue #3419, opened 2024-06-24
"error" is not allowed as `ABIFunction.type`


avoiney-ledger:
### What feature should we add?

Some ABI, like the one for smart contract [0x48C7f48622a2C31FC72B326332cF1e2357C8EF30](https://etherscan.io/address/0x48C7f48622a2C31FC72B326332cF1e2357C8EF30), fetched using Etherscan API, contains the defined errors objects. So the JSON abi contains this:

```json
[
  {
    "inputs": [
      {
        "internalType": "contract IgmRenderer",
        "name": "_renderer",
        "type": "address"
      }
    ],
    "stateMutability": "nonpayable",
    "type": "constructor"
  },
  {
    "inputs": [
      
    ],
    "name": "IncorrectAmountForMint",
    "type": "error"
  },
  {
    "inputs": [
      
    ],
    "name": "LimitReachedForWallet",
    "type": "error"
  },
  {
    "inputs": [
      
    ],
    "name": "MaxSupplyReached",
    "type": "error"
  },
  {
    "inputs": [
      {
        "internalType": "uint256",
        "name": "available",
        "type": "uint256"
      },
      {
        "internalType": "uint256",
        "name": "required",
        "type": "uint256"
      }
    ],
    "name": "NotEnoughAllowance",
    "type": "error"
  },
  {
    "inputs": [
      {
        "internalType": "address",
        "name": "operator",
        "type": "address"
      }
    ],
    "name": "OperatorNotAllowed",
    "type": "error"
  },
  {
    "inputs": [
      
    ],
    "name": "SaleNotActive",
    "type": "error"
  },
  {
    "inputs": [
      
    ],
    "name": "SendToAddressZero",
    "type": "error"
  },
  {
    "inputs": [
      {
        "internalType": "uint256",
        "name": "id",
        "type": "uint256"
      }
    ],
    "name": "TokenDoesNotExist",
    "type": "error"
  },
  {
    "inputs": [
      
    ],
    "name": "WithdrawSendFailed",
    "type": "error"
  },
  {
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "internalType": "address",
        "name": "owner",
        "type": "address"
      },
      {
        "indexed": true,
        "internalType": "address",
        "name": "approved",
        "type": "address"
      },
      {
        "indexed": true,
        "internalType": "uint256",
        "name": "tokenId",
        "type": "uint256"
      }
    ],
    "name": "Approval",
    "type": "event"
[...]

PR #3420, opened 2024-06-25
`v6` `ABIError` type

reedsa:
### What was wrong?

Encountering type errors when parsing contracts with `"error"` type ABI elements.

Related to Issue #3419

### How was it fixed?

Created ``ABIError`` type in the ``web3.types`` module and added as a valid type of ``ABIElement``.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/web3/types.py
+++ b/web3/types.py
@@ -112,6 +112,19 @@ class ABIFunction(TypedDict, total=False):
     type: Literal["function", "constructor", "fallback", "receive"]
 
 
+class ABIComponent(TypedDict):
+    """
+    TypedDict representing an `ABIElement` component.
+    """
+
+    name: str
+    """Name of the component."""
+    type: str
+    """Type of the component."""
+    components: NotRequired[Sequence["ABIComponent"]]
+    """List of nested `ABI` components for ABI types."""
+
+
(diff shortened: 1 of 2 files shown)


web3/types.py
```diff
@@ -112,7 +112,20 @@ class ABIFunction(TypedDict, total=False):
...
+class ABIError(TypedDict):
+    """
+    TypedDict representing the `ABI` for an error.
+    """
+
+    type: Literal["error"]
+    """Type of the error."""
+    name: str
+    """Name of the error."""
+    inputs: NotRequired[Sequence["ABIFunctionParams"]]
```
fselmo: It doesn't feel fantastic to use `ABIFunctionParams` here. I could be swayed to let it slide but it feels pretty minimal to also introduce the new `ABIComponent` to v6 and use it here instead? Thoughts on that?

fselmo (approved):
lgtm once all tests pass 👍🏼

web3/types.py
reedsa: No complaints from me, it is better to use the correct types from the start. I considered using it but left it out to do the bare minimum change. Added!

PR #3421, opened 2024-06-25
Fix Beacon API after internal cache rehaul:

fselmo:
### What was wrong?

We moved session caching as internal parts of HTTP providers. Since that change, the ``Beacon`` API was receiving a bad bath for session caching.

### How was it fixed?

These updates fix that (by setting the proper path to the `HTTPSessionManager` class) and raise the timeout for the async tests since some seem to be timing out.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


reedsa (approved):
lgtm

kclowes (approved):
🚀

PR #3424, opened 2024-06-27
Fix websocket provider receives nothing forever when the peer server is closed or gone

kaiix:
This should fix websocket provider receives nothing forever when the peer server is closed or gone.

When receiving websocket messages using iterator pattern, [websockets.py suppresses `ConnectionClosedOK` errors](https://github.com/python-websockets/websockets/blob/main/src/websockets/legacy/protocol.py#L496-L500). so [`_provider_specific_message_listener`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/providers/websocket/websocket_v2.py#L139) will stop iterating without an error, and [`_message_listener`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/providers/persistent.py#L139) will continue the `while` loop without knowing that the connection is closed and will not receive any messages.

The fix use `recv()` function to allow the websocket to throw the `ConnectionClosedOK` exception.

(After merging this PR, I'll port the fix to the main branch)

### DIFF
--- a/web3/_utils/module_testing/module_testing_utils.py
+++ b/web3/_utils/module_testing/module_testing_utils.py
@@ -1,6 +1,3 @@
-from collections import (
-    deque,
-)
 import pytest
 import time
 from typing import (
@@ -9,6 +6,7 @@ from typing import (
     Collection,
     Dict,
     Generator,
+    Optional,
     Sequence,
     Union,
 )
@@ -35,6 +33,7 @@ from web3._utils.compat import (
 )
 from web3._utils.request import (
     async_cache_and_return_session,
+    asyncio,
     cache_and_return_session,
 )
 from web3.types import (
@@ -188,9 +187,13 @@ class WebsocketMessageStreamMock:
     closed: bool = False
 
     def __init__(
-        self, messages: Collection[bytes] = None, raise_exception: Exception = None
+        self,
+        messages: Optional[Collection[bytes]] = None,
+        raise_exception: Optional[Exception] = None,
     ) -> None:
-        self.messages = deque(messages) if messages else deque()
+        self.queue = asyncio.Queue[bytes]()
+        for msg in messages or []:
+            self.queue.put_nowait(msg)
         self.raise_exception = raise_exception
 
     def __await__(self) -> Generator[Any, Any, "Self"]:
@@ -203,13 +206,13 @@ class WebsocketMessageStreamMock:
         return self
 
     async def __anext__(self) -> bytes:
+        return await self.recv()
+
+    async def recv(self) -> bytes:
         if self.raise_exception:
             raise self.raise_exception
 
-        elif len(self.messages) == 0:
-            raise StopAsyncIteration
-
-        return self.messages.popleft()
+        return await self.queue.get()
 
     @staticmethod
     async def pong() -> Literal[False]:
--- a/web3/_utils/module_testing/module_testing_utils.py
+++ b/web3/_utils/module_testing/module_testing_utils.py
@@ -191,7 +191,7 @@ class WebsocketMessageStreamMock:
         messages: Optional[Collection[bytes]] = None,
         raise_exception: Optional[Exception] = None,
     ) -> None:
-        self.queue = asyncio.Queue[bytes]()
+        self.queue = asyncio.Queue()  # type: ignore  # py38 issue
         for msg in messages or []:
             self.queue.put_nowait(msg)
         self.raise_exception = raise_exception
--- a/web3/providers/persistent.py
+++ b/web3/providers/persistent.py
@@ -191,10 +191,6 @@ class PersistentConnectionProvider(AsyncJSONBaseProvider, ABC):
             request_cache_key = generate_cache_key(request_id)
 
             while True:
-                # check if an exception was recorded in the listener task and raise it
-                # in the main loop if so
-                self._handle_listener_task_exceptions()
-
                 if request_cache_key in self._request_processor._request_response_cache:
                     self.logger.debug(
                         f"Popping response for id {request_id} from cache."
@@ -204,6 +200,9 @@ class PersistentConnectionProvider(AsyncJSONBaseProvider, ABC):
                     )
                     return popped_response
                 else:
+                    # check if an exception was recorded in the listener task and raise
+                    # it in the main loop if so
+                    self._handle_listener_task_exceptions()
                     await asyncio.sleep(0)
 
         try:
--- a/web3/providers/persistent.py
+++ b/web3/providers/persistent.py
@@ -9,6 +9,7 @@ from typing import (
 
 from websockets import (
     ConnectionClosed,
+    ConnectionClosedOK,
     WebSocketClientProtocol,
     WebSocketException,
 )
@@ -142,6 +143,12 @@ class PersistentConnectionProvider(AsyncJSONBaseProvider, ABC):
             await asyncio.sleep(0)
             try:
                 await self._provider_specific_message_listener()
+            except ConnectionClosedOK:
+                # RequestManager._ws_message_stream() will rethrow this exception and
+                # cause WebsocketConnection to stop the process_subscriptions() loop,
+                # or raise ConnectionClosedOK when recv() is called.
+                # see: https://github.com/ethereum/web3.py/pull/3424
+                raise
             except Exception as e:
                 if not self.silence_listener_task_exceptions:
                     raise e
--- a/newsfragments/3424.bugfix.rst
+++ b/newsfragments/3424.bugfix.rst
@@ -1,0 +1,1 @@
+Propagate ``ConnectionClosedOK`` exception for ``WebsocketProviderV2`` to properly break the loop when listening for messages over the socket.
--- a/web3/providers/websocket/websocket_v2.py
+++ b/web3/providers/websocket/websocket_v2.py
@@ -136,9 +136,8 @@ class WebsocketProviderV2(PersistentConnectionProvider):
         return response
 
     async def _provider_specific_message_listener(self) -> None:
-        async for raw_message in self._ws:
-            await asyncio.sleep(0)
-
+        while True:
+            raw_message = await self._ws.recv()
             response = json.loads(raw_message)
             subscription = response.get("method") == "eth_subscription"
             await self._request_processor.cache_raw_response(


fselmo:
Thanks @kaiix, I do like this change and I think it makes a lot of sense. Since the `ConnectionClosedOK` isn't used as an error exception, but rather a signal that the connection was closed with a purpose and a successful callback, do you think this should be handled separately from `silence_listener_task_exceptions`?

In my mind, this could go one of two ways. We can handle the `ConnectionClosedOK` exceptions as a `break` in the listener task and shut it down, or we can let the user handle that. I think it may be quite noisy to always propagate that exception and leave it to the user to handle. One idea is that we could use a separate configuration for this, default it to be silently handled, but if a user wants full control they can choose their own logic for when a connection is successfully closed on purpose by the server.

Thoughts on that?

cc: @kclowes, @pacrob, @reedsa

fselmo:
> One idea is that we could use a separate configuration for this, default it to be silently handled, but if a user wants full control they can choose their own logic for when a connection is successfully closed on purpose by the server.


I imagine this exists in some shape of form like this:

*websocket_v2.py* (untouched from your code)
```py
    async def _provider_specific_message_listener(self) -> None:
        while True:
            raw_message = await self._ws.recv()
            response = json.loads(raw_message)
            subscription = response.get("method") == "eth_subscription"
            await self._request_processor.cache_raw_response(
                response, subscription=subscription
            )
```

*persistent.py*
```py
    async def _message_listener(self) -> None:
        self.logger.info(
            f"{self.__class__.__qualname__} listener background task started. Storing "
            "all messages in appropriate request processor queues / caches to be "
            "processed."
        )
        while True:
            # the use of sleep(0) seems to be the most efficient way to yield control
            # back to the event loop to share the loop with other tasks.
            await asyncio.sleep(0)
            try:
                await self._provider_specific_message_listener()
            except ConnectionClosedOK as e:
                self.logger.info("Connection closed successfully, disconnecting...")
                if self.raise_on_successful_close:
                    raise e
                else:
                    await self.disconnect()
                    break
            except Exception as e:
                if not self.silence_listener_task_exceptions:
                    raise e
                else:
                    self._error_log_listener_task_exception(e)
```

kaiix:
I've seen many projects (include mine) use the async iterator pattern for websocket provider from the official document to catch the `ConnectionClosed` and try to reconnect.
https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/docs/providers.rst?plain=1#L354-L360

Under this pattern, when `_ws.recv()` raises the `ConnectionClosedOK`, [`persistent_websocket()`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/main.py#L575) will try to reconnect, and after 5 retries it will throw a [`ProviderConnectionError`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/providers/persistent.py#L86), which seems a very reasonable way to handle the case.

As to whether `_silence_listener_task_exceptions` should be changed, I think that can be discussed separately (maybe another PR?).

kaiix:
@fselmo Understand what you mean, I think just rethrowing `ConnectionClosedOK` from `message_listener_task` is enough and don't need to introduce any other options to control the behavior.

The modified code looks like this:

*persistent.py*

```py
    async def _message_listener(self) -> None:
        self.logger.info(
            f"{self.__class__.__qualname__} listener background task started. Storing "
            "all messages in appropriate request processor queues / caches to be "
            "processed."
        )
        while True:
            # the use of sleep(0) seems to be the most efficient way to yield control
            # back to the event loop to share the loop with other tasks.
            await asyncio.sleep(0)
            try:
                await self._provider_specific_message_listener()
            except ConnectionClosedOK:
                raise
            except Exception as e:
                if not self.silence_listener_task_exceptions:
                    raise e
                else:
                    self._error_log_listener_task_exception(e)
```

When the message listener encounters the `ConnectionClosedOK`, the listener task will end with the exception, [`_get_next_ws_message`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/manager.py#L367) ([`ws.recv()`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/providers/websocket/websocket_connection.py#L39) and [`ws.process_subscriptions()`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/providers/websocket/websocket_connection.py#L42) use it) will rethrowing it in [`_handle_listener_task_exceptions`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/manager.py#L393).

If the user uses async iterator to process websocket subscriptions (which is the [recommended](https://web3py.readthedocs.io/en/stable/providers.html#usage) in the doc) like this

```py
async for response in w3.ws.process_subscriptions():
  ...
```

the rethrowed `ConnectionClosedOK` will cause a [`StopAsyncIteration`](https://github.com/ethereum/web3.py/blob/616212e3bf9d4ff7c24edc7db2a98d03e5616626/web3/manager.py#L467) in `_AsyncPersistentMessageStream`, which stopped the for loop as there are no more responses. It looks very reasonable.

Only when the user use [`recv()`](https://web3py.readthedocs.io/en/stable/providers.html#web3.providers.websocket.WebsocketConnection.recv) directly (which is not the recommended usage), they will need to handle `ConnectionClosedOK` exception manually like this:

```py
while True:
  try:
    response =  w3.ws.recv()
  except ConnectionClosed:  # ConnectionClosed is the base class of ConnectionClosedOK
    break
```

How do you think?

kaiix:
I first noticed this issue when listening for new blocks using a public node (and it turns out to be unstable), and I found that at some point, the CPU usage is **100%** while the provider received nothing.

Just some additional information to help developers with the same problem find this issue when searching for high CPU usage.

fselmo:
@kaiix I think that makes good sense. Have you had a chance to test this out? Do you mind updating this PR to reflect these changes?

web3/providers/persistent.py
```diff
@@ -197,6 +200,9 @@ async def _match_response_id_to_request_id() -> RPCResponse:
                     )
                     return popped_response
                 else:
+                    # check if an exception was recorded in the listener task and raise
```
kaiix, line 203: This change allows for getting the response which has already been received before handling the message listener exceptions.

As shown in the test change, the server sends a series of responses and closes the connection (which raises an `ConnectionClosedOK`); before this change, although the client had received the responses in the buffer (`_request_response_cache`), it did not return the response, but instead raised a "ConnectionClosed" error.

As a reference, we can see that the ethers.js [implementation](https://github.com/ethers-io/ethers.js/blob/fc66b8ad405df9e703d42a4b23bc452ec3be118f/src.ts/providers/provider-socket.ts#L309) prioritizes the return of a response before handling the exception.

kaiix:
@fselmo PR updated, please review

tests/core/providers/test_wsv2_provider.py
```diff
@@ -129,16 +130,19 @@ async def test_async_make_request_returns_desired_response():
...
         cached_response = (
             await provider._request_processor._subscription_response_queue.get()
         )
         # assert all cached responses are in the list of responses we received
         assert to_bytes(text=json.dumps(cached_response)) in ws_messages
 
+    # due to different asyncio scheduler implementation, the last item could be a
+    # TaskNotRunning() if scheduler give control to `_handle_listener_task_exceptions`
+    # before `_get_response_for_request_id`
+    if not provider._request_processor._subscription_response_queue.empty():
```
fselmo: I don't think we should have conditionals in tests. This isn't immediately obvious when it triggers or not. I don't think we should really force a connection closed on the `WebsocketMessageStreamMock` class. What if we defined that class with a proper `asyncio.Queue` instead of a `deque` so that we can properly await messages? I think I used a `deque` solely because it was a bit quicker to implement but I don't think we should try to simulate a connection closing at the end of the async iteration anyhow.

So, in `module_testing_utils.py`:

```py
class WebsocketMessageStreamMock:
    closed: bool = False

    def __init__(
        self, messages: Collection[bytes] = None, raise_exception: Exception = None
    ) -> None:
        self.queue = asyncio.Queue()
        for msg in messages or []:
            self.queue.put_nowait(msg)
        self.raise_exception = raise_exception

        return __async_init__().__await__()
        
    # ... untouched methods

    async def __anext__(self) -> bytes:
        return await self.recv()

    async def recv(self) -> bytes:
        if self.raise_exception:
            raise self.raise_exception

        return await self.queue.get()
        
    # ... untouched methods
```

We can just leave this test alone in that case. I was actually not seeing any errors even without that change, but this enforces the mock to behave more properly anyhow. Does this work?

web3/_utils/module_testing/module_testing_utils.py
```diff
@@ -211,6 +214,15 @@ async def __anext__(self) -> bytes:
 
         return self.messages.popleft()
 
+    async def recv(self) -> bytes:
```
fselmo, line 211: See my previous comment, I think we can better mock the stream for the tests instead. It was a bit lazy on my part for the the original implementation of the stream mock to not use a proper `asyncio.Queue` from the beginning.

web3/providers/persistent.py
```diff
@@ -142,6 +143,12 @@ async def _message_listener(self) -> None:
             await asyncio.sleep(0)
             try:
                 await self._provider_specific_message_listener()
+            except ConnectionClosedOK:
+                # RequestManager._ws_message_stream() will rethrow this exception and
+                # cause WebsocketConnection stop the process_subscriptions() loop, or
```
fselmo:
```suggestion
                # cause WebsocketConnection to stop the process_subscriptions() loop, or
```

web3/providers/persistent.py
fselmo: Regardless of the test change, which I don't believe needs to happen if we refactor the message queue in the mock class a bit, I think this change is ideal so we should keep it 👌🏼.

fselmo (commented):
I took another pass at it. I think we can mock the message stream a bit better but other than that this looks good! Let me know your thoughts on that and if I missed any point you were trying to make with these changes 👀. Thanks again for the contribution!

tests/core/providers/test_wsv2_provider.py
kaiix: Great, this way we only need to update the mock without changing the test case.

fselmo (approved):
lgtm, thanks again for the contribution 👍🏼

vitor-alves:
Thank you for working on this @kaiix . I was having this issue and didn't know what the problem was exactly.
After 2 or 3 days of running 24/7 I was facing this issue consistently. The websocket connection dies, no exception is thrown, and it ends up stuck in an infinite loop.

Thank you for your contribution!

Issue #3418, opened 2024-06-21
HTTPProvider timeout is not working for chunked responses


Uxio0:
### What happened?

This is a problem with urllib3 and not with web3.py. If server answers with a chunked response, timeout will not be respected. More info:

- https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Timeout
- https://github.com/psf/requests/issues/5668

### Code that produced the error

```python
from web3 import Web3
w3 = Web3(Web3.HTTPProvider("http://127.0.0.1:8545", request_kwargs={'timeout': 60}))


A big query to `tracing.trace_filter` using an Erigon node. Erigon answers with a `Transfer-Encoding: chunked` header and starts slowly streaming the result, not letting the timeout trigger, query takes more than 10 minutes.
```

### Full error output

```shell
No error. Code hangs when `Web3.py` does the request until Erigon (or other RPC) ends streaming the response.
```

### Fill this section in if you know how this could or should be fixed

Using `requests.post(url, json=data, stream=True)` would allow Web3.py to control for how long the request is stuck, and process the chunked data in a more optimal way. Also, if the data is not chunked it shouldn't be an issue at all.

### web3 Version

6.19.0

### Python Version

3.12.4

### Operating System

linux

### Output from `pip freeze`

```shell
aiohttp==3.9.5
aiosignal==1.3.1
amqp==5.2.0
asgiref==3.7.2
astroid==3.1.0
asttokens==2.4.1
attrs==23.2.0
billiard==4.2.0
bitarray==2.9.2
black==24.4.2
boto3==1.34.117
botocore==1.34.126
CacheControl==0.14.0
cached-property==1.5.2
cachetools==5.3.3
celery==5.4.0
certifi==2024.2.2
cffi==1.16.0
cfgv==3.4.0
charset-normalizer==3.3.2
ckzg==1.0.1
click==8.1.7
click-didyoumean==0.3.1
click-plugins==1.1.1
click-repl==0.3.0
coverage==7.5.3
cron-descriptor==1.4.3
cryptography==42.0.5
cytoolz==0.12.3
debugpy==1.8.1
decorator==5.1.1
dill==0.3.8
distlib==0.3.8
Django==5.0.6
django-appconf==1.0.6
django-cache-memoize==0.2.0
django-celery-beat==2.6.0
django-cors-headers==4.3.1
django-db-geventpool==4.0.5
django-debug-toolbar==4.3.0
[...]

PR #3428, opened 2024-07-03
Allow timeout for chunked responses

Uxio0:
- Related to #3418


kclowes:
@Uxio0 does this solution work for your use case?

Uxio0 (commented):
It looks awesome, thanks for taking time to implement it!

fselmo (approved):
lgtm w/ adding a newsfragment 👍🏼

Issue #3384, opened 2024-05-08
Remove filtering by pending blocks


kclowes:
### What feature should we add?

Geth has removed the ability to filter by pending blocks. Reasoning from Geth 1.14.0 release notes here: 

> We see next to no value in monitoring the logs of a random subset of transactions that might (but probably mostly will not) end up in the next block. Power users who wish to monitor the transaction pool for MEV (or similar) purposes, should hook into Geth directly and extract the exact data they need, rather than rely on insufficient API endpoints.

I think we can fairly safely remove it too. TBD if this needs to go in v7 before breaking changes are wrapped.

pacrob:
Seems like this should be considered breaking, even if the endpoint isn't actually that useful. Adding to `v8` stuff.

vicmik:
Why remove this? It's arguably a mistake by the geth team to remove pending logs. I have had to bring them back manually by forking and reverting the commit.

Removing pending blocks has no value to web3py. Just because a client has stopped supporting it, doesn't mean that all other geth-like-API clients will stop supporting them.

PR #3430, opened 2024-07-09
Removed the ability to filter by pending blocks

BZAghalarov:
### What was wrong?

Related to Issue #3384 

Geth no longer supports a 'pending' flag. So it is no longer possible to filter by the pending flag.

### How was it fixed?
RPC method called eth_filter, some filter middlewares and the corresponding tests were updated.

### Todo:

#### Cute Animal Picture

image


BZAghalarov:
@kclowes , could you please take a look to this PR?

vicmik:
Why remove this? It's arguably a mistake by the geth team to remove pending logs. I have had to bring them back manually by forking and reverting the commit.

Removing pending blocks has no value to web3py. Just because a client has stopped supporting it, doesn't mean that all other geth-like-API clients will stop supporting them.

PR #3431, opened 2024-07-11
Use context manager for post request

kclowes:
@Uxio0 - I came up with something similar but without the manual management of the timeout. Does this do what you need it to? I can port to v7 if that's easier too.

### What was wrong?

Related to Issue #
Closes #

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/request.py
+++ b/web3/_utils/request.py
@@ -102,7 +102,6 @@ def json_make_get_request(
 def get_response_from_post_request(
     endpoint_uri: URI, *args: Any, **kwargs: Any
 ) -> requests.Response:
-    kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
     session = cache_and_return_session(endpoint_uri)
     response = session.post(endpoint_uri, *args, **kwargs)
     return response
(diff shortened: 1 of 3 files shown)


web3/_utils/request.py
```diff
@@ -111,9 +110,16 @@ def get_response_from_post_request(
 def make_post_request(
     endpoint_uri: URI, data: Union[bytes, Dict[str, Any]], *args: Any, **kwargs: Any
 ) -> bytes:
-    response = get_response_from_post_request(endpoint_uri, data=data, *args, **kwargs)
-    response.raise_for_status()
-    return response.content
+    kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
+    kwargs.setdefault("stream", True)
+    with get_response_from_post_request(
```
Uxio0, line 115: Hi @kclowes . Looks good, but you are still missing the manual control of the `timeout`. In this case, even if you are streaming, if let's say, you set the timeout to `60 seconds`, and the RPC returns `1 byte` every `59 seconds`, the timeout will not trigger and this loop can be stuck until the RPC stops sending data

kclowes: Gotcha. I'll close this is favor of yours then. On the one hand, it seems like this is desired behavior though? Could you decrease the value of your timeout?

kclowes: Curious to get your thoughts here too @fselmo. Related to #3418

Uxio0: Reducing the timeout is not helping if the server is slowly streaming a response (as Erigon does, for example)

PR #3432, opened 2024-07-12
Port similar changes from #3424 to main / v7:

fselmo:
### What was wrong?

Implement similar logic made in #3424 for `v7` / `main` branch, gracefully handling the case of `ConnectionClosedOK` for websocket connections.

### How was it fixed?

- If a persistent connection is closed gracefully, log and raise a silent ``PersistentConnectionClosedOK`` exception, triggering an end to the message listener task and breaking out of the ``process_subscriptions()`` iterator.

Since we have very different socket logic between `AsyncIPCProvider` and `WebSocketProvider`, handle the specific websocket case inside the `_provider_specific_message_listener()` method. For WebSocket case, if a `ConnectionClosedOK` is raised, silence it and re-raise a `PersistentConnectionClosedOK`.

I don't currently see an equivalent case for `AsyncIPCProvider` so this pattern isn't utilized for that specific provider.

### Todo:

- [x] Add specific test


pacrob (approved):
lgtm, :sheep: it!

PR #3433, opened 2024-07-16
Refactor public API for persistent connection provider socket interactions

fselmo:
- Refactor shared logic for persistent connection providers back into the `PersistentConnectionProvider` base class.
- Refactor reading from and sending to the socket into their own methods. Add these as abstract methods on the base class and force subclasses to implement them.

Important:

- Refactor the public `socket` API (`PersistentConnection` class) to properly define a `send()`, `recv()`, and `make_request()` which do what they say they do. Previously, `send()` was actually a "send_and_recv()" method. This was confusing and not a good API design. Note, these methods don't and cannot process the response in the way that invoking a method via its module within the web3 class does.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/web3/providers/persistent/persistent_connection.py
+++ b/web3/providers/persistent/persistent_connection.py
@@ -2,7 +2,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
-    Optional,
 )
 
 from web3.types import (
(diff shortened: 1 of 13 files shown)


web3/providers/persistent/persistent_connection.py
```diff
@@ -28,16 +27,56 @@ class PersistentConnection:
...
+        :return: The active subscriptions on the persistent connection.
+        :rtype: Dict[str, Any]
+        """
         return self._manager._request_processor.active_subscriptions
 
-    async def send(self, method: RPCEndpoint, params: Any) -> RPCResponse:
-        return await self._manager.send(method, params)
+    async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
+        """
+        Make a request to the persistent connection, returning the response. This method
```
reedsa:
```suggestion
        Make a request to the persistent connection and return the response. This method
```

reedsa (commented):
First pass, I like the naming updates and cleaner organization. Good refactor!

web3/providers/persistent/persistent_connection.py
```diff
@@ -28,16 +27,56 @@ class PersistentConnection:
...
+        """
+        await self._manager.send(method, params)
+
+    async def recv(self) -> RPCResponse:
+        """
+        Receive the next unprocessed response for a request from the persistent
+        connection.
+
+        :return: RPCResponse: The raw, unprocessed message from the
+                                        persistent connection.
```
pacrob: weird indentation

pacrob:
When you're doing docs, I noticed a typo [here](https://github.com/ethereum/web3.py/blob/e9ecd9db4d9f313761f0efbfcb6afdba4269c3fc/docs/providers.rst?plain=1#L363) - function name has an "s", but calling it does not. Seems appropriate to add here rather than separate PR.

pacrob (approved):
oop, looks good but failing tests!

fselmo:
> oop, looks good but failing tests!

The async IPC `socket_recv()` implementation was not correct which was leading to intermittently failed tests, especially for the `asyncio.gather()` test. We needed to check that `self._raw_message` wasn't already a well-formed message before reading from the socket. I updated it, please [take a look here](https://github.com/ethereum/web3.py/pull/3433/files#diff-e313da5acd150220a9d246a9f5341075d0fa4635df8215718226da8819453509R103-R115).

pacrob (approved):
:sheep:

PR #3434, opened 2024-07-17
coerce accessList storageKeys to HexStr, not HexBytes

pacrob:
### What was wrong?

`eth-account` expects `storageKeys` to be a sequence of `HexStr`s, as does `web3` in its defined type, but in `method_formatters.py` we were coercing it to `HexBytes`

Related to Issue https://github.com/ethereum/eth-account/issues/286

### How was it fixed?

Created new `storage_key_to_hexstr` to properly format `accessList`s.

@fselmo noted that the test for default exceptions in `test_library_files.py` had the wrong path. Updated and fixed the one new resulting error.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


web3/_utils/method_formatters.py
```diff
@@ -187,10 +187,26 @@ def apply_list_to_array_formatter(formatter: Any) -> Callable[..., Any]:
     return to_list(apply_formatter_to_array(formatter))
 
 
+def storage_key_to_hexstr(value: Union[bytes, int, str]) -> HexStr:
+    if isinstance(value, str):
+        if value.startswith("0x") and len(value) == 66:
```
fselmo, line 196: Ideally we can simplify this, especially the two messages that are exactly the same. They should also be `Web3ValueError` instances... we should check the test [here](https://github.com/ethereum/web3.py/blob/main/tests/core/test_library_files.py#L12-L30) and make sure this gets picked up and fails that test before changing it. Not sure why it didn't pick this up.

Maybe something like:

```py
def storage_key_to_hexstr(value: Union[bytes, int, str]) -> HexStr:
    if not isinstance(value, (str, int, bytes)):
        # initial type validation with early exit
        raise Web3ValueError(f"Storage key must be one of bytes, int, str, got {type(value)}")
    
    if isinstance(value, str):
        if len(value) == 66 and value.startswith("0x"):
            return HexStr(value)
        elif len(value) == 64:
            return HexStr(f"0x{value}")
    
    elif isinstance(value, bytes):
        if len(value) == 32:
            return HexBytes(value).to_0x_hex()
    
    elif isinstance(value, int):
        return storage_key_to_hexstr(hex(value))
    
    raise Web3ValueError(f"Storage key must be a 32-byte value, got {value!r}")
```

fselmo: Alternatively, remove the initial type validation altogether and only raise the exception in the last line with an all-encompassing message like `"Storage key must be a 32-byte value represented as int, str, or bytes, got {value!r}`"

fselmo: Perhaps [this conditional](https://github.com/ethereum/web3.py/blob/main/tests/core/test_library_files.py#L14-L16) might be removing the entirety of the `_utils` folder, not just the module testing utils as intended?

---
edit: nm, I don't think that would happen. Let me know if you need another set of eyes on that.

fselmo (approved):
lgtm!

PR #3443, opened 2024-07-26
Remove Web3ValidationError dependency on eth-utils ValidationError

### LINKED ISSUE #3441: Remove dependence on eth-utils `ValidationError` for `Web3ValidationError`
All *web3.py* exceptions extend from `Web3Exception`. This will suffice for fine-tuning all exceptions raised within *web3.py*. 

The current approach to the tooling libraries is to internalize all exceptions as internal exception classes to each library, e.g. `Web3Exception`, `ENSException`, `PyGethException`, etc...

One day, if all libraries are caught up, perhaps it will make sense to inherit all of them from *eth-utils* base exception classes for better control over exception handling. For now, let's remove any ambiguity about how they should behave by removing this inheritance from `Web3ValidationError`.

### Tasks

- [ ] Remove ``ValidationError`` from ``Web3ValidationError`` inherited classes [here](https://github.com/ethereum/web3.py/blob/main/web3/exceptions.py#L188).
- [ ] Update the migration guide to reflect this breaking change.

fselmo:
### What was wrong?

Closes #3441 

### Todo:

### DIFF
--- a/ens/exceptions.py
+++ b/ens/exceptions.py
@@ -1,6 +1,3 @@
-from eth_utils import (
-    ValidationError,
-)
 import idna
 
 
@@ -104,8 +101,7 @@ class UnderfundedBid(ENSException):
     """
 
 
-# type ignored because subclassing ValidationError which has type Any
-class ENSValidationError(ENSException, ValidationError):  # type: ignore[misc]
+class ENSValidationError(ENSException):
     """
     Raised if there is a validation error
     """
--- a/newsfragments/3443.breaking.rst
+++ b/newsfragments/3443.breaking.rst
@@ -1,0 +1,1 @@
+Remove ``Web3ValidationError`` dependence / inheritance from `eth-utils` ``ValidationError``.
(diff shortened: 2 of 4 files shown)


pacrob:
Looks like there's still import of `ValidationError` in  `ens/exceptions.py`, `tests/core/middleware/test_transaction_signing.py`, and `tests/ens/test_get_text.py`. Are those different, or are we looking to clear all of them?

kclowes (approved):
🚀 

edit: Approved once @pacrob's comments are addressed :)

fselmo:
> Looks like there's still import of ValidationError in ens/exceptions.py, tests/core/middleware/test_transaction_signing.py, and tests/ens/test_get_text.py. Are those different, or are we looking to clear all of them?

Good catch. We should remove the dependency from `ENSValidationError` for sure. In regards to the tests, I had seen those but they're raised within `py-evm` so I think that part's good. It's coming from another library and we're just making sure that it errors out.

Removed the dependency from `ENSValidationError` in the last commit. Re-review @pacrob.

pacrob (approved):
:sheep:

PR #3452, opened 2024-08-19
Updates to ci runs

fselmo:
### What was wrong?

- We have a few places to update `py-geth` versions that can be out of sync if we aren't careful. Update the version in all relevant places when we generate the fixture and take the version directly from the `GETH_BINARY` that needs to be set when the test fixture is generated. 

- Along the way, parametrize and DRY up the circleci config.

- We should also figure out a good way to `pip install --pre` (include pre-releases) for all of the ethereum libraries we maintain so we can get pre-releases building in the nightly runs and catch issues sooner (before they get to "stable").

^ I implemented the best approach I saw for installing pre-releases of our libs. The shortcoming is that the libs need to be added to the script but if we document it well enough in the internal handbook it seemed the least intrusive way into other parts of the codebase that I saw.

### Todo:

- [x] pip install pre-release versions of only our deps.
- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,5 @@
 version: 2.1
 
-# heavily inspired by https://raw.githubusercontent.com/pinax/pinax-wiki/6bd2a99ab6f702e300d708532a6d1d9aa638b9f8/.circleci/config.yml
-
 parameters:
   geth_version:
     # update default value when updating geth integration test fixture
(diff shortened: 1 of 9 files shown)


reedsa (approved):
nice cleanup! I like the single source of truth for the version and documenting like you have looks great. 👍

pacrob (approved):
like the dry CI
:sheep:

kclowes (approved):
This is very nice! 🤩

PR #3453, opened 2024-08-19
Descriptive Event `ABI` Error Messages

### LINKED ISSUE #964: Improve error messages on mismatched argument types
### What was wrong?

When sending an argument with the wrong type to a contract function, people find it difficult to debug and end up frustrated, like in #625

### How can it be fixed?

We largely solve for the *general* case: that multiple functions can exist with the same name. Obviously we have to support that, but it's incredibly common to have names that are unique even without the argument types. In these cases, we can give clearer error messages about exactly which types are expected for which arguments.

For example, we could raise the function:
> ValidationError: a function with name testFunction is defined by the contract, but some argument types don't match:
> -  address _location :heavy_check_mark:
> -  uint256 _reading :heavy_check_mark:
> -  uint256 _minParameter :heavy_check_mark:
> -  uint256 _maxParameter :heavy_check_mark:
> -  bytes32 _codeHash :warning: This argument was supplied as a string that was too long. If you want to supply a hex string, this function only accepts values that are exactly 64 characters (plus 2 for 0x). Alternatively, supply this argument as python native bytes, with a length of 32 bytes. Actually supplied argument: "0xFFFFF..." (length: 67)

reedsa:
### What was wrong?

Closes #964 

A `Web3ValueError` is raised by `get_event_abi` if a match is not found. This error is not very useful, but could include more information. `get_element_abi` returns a meaningful exception message which lists the arguments and potential matches. 

### How was it fixed?

Raise the same `MismatchedABI` exception from `get_event_abi`.

### Todo:

- [X] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -626,7 +626,7 @@ Taking the following contract code as an example:
     Traceback (most recent call last):
        ...
     web3.exceptions.MismatchedABI:
-    Could not identify the intended function with name
+    Could not identify the intended ABI with name
     >>> # check value is still b'aa'
     >>> arrays_contract.functions.getBytes2Value().call()
     [b'aa']
@@ -651,7 +651,7 @@ Taking the following contract code as an example:
     Traceback (most recent call last):
        ...
     web3.exceptions.MismatchedABI:
-    Could not identify the intended function with name
+    Could not identify the intended ABI with name
 
 .. _contract-functions:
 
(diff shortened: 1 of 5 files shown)


pacrob (approved):
:sheep:

reedsa:
In the context of https://github.com/ethereum/web3.py/pull/3457/files, exception messages will be raised in cases where the "event" ABI is not matched. Instead of merging this, I'll apply the message updates in the other PR.

PR #3455, opened 2024-08-19
Update filters.rst

bibo7086:
### What was wrong?
In the current implementation, estimated_end_block is calculated as current_block + chunk_size. This value could exceed end_block if the chunk size is large enough or if you’re near the end of the scan range. If estimated_end_block exceeds end_block, the scan might attempt to process blocks beyond the specified range, leading to issues.

### How was it fixed?
Capped estimated_end_block to end_block by updating the calculation to estimated_end_block = min(current_block + chunk_size, end_block).


pacrob:
Thanks @bibo7086 !

PR #3456, opened 2024-08-21
Pin ckzg library

kclowes:
### What was wrong?
ckzg updated something in v2.0, and our builds started failing. I haven't checked to see what it is yet, but figured I'd get this up so we're passing again.

### How was it fixed?
Added an upper pin here. It probably makes more sense to actually pin this in eth-account, though long term. Once we're pinned in eth-account and that change is released, we can remove the requirement here. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3456.bugfix.rst
+++ b/newsfragments/3456.bugfix.rst
@@ -1,0 +1,1 @@
+Pin ckzg library to <2.0 so core tests pass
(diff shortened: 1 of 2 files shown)


kclowes:
The actual fix is up in ethereum/eth-account#292. We can close this if we want to release eth-account before v7 stable, or keep it if we want to release v7 and then muck around with eth-account.

reedsa (approved):
🚀 Looks good if we decide this is needed!

kclowes:
Closing. fixing in eth-account instead.

PR #3459, opened 2024-08-21
Pin ckzg <2

kclowes:
### What was wrong?
ckzg released a breaking change in v2 for our blob transactions. 

### How was it fixed?
First I tried to pin eth-account to the lowest version that had the ckzg pin, but the 0.11.x line of eth-account dropped support for python 3.7, and so all of our Python 3.7 tests started failing. I think it makes the most sense to just require ckzg<2 here after all. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3459.bugfix.rst
+++ b/newsfragments/3459.bugfix.rst
@@ -1,0 +1,1 @@
+Pin ckzg dependency to <2 so that blob transactions work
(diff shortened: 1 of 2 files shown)


reedsa (approved):
sounds good! 👍

fselmo (approved):
yep, makes sense 👍🏼

PR #3463, opened 2024-08-27
(bugfix) Use a defined __hash__() for unhashable middleware:

### LINKED ISSUE #3461: web3 version 7.0.0: what happened to web3.middleware construct_sign_and_send_raw_middleware and pythonic_middleware?
### What happened?

Today we tried updating our Python project's requirements.txt files to use the latest version of 3rd party dependencies.

One of the dependencies that got updated was web3 went from version 6.20.2 --> 7.0.0

Our code had this import line that used to work in version 6.20.2:
`from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware, pythonic_middleware
`

**After the update to version 7.0.0, PyCharm says that all 3 of those items on that import line cannot be found.**

[From this section of the v6 to v7 Migration Guide](https://web3py.readthedocs.io/en/stable/migration.html#middleware-renaming-and-removals), I see that I need to replace
`geth_poa_middleware --> ExtraDataToPOAMiddleware`

_That works._

**The problem is those other 2 middleware:**
`construct_sign_and_send_raw_middleware    and    pythonic_middleware`

I can find no mention of them in the v6 to v7 Migration Guide, nor did a web search find anything related to how they changed in version 7.0.0.

_What happened to them?  Renamed?  Totally removed?_

This change has broke my code.

At a minimum, the v6 to v7 Migration Guide needs to tell me how to cope, and it is a documentation bug that it is currently unmentioned.

```python
from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware, pythonic_middleware
```

### Full error output

_No response_

Update the v6 to v7 Migration Guide

### web3 Version

7.0.0

### Python Version

3.12.4

### Operating System

win

### Output from `pip freeze`

```shell
I do not feel comfortable posting my project's complete list of dependencies here.

I hope that it is sufficient to note that this is what pip freeze prints regarding web3:

web3==7.0.0
```

fselmo:
### What was wrong?

- Closes #3461
- There is a bug where the middleware builders, which return curry-wrapped ``build`` functions, end up being given the same identifier when a name is not provided for the middleware in the onion.

### How was it fixed?

This fixes the issue by defining a unique hash for the base ``Web3Middleware`` class and using that hash in the case where the middleware is unhashable and therefore indistinguishable from another "built" middleware. This also uses the `__hash__()` by default for any of the class-based middleware if a name isn't provided, rather than using a `repr()` as before - which is not unique enough, leading to the complications.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/tests/core/middleware/test_middleware.py
+++ b/tests/core/middleware/test_middleware.py
@@ -12,17 +12,25 @@ from web3.middleware import (
 )
 
 
+class TestMiddleware(Web3Middleware):
+    def response_processor(self, method, response):
+        if method == "eth_blockNumber":
+            response["result"] = 1234
+
+        return response
+
+
+class TestMiddleware2(Web3Middleware):
+    def response_processor(self, method, response):
+        if method == "eth_blockNumber":
+            response["result"] = 4321
+
+        return response
+
+
 def test_middleware_class_eq_magic_method():
     w3_a = Web3()
     w3_b = Web3()
-
-    class TestMiddleware(Web3Middleware):
-        def request_processor(self, method, params):
-            return 1234
-
-    class TestMiddleware2(Web3Middleware):
-        def request_processor(self, method, params):
-            return 4321
 
     mw1w3_a = TestMiddleware(w3_a)
     assert mw1w3_a is not None
@@ -61,3 +69,13 @@ def test_unnamed_middleware_are_given_unique_keys(w3):
     with pytest.raises(Web3ValueError):
         # adding the same middleware again should cause an error
         w3.middleware_onion.add(request_formatting_middleware)
+
+
+def test_unnamed_class_middleware_are_given_unique_keys(w3):
+    w3.middleware_onion.add(TestMiddleware)
+    w3.middleware_onion.add(TestMiddleware2)
+    assert isinstance(w3.eth.block_number, int)
+
+    with pytest.raises(Web3ValueError):
+        # adding the same middleware again should cause an error
+        w3.middleware_onion.add(TestMiddleware)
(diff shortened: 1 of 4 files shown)


tests/core/middleware/test_middleware.py
```diff
@@ -0,0 +1,63 @@
...
+    mw1w3_a_equal = TestMiddleware(w3_a)
+    assert mw1w3_a == mw1w3_a_equal
+
+    mw2w3_a = TestMiddleware2(w3_a)
+    mw1w3_b = TestMiddleware(w3_b)
+    assert mw1w3_a != mw2w3_a
+    assert mw1w3_a != mw1w3_b
+
+
+def test_unnamed_middleware_are_given_unique_keys(w3):
```
kclowes, line 48: Was this happening when the middleware got passed in as a class too, rather than the build method? Do you think it's worth adding a test for that case?

web3/datastructures.py
```diff
@@ -248,15 +248,25 @@ def replace(self, old: TKey, new: TKey) -> TValue:
             self._queue[old_name] = new
         return to_be_replaced
 
-    def _repr_if_not_hashable(self, value: TKey) -> TKey:
+    @staticmethod
+    def _build_tkey(value: TKey) -> TKey:
```
kclowes: what do you think about calling this function something like `_build_name`?

kclowes (approved):
I left a comment about naming, and adding another test if appropriate, but feel free to take or leave. Thanks for getting this in!

web3/datastructures.py
fselmo: That works 👍🏼

tests/core/middleware/test_middleware.py
fselmo: I definitely think it's worth it to add that, will do.

pacrob (approved):
:sheep:

PR #3467, opened 2024-08-28
get_request_headers combomethod

### LINKED ISSUE #3466: Make `HTTPProvider.get_request_headers()` a class-level method

At the time of writing this issue, the method is defined here:
https://github.com/ethereum/web3.py/blob/main/web3/providers/rpc/rpc.py#L119-L123

It is:

```python
    def get_request_headers(self) -> Dict[str, str]:
        return {
            "Content-Type": "application/json",
            "User-Agent": construct_user_agent(str(type(self))),
        }
```

I am requesting this method be exposed at the class-level, to be re-written as:

```python
    @classmethod
    def get_request_headers(cls) -> Dict[str, str]:
        return {
            "Content-Type": "application/json",
            "User-Agent": construct_user_agent(str(cls)),
        }
```

Use-case: This allows me to grab the default headers _before_ constructing an `HTTPProvider` object for the purpose of modifying them / appending to them, without having to redefine.

kclowes:
### What was wrong?

@antazoey requested that get_request_headers be exposed as a class method. 

Closes #3466 

### How was it fixed?
Added a check to see where to get the class name from, and a combomethod decorator, so it can be called either way.  

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3467.feature.rst
+++ b/newsfragments/3467.feature.rst
@@ -1,0 +1,1 @@
+HTTPProvider and AsyncHTTPProvider's get_request_headers is now available on both the class and the instance
--- a/tests/core/providers/test_async_http_provider.py
+++ b/tests/core/providers/test_async_http_provider.py
@@ -94,8 +94,8 @@ async def test_async_user_provided_session() -> None:
     assert cached_session == session
 
 
-def test_get_request_headers():
-    provider = AsyncHTTPProvider()
+@pytest.mark.parametrize("provider", (AsyncHTTPProvider(), AsyncHTTPProvider))
+def test_get_request_headers(provider):
     headers = provider.get_request_headers()
     assert len(headers) == 2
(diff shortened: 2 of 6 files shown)


fselmo (approved):
lgtm 👍🏼

pacrob (approved):
:sheep:

souliane:
Hi, I am wondering why you chose to use `@combomethod`  instead of the stdlib `@classmethod`? It would also work with `self.get_request_headers()`:
```python
In [1]: class A:
   ...:     @classmethod
   ...:     def test(cls):
   ...:         print(cls)
   ...: 

In [2]: A.test()
<class '__main__.A'>

In [3]: A().test()
<class '__main__.A'>
```
 Did I miss something?

kclowes:
Mostly codebase consistency, and to make it explicit that calling it either way is fine and by design.

PR #3469, opened 2024-08-29
base default maxFeePerGas value off of maxPriorityFeePerGas

### LINKED ISSUE #3052: A race condition sometimes causes a transaction to have an invalid pair of maxFeePerGas/maxPriorityFeePerGas
When `Web3` is asked to construct a transaction without gas fees set explicitly, the values of the `maxFeePerGas` and `maxPriorityFeePerGas` fields are filled as part of the defaulting process, governed by this code:

https://github.com/ethereum/web3.py/blob/4b509a7d5fce0b9a67dbe93151e8b8a01e83b3cc/web3/_utils/transactions.py#L69-L80

https://github.com/ethereum/web3.py/blob/4b509a7d5fce0b9a67dbe93151e8b8a01e83b3cc/web3/_utils/transactions.py#L103-L137

If a new block is created right as the defaulting process takes place, then `w3.eth.max_priority_fee` will return different values for different fields of the same transaction. This can cause the transaction to end up with `tx['maxFeePerGas'] < tx['maxPriorityFeePerGas']`, which is invalid and will cause it to fail further down the line with something like
```json
{ "jsonrpc": "2.0"
, "id": 1383832
, "error":
    { "code": -32000
    , "message": "max priority fee per gas higher than max fee per gas"
    }
}
```

As for the fix, I propose to change the behavior like this:
1. Swap the order of the fields in the dictionary, so that `"maxPriorityFeePerGas"` goes before `"maxFeePerGas "`. Since Python dictionaries remember the insertion order of their keys and respect it during iteration, changing their order in the `TRANSACTION_DEFAULTS` map will correspondingly change the order the initializers are actually run.
2. In the loop, pass the `defaults` accumulator to the initializers together with `w3` and the initial transaction. Again, since Python guarantees to iterate over maps in the key insertion order, it's perfectly fine to carry state in-between iterations and rely on some actions being done before others.
3. In the `"maxFeePerGas"` initializer, instead of asking for `w3.eth.max_priority_fee` again, use `tx.get('maxPriorityFeePerGas', defaults['maxPriorityFeePerGas'])`.

So, with these considerations, the fix in code would look somewhat like this:
```py
# >>> add 3rd parameter `defs` to the lambda
TRANSACTION_DEFAULTS = {
    "value": 0,
    "data": b"",
    "gas": lambda w3, tx, defs: w3.eth.estimate_gas(tx),
    "gasPrice": lambda w3, tx, defs: w3.eth.generate_gas_price(tx),
# >>> swap the order of `maxFeePerGas` and `maxPriorityFeePerGas`
    "maxPriorityFeePerGas": lambda w3, tx, defs: w3.eth.max_priority_fee,
    # maxFeePerGas depends on maxPriorityFeePerGas having already run
    "maxFeePerGas": lambda w3, tx, defs: (
        tx.get("maxPriorityFeePerGas", defs["maxPriorityFeePerGas"])
        + (2 * w3.eth.get_block("latest")["baseFeePerGas"])
    ),
    "chainId": lambda w3, tx: w3.eth.chain_id,
}

<...>

@curry
def fill_transaction_defaults(w3: "Web3", transaction: TxParams) -> TxParams:
    """
    if w3 is None, fill as much as possible while offline
    """
    strategy_based_gas_price = w3.eth.generate_gas_price(transaction)
    is_dynamic_fee_transaction = strategy_based_gas_price is None and (
        "gasPrice" not in transaction  # default to dynamic fee transaction
        or any_in_dict(DYNAMIC_FEE_TXN_PARAMS, transaction)
    )

    defaults = {}
    for key, default_getter in TRANSACTION_DEFAULTS.items():
        if key not in transaction:
            if (
                is_dynamic_fee_transaction
                and key == "gasPrice"
                or not is_dynamic_fee_transaction
                and key in DYNAMIC_FEE_TXN_PARAMS
            ):
                # do not set default max fees if legacy txn or
                # gas price if dynamic fee txn
                continue

            if callable(default_getter):
                if w3 is None:
                    raise ValueError(
                        f"You must specify a '{key}' value in the transaction"
                    )
# >>> pass current `defaults` to the getter
                default_val = default_getter(w3, transaction, defaults)
            else:
                default_val = default_getter

            defaults[key] = default_val
    return merge(defaults, transaction)
```

pacrob:
### What was wrong?

Default values for `maxFeePerGas` and `maxPriorityFeePerGas` each relied on a separate RPC call to `eth.max_priority_fee`. This causes problems if a new block appears in between the 2 calls. See #3052 

Closes #3052

### How was it fixed?

Make a single call to `eth.max_priority_fee` and base the `maxFeePerGas` calculation off of that value. Additionally, our previous defaults would still make the second call to calculate `maxFeePerGas` even if `maxPriorityFeePerGas`  already existed in the transaction. This now uses that key if it exists in the transaction being filled.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/web3/_utils/async_transactions.py
+++ b/web3/_utils/async_transactions.py
@@ -1,6 +1,8 @@
 from typing import (
     TYPE_CHECKING,
+    Dict,
     Optional,
+    Union,
     cast,
 )
 
@@ -42,21 +44,32 @@ if TYPE_CHECKING:
     )
 
 
-async def _estimate_gas(async_w3: "AsyncWeb3", tx: TxParams) -> int:
+# unused vars present in these funcs because they all need to have the same signature
+async def _estimate_gas(
+    async_w3: "AsyncWeb3", tx: TxParams, _defaults: Dict[str, Union[bytes, int]]
+) -> int:
     return await async_w3.eth.estimate_gas(tx)
 
 
-async def _max_fee_per_gas(async_w3: "AsyncWeb3", _tx: TxParams) -> Wei:
+async def _max_fee_per_gas(
+    async_w3: "AsyncWeb3", tx: TxParams, defaults: Dict[str, Union[bytes, int]]
+) -> Wei:
     block = await async_w3.eth.get_block("latest")
-    max_priority_fee = await async_w3.eth.max_priority_fee
-    return Wei(max_priority_fee + (2 * block["baseFeePerGas"]))
+    max_priority_fee = tx.get(
+        "maxPriorityFeePerGas", defaults.get("maxPriorityFeePerGas")
+    )
+    return Wei(int(max_priority_fee) + (2 * int(block["baseFeePerGas"])))
 
 
-async def _max_priority_fee_gas(async_w3: "AsyncWeb3", _tx: TxParams) -> Wei:
+async def _max_priority_fee_gas(
+    async_w3: "AsyncWeb3", _tx: TxParams, _defaults: Dict[str, Union[bytes, int]]
+) -> Wei:
     return await async_w3.eth.max_priority_fee
 
 
-async def _chain_id(async_w3: "AsyncWeb3", _tx: TxParams) -> int:
+async def _chain_id(
+    async_w3: "AsyncWeb3", _tx: TxParams, _defaults: Dict[str, Union[bytes, int]]
+) -> int:
     return await async_w3.eth.chain_id
 
 
@@ -64,9 +77,9 @@ TRANSACTION_DEFAULTS = {
     "value": 0,
     "data": b"",
     "gas": _estimate_gas,
-    "gasPrice": lambda async_w3, tx: async_w3.eth.generate_gas_price(tx),
-    "maxFeePerGas": _max_fee_per_gas,
+    "gasPrice": lambda async_w3, tx, _defaults: async_w3.eth.generate_gas_price(tx),
     "maxPriorityFeePerGas": _max_priority_fee_gas,
+    "maxFeePerGas": _max_fee_per_gas,
     "chainId": _chain_id,
 }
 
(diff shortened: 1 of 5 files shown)


fselmo (commented):
This makes a whole lot of sense. We should add similar logic to the async transaction defaults as well.

reedsa (approved):
This is great! Lgtm after async is in.

web3/_utils/async_transactions.py
```diff
@@ -42,31 +44,39 @@
...
-async def _max_priority_fee_gas(async_w3: "AsyncWeb3", _tx: TxParams) -> Wei:
+async def _max_priority_fee_gas(
+    async_w3: "AsyncWeb3", _tx: TxParams, _defs: Dict[str, Union[bytes, int]]
+) -> Wei:
     return await async_w3.eth.max_priority_fee
 
 
-async def _chain_id(async_w3: "AsyncWeb3", _tx: TxParams) -> int:
+async def _chain_id(
+    async_w3: "AsyncWeb3", _tx: TxParams, _defs: Dict[str, Union[bytes, int]]
```
kclowes: Is this new `_defs` arg because all of these methods need to have the same signature? Might be worth a comment to explain why we need all of these args if they aren't ever used.

kclowes (approved):
nit: I'd vote to make `_defs` -> `_defaults`, but feel free to take or leave :) I left one comment about adding a comment, but feel free to take or leave that one too.

fselmo (approved):
lgtm!

web3/_utils/async_transactions.py
pacrob: Exactly that. Comment added in both locations.

PR #3472, opened 2024-08-30
Contract Events APIs

reedsa:
### What was wrong?

Contract function classes provide functions for extracting contract components. Contract event API methods do not yet exist. The first set of methods for `find_events_by_identifier` and `get_event_by_identifier` should be added.

Contract Event APIs should be provided, namely `all_events`, `find_events_by_name` and `get_event_by_name`. 

### How was it fixed?

Implement functions for `Contract` classes to find many events or getting a single event by identifier, `find_events_by_identifier` and `get_event_by_identifier`.

Additional Event APIs are now available (`all_events`, `find_events_by_name`, `get_event_by_name`, `find_events_by_signature`, `get_event_by_signature`, `find_events_by_selector`, `get_event_by_selector`, `find_events_by_topic`, and `get_event_by_topic`)

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/web3.contract.rst
+++ b/docs/web3.contract.rst
@@ -210,7 +210,23 @@ Each Contract Factory exposes the following properties.
 Methods
 -------
 
+Method doctests use the following ABI and bytecode.
+
+.. code-block:: python
+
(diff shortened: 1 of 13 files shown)


docs/web3.contract.rst
```diff
@@ -232,12 +248,13 @@ Each Contract Factory exposes the following methods.
 
     Returns the transaction hash for the deploy transaction.
 
-    .. code-block:: python
+    .. doctest:: contractmethods
```
kclowes, line 251: Nice work with this!

web3/contract/utils.py
```diff
@@ -365,6 +373,46 @@ def get_function_by_identifier(
     return fns[0]
 
 
+def find_events_by_identifier(
```
fselmo, line 376: nit: Should we rename this to `get_events_by_identifier` since the singular is `get_event_by_identifier`?

web3/contract/utils.py
```diff
@@ -365,6 +373,46 @@ def get_function_by_identifier(
...
+def get_event_by_identifier(
+    events: Sequence[TContractEvent], identifier: str
+) -> TContractEvent:
+    """
+    Check that the provided list of TContractEvent instances contains one element and
+    return it.
+    """
+    if len(events) > 1:
+        raise Web3ValueError(
+            f"Found multiple events with matching {identifier}. " f"Found: {events!r}"
```
fselmo, line 409: nit: use "identifier" in the message for clarity 

```suggestion
            f"Found multiple events with matching identifier `{identifier}`: {events!r}"
```

web3/contract/utils.py
```diff
@@ -365,6 +373,46 @@ def get_function_by_identifier(
...
+        )
+    elif len(events) == 0:
+        raise Web3ValueError(f"Could not find any event with matching {identifier}")
```
fselmo, line 412: nit

```suggestion
        raise Web3ValueError(f"Could not find any event with matching identifier `{identifier}`")
```

web3/contract/utils.py
```diff
@@ -356,6 +360,10 @@ def find_functions_by_identifier(
 def get_function_by_identifier(
     fns: Sequence[TContractFn], identifier: str
 ) -> TContractFn:
+    """
+    Check that the provided list of TContractFunction instances contains one element and
+    return it.
+    """
     if len(fns) > 1:
         raise Web3ValueError(
             f"Found multiple functions with matching {identifier}. " f"Found: {fns!r}"
```
fselmo, line 369: I see now we already phrase it this way with functions... take or leave really. Tiniest of nits.

```suggestion
            f"Found multiple functions with matching identifier `{identifier}`. Found: {fns!r}"
```

fselmo (commented):
This will be a great addition to get in. Currently as it stands, if I have an ABI that has both a `Deposit` event and a `Deposited` event, and I try to use `contract.get_event_by_name()` with either "Deposit" or "Deposited", I can get neither event. Instead, I get:

```
Web3ValueError: Could not find any event with matching name
```

I don't think this is desired behavior. Those are not ambiguous but distinct names. Can we add a test for something like this and make sure it passes? I saw a few nits in message wording, feel free to take or leave those as I see it in the function messages as well. I'll take another pass here once we get that test in.

tests/core/contracts/test_extracting_event_data.py
```diff
@@ -545,14 +545,14 @@ def test_contract_event_get_logs_sorted_by_log_index(w3, emitter, request_mocker
     ]
 
     with request_mocker(w3, mock_results={"eth_getLogs": get_logs_response}):
-        logs = emitter.events.LogNoArguments().get_logs()
```
kclowes, line 548: Do we have any tests to make sure that an event called with parens still passes? I'm fine taking them out here, but it might be good to have an explicit test to make sure it still works since it's documented. And the parens/no parens might be worth removing in v8. It feels antithetical to the zen of python, and I don't think there is a whole lot of value add in being able to call events with both parentheses and without. Although I can see the argument for maintaining consistency with `ContractFunctions` /shrug

web3/contract/base_contract.py
```diff
@@ -865,9 +918,125 @@ def callable_check(fn_abi: ABIFunction) -> bool:
...
+        )
+
+    @combomethod
+    def get_event_by_signature(self, signature: str) -> "BaseContractEvent":
+        """
+        Return a distinct event with matching signature.
+        Raises a Web3ValueError if the signature is invalid or if there is no match or
+        more than one is found.
+        """
+        if " " in signature:
```
kclowes: Should we just strip out the spaces instead of raising here?

docs/web3.contract.rst
```diff
@@ -296,54 +314,156 @@ Each Contract Factory exposes the following methods.
...
+    Searches for a distinct event matching the name. Returns an instance of
+    :py:class:`ContractEvent` upon finding a match. Raises ``Web3ValueError`` if no
+    match is found or if multiple matches are found.
+
+    .. doctest:: contractmethods
+
+        >>> contract.get_event_by_name('Approval')
+        <Event Approval(address,address,uint256)>
+
+.. py:classmethod:: Contract.find_events_by_selector(selector)
```
kclowes, line 408: Does this work with the 4byte selector too or just the whole thing?

kclowes (commented):
This is looking good to me once the Deposit/Deposited edge case that @fselmo found is cleaned up. Those doctests are super nice, thanks for getting those in. I just left a few comments!

docs/web3.contract.rst
reedsa: I'm not certain I've come across events that are represented by a 4byte selector, so this is implemented to match the whole event signature encoding. If it makes sense to use the 4byte selector I can update this to match the function selector.

kclowes: Oh, right. This is events. I think we can just leave it as-is. If we get a request to change it later, we can.

web3/contract/base_contract.py
reedsa: I pulled this logic from `get_function_by_signature` but it might be better to strip out spaces in the long run. I can strip those in `get_event_by_signature` for now.

web3/contract/utils.py
reedsa: There's a convention that follows the `find_*` used for lookups which return a list of items, and `get_*` when expecting a single result. The get method takes the resulting list from the find and just returns a single element (if there is not a single result in the list, the get method raises an exception).

web3/contract/utils.py
reedsa: I think `identifier` is a little confusing because we use `abi_element_identifier` to represent the name of the `ABIFunction` or `ABIEvent`.

In this context, `identifier` is a string that represents the type of "check" passed in. For example, `get_function_by_signature` uses `find_functions_by_identifier` by passing a "check" function that retrieves all functions from the ABI which passes the check. Then the resulting list of ABI components is passed to `get_function_by_identifier` to ensure there is only one result in the list and returns the ABI component itself. If not, an exception is raised, and the `identifier` is `"signature"`, indicating the search type.

reedsa:
@fselmo I havent reproduced the error you saw but added tests to cover those cases for both functions and events. Please let me know what the ABI was that you used and maybe there's an edge case I'm missing. Thanks!

web3/contract/async_contract.py
```diff
@@ -556,6 +581,27 @@ def get_function_by_identifier(
...
 
+    @combomethod
+    def find_events_by_identifier(
+        cls,
+        contract_abi: ABI,
+        w3: "AsyncWeb3",
+        address: ChecksumAddress,
+        callable_check: Callable[..., Any],
+    ) -> List["AsyncContractEvent"]:
+        return cast(
```
kclowes: I think this is a redundant cast

web3/contract/base_contract.py
```diff
@@ -258,8 +272,12 @@ def _get_event_filter_params(
         return event_filter_params
 
     @classmethod
-    def factory(cls, class_name: str, **kwargs: Any) -> PropertyCheckingFactory:
-        return PropertyCheckingFactory(class_name, (cls,), kwargs)
+    def factory(
+        cls, class_name: str, **kwargs: Any
+    ) -> Union["ContractEvent", "AsyncContractEvent"]:
```
kclowes, line 277: I wonder if this should be `TContractEvent`?

web3/contract/base_contract.py
```diff
@@ -956,6 +1121,26 @@ def get_function_by_identifier(
             "This method should be implemented in the inherited class"
         )
 
+    @combomethod
+    def find_events_by_identifier(
```
kclowes, line 1056: nit:`find/get_functions/events_by_id` are all beneath the `# Private helpers` comment. Consider moving above 🤷

web3/contract/contract.py
```diff
@@ -562,6 +585,27 @@ def get_function_by_identifier(
...
 
+    @combomethod
+    def find_events_by_identifier(
+        cls,
+        contract_abi: ABI,
+        w3: "Web3",
+        address: ChecksumAddress,
+        callable_check: Callable[..., Any],
+    ) -> List["ContractEvent"]:
+        return cast(
```
kclowes: redundant cast

kclowes (approved):
Looks good to me! Just left a nit or two, and some typing suggestions/clarifications 🚀

web3/contract/base_contract.py
reedsa: I found that using `TContractEvent` requires one argument for `factory` to be of that same type.

kclowes: Ah, okay. Thanks for looking into it!

PR #3473, opened 2024-08-31
Correct the .encode_abi code example snippet in the docs (web3.contract.rst)

flanagansteve:
Previously, `fn_name` was a named argument in the example snippet, used like:

```
contract.encode_abi(fn_name="register", ...)
```

But now, `fn_name` is the unnamed first argument, used like:

```
contract.encode_abi("register", ...)
```

(I believe, at least - I note that line 326, which defines the class method, uses it this way:

```
.. py:classmethod:: Contract.encode_abi(fn_name, args=None, kwargs=None, data=None)
```

And in my local project `encode_abi("getNames", ...)` worked where `encode_abi(fn_name="getNames", ...)` did not, which makes me think `encode_abi("register", ...)` is indeed the correct one.

#### Cute Animal Picture


pacrob:
Thanks, @flanagansteve! I've updated related places, otherwise should be gtg.

fselmo (approved):
👍🏼

PR #3476, opened 2024-09-04
Initialize contracts having function or event overrides

### LINKED ISSUE #1704: Events cannot retreive their ABI if multiple events have the same name

reedsa:
### What was wrong?

Closes #1704, #3482

The `Contract` class cannot initialize if the contract contains overloaded functions or events. Currently only the function/event name is used to filter ABI events from the contract ABI, so an exception is raised. Currently, events raise a ValueError when `_get_event_abi` is called on the Contract or ContractEvent class. Functions raise a MismatchedABI error.

Example:
`BaseContractEvent._get_event_abi()` calls eth-utils `get_event_abi()` and expects to return the ABI for the event class. The event classes are not created for every overloaded function. When there are contract events with the same name, only one is ever created and used.

### How was it fixed?

Initialize the `ContractFunctions` or `ContractEvents` class with every ABI found in the contract, storing the abi signature in the class name. Along with the signature, the `abi` property is set for each class within the `__init__` method. This ensures the correct instance is found and used for subsequent functions like `get_logs()`.

`get_abi_element` now handles both functions and events using the ABI signature to find a unique ABI. The signature is represented by a string that includes it's name and list of input types within parentheses. `name(arg1Type,arg2Type,...)`

Exceptions are now shared between functions and events for consistency.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


tests/core/contracts/conftest.py
```diff
@@ -760,3 +785,4 @@ def async_call(request):
 @pytest.fixture
 def async_estimate_gas(request):
     return async_partial(async_invoke_contract, api_call_desig="estimate_gas")
+    return async_partial(async_invoke_contract, api_call_desig="estimate_gas")
```
kclowes: copy pasta?

tests/core/contracts/test_contract_ambiguous_events.py
```diff
@@ -0,0 +1,260 @@
...
+            "0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7",
+            b"0x0",
+            1,
+        ],
+    )
+
+    assert event_abi == ABI_EVENT_DEPOSIT
+
+
+def test_get_abi_element_with_signature_for_amibguous_tuple_events() -> None:
```
kclowes: nit: 
```suggestion
def test_get_abi_element_with_signature_for_amibiguous_tuple_events() -> None:
```

tests/core/utilities/test_abi.py
```diff
@@ -510,14 +535,6 @@ def test_get_abi_element(
             TypeError,
             "Unsupported function identifier",
         ),
-        (
-            CONTRACT_ABI,
```
kclowes, line 514: Curious about this removal?

web3/contract/base_contract.py
```diff
@@ -148,19 +152,27 @@ class BaseContractEvent:
     w3: Union["Web3", "AsyncWeb3"] = None
     contract_abi: ABI = None
     abi: ABIEvent = None
+    argument_types: Tuple[str] = None
 
-    def __init__(self, *argument_names: Tuple[str]) -> None:
+    def __init__(
+        self,
+        *argument_names: Tuple[str],
+        abi: ABIEvent = None,
```
kclowes: nit?  
```suggestion
        abi: Optional[ABIEvent] = None,
```

web3/contract/base_contract.py
```diff
@@ -422,30 +432,40 @@ def __init__(
...
+                setattr(self, abi_to_signature(event), event_factory)
 
     def __getattr__(self, event_name: str) -> Type["BaseContractEvent"]:
         if "_events" not in self.__dict__:
             raise NoABIEventsFound(
                 "The abi for this contract contains no event definitions. ",
                 "Are you sure you provided the correct contract abi?",
             )
-        elif event_name not in self.__dict__["_events"]:
+        elif event_name.split("(")[0] not in [
```
kclowes: this `event_name.split...` may deserve to be its own method since you're using it so much in here. Feel free to take or leave :)

newsfragments/3476.feature.rst
```diff
@@ -0,0 +1 @@
+Contracts which overloaded functions or events are now supported. The Contract initializes functions and events using an identifier to distinguish between them. The identifier is the function or event signature, which consists of the name and the parameter types.
```
kclowes: nit:
```suggestion
Contracts with overloaded functions or events are now supported. The Contract initializes functions and events using an identifier to distinguish between them. The identifier is the function or event signature, which consists of the name and the parameter types.
```

web3/contract/base_contract.py
```diff
@@ -490,19 +510,39 @@ def __init__(self, abi: Optional[ABIFunction] = None) -> None:
...
+        if not args and not kwargs:
+            return cast(
                 ABIFunction,
                 get_abi_element(
-                    self.contract_abi,
-                    self.abi_element_identifier,
-                    *self.args,
-                    abi_codec=self.w3.codec,
-                    **self.kwargs,
+                    cls.contract_abi, get_abi_element_identifier(abi_element_identifier)
```
kclowes: Does this need the codec passed too?

kclowes (approved):
LGTM! Just left a few nits!

tests/core/utilities/test_abi.py
reedsa: Meant to move this to the `test_get_abi_element`, since now `get_abi_element` will determine the function by matching it's name (without arguments). Since `logTwoEvents` is the only function with that name, this case works.

I'll add a new failing case for an ambiguous function name that raises a MismatchedABI.

reedsa:
The ambiguity with functions is also an issue unfortunately. Taking a look at how to disambiguate since right now it uses the "first" one.

tests/ens/test_offchain_resolution.py
```diff
@@ -173,14 +173,14 @@ def test_offchain_resolver_function_call_raises_with_ccip_read_disabled(
     with pytest.raises(OffchainLookup):
         offchain_resolver.functions.resolve(
             ens_encode_name("offchainexample.eth"),
-            ENCODED_ADDR_CALLDATA,
+            ENCODED_ADDR_CALLDATA.encode(),
```
pacrob: Why the added `encode` here? Test passes for me with or without it.

web3/contract/contract.py
```diff
@@ -245,28 +255,55 @@ def __init__(
...
+            argument_types = get_abi_input_types(event_abi)
+            event_signature = str(get_abi_element_signature(event_name, argument_types))
+            return super().__getattribute__(event_signature)
+
+    def __getitem__(self, event_name: str) -> "ContractEvent":
+        return getattr(self, event_name)
+
+    def __iter__(self) -> Iterable["ContractEvent"]:
+        """
+        Iterate over supported
```
pacrob: nit - docstring should state what the function does (not sure if it's necessary for a magic method?). Also should match for the same method in `AsyncContractEvents`.

web3/utils/abi.py
```diff
@@ -306,24 +449,27 @@ def get_abi_element_info(
...
-    *args: Optional[Sequence[Any]],
+    *args: Optional[Any],
     abi_codec: Optional[Any] = None,
-    **kwargs: Optional[Dict[str, Any]],
+    **kwargs: Optional[Any],
 ) -> ABIElement:
     """
-    Return the interface for an ``ABIElement`` which matches the provided identifier
-    and arguments.
+    Return the interface for an ``ABIElement`` from the `abi` that matches the provided
```
pacrob:
```suggestion
    Return the interface for an ``ABIElement`` from the ``abi`` that matches the provided
```

web3/utils/abi.py
```diff
@@ -472,12 +598,17 @@ def check_if_arguments_can_be_encoded(
...
+@deprecated_for("get_abi_element")
 def get_event_abi(
     abi: ABI,
     event_name: str,
     argument_names: Optional[Sequence[str]] = None,
 ) -> ABIEvent:
     """
+    .. warning::
+        This function is deprecated. It is unable to distinguish between
+        overloaded events. Use `get_abi_element` instead.
```
pacrob:
```suggestion
        overloaded events. Use ``get_abi_element`` instead.
```

web3/contract/contract.py
reedsa: Good point, there's no practical information I provided there anyway. Removing!

tests/ens/test_offchain_resolution.py
reedsa: No need for this after all, good callout.

web3/utils/abi.py
```diff
@@ -231,6 +274,106 @@ def _get_argument_readable_type(arg: Any) -> str:
...
+    argument_types: Optional[Sequence[str]] = None,
+    abi_codec: Optional[Any] = None,
+    **kwargs: Optional[Any],
+) -> List[Callable[..., Sequence[ABIElement]]]:
+    """
+    Build a list of ABI filters to find an ABI element within a contract ABI. Each
+    filter is a partial function that takes a contract ABI and returns a filtered list.
+    Each parameter is checked before applying the relevant filter.
+
+    When the `abi_element_identifier` is a function name or signature and no arguments
```
pacrob:
```suggestion
    When the ``abi_element_identifier`` is a function name or signature and no arguments
```

web3/utils/abi.py
```diff
@@ -231,6 +274,106 @@ def _get_argument_readable_type(arg: Any) -> str:
...
+    are provided, the returned filters include the function name or signature.
+
+    A function ABI may take arguments and keyword arguments. When the ``args`` and
+    ``kwargs`` values are passed, several filters are combined together. Available
+    filters include the function name, argument count, argument name, argument type,
+    and argument encodability.
+
+    `constructor`, `fallback`, and `receive` ABI elements are handled only with a
```
pacrob:
```suggestion
    ``constructor``, ``fallback``, and ``receive`` ABI elements are handled only with a
```

web3/utils/abi.py
```diff
@@ -306,24 +449,27 @@ def get_abi_element_info(
...
     The `abi_codec` may be overridden if custom encoding and decoding is required. The
     default is used if no codec is provided. More details about customizations are in
     the `eth-abi Codecs Doc <https://eth-abi.readthedocs.io/en/latest/codecs.html>`__.
 
     :param abi: Contract ABI.
     :type abi: `ABI`
-    :param abi_element_identifier: Find an element ABI with matching identifier.
+    :param abi_element_identifier: Find an element ABI with matching identifier. The \
+    identifier may be a function name, signature, or ``FallbackFn`` or ``ReceiveFn``. \
+    A function signature is in the form `name(arg1Type,arg2Type,...)`.
```
pacrob:
```suggestion
    A function signature is in the form ``name(arg1Type,arg2Type,...)``.
```

pacrob (approved):
:sheep:

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
     assert math.abi == math_contract_abi
 
 
+def test_contract_init_with_abi_function_name(
+    w3,
+    function_name_tester_contract_abi,
+    function_name_tester_contract,
+):
+    # test `w3` function name does not throw when creating the contract factory
```
kclowes:
```suggestion
    # test `abi` function name does not throw when creating the contract factory
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    contract_factory = w3.eth.contract(abi=function_name_tester_contract_abi)
+
+    # re-instantiate the contract
+    contract = contract_factory(function_name_tester_contract.address)
+
+    # Contract w3 function should not override web3 instance
```
kclowes: Maybe something to test that `contract.abi` doesn't get overridden here instead?

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    with pytest.raises(AttributeError):
+        contract.functions.abi.get_block("latest")
+
+    assert contract.abi == function_name_tester_contract_abi
+
+    # assert the `w3` function returns true when called
```
kclowes:
```suggestion
    # assert the `abi` function returns true when called
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    assert result is True
+
+
+@pytest.mark.asyncio
+async def test_async_contract_init_with_abi_function_name(
+    async_w3,
+    function_name_tester_contract_abi,
+    async_function_name_tester_contract,
+):
+    # test `w3` function name does not throw when creating the contract factory
```
kclowes:
```suggestion
    # test `abi` function name does not throw when creating the contract factory
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    contract_factory = async_w3.eth.contract(abi=function_name_tester_contract_abi)
+
+    # re-instantiate the contract
+    contract = contract_factory(async_function_name_tester_contract.address)
+
+    # Contract w3 function should not override web3 instance
+    with pytest.raises(AttributeError):
+        contract.functions.abi.get_block("latest")
```
kclowes: Maybe a test to make sure `abi` doesn't get overridden here instead?

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +43,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+
+    assert contract.abi == function_name_tester_contract_abi
+
+    # assert the `w3` function returns true when called
```
kclowes:
```suggestion
    # assert the `abi` function returns true when called
```

web3/contract/base_contract.py
```diff
@@ -486,23 +516,49 @@ class BaseContractFunction:
     args: Any = None
     kwargs: Any = None
 
-    def __init__(self, abi: Optional[ABIFunction] = None) -> None:
+    def __init__(self, *arguments: Any, abi: Optional[ABIFunction] = None) -> None:
```
kclowes: This is a breaking change, since I believe now abi has to be passed in as a kwarg. Does that change anything for the UI?

web3/contract/base_contract.py
```diff
@@ -148,19 +154,55 @@ class BaseContractEvent:
...
-            self.argument_names = tuple()  # type: ignore
-        else:
-            self.argument_names = argument_names
+    def __init__(
+        self,
+        *argument_names: Tuple[str],
+        abi: Optional[ABIEvent] = None,
+        abi_element_identifier: Optional[ABIElementIdentifier] = None,
+    ) -> None:
+        # event was referenced with parens, juts return the event
```
kclowes: nit:
```suggestion
        # event was referenced with parens, just return the event
```

web3/contract/utils.py
```diff
@@ -119,6 +122,34 @@ def format_contract_call_return_data_curried(
     return normalized_data[0] if len(normalized_data) == 1 else normalized_data
 
 
+def copy_contract_function(
```
kclowes: I'm not sure we gain anything by exposing this publicly. Consider moving these two methods to `web3/_utils/contracts.py`

tests/core/contracts/test_contract_ambiguous_events.py
```diff
@@ -0,0 +1,260 @@
...
+                "type": "uint256",
+            }
+        ],
+        "name": "LogSingleArg",
+        "type": "event",
+    }
+
+    event_get_abi_result = ambiguous_event_contract.events[
+        "LogSingleArg(uint256)"
+    ]._get_event_abi()
```
kclowes: This is a stylistic nitpick, but it seems a little smelly to me to test a private method. It looks like `_get_event_abi` is used to assign the abi attribute on `ContractEvent`. Would it be possible to do something like: 

```suggestion
event_get_abi_result = ambiguous_event_contract.events[
    "LogSingleArg(uint256)"
].abi
```

Feel free to take or leave if that doesn't make sense.

kclowes (commented):
Sweet! I didn't see anything major, just left a few small comments.

web3/contract/base_contract.py
reedsa: This was originally intended to bring the same interface to `BaseContractFunction` and `BaseContractEvent`, but the `arugments` are never actually used.

I've also taken further steps to eliminate extraneous arguments passed to the factory function. Passing the signature is enough to find the ABI that is needed later on. I've made updates to the async/sync contracts so initialization of the ABI still works as intended.

web3/contract/utils.py
reedsa: Good point, moved!

tests/core/contracts/test_contract_ambiguous_events.py
reedsa: This is definitely better as `.abi` 👍

tests/core/contracts/test_contract_class_construction.py
reedsa: I made a slight adjustment to the order of the assertion/exception cases and updated the inline comments to be accurate. Let me know if there's further clarity needed there!

pacrob (approved):
:sheep:

kclowes (approved):
LGTM! Since this is such a core component of our library, it's probably worth some manual testing to make sure old functionality still works as expected. I did some myself and have fairly high confidence in the tests, but it's probably worth one more run through!

PR #3483, opened 2024-09-11
Bugfix tx caching

### LINKED ISSUE #1086: The simple cache caches unmined transactions
### What was wrong?

`w3.eth.getTransactionByHash` is caching the result, if the transaction is unmined. I believe some clients return `None` on an unmined transaction, and others return a transaction without a block number or hash.

Discovered by @palango in https://github.com/ethereum/web3.py/issues/1066#issuecomment-427321034

### How can it be fixed?

We can introspect the transaction in the simple cache and skip caching it when the block number or block hash is missing.

Should be backported to v4.

fselmo:
### What was wrong?

Closes #1086

### How was it fixed?

Validate certain cached requests against a validation threshold:

- Validate requests deemed unsafe to cache when they are not yet
  within a validation threshold of the chain. These requests
  rely on data from blocks and are only cached if the blocks they
  rely on are within this configured threshold.

- Allow threshold to be configurable to `finalized`, `safe`, or
  turned off by setting to `None`. Default this value to `finalized` 
  (safest option). Configuration is done via an enum.

- Never cache requests that make use of block identifiers:
  `finalized`, `safe`, `latest`, and `pending`. These are moving
  targets and will never be static, thus should never be cached.
  `earliest` should not be an issue.

- Add tests around the validation threshold boundary to make sure
  we cache only when appropriate, based on the configured options.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/tests/core/caching-utils/test_request_caching.py
+++ b/tests/core/caching-utils/test_request_caching.py
@@ -240,7 +240,8 @@ def test_blocknum_validation_against_validation_threshold_when_caching(
     ):
         assert len(w3.provider._request_cache.items()) == 0
         w3.manager.request_blocking(endpoint, [blocknum, False])
-        assert len(w3.provider._request_cache.items()) == int(should_cache)
+        cached_items = len(w3.provider._request_cache.items())
+        assert cached_items == 1 if should_cache else cached_items == 0
 
 
 @pytest.mark.parametrize(
@@ -281,7 +282,8 @@ def test_block_id_param_caching(
     ):
         assert len(w3.provider._request_cache.items()) == 0
         w3.manager.request_blocking(RPCEndpoint(endpoint), [block_id, False])
-        assert len(w3.provider._request_cache.items()) == int(should_cache)
+        cached_items = len(w3.provider._request_cache.items())
+        assert cached_items == 1 if should_cache else cached_items == 0
 
 
 @pytest.mark.parametrize(
@@ -530,7 +532,8 @@ async def test_async_blocknum_validation_against_validation_threshold(
     ):
         assert len(async_w3.provider._request_cache.items()) == 0
         await async_w3.manager.coro_request(endpoint, [blocknum, False])
-        assert len(async_w3.provider._request_cache.items()) == int(should_cache)
+        cached_items = len(async_w3.provider._request_cache.items())
+        assert cached_items == 1 if should_cache else cached_items == 0
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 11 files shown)


web3/_utils/caching/request_caching_validation.py
```diff
@@ -0,0 +1,133 @@
...
+        return False
+
+    blocknum = to_int(hexstr=response["result"]["number"])
+    return is_beyond_validation_threshold(provider, blocknum)
+
+
+# -- async -- #
+
+
+async def async_always_cache_request(*_args: Any, **_kwargs: Any) -> bool:
```
kclowes: nit: could just use `always_cache_request`?

kclowes (approved):
:shipit: Left one nit but looks good otherwise!

web3/_utils/caching/request_caching_validation.py
fselmo: Yeah, the validation methods are all `async` so this would break when we await it [here](https://github.com/ethereum/web3.py/pull/3483/files#diff-484ba843299ce80fc945089d4bae6683dfdfa7356a43c27432b644aad00827d8R228). But what we can do is add a conditional that checks whether the validation function is a coroutine and if it is call `await` and if it's not just run the func? Thoughts there?

fselmo: Actually, it's even less complicated since we define the `ALWAYS_CACHE` methods. I'll update it.

fselmo: ~~@kclowes I had to use a `type: ignore`. Thoughts? [commit is here](https://github.com/ethereum/web3.py/pull/3483/commits/70d9086219a44fcc526a025063e9b673e06a8fcc).~~

fselmo: Nevermind, I was able to resolve the typing. I like the change 👍🏼

tests/core/caching-utils/test_request_caching.py
```diff
@@ -187,9 +195,176 @@ def test_all_providers_do_not_cache_by_default_and_can_set_caching_properties(pr
...
+                # mock the threshold block to be blocknum "0x2" for all test cases
+                {"number": "0x2"}
+                if params[0] == threshold.value
+                else {"number": blocknum}
+            ),
+        },
+    ):
+        assert len(w3.provider._request_cache.items()) == 0
+        w3.manager.request_blocking(RPCEndpoint(endpoint), [block_id, False])
+        assert len(w3.provider._request_cache.items()) == int(should_cache)
```
reedsa: Nit: The assert here is a little awkward with `should_cache` as an int. Would suggest the ternary as you have in the other test below. Take or leave, slight readability improvement.
```suggestion
        cached_items = len(w3.provider._request_cache.items())
        assert cached_items == 1 if should_cache else cached_items == 0
```

reedsa (approved):
Tiny nit, no concerns overall! 👍

PR #3488, opened 2024-09-17
Encourage signing middleware injection before txn modifying middleware

### LINKED ISSUE #2167: construct_sign_and_send_raw_middleware uses unbuffered gas estimation following PR #1999.
After the change that made buffered gas estimate a middleware 
https://github.com/ethereum/web3.py/commit/17264b423f24a6c4e88f690391d5e9fc56bb0714
Adding the construct_sign_and_send_raw_middleware using w3.middleware_onion.add like the example
in the documentation:
https://web3py.readthedocs.io/en/stable/middleware.html#web3.middleware.construct_sign_and_send_raw_middleware
causes the signing middleware to run before the buffered gas estimate middleware and use unbuffered gas estimation that is computed here:
https://github.com/ethereum/web3.py/blob/v5.23.1/web3/_utils/transactions.py#L67

This causes sporadic out of gas failures.

injecting construct_sign_and_send_raw_middleware as the innermost middleware does not work well with the exception_retry_middleware.

fselmo:
### What was wrong?

Closes #2167

### How was it fixed?

- Encourage injecting the sign-and-send raw middleware at the 0th layer. Any middleware that modifies transaction params should be run before the signing middleware in order to sign the modified values.
- Add test for sign-and-send along with buffered gas estimate middleware

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

boxwink

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -319,9 +319,13 @@ Proof of Authority
 
 .. py:class:: web3.middleware.ExtraDataToPOAMiddleware
 
-.. note::
-    It's important to inject the middleware at the 0th layer of the middleware onion:
-    ``w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)``
+.. important::
+    It is **crucial** that this middleware is injected at the 0th layer of the
+    middleware onion, using
+    ``w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)``, to guarantee
+    it is the *first* middleware to process the response and modify the ``extraData``
+    field. This ensures it processes the field before any other middleware attempts
+    to validate it.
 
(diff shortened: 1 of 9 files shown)


pacrob:
We have the same suggestion for ExtraDataToPOAMiddleware:

It’s important to inject the middleware at the 0th layer of the middleware onion: w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)

so should probably clarify if using both.

kclowes (approved):
This makes sense to me! What happens if we have POA and SignAndSend middleware? Might be worth a clarifying note (maybe in the POA docs?) about which should go at `layer=0` in that case.

Edit: ha, @pacrob beat me to it 😆 🐌

PR #3490, opened 2024-09-18
Validate provider is correct type for the w3 class on init:

### LINKED ISSUE #3460: Add nice error message when a sync Web3 instance is used with an async provider and vice versa

When a web3 instance is instantiated with the `AsyncWeb3` class and a sync provider a TypeError gets raised: 
```
>>> from web3 import AsyncWeb3, HTTPProvider
>>> w3 = AsyncWeb3(HTTPProvider(...))
>>> await w3.is_connected()
TypeError: object bool can't be used in 'await' expression
```
When a provider gets passed in, we should check to make sure it's passed to an async web3 instance if it's an async provider or a sync web3 instance if a sync provider gets passed.

fselmo:
### What was wrong?

Closes #3460

### How was it fixed?

Disallow sync providers on `AsyncWeb3` init; disallow async providers on `Web3` init. This will provide better messaging to users when they try to use the wrong provider type.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

SSPX0264

### DIFF
--- a/newsfragments/3490.misc.rst
+++ b/newsfragments/3490.misc.rst
@@ -1,0 +1,2 @@
+
+Validate against sync providers on ``AsyncWeb3`` init; validate against async providers on ``Web3`` init. This will provide better messaging when a web3 class is instantiated with an incompatible provider type.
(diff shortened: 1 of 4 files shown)


pacrob (approved):
:sheep:

Issue #3482, opened 2024-09-11
Error when invoking ambiguous contract functions


falvaradorodriguez:
### What happened?

Migrating from version 6 to version 7 of the [`safe-eth-py`](https://github.com/safe-global/safe-eth-py) library we have detected a problem invoking a function of a contract with the same name but ambiguous arguments. 

Specifically, the function names are:

`isValidSignature(bytes32,bytes)` and `isValidSignature(bytes,bytes)`.

The values I am passing as an argument (for example) are:

- `self.safe_hash_preimage = b'\x9c’\xff_!\xf0\xb8\x1b11>c\xf7\xdbm\xa9O\xedxedxef\x11\xb2\x11\x9b@\x88\xb8\x96d\xfb9a<\xb6X‘` 
- `self.contract_signature = b''`.

The code we execute is similar to the one indicated in the [guide](https://web3py.readthedocs.io/en/stable/web3.contract.html#invoke-ambiguous-contract-functions) for this case and in the [migration guide](https://web3py.readthedocs.io/en/latest/migration.html#migrating-from-v6-to-v7) from version 6 to 7 I don't see any entry related to this problem. 

### Code that produced the error

```python
def is_valid(self, ethereum_client: EthereumClient, *args) -> bool:
        compatibility_fallback_handler = get_compatibility_fallback_handler_contract(
            ethereum_client.w3, self.owner
        )
        is_valid_signature_fn = (
            compatibility_fallback_handler.get_function_by_signature(
                "isValidSignature(bytes,bytes)"
            )
        )
        try:
            return is_valid_signature_fn(
                bytes(self.safe_hash_preimage), bytes(self.contract_signature)
            ).call() in (
                self.EIP1271_MAGIC_VALUE,
                self.EIP1271_MAGIC_VALUE_UPDATED,
            )
        except (Web3Exception, DecodingError, Web3ValueError, Web3RPCError):
            # Error using `pending` block identifier or contract does not exist
            logger.warning(
                "Cannot check EIP1271 signature from contract %s", self.owner
            )
        return False
```

### Full error output

```shell
Could not identify the intended function with name `isValidSignature`, positional arguments with type(s) `(bytes,bytes)` and keyword arguments with type(s) `{}`.
Found 2 function(s) with the name `isValidSignature`: ['isValidSignature(bytes32,bytes)', 'isValidSignature(bytes,bytes)']
Ambiguous argument encoding. Provided arguments can be encoded to multiple functions matching this call.
```

[...]

PR #3491, opened 2024-09-19
Raise `MismatchedABI` exceptions with detailed error messaging

### LINKED ISSUE #964: Improve error messages on mismatched argument types

reedsa:
### What was wrong?

In `get_abi_element`, exception handling between functions and events is now shared. The messages reference `functions`, but should instead say `element` and provide additional details about any element that matches the name. The provided arguments and their types should be checked with the encodability for each element ABI in order to list each argument error.

Closes #964, #3482

### How was it fixed?

Add logic for mismatched ABI errors to invoke a new private method that generates a list of errors for each ABI that matches the provided `abi_element_identifier`. The messages returned depend on the number of matches and if there are arguments, the number of matching ABIs that have the same number of arguments is also considered.

The messages will include all potential matches (by name) so that users can identify the function or event they intended to reference. The specific messages for each match depend on the arguments provided.

When an argument is not encodable to a certain type, the message appends information about the argument position, the provided value, and the expected type.

### Todo:

- [X] Clean up commit history

#### Cute Animal Picture


tests/core/contracts/test_contract_method_to_argument_matching.py
```diff
@@ -168,22 +169,69 @@ def test_finds_function_with_matching_args_non_strict(
...
     contract = w3.eth.contract(abi=MULTIPLE_FUNCTIONS)
-    with pytest.raises(MismatchedABI):
+    with pytest.raises(
+        MismatchedABI,
+        match=re.escape(
+            "\nABI Not Found!\n"
+            "Multiple elements were found matching 1 arguments.\n"
+            "Provided argument types: (str)\n"
+            "Provided keyword argument types: {}\n\n"
+            "Encountered problems with the following elements named `a`.\n"
```
kclowes: It took me a minute to figure out what this error message was telling me. Maybe something along the lines of: `Tried to find a matching ABI Element `a`, but encountered the following problems:` 🤷

tests/core/contracts/test_contract_method_to_argument_matching.py
```diff
@@ -168,22 +169,69 @@ def test_finds_function_with_matching_args_non_strict(
...
 
 def test_error_when_duplicate_match(w3):
     Contract = w3.eth.contract(abi=MULTIPLE_FUNCTIONS)
 
-    with pytest.raises(MismatchedABI):
+    with pytest.raises(
+        MismatchedABI,
+        match=re.escape(
+            "\nABI Not Found!\n"
+            "Multiple elements were found matching 1 arguments.\n"
```
kclowes: I'm not sure if you can make this wording work throughout, but maybe something like: `Multiple elements were found that accept 1 arguments`

tests/core/contracts/test_contract_method_to_argument_matching.py
kclowes: /cc our resident words guy ™️ @wolovim

kclowes (approved):
This is awesome! I left a few comments on the wording of the messages, but I think this will really help!

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
+                )
+            else:
+                error += (
+                    f"Error! Could not encode argument {position} value `{arg}` "
+                    f"as `{_type}`.\n"
+                )
+    else:
+        error += (
+            f"Function `{abi_element_name}` expects "
+            f"{len(abi_element_input_types)} arguments but {num_args} were given.\n"
```
fselmo:
```suggestion
            f"{len(abi_element_input_types)} argument(s) but received {num_args} argument(s).\n"
```

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
-        f"`{abi_element_identifier}`: {matching_function_signatures}{diagnosis}"
-    )
+    error = "ABI Not Found!\n"
+    if num_matches == 0 and num_abis_matching_arg_count == 0:
+        error += (
+            f"No declaration found for `{str(abi_element_identifier)}` with "
+            f"{num_args} arguments.\n"
+        )
+    elif num_matches > 1 or num_abis_matching_arg_count > 1:
+        error += f"Multiple elements were found matching {num_args} arguments.\n"
```
fselmo:
```suggestion
        error += f"Multiple elements were found matching {num_args} argument(s).\n"
```

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
-        f"`({collapsed_args})` and keyword arguments with type(s) "
-        f"`{collapsed_kwargs}`."
-        f"\nFound {len(matching_function_signatures)} function(s) with the name "
-        f"`{abi_element_identifier}`: {matching_function_signatures}{diagnosis}"
-    )
+    error = "ABI Not Found!\n"
+    if num_matches == 0 and num_abis_matching_arg_count == 0:
+        error += (
+            f"No declaration found for `{str(abi_element_identifier)}` with "
+            f"{num_args} arguments.\n"
```
fselmo:
```suggestion
            f"{num_args} argument(s).\n"
```

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
+        )
+    elif num_matches > 1 or num_abis_matching_arg_count > 1:
+        error += f"Multiple elements were found matching {num_args} arguments.\n"
+    elif num_abis_matching_arg_count == 1:
+        error += (
+            f"Found {num_abis_matching_arg_count} elements named `{name}` that takes "
```
fselmo:
```suggestion
            f"Found {num_abis_matching_arg_count} element(s) named `{name}` that takes "
```

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
+            f"{num_args} arguments.\n"
```
fselmo:
```suggestion
            f"{num_args} argument(s).\n"
```

web3/utils/abi.py
```diff
@@ -158,45 +164,148 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
+    elif num_matches == 0:
+        error += (
+            "Unable to find an element that matches the provided identifier and "
+            "argument types.\n"
+        )
+    arg_types = _extract_argument_types(*args)
+    kwarg_types = dict({(k, _extract_argument_types([v])) for k, v in kwargs.items()})
+    error += f"Provided argument types: ({arg_types})\n"
+    error += f"Provided keyword argument types: {kwarg_types}\n\n"
+    error += f"Encountered problems with the following elements named `{name}`.\n"
```
fselmo:
```suggestion
    error += f"Encountered problems with the following element(s) named `{name}`.\n"
```

tests/core/contracts/test_contract_method_to_argument_matching.py
fselmo: @kclowes I think it may be as simple as adding `(s)`. I made a comment in the actual messages under the abi utils. Do you think that clarifies well enough?

fselmo (commented):
I made an initial pass to contribute to the messaging conversation. I think we can probably add `(s)` around the descriptors for dynamic singular or multiple values, e.g. `1 elements(s)`, `2 argument(s)`, etc...

I'll take a more careful pass at the rest on Monday 👀

web3/utils/abi.py
fselmo: Is this always a singular element? If so, we can drop the `s` altogether.

tests/core/contracts/test_contract_method_to_argument_matching.py
kclowes: The word `matching` is what's confusing to me. To me, this says the element matches the argument (which I know is not the intent) but changing to to `accepts` clears it up. Either way, feel free to take or leave.

tests/core/utilities/test_event_interface.py
```diff
@@ -18,7 +19,7 @@
 
 def test_access_event_with_no_abi(w3):
     contract = w3.eth.contract()
-    with pytest.raises(NoABIEventsFound):
+    with pytest.raises(NoABIFound):
```
fselmo, line 22: Is this desirable? If we're under the `events` namespace, and for the sake of not breaking anything, we should probably keep the same exception. Thoughts there?

edit: We should also look at `ABIEventNotFound` cases.

web3/utils/abi.py
```diff
@@ -472,12 +661,17 @@ def check_if_arguments_can_be_encoded(
     )
 
 
+@deprecated_for("get_abi_element")
```
fselmo, line 702: We should add a `with pytest.warns()` test for this and assert it doesn't warn when we use `get_abi_element()`. We've had a few issues where the warning wasn't quite working as expected recently.

tests/core/utilities/test_event_interface.py
fselmo: We should also check that `NoABIFunctionsFound` and `ABIFunctionNotFound` are still raised where they used to be raised if we want to go that route.

fselmo (commented):
This looks great. I left some comments on error message wording. I think they could be good changes for clarity, including @kclowes comment.

I have some small outstanding questions on possible breaking changes as well. Exceptions that were raised before this PR and new classes seem to be raised now. Let's make sure we're not braking anything there.

web3/contract/contract.py
```diff
@@ -168,29 +182,28 @@ def get_logs(
           same time as ``from_block`` or ``to_block``
         :yield: Tuple of :class:`AttributeDict` instances
         """
-        event_abi = self._get_event_abi()
-
+        abi = self._get_event_abi()
```
fselmo: nit: this seems like a less clarifying change. We could keep `event_abi` here instead of the more generic "abi".

web3/utils/abi.py
reedsa: Keeping plural here. The error lists one to many elements with the same name after this message.

tests/core/utilities/test_event_interface.py
reedsa: In the [`__getattribute__` function of the `ContractFunctions` class](https://github.com/ethereum/web3.py/blob/9baae32574e5dd655f3be50aee0f31c46548cc16/web3/contract/contract.py#L417) there is logic in raising exceptions that separates whether a contract was intialized with an `ABI` or the contract has an `ABI` but contains no functions. Functions have always raised `NoABIFound` when the ABI is not available.

Events now behave similar to functions, so the logic for the exceptions match what functions have. For functions, any attempt to use functions would raise the `NoABIFound` or `NoABIFunctionsFound` exception.

I agree we should not have breaking changes for this. Maybe one day this could be addressed as a breaking change. I can create a new issue to track that.

For this, I will change the logic for events so that `NoABIEventsFound` is raised and it wont be breaking. Would it be better to combine the logic so the exception and message is the same for both cases?

Alternatively, the same exception could be raised with different messages based on either the ABI not being present vs. the contract having no events. Something like:

```python
        if super().__getattribute__("abi") is None:
            raise NoABIEventsFound(
                "There is no ABI found for this contract.",
            )
        if "_events" not in self.__dict__:
            raise NoABIEventsFound(
                "The abi for this contract contains no event definitions. ",
                "Are you sure you provided the correct contract abi?",
            )
```

fselmo: Going to mark this resolved. We cleared this up that a contract with no ABI at all should indeed raise the `NoABIFound` exception. This test case seems to be expecting the wrong exception here so this is technically a bugfix.

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -86,3 +148,53 @@ def test_error_to_call_non_existent_fallback(
...
+    (
+        (None, "functions", NoABIFound),
+        (None, "events", NoABIFound),
+        ([{"type": "event", "name": "AnEvent"}], "functions", NoABIFunctionsFound),
+        ([{"type": "function", "name": "aFunction"}], "events", NoABIEventsFound),
+        ([{"type": "function", "name": "aFunction"}], "functions", ABIFunctionNotFound),
+        ([{"type": "event", "name": "AnEvent"}], "events", ABIEventNotFound),
+    ),
+)
+def test_appropriate_exceptions_based_on_namespaces(
```
fselmo, line 91: ❤️

fselmo (dismissed):
~~lgtm!~~ I retracted the accept for now... I think we need to solve for some edge cases in messaging, etc. See comments below.

web3/utils/abi.py
```diff
@@ -158,46 +164,153 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
...
+    kwarg_types = dict({(k, _extract_argument_types([v])) for k, v in kwargs.items()})
+    error += (
+        f"Provided argument types: ({arg_types})\n"
+        f"Provided keyword argument types: {kwarg_types}\n\n"
+        f"Tried to find a matching ABI element named `{name}`, but encountered "
+        "the following problems:\n"
     )
 
+    for abi_signature in abi_signatures_matching_names:
+        error += _build_abi_input_error(
```
fselmo, line 332: @reedsa I approved but I think we need to tighten up the messaging a little bit. It looks like we only append problems when there are matching names. Perhaps we should make this code block a bit more like:

```py
if abi_signatures_matching_names:
    error += f"Tried to find a matching ABI element named `{name}`, but encountered the following problems:\n"

    for abi_signature in abi_signatures_matching_names:
        error += _build_abi_input_error(...)
```

Because if I run something like the following:

```bash
$ python -c "from web3.utils import get_abi_element;get_abi_element({}, 'nonexistent')"
```

I will get "encountered the following problems:" with no "problems" after it, just a blank.

fselmo:
@reedsa I'm also seeing a recursion issue with simplistic ABI structures. If I try to `get_abi_element()` with an ABI that has a simple function and event with the same name, I enter into `_build_abi_input_error()` which tries to `get_abi_element()` with a signature and re-enters `_build_abi_input_error()`, etc...

reproduce with:

```bash
python -c "from web3.utils import get_abi_element;get_abi_element([{'type': 'function', 'name': 'Nonexistent'},{'type': 'event', 'name': 'Nonexistent'}], 'Nonexistent')"
```

reedsa:
@fselmo Interesting to see the function/event name overlap. I think the general rule of thumb in solidity is using TitleCase for events, camelCase for functions. Though that is not explicitly enforced by Solidity as far as I know.

I'll have a look at the empty messages and recursive issue, good finds!

fselmo:
> @fselmo Interesting to see the function/event name overlap. I think the general rule of thumb in solidity is using TitleCase for events, camelCase for functions. Though that is not explicitly enforced by Solidity as far as I know.

Yeah that was more of a lazy quick example to test. I think the important takeaway is a graceful handling of the recursion / edge case rather than just letting it recurse.

reedsa:
@fselmo Found a better approach that eliminates recursive calls entirely. Also doing validations up front in `get_abi_element` by calling `validate_abi` and makes those type of errors much clearer.

fselmo:
> @fselmo Found a better approach that eliminates recursive calls entirely. Also doing validations up front in `get_abi_element` by calling `validate_abi` and makes those type of errors much clearer.

I think we can probably still trim this messaging a bit. It seems quite lengthy when I think we can convey the same message more briefly. I'm seeing double errors as well, one for each of the identifying elements. Was that intentional?

```shell
$ python -c "from web3.utils import get_abi_element;get_abi_element([{'type': 'event', 'name': 'MyElement'}, {'type': 'function', 'name': 'MyElement'}], 'MyElement')"

  File "<string>", line 1, in <module>
  File "/Users/fselmo/codigo/e/py/web3.py/web3/utils/abi.py", line 611, in get_abi_element
    raise MismatchedABI(error_diagnosis)
web3.exceptions.MismatchedABI: 
ABI Not Found!
Multiple elements were found that accept 0 argument(s).
Provided argument types: ()
Provided keyword argument types: {}

Tried to find a matching ABI element named `MyElement`, but encountered the following problems:
MyElement()
The provided identifier matches multiple elements.
If you meant to call `MyElement()`, please specify the full signature.
MyElement()
The provided identifier matches multiple elements.
If you meant to call `MyElement()`, please specify the full signature.
```

Doubled part:

```
MyElement()
The provided identifier matches multiple elements.
If you meant to call `MyElement()`, please specify the full signature.
```

What do you think? I can consider this a nit on my part if we think this is desirable. I'm not sure how much of an overhead it would be to not repeat the messaging per item found at the end.

reedsa:
Oh that does look strange for that case, since it lists both ABI elements in the abi that match the name. Maybe the error message should include the type of each `function` or `event` to make that clear. Would make it even more verbose so may not be much better. Something like:

```sh
MyElement() - Type: `function`
The provided identifier matches multiple elements.
If you meant to call `MyElement()`, please specify the full signature.
MyElement() - Type: `event`
The provided identifier matches multiple elements.
If you meant to call `MyElement()`, please specify the full signature.
```

fselmo:
@reedsa I think I see the point of the repeat (in this case). I'm also breaking the API on purpose... an ABI should never really define an event or function like that. I think this looks good. Messaging seems pretty clear to me and I'm not sure I mind the verbosity as much. It is at least very descriptive of the issues.

> Maybe the error message should include the type of each function or event to make that clear. 

I think including the type would be a great clarifying addition 👌🏼

fselmo:
@reedsa it would be ideal if we could avoid repeating the obvious parts. Something more along the lines of:

```
The provided identifier matches multiple elements. If you meant to call `MyElement`, please specify the full signature. 
    - signature: "MyElement()", type: "event"
    - signature: "MyElement()", type: "function"
```

Or something along those lines. Specifically removing the unnecessary repeat of "The provided identifier matches multiple elements. If you meant to call `MyElement()`, please specify the full signature. "

fselmo (approved):
Good stuff. There's a lot of surface area here, maybe one more approval couldn't hurt but lgtm!

tests/core/contracts/test_contract_ambiguous_functions.py
```diff
@@ -83,6 +83,12 @@ def string_contract(w3, string_contract_factory, address_conversion_func):
             repr,
             "<Function identity(uint256,bool)>",
         ),
+        (
```
kclowes, line 115: This looks like it's exactly the same as lines 81-85?

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +48,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    function_name_tester_contract_abi,
+    function_name_tester_contract,
+):
+    # test `abi` function name does not throw when creating the contract factory
+    contract_factory = w3.eth.contract(abi=function_name_tester_contract_abi)
+
+    # re-instantiate the contract
+    contract = contract_factory(function_name_tester_contract.address)
+
+    # Contract `abi`` function should not override `web3` attribute
```
kclowes: nit:
```suggestion
    # Contract `abi`` function should not override `abi` attribute
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +48,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    assert contract.abi == function_name_tester_contract_abi
+
+    with pytest.raises(TypeError):
+        contract.functions.abi[0]
+
+    # assert the `abi` function returns true when called
+    result = contract.functions.abi().call()
```
kclowes: It might also be worth adding a few of these tests on the ContractCaller class. I can't remember how the factory is instantiated over there. The test would look something like: 
```
assert contract.caller.abi() is True
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -43,6 +48,51 @@ def test_abi_as_json_string(w3, math_contract_abi, some_address):
...
+    function_name_tester_contract_abi,
+    async_function_name_tester_contract,
+):
+    # test `abi` function name does not throw when creating the contract factory
+    contract_factory = async_w3.eth.contract(abi=function_name_tester_contract_abi)
+
+    # re-instantiate the contract
+    contract = contract_factory(async_function_name_tester_contract.address)
+
+    # Contract `abi` function should not override `web3` attribute
```
kclowes: nit:
```suggestion
    # Contract `abi` function should not override `abi` attribute
```

tests/core/contracts/test_contract_class_construction.py
```diff
@@ -86,3 +148,53 @@ def test_error_to_call_non_existent_fallback(
...
+        ([{"type": "event", "name": "AnEvent"}], "functions", NoABIFunctionsFound),
+        ([{"type": "function", "name": "aFunction"}], "events", NoABIEventsFound),
+        ([{"type": "function", "name": "aFunction"}], "functions", ABIFunctionNotFound),
+        ([{"type": "event", "name": "AnEvent"}], "events", ABIEventNotFound),
+    ),
+)
+def test_appropriate_exceptions_based_on_namespaces(
+    w3, abi, namespace, expected_exception
+):
+    if abi is None:
```
kclowes: Feel free to take or leave, but we've been moving towards having separate tests instead of having `if` branches within the test. For example, having a test like `test_appropriate_exceptions_based_on_namespaces_with_no_abi` or something and moving those tests to a new test function. I can also see keeping them here since it's the same assertion. I'll leave it up to you :)

web3/_utils/contracts.py
```diff
@@ -47,13 +48,11 @@
 
 from web3._utils.abi import (
     filter_by_argument_name,
+    get_abi_element_signature,
+    get_name_from_abi_element_identifier,
     map_abi_data,
     named_tree,
 )
-from web3._utils.abi_element_identifiers import (
-    FallbackFn,
```
kclowes, line 54: ~~Are these being used anywhere now? We may be able to get rid of them entirely~~ For some reason Github won't let me delete. I see they're being used later on. Don't mind me :)

web3/utils/abi.py
```diff
@@ -158,46 +167,179 @@ def _get_fallback_function_abi(contract_abi: ABI) -> ABIFallback:
         raise ABIFallbackNotFound("No fallback function was found in the contract ABI.")
 
 
+def _get_any_abi_signature_with_name(element_name: str, contract_abi: ABI) -> str:
+    """
+    Find an ABI identifier signature by element name. A signature identifier is
+    returned, "name(arg1Type,arg2Type,...)".
+
+    This function forces a result to be returned even if multiple are found. Depending
```
kclowes: Might be worth putting here that the first match gets returned. Is it the first match based on the ABI order?

kclowes (commented):
Looking good! I didn't see anything major really. Just a few small things I noticed, mostly in the testing department: 
- I didn't comment on all of the instances I saw, but left a comment around splitting out branching `if` statements in tests into separate tests. Feel free to take or leave. 
- I also left a few nits and see a few warnings in the tests that weren't there before that could be cleaned up 🧹 (UserWarnings in the extracting_event_data test, and some DeprecationWarnings in the test_abi test)

tests/core/contracts/test_contract_class_construction.py
reedsa: Yikes, this is really not working. I'm trying to figure out where the collisions are occurring. It seems to be intertwined with the `ContractFunctions` class.

reedsa: @kclowes in my debugging, it appears that the `abi` property gets overridden when there is an `abi()` function in the contract. I assume this is because the special characters are ignored, causing both properties to be set with the function as the value. It should instead set `abi` to the ABI, and `abi()` to the function. It doesnt seem like there is a way to force the value to be the actual ABI. From what I see, setting the function to the `abi()` property also sets `abi`.

It is wildly confusing to trace this edge case through the code and perhaps doesnt make sense anyway. It is also an issue with the other class properties like `w3`/`w3()` and `address`/`address()` cases.

I am thinking it might just be better to disallow these specific function names to be used in a contract altogether. Would that be reasonable, especially since I was trying to do this on top of the other overloaded contract functions?

kclowes: Is this just in the `ContractCaller` class? If so, think it would be fine to disallow functions with the `w3`/`abi`/`address` names via the `ContractCaller`, as long as it gets documented well. I wonder if we could raise an error or something if someone tries to access the function name? I don't necessarily think that we should block a contract abi with those method names in there, but I could be convinced if you feel strongly. If we do go that path (not allowing w3/abi/address function names on ContractCaller), would you add an issue so that we can track it?

reedsa: I'm going to timebox this to see if there is anything I can isolate with these issues. I agree these names may not be widely used so it may be better to resolve in a later issue. I'll follow up with a new issue if it cannot be resolved in the next day or two.

reedsa: @kclowes I am tinkering with the idea of using an underscore prefix for the class attributes. `_w3` / `_abi` / `_address` would be the new names. This is working at least to avoid collisions with `w3` / `abi` / `address`. If the contract contains the underscored name, the variable itself is no longer overwritten but the code does not access the function when called. It instead uses the class variable itself and throws an exception.

A couple questions:
1. Is changing these to private variables considered breaking?
2. Is it acceptable to fail initializing the contract if a contract uses this convention with the `_` prefix with the same name?

kclowes: > Is changing these to private variables considered breaking?

As long as they're private, it shouldn't be breaking. A good smoke test would be to see what tests break. I can't remember if those really are private variables or not, but am happy to dig in further if you want a second pair of eyes.

> Is it acceptable to fail initializing the contract if a contract uses this convention with the _ prefix with the same name?

I think that's okay here. Looking into this more, `abi` and `address` are reserved in Solidity, so I don't think that we'll run into a problem with those two, and I think `w3` is enough of an edge case that we probably don't need to worry too much about it. Hopefully someone would raise an issue if they ran into problems.

reedsa:
@kclowes I noticed that `UserWarning` has occurred for a while now but it's hard to pinpoint exactly what happened. There have been changes in late v6 and into v7 that may be the culprit. However I reverted a couple of those without success. I have created an [issue](https://github.com/ethereum/web3.py/issues/3521) to resolve that warning at some point in the near future.

web3/_utils/contract_sources/contract_data/event_contracts.py
```diff
@@ -1,11 +1,11 @@
 """
 Generated by `compile_contracts.py` script.
-Compiled with Solidity v0.8.28.
```
kclowes: We should maybe recompile here with `v0.8.28` before merge, assuming that's the latest version

kclowes (approved):
LGTM! I left one comment about recompiling with solidity 0.8.28 but then I think it's ready when you're ready to pull the trigger!

jurekovac:
This changes to async_contract.py and contract.py (especially (Async)ContractFunction.__call__) slows down the performance of obtaining the call of contract functions e.g. contract.function_name(*args) considerably. ~ 5x. This is a problem if having a lot of such calls.

PR #3480, opened 2024-09-09
docs: fixed typos in pip install "web3[test]"

LuizDMM:
### What was wrong?

I tried to install the project with "tester", following the current documentation in a project and the dependencies for testing (eth_tester) were not installed.

When checking the current [setup.py](https://github.com/ethereum/web3.py/blob/c4d07a323d92995c5bfefb1763dc0587fa47771f/setup.py#L30) I found that the correct command to do would be `pip install web3[test]` and not `pip installl web3[tester]`.

### How was it fixed?

I changed the documentation in every reference I could find from `pip install web3[tester]` to `pip install web3[test]`.

### Todo:

- [x] Clean up commit history
- [x] Add or update documentation related to these changes
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/main/newsfragments/README.md)

#### Cute Animal Picture

hedgehog-1215140_1920

PR #3495, opened 2024-09-24
Refactor install extras

fselmo:
### What was fixed

- We [recently](https://github.com/ethereum/web3.py/pull/3480/files) added back the `tester` install extra. The `test` extra, however, will always require `eth-tester` and `py-geth` to run. Re-define the `test` extra as `test` + `tester` below.

- Leave dependencies needed for test runs only on the `test` extra,  removing them from the `dev` extra. The `dev` extra is re-defined below and pulls in all of the `test` extra dependencies.

- Use the ``test`` install extra for running the tests with CI, rather than the whole of the ``dev`` extra. This should help us catch any issues with the ``test`` extra as well. Use ``docs`` + ``test`` for the docs run.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

149051_559272406631_18002879_32491667_4977974_n

### DIFF
--- a/newsfragments/3495.internal.rst
+++ b/newsfragments/3495.internal.rst
@@ -1,0 +1,1 @@
+Re-organize the install extras. Re-define the ``test`` extra to always include the ``tester`` extra since it's needed for testing.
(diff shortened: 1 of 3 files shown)


reedsa (approved):
lgtm

PR #3498, opened 2024-09-24
Some optimizations for JSON

### LINKED ISSUE #3496: make json improvements

i am debugging an issue, and I had to log the json for the best results.  what was annoying, is that I couldn't just turn on websockets debugging with their WEBSOCKETS_MAX_LOG_SIZE env variable, because web3.py is sending binary so it's log messages , so I decided to take a look.  

for reference this is what logging in websockets looks like when you send in bytes instead of a str.
```text
websockets.client         DEBUG    > BINARY 7b 22 6a 73 6f 6e 72 70 63 22 3a 20 22 32 2e 30 ... 22 69 64 22 3a 20 30 7d [222 bytes]
websockets.client         DEBUG    < TEXT '{"jsonrpc":"2.0","id":"0","result":"0x10ff14c63...ae4c6e70bce401fd6f8c4"}' [104 bytes]
```

While looking around I found some room for improvement, happy to help if you think any of this is something you'd consider.

1. have provider.socket_send take a string, and let async_ipc encode to bytes, WebSocketProvider can just pass the string to websockets, which will do it's own encoding, and log in text instead of hex bytes which is what it does if you send bytes.
2. json.dumps should have the separartor=(',',':').  in _friendly_json_encode.  no need to pad the network with 0x20's
3. FriendlyJsonSerde should just be turned into @classmethod, since there's no instance state.  save the churn of creating these objects over and over as encode_rpc_request does.
4. since I think the point of FriendlyJsonSerde is to pinpoint failures _if_ you happen to be encoding a dict like/list like object, shouldn't the encoder `cls` be passed to the recursive errors/encode calls?
5. encode_batch_rpc_request .  it's handrolling json (in binary, with spaces :) ) because we do the  transformation of RPCEndpoint/params into rpc_dict and json encoding in one step. if that rpc_dict was separated out, and especially since FriendlyJsonSerde does this recursive error mapping, it seems like we could just give FriendlyJsonSerde.json_encode  a list of rpc_dicts and it would be fine. and the had coded json '[' ']' can be removed.

if any of this stuff is of interest for this project, I'm happy to submit a PR.

toppk:
### What was wrong?

Closes https://github.com/ethereum/web3.py/issues/3496

### How was it fixed?

This makes some modest changes to JSON, to squeeze out some spaces, get rid of some unnecessary object instantiation.  added some test cases because JSONBaseProvider was tested, not AsyncJSONBaseProvider.   The big issue I wanted to solve was to send text to websockets package, so the logging there would be in text, and not binary (see issue).  But i didn't change Legacy for that, which still is bytes.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/web3/_utils/encoding.py
+++ b/web3/_utils/encoding.py
@@ -191,41 +191,57 @@ class FriendlyJsonSerde:
     helpful information in the raised error messages.
     """
 
-    def _json_mapping_errors(self, mapping: Dict[Any, Any]) -> Iterable[str]:
+    @classmethod
+    def _json_mapping_errors(
+        self,
+        mapping: Dict[Any, Any],
+        encoder_cls: Optional[Type[json.JSONEncoder]] = None,
+    ) -> Iterable[str]:
         for key, val in mapping.items():
             try:
-                self._friendly_json_encode(val)
+                self._friendly_json_encode(val, encoder_cls=encoder_cls)
             except TypeError as exc:
                 yield f"{key!r}: because ({exc})"
 
-    def _json_list_errors(self, iterable: Iterable[Any]) -> Iterable[str]:
+    @classmethod
+    def _json_list_errors(
+        self,
+        iterable: Iterable[Any],
+        encoder_cls: Optional[Type[json.JSONEncoder]] = None,
+    ) -> Iterable[str]:
         for index, element in enumerate(iterable):
             try:
-                self._friendly_json_encode(element)
+                self._friendly_json_encode(element, encoder_cls=encoder_cls)
             except TypeError as exc:
                 yield f"{index}: because ({exc})"
 
+    @classmethod
     def _friendly_json_encode(
-        self, obj: Dict[Any, Any], cls: Optional[Type[json.JSONEncoder]] = None
+        cls, obj: Dict[Any, Any], encoder_cls: Optional[Type[json.JSONEncoder]] = None
     ) -> str:
         try:
-            encoded = json.dumps(obj, cls=cls)
+            encoded = json.dumps(obj, cls=encoder_cls, separators=(",", ":"))
             return encoded
         except TypeError as full_exception:
             if hasattr(obj, "items"):
-                item_errors = "; ".join(self._json_mapping_errors(obj))
+                item_errors = "; ".join(
+                    cls._json_mapping_errors(obj, encoder_cls=encoder_cls)
+                )
                 raise Web3TypeError(
                     f"dict had unencodable value at keys: {{{item_errors}}}"
                 )
             elif is_list_like(obj):
-                element_errors = "; ".join(self._json_list_errors(obj))
+                element_errors = "; ".join(
+                    cls._json_list_errors(obj, encoder_cls=encoder_cls)
+                )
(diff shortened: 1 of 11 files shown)


toppk:
i'll fix up the linting if there's interested in the direction this patch is heading into.

toppk:
fwiw, i think there are some things in this PR  that might want to be picked up here, like making the web3.providers.auto not circular importing on web3.providers, even if some of the other bits are not wanted.

web3/_utils/encoding.py
```diff
@@ -191,41 +195,59 @@ class FriendlyJsonSerde:
...
             try:
-                self._friendly_json_encode(val)
+                self._friendly_json_encode(val, encoder_cls=encoder_cls)
             except TypeError as exc:
                 yield f"{key!r}: because ({exc})"
 
-    def _json_list_errors(self, iterable: Iterable[Any]) -> Iterable[str]:
+    @classmethod
+    def _json_list_errors(
+        self,
```
fselmo, line 212:
```suggestion
        cls,
```

web3/_utils/encoding.py
```diff
@@ -191,41 +195,59 @@ class FriendlyJsonSerde:
     helpful information in the raised error messages.
     """
 
-    def _json_mapping_errors(self, mapping: Dict[Any, Any]) -> Iterable[str]:
+    @classmethod
+    def _json_mapping_errors(
+        self,
```
fselmo, line 200:
```suggestion
        cls,
```

web3/providers/persistent/persistent.py
```diff
@@ -164,7 +164,7 @@ async def make_batch_request(
     # -- abstract methods -- #
 
     @abstractmethod
-    async def socket_send(self, request_data: bytes) -> None:
+    async def socket_send(self, request_data: str) -> None:
```
fselmo, line 167: I think I'm good with opening up this type but this is a breaking change otherwise and I'm not keen on keeping it as string only. If you want to handle the string case, encoding the data if it's a string, I think that would be the best approach here. 

We should make sure we update the docstring to reflect the change as well.

web3/providers/rpc/async_rpc.py
```diff
@@ -158,7 +158,7 @@ async def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
             f"Making request HTTP. URI: {self.endpoint_uri}, Method: {method}"
         )
         request_data = self.encode_rpc_request(method, params)
```
fselmo, line 160: This flow ends up in a strange "encoding" limbo where we have an `encode_rpc_request` that generates the `request_data` and yet we call `request_data.encode()` below. We should properly handle encoding within the `encode_rpc_request` method.

fselmo (commented):
Thanks for the PR @toppk. There's definitely some good changes in here! I don't think we should break these existing methods and I don't think strictly passing a string to the socket is really the way to go. I'm not against opening the type up there though and encoding when a string is passed.

Thoughts on the comments here? We can take these changes on eventually, otherwise feel free to update here and I'll give this another pass.

PR #3500, opened 2024-09-26
update EthereumTesterProvider docs

pacrob:
### What was wrong?

`EthereumTesterProvder` docs had arg name `eth_tester`, should be `ethereum_tester`.

### How was it fixed?
 - corrected arg name
 - added missing `api_endpoints` arg + explanation
 - added existence of `AsyncEthereumTesterProvider
 - cleaned up line length for readability
 - added double-quotes around extras install for zsh compatibility

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -568,22 +568,26 @@ explicitly.
 EthereumTesterProvider
 ~~~~~~~~~~~~~~~~~~~~~~
 
-.. warning:: Experimental:  This provider is experimental. There are still significant gaps in
-    functionality. However it is being actively developed and supported.
+.. warning:: Experimental:  This provider is experimental. There are still significant
+    gaps in functionality. However it is being actively developed and supported.
 
-.. py:class:: EthereumTesterProvider(eth_tester=None)
+.. py:class:: EthereumTesterProvider(ethereum_tester=None, api_endpoints=None)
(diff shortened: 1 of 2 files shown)


docs/providers.rst
```diff
@@ -568,22 +568,32 @@ explicitly.
...
         >>> w3 = Web3(EthereumTesterProvider())
 
-.. NOTE:: To install the needed dependencies to use EthereumTesterProvider, you can install the
-    pip extras package that has the correct interoperable versions of the ``eth-tester``
-    and ``py-evm`` dependencies needed: e.g. ``pip install web3[tester]``
+    .. code-block:: python
+
+        >>> from web3 import Web3, EthereumTesterProvider
+        >>> from eth_tester import MockBackend
+        >>> w3 = Web3(EthereumTesterProvider(MockBackend()))
```
kclowes: I'd be curious to get @fselmo's opinion here, but I'm not sure we want to explicitly encourage users to use the mock backend? I think py-evm (the default) would be better?

fselmo: +1 we definitely don't want to encourage it. It's not very well maintained. We should probably begin to deprecate it or write some good docs around that. Let's keep to use pyevm everywhere.

fselmo (approved):
lgtm!

PR #3501, opened 2024-09-30
Update ensip15 to latest spec

fselmo:
### What was wrong?

- Update specs and tests to latest generated from the normalization outputs at: https://github.com/adraffy/ens-normalize.js
    - nf and spec: https://github.com/adraffy/ens-normalize.js/tree/7cd7cef7fbf73726dae93673f0691f6a49ca39b6/derive/output
    - tests is the tests.json here: https://github.com/adraffy/ens-normalize.js/tree/7cd7cef7fbf73726dae93673f0691f6a49ca39b6/validate
- Fix empty name case. This should be valid, just not normalizable.
- If the name is empty after removing 65039 (0xFE0F) then the name is invalid (empty ignorable characters).

### Todo:

- [x] Clean up commit history


reedsa (approved):
lgtm! 🛥️

pacrob (approved):
lgtm! Just curious, where do you find the updates to the standard? I poked through the reference JS library but couldn't find anything relevant.

fselmo:
> lgtm! Just curious, where do you find the updates to the standard? I poked through the reference JS library but couldn't find anything relevant.

I was going to point straight to it but the `outputs` folder in particular. I figured that path may change in the future so figured I'd link to the repo instead but I can do a reference permalink instead. That would help anyone looking for it directly.

PR #3508, opened 2024-10-10
Add support for non-mainnet request cache validation thresholds

### LINKED ISSUE #3506: get_transaction failed when cache_allowed_requests is set to True.
### What happened?

When set `cache_allowed_requests`  in `HTTPProvider`, and assign get_transaction  result to a variable, there will be an exception.

```python
from web3 import Web3
# test on bsc chain
w3 = Web3(
    Web3.HTTPProvider(
        "http://127.0.0.1:8545",  cache_allowed_requests=True
    )
)

txhash = 0xd7caec490f96319fee3b06ac2ad0f7bb03ecf8cc464304b256a51dcd34ea50cc
txdata = w3.eth.get_transaction(txhash)
```

### Full error output

```shell
There was an exception while caching the request.
  File "lib/python3.12/site-packages/web3/_utils/caching/request_caching_validation.py", line 39, in is_beyond_validation_threshold
    return blocknum <= to_int(hexstr=response["result"]["number"])
                                     ~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

_No response_

### web3 Version

7.3.0

### Python Version

3.12.3

### Operating System

osx

### Output from `pip freeze`

_No response_

fselmo:
### What was wrong?

closes #3506

### How was it fixed?

- If non-mainnet ethereum, allow the use of an integer value for the request cache validation threshold. Pre-configure "safe" default values for some common non-mainnet chain ids based on their varied finality mechanisms.
- This integer value represents the number of seconds from time.now() that the request cache deems as a safe enough time window to allow the request to be cached.
- Update the tests to reflect these changes.
- Added comprehensive testing around the caching logic, affected endpoints, and all relevant providers.

Unrelated:

- Add a note to the ``request_mocker`` to make sure that mocked results
  are of the correct expected types based on JSON-RPC spec (e.g. hex
  strings instead of ints for numbers, etc.)
- Removed request caching decorator from base providers since they don't implement `make_request`. This makes typing tighter around the code base as well.
- IPCProvider did not have request caching turned on. This PR turns that on.

### Todo:

- [x] Add or update documentation related to these changes. This documentation should detail the underlying mechanism of how (and why) to check certain endpoints against a validation threshold so that users know:

  1. When to reach for caching in the first place
  2. If they reach for caching, clarify internal calls can be made to get the necessary validation data
  3. Know how they can tweak the validation threshold to lower the number of internal calls made (or set it to ``None`` to cache any and all requests)
- [x] Clean up commit history

#### Cute Animal Picture

```
     __      __
    /  \____/  \
   (    o  o    )
    \    ∇     /
     \  \___/  /
      \_______/
        |   |
```

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -119,12 +119,21 @@ Provider Configurations
 Request Caching
 ```````````````
 
+.. important::
+    Familiarize yourself with the validation logic for request caching before
+    enabling it. Since this feature often requires making additional requests under the
+    hood to try to guarantee the validity of the data, it may create unnecessary
+    overhead for your use case. Validation can be turned off by setting the
+    ``request_cache_validation_threshold`` option to ``None``, caching all allowed
+    requests, or configured for adjusting performance to your needs.
+
+
 Request caching can be configured at the provider level via the following configuration
 options on the provider instance:
 
 - ``cache_allowed_requests: bool = False``
-- ``cacheable_requests: Set[RPCEndpoint] = CACHEABLE_REQUESTS``
-- ``request_cache_validation_threshold: RequestCacheValidationThreshold = RequestCacheValidationThreshold.FINALIZED``
+- ``cacheable_requests: Optional[Set[RPCEndpoint]]``
+- ``request_cache_validation_threshold: Optional[Union[RequestCacheValidationThreshold, int]]``
 
 For requests that don't rely on block data (e.g., ``eth_chainId``), enabling request
 caching by setting the ``cache_allowed_requests`` option to ``True`` will cache all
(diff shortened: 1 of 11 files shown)


web3/_utils/caching/caching_utils.py
```diff
@@ -142,16 +173,60 @@ def is_cacheable_request(
...
+                    current_threshold,
+                    RequestCacheValidationThreshold,
+                )
+                and chain_id != 1
+            ):
+                provider.logger.debug(
+                    "Request cache validation threshold is set to "
+                    f"{current_threshold.value} "
+                    f"for chain with chain_id `{chain_id}` but this value only works "
+                    "on chain_id `1`. Setting to default value for chain_id "
```
kclowes: This won't work with other POS chains? testnets?

web3/_utils/caching/request_caching_validation.py
```diff
@@ -31,99 +32,244 @@ def _error_log(
...
+            provider.logger.error(
+                "Invalid request_cache_validation_threshold value. This should not "
+                f"have happened. Request not cached.\n    threshold: {threshold}"
+            )
+            return False
     except Exception as e:
         _error_log(provider, e)
         return False
+    finally:
+        provider.cache_allowed_requests = True
```
kclowes: lazy question: instead of always setting this back to true, I wonder if it makes sense to store the value of cache_allowed_requests coming in? for example, something like:
on line 46:
`cache_allowed_requests = provider.cache_allowed_requests`
and then here:
`provider.cache_allowed_requests = cache_allowed_requests`

web3/_utils/caching/request_caching_validation.py
```diff
@@ -31,99 +32,244 @@ def _error_log(
...
+    _result: Dict[str, Any],
 ) -> bool:
     block_id = params[0]
     if block_id == "earliest":
         # `earliest` should always be cacheable
         return True
-    blocknum = to_int(hexstr=block_id)
-    return is_beyond_validation_threshold(provider, blocknum)
 
+    blocknum = int(block_id, 16)
```
kclowes, line 95: another lazy question: Is there a difference between what `to_int` is doing and `int` here?

web3/_utils/module_testing/utils.py
```diff
@@ -35,6 +35,14 @@ class RequestMocker:
     Context manager to mock requests made by a web3 instance. This is meant to be used
     via a ``request_mocker`` fixture defined within the appropriate context.
 
+    ************************************************************************************
+    Important: When mocking results, it's important to keep in mind the types that
+        clients return. For example, what we commonly translate to integers are returned
+        as hex strings in the RPC response and should be mocked as such for more
+        accurate testing.
+    ************************************************************************************
```
kclowes, line 43: 👍

tests/core/caching-utils/test_request_caching.py
```diff
@@ -260,30 +265,31 @@ def test_blocknum_validation_against_validation_threshold_when_caching(
...
-                else {"number": blocknum}
+                else {"number": blocknum, "timestamp": "0x0"}
             ),
         },
     ):
         assert len(w3.provider._request_cache.items()) == 0
         w3.manager.request_blocking(RPCEndpoint(endpoint), [block_id, False])
         cached_items = len(w3.provider._request_cache.items())
-        assert cached_items == 1 if should_cache else cached_items == 0
+        assert cached_items > 0 if should_cache else cached_items == 0
```
kclowes: curious about this change?

docs/internals.rst
```diff
@@ -119,48 +119,100 @@ Provider Configurations
...
-default, the ``finalized`` block number is used as the validation threshold, meaning
-that a request's response will be cached if the block number it relies on is less than
-or equal to the ``finalized`` block number. If the block number exceeds the
-``finalized`` block number, the response won't be cached.
+chain reorganization or while finality has not been reached, for example. The
+``request_cache_validation_threshold`` option allows configuring a safe threshold for
+caching responses that depend on block data. By default, this option is configured
+to internal values deemed "safe" for the chain id you are connected to. If you are
+connected to mainnet Ethereum, this value is set to the ``finalized`` block number.
+If you are connected to another chain, this value is set to a time internal in seconds,
```
pacrob: did you mean "is set internally to a time in seconds" or something like that?

fselmo: Womp... "time interval". Will fix

tests/core/caching-utils/test_request_caching.py
```diff
@@ -211,11 +215,11 @@ def test_all_providers_do_not_cache_by_default_and_can_set_caching_properties(pr
         ("0x5", False),
     ),
 )
-def test_blocknum_validation_against_validation_threshold_when_caching(
+def test_blocknum_validation_against_validation_threshold_when_caching_mainnet(
     threshold, endpoint, blocknum, should_cache, request_mocker
 ):
     w3 = Web3(
-        HTTPProvider(
+        BaseProvider(
```
kclowes: more lazy questions: Is it just the HTTPProvider that caches? or should we maybe have a test that parametrizes the other providers too?

kclowes (approved):
Mostly left lazy questions and small things, but looks good otherwise!

web3/_utils/caching/request_caching_validation.py
fselmo: I think we can for safety reasons but we should technically only get here if we are caching in the first place. Better to be safe though?

web3/_utils/caching/request_caching_validation.py
fselmo: I think it's more of an overhead than necessary? If we know that we're getting back a hex string from the client I though it'd be best to just make use of the `int` directly. Thoughts?

tests/core/caching-utils/test_request_caching.py
fselmo: oops, that was me playing around with some different ideas I think. Thought I had changed them all back. This was a result of not turning off caching when making requests to get validation blocks, etc. But it's not a good approach since it can enter recursion. I'll change this back, good catch 👍🏼

tests/core/caching-utils/test_request_caching.py
fselmo: Nope, all providers should cache. I think that would be a good idea. Should we track it separately though?

web3/_utils/caching/caching_utils.py
fselmo: Good catch. It should work with others as well. I think this messaging should be improved. Will fix 👍🏼

fselmo: I think we can just remove that validation entirely. I was trying to issue a warning related to that value not being available, for friendliness, but that responsibility may have to be on the user since we can't realistically keep track of all chains that have it available.

tests/core/caching-utils/test_request_caching.py
fselmo: We have a [test for caching properties](https://github.com/ethereum/web3.py/blob/46913500aba08113aad07c50b858aca18cc70f46/tests/core/caching-utils/test_request_caching.py#L167-L199) but we don't run all providers through the test suite currently.

kclowes: > Nope, all providers should cache. I think that would be a good idea. Should we track it separately though?

Sure, that's fine by me if you'd rather track separately.

web3/_utils/caching/request_caching_validation.py
kclowes: Cool, that makes sense to me!

tests/core/caching-utils/test_request_caching.py
fselmo: I added it in here instead, as a different commit.

web3/_utils/module_testing/utils.py
```diff
@@ -99,7 +107,7 @@ def __init__(
 
     def __enter__(self) -> "Self":
         # mypy error: Cannot assign to a method
-        self.w3.provider.make_request = self._mock_request_handler  # type: ignore[method-assign]  # noqa: E501
+        self.w3.provider.make_request = self._mock_request_handler  # type: ignore
```
pacrob: Why remove the type of ignore here?

pacrob (approved):
just one q, otherwise lgtm!

web3/_utils/module_testing/utils.py
fselmo: Good point I think `__enter__` can keep `[method-assignment]` but I think `__exit__` had the issue. I think we can be more specific there and use `__assignment__`. This happened when I removed the request_caching decorators from the base provider classes. I think the typing ends up slightly off with `Arg()` wrappers around the args even though the type is essentially the same.

@pacrob are you good with narrowing the `__exit__` ignores to `[assignment]` and keeping `[method-assign]` here?

pacrob: yup, sgtm

PR #3514, opened 2024-10-17
Adds typing hints documentation

### LINKED ISSUE #1628: Provide `TypedDict` types for return values of methods usable by 3rd parties.
### What was wrong?

While making an application that uses the web3.py library I found myself really wishing there were strong return types for the various methods like `w3.eth.get_block(...)`.  Basically, I want to have something like a `TypedDict` which strongly defines the return types of the various web3 methods.

### How can it be fixed?

The `web3` library should have `TypedDict` types for all of the various return types that return a mapping.  These types should be importable by 3rd party libraries (probably via `web3.typing`) and should also be documented (probably in a new section of the documentation focused on how to do typing for 3rd party applications).

flamiedev:
### What was wrong?

Related to Issue #1628
Closes #1628

### How was it fixed?

Created a new web3.types.rst file in the docs to list all the possible type hints available in the types file.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/web3.types.rst
+++ b/docs/web3.types.rst
@@ -1,0 +1,81 @@
+TYPES
+=====
+
+.. py:module:: web3.types
+
+The ``web3.types`` exposes the available modules that enables you to define the data types for the various web3 methods
+
+The following types are available on the ``web3.types``:
+
+.. py:class:: TFunc
+.. py:class:: TParams 
+.. py:class:: TReturn
+.. py:class:: TValue
+.. py:class:: BlockParams
+.. py:class:: BlockIdentifier
+.. py:class:: LatestBlockParam
+.. py:class:: ABIElementIdentifier
+.. py:class:: _Hash32
+.. py:class:: EnodeURI
+.. py:class:: ENS
+.. py:class:: Nonce
+.. py:class:: RPCEndpoint
+.. py:class:: Timestamp
+.. py:class:: Wei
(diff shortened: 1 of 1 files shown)


pacrob:
Hey @flamiedev , thanks for getting this started! I'd rather we went in an auto-generated direction, that way we don't have to remember to update across both places when changes happen in the future.

Just to get you started:

1) Add `types` to `docs/toc.rst`, probably in the `API` section, under `constants`

```
...
.. toctree::
    :maxdepth: 1
    :caption: API

    web3.main
    web3.eth
    web3.beacon
    web3.net
    web3.geth
    web3.tracing
    web3.utils
    gas_price
    ens
    constants
    types
...
```

2) Create `docs/types.rst` with the following in it:

```
Types
=====

.. automodule:: web3.types
    :members:
```

Then when you build the docs, you'll see all the classes from `types.py` in there. The initial request specifically mentions mappings, so I'd like to see the members of each of these `TypedDicts`, along with their types. I'm not sure how to pull those out directly,  or you may just need to add docstrings to the classes themselves. Docstrings would be helpful to give a brief explanation of what each `TypedDict` is used for, especially if it's not clear from the name.

pacrob:
Hey, @flamiedev, are you planning to continue work here? Happy to leave it open if you are, but it can be closed if you've moved on.

PR #3525, opened 2024-11-07
add check bytes length in get_create2_address

artemki2077:
### What was wrong?

I did not correctly determine the address through the create 2 opcode, I thought I had a problem with the smart contract, but it turns out that the length of the salt should be 32 bytes, this was not checked in web3 py

### How was it fixed?

added a check

### Todo:

- [x] Clean up commit history


fselmo:
This is a nice addition. Thanks @artemki2077. I added some tests and a newsfragment and made some slight tweaks.

PR #3529, opened 2024-11-13
pin websockets <14 due to breaking changes

pacrob:
### What was wrong?

`websockets v14` has breaking changes - https://websockets.readthedocs.io/en/stable/project/changelog.html#backwards-incompatible-changes

### How was it fixed?

Pin to `<14` for now. They also dropped `py38` with the same release, so change to `>=14` once we do the same.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/3529.internal.rst
+++ b/newsfragments/3529.internal.rst
@@ -1,0 +1,1 @@
+Pin ``websockets<14`` due to breaking changes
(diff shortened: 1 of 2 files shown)


fselmo (approved):
lgtm!

PR #3542, opened 2024-11-27
Update outdated Python 2 API reference to reflect Python 3 usage docs

mdqst:
### What was wrong?

I noticed that in the documentation, there was a reference to `execfile()d`, which is an old Python 2 API that is no longer used in Python 3. In Python 3, the correct approach is to use `exec()` instead. To ensure the documentation stays up to date and accurate, I’ve updated the comment to reflect the correct usage of `exec()` in Python 3. This change is important for clarity, as it helps avoid confusion for users working with modern Python versions.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Cute Animal


pacrob:
Thanks @mdqst

PR #3546, opened 2024-12-01
remove verbose RPC error logging

simone1999:
Currently RPC error responses are logged and thrown. The application should decide whether the thrown error should be logged or not. All logged information are part of the thrown exception and often RPC errors are handled gracefully by the application logic. The current situation is that even handled errors are spamming the console and the only way to get rid of them is to completely disable the web3.manager.RequestManager logger.


fselmo (approved):
This seems fine. We do have the response already debug logged and the clarifying user message is available on any caught `Web3RPCError` exceptions.

PR #3548, opened 2024-12-02
Contract Functions utils

reedsa:
### What was wrong?

Related to Issue #3036

### How was it fixed?

Add contract functions methods for `find_by_name`, `find_by_signature` and `find_by_selector`

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/contracts/test_contract_functions.py
+++ b/tests/core/contracts/test_contract_functions.py
@@ -1,0 +1,157 @@
+import pytest
+
+from web3.contract.async_contract import (
+    AsyncContract,
+)
+from web3.contract.contract import (
+    Contract,
+)
+from web3.exceptions import (
+    Web3ValueError,
+)
+
+
+def test_functions_find_by_name(event_contract: "Contract") -> None:
+    functions = event_contract.functions.find_by_name("logTwoEvents")
+
+    expected_functions = [repr(fn) for fn in functions]
+    assert expected_functions == ["<Function logTwoEvents(uint256)>"]
+
+
+def test_functions_find_by_name_with_ambiguous_name(math_contract: "Contract") -> None:
+    functions = math_contract.functions.find_by_name("incrementCounter")
+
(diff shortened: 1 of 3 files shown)


reedsa:
It is a little awkward to add functions to the API where collisions may occur. These new utilities will exist in the `functions` namespace on a contract, which already contains functions from the contract itself. This is a similar issue we encountered when existing properties could have name collisions with any functions with the same name in the contract.

We could allow these as "reserved" APIs that would not invoke any function on the contract, since the names seem intuitive enough but I'm open to any other thoughts here. I could see us adding `web3` to the name or even finding another namespace to further distinguish this from actual contract functions/events, though I think it's useful to have the ability to filter functions in a similar fashion to what we implemented on the Contract class itself.

kclowes:
> We could allow these as "reserved" APIs that would not invoke any function on the contract, since the names seem intuitive enough but I'm open to any other thoughts here.

Good point. I think this is a good way to do it as long as there is a workaround, just in case someone has a contract method that is called `find_by_name` as unlikely as it may be. But I also think that a different namespace is fine too, whether with the addition of `web3_` on the front, or some other namespace.

reedsa:
@kclowes I'm inclined to leave these and the event APIs out. We talked about the redundancy with the Contract APIs. I would rather not clutter the functions/events with separate capabilities.

Curious if anyone feels strongly about these being available, also would like @fselmo to weigh in here.

kclowes:
Yeah, I honestly think that makes the most sense, but also curious to hear from @fselmo if there is something I'm not thinking of

fselmo:
I'm on board. I think the goal of the original issue was to expose _any_ public utils for getting functions, events, etc. If we did it well enough through the contract instance, I do think that makes sense. We can eventually assess the demand but I think your recent work scratches the itch and is enough to close the original issue @reedsa. Let's work on getting those working well and work on parity across `contract.events` and `contract.functions` and we should be able to close the original issue.

Issue #3397, opened 2024-05-15
`PersistentConnectionProvider` UX Improvements Discussion


fselmo:
The newer `PersistentConnectionProvider` implementations, mostly the `WebSocketProvider`, seem to be getting a lot of use and a lot of good feedback is coming back from users, whether through issues or Discord. It's time to start thinking about some good UX abstractions that can help facilitate subscription management, re-connection logic, and anything else that might make sense to _bake into_ the library.

Some ideas that have been brewing:

- [ ] Configurable re-connection logic. This should be able to be turned off and also configured for different exceptions, etc. Re-subscribe to existing subscriptions with new subscription ids when this re-connection is triggered.
- [x] (#3554) Handling subscriptions via specific callbacks. Could this be an argument to `w3.eth.subscribe` method perhaps? Likely managed within a `SubscriptionManager` class strapped to persistent connection providers.

---

If you have UX improvements, pain points, or any comments on anything discussed in this issue, please feel free to join the discussion below!

DefiDebauchery:
You hit on the two that I'm most interested in! As a user, I don't want to have to manage `process_subscriptions()`.

Not specific to the Persistent Provider itself, but as far as overall UX - and to tack onto the mention of `.subscribe()`, do you think it would be worthwhile to add optional kwargs to the method instead of (or in addition to) relying on `LogsSubscriptionArg`? 

```py
    async def subscribe(
        self,
        subscription_type: SubscriptionType,  # Meh
        subscription_arg: Optional[
            Union[
                LogsSubscriptionArg,  # logs, optional filter params
                bool,  # newPendingTransactions, full_transactions
            ]
        ] = None,
        callback: Optional[Callable] = None,  # Sneaking that in here
        *,
        address: Optional[
            Union[
                Address,
                ChecksumAddress,
                ENS,
                Sequence[Union[Address, ChecksumAddress, ENS]]
            ],
        ]= None,
        topics: Optional[Sequence[Union[HexStr, Sequence[HexStr]]]] = None,
    ) -> HexStr:
```

This will certainly require some additional sanity-checking, and may even require separate methods for subscriptions between newHeads and logs ([et al](https://github.com/ethereum/web3.py/blob/main/web3/types.py#L545); which wouldn't be terrible, as I don't necessarily love `subscription_type`), but it may be more expressive to have a defined set of arguments, while still allowing people to use pre-packaged log structures.

clover-es:
[...]

PR #3554, opened 2024-12-09
Subscription management implementation

fselmo:
### What was wrong?

Related to Issue #3397

Working with many subscriptions is not ideal at the moment. In web3.py v7, we added significant support for `eth_subscribe` via the new `PersistentConnectionProvider` implementation. Though we've added support with parity to how `websockets` works, this still only left us with a raw API to handle messages as they come through the socket. We have JSON-RPC specifications of how subscriptions should be formed and the client returns a unique `id` per subscription. Since `PersistentConenctionProvider` classes are fully asynchronous, as they should be, we should implement an API that can leverage all of this in order to handle subscriptions asynchronously as they come in.

### How was it fixed?

- Implement a subscription manager class to manage all subscriptions under-the-hood. This manager can be accessed directly to deal with subscriptions as objects, rather than free-floating ids that are tracked. Each `EthSubscription` class can, and should be encouraged to, have a `handler` function passed to it so that it can properly handle the subscription as it comes in, rather than having the burden of listening and parsing based on message `id` be on the user.
- `w3.eth.subscribe` can now also accept a `handler` kwarg since all subscriptions are managed by the manager under-the-hood. It can also accept an `event` for "logs" subscriptions so that it may be easily parsed by the handler.
- Subscriptions can have descriptive labels for clearer log messages, if desired, instead of relying on an abstract subscription `id` to parse logs.
- The manager counts how many total handler calls have been issued and each subscription counts how many times its handler has been called.

---

In a separate PR we should add a significant docs overhaul for clearer communication on `PersistentConnectionProvider` usage as well as subscription management usage

### Todo:

- [x] Clean up commit history
- [x] Early documentation before a significant docs overhaul to highlight this usage more appropriately
- [x] Add tests

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -445,12 +445,12 @@ the user is done with it.
 .. code-block:: python
 
     >>> async def new_heads_handler(
-    ...     context: NewHeadsSubscriptionContext,
+    ...     handler_context: NewHeadsSubscriptionContext,
     ... ) -> None:
-    ...     result = context.result
+    ...     result = handler_context.result
     ...     print(f"New block header: {result}\n")
     ...     if result["number"] > 1234567:
-    ...         await context.subscription.unsubscribe()
+    ...         await handler_context.subscription.unsubscribe()
 
     >>> async def ws_subscription_example():
     ...     async with AsyncWeb3(WebSocketProvider(f"ws://127.0.0.1:8546")) as w3:
@@ -470,7 +470,17 @@ The manager can also subscribe to many subscriptions at one time. The
 friendly API for managing subscriptions. Since each connection and provider instance
 has its own message listener task and subscription manager instance, you can subscribe
 to many subscriptions at once and handle the many responses that come in over the socket
-connections.
+connections via handlers. The handlers contain:
+
+- ``async_w3``: The ``AsyncWeb3`` instance that the subscription was made on.
+- ``subscription``: The subscription instance that the handler is attached to.
+- ``result``: The response that came in over the socket connection for the subscription.
+
+Subscriptions also accept a ``handler_context`` argument that can be used to pass
+additional information to the handler when subscribing to a subscription. This can be
+used to pass in an event object, for example, that can be used to parse a log event
+when it comes in.
+
 
 .. code-block:: python
 
@@ -490,23 +500,28 @@ connections.
     ... )
 
     >>> async def new_heads_handler(
-    ...     context: NewHeadsSubscriptionContext,
+    ...     handler_context: NewHeadsSubscriptionContext,
     ... ) -> None:
-    ...     print(f"New block header: {context.result}\n")
-    ...     if context.result["number"] > 1234567:
-    ...         await context.subscription.unsubscribe()
+    ...     header = handler_context.result
+    ...     print(f"New block header: {header}\n")
+    ...     if header["number"] > 1234567:
+    ...         await handler_context.subscription.unsubscribe()
 
     >>> async def pending_txs_handler(
-    ...     context: PendingTxSubscriptionContext,
+    ...     handler_context: PendingTxSubscriptionContext,
     ... ) -> None:
     ...     ...
 
     >>> async def log_handler(
-    ...     context: LogsSubscriptionContext,
+    ...     handler_context: LogsSubscriptionContext,
     ... ) -> None:
-    ...     ...
+    ...     log_receipt = handler_context.result
+    ...     # event is now available in the handler context, because we pass it to in the
+    ...     # ``handler_context`` when subscribing to the log
+    ...     event_data = handler_context.event.process_log(log_receipt)
+    ...     print(f"Log event data: {event_data}\n")
 
-    >>> async def sx_manager():
+    >>> async def sub_manager():
     ...     local_w3 = await AsyncWeb3(AsyncIPCProvider(LOCAL_IPC, label="mainnet-ipc"))
     ...     # subscribe to many subscriptions via the subscription manager with handlers
     ...     await local_w3.subscription_manager.subscribe(
@@ -522,6 +537,8 @@ connections.
     ...                 address=local_w3.to_checksum_address("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"),
     ...                 topics=[HexStr("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef")],
     ...                 handler=log_handler,
+    ...                 # optional ``handler_context`` args to help parse a response
+    ...                 handler_context={"event": my_event},
     ...             ),
     ...         ]
     ...     )
@@ -538,7 +555,7 @@ connections.
     ...         local_w3.subscription_manager.handle_subscriptions(),
     ...     )
 
-    >>> asyncio.run(sx_manager())
+    >>> asyncio.run(sub_manager())
 
 
 The ``process_subscriptions()`` method on the
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -445,13 +445,12 @@ the user is done with it.
 .. code-block:: python
 
     >>> async def new_heads_handler(
-    ...     async_w3: AsyncWeb3,
-    ...     sx: EthSubscription,
-    ...     response: BlockData,
+    ...     context: NewHeadsSubscriptionContext,
     ... ) -> None:
-    ...     print(f"New block header: {response}\n")
-    ...     if response["number"] > 1234567:
-    ...         await sx.unsubscribe()
+    ...     result = context.result
+    ...     print(f"New block header: {result}\n")
+    ...     if result["number"] > 1234567:
+    ...         await context.subscription.unsubscribe()
 
     >>> async def ws_subscription_example():
     ...     async with AsyncWeb3(WebSocketProvider(f"ws://127.0.0.1:8546")) as w3:
@@ -483,30 +482,27 @@ connections.
     >>> from web3.utils.subscriptions import (
     ...     EthSubscription,
     ...     NewHeadsSubscription,
+    ...     NewHeadsSubscriptionContext,
     ...     PendingTxSubscription,
+    ...     PendingTxSubscriptionContext,
     ...     LogsSubscription,
+    ...     LogsSubscriptionContext,
     ... )
 
     >>> async def new_heads_handler(
-    ...     async_w3: AsyncWeb3,
-    ...     sx: EthSubscription,
-    ...     response: BlockData,
+    ...     context: NewHeadsSubscriptionContext,
     ... ) -> None:
-    ...     print(f"New block header: {response}\n")
-    ...     if response["number"] > 1234567:
-    ...         await sx.unsubscribe()
+    ...     print(f"New block header: {context.result}\n")
+    ...     if context.result["number"] > 1234567:
+    ...         await context.subscription.unsubscribe()
 
     >>> async def pending_txs_handler(
-    ...     async_w3: AsyncWeb3,
-    ...     sx: EthSubscription,
-    ...     response: TxData,
+    ...     context: PendingTxSubscriptionContext,
     ... ) -> None:
     ...     ...
 
     >>> async def log_handler(
-    ...     async_w3: AsyncWeb3,
-    ...     sx: EthSubscription,
-    ...     response: LogData,
+    ...     context: LogsSubscriptionContext,
     ... ) -> None:
     ...     ...
 
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -430,10 +430,126 @@ request. The initial ``eth_subscribe`` request expects only one response, the
 subscription *id* value, but it also expects to receive many ``eth_subscription``
 messages if and when the request is successful. For this reason, the original request
 is considered a one-to-one request so that a subscription *id* can be returned to the
-user on the same line, but the ``process_subscriptions()`` method on the
+user on the same line. The many responses this call will produce can be handled in one
+of a few ways.
+
+The recommended way to handle one-to-many responses is to use the subscription manager
+API. The subscription manager API is a public API on the ``AsyncWeb3`` class, when
+connected to a ``PersistentConnectionProvider`` instance, that allows the user to
+subscribe to a subscription and handle the many responses asynchronously. The
+``subscription_manager`` instance is responsible for handling the many responses that
+come in over the socket connection, as long as handlers are passed to each subscription
+call. The subscription manager can also be used to unsubscribe from a subscription when
+the user is done with it.
+
+.. code-block:: python
+
+    >>> async def new_heads_handler(
+    ...     async_w3: AsyncWeb3,
+    ...     sx: EthSubscription,
+    ...     response: BlockData,
+    ... ) -> None:
+    ...     print(f"New block header: {response}\n")
+    ...     if response["number"] > 1234567:
+    ...         await sx.unsubscribe()
+
+    >>> async def ws_subscription_example():
+    ...     async with AsyncWeb3(WebSocketProvider(f"ws://127.0.0.1:8546")) as w3:
+    ...         # Subscribe to new block headers and receive the subscription_id.
+    ...         # A one-to-one call with a trigger for many responses
+    ...         subscription_id = await w3.eth.subscribe("newHeads", handler=new_heads_handler)
+    ...
+    ...         # Handle the subscription messages asynchronously using the subscription
+    ...         # manager. This will continue until no more subscriptions are present in
+    ...         # the subscription manager, or indefinitely if the `run_forever` flag
+    ...         # is set to `True`.
+    ...         await w3.subscription_manager.handle_subscriptions(run_forever=False)
+    >>> asyncio.run(ws_subscription_example())
+
+The manager can also subscribe to many subscriptions at one time. The
+``EthSubscription`` classes, available via ``web3.utils.subscriptions``, provide a
+friendly API for managing subscriptions. Since each connection and provider instance
+has its own message listener task and subscription manager instance, you can subscribe
+to many subscriptions at once and handle the many responses that come in over the socket
+connections.
+
+.. code-block:: python
+
+    >>> from web3 import (
+    ...     AsyncWeb3,
+    ...     WebSocketProvider,
+    ...     AsyncIPCProvider,
+    ... )
+    >>> from web3.utils.subscriptions import (
+    ...     EthSubscription,
+    ...     NewHeadsSubscription,
+    ...     PendingTxSubscription,
+    ...     LogsSubscription,
+    ... )
+
+    >>> async def new_heads_handler(
+    ...     async_w3: AsyncWeb3,
+    ...     sx: EthSubscription,
+    ...     response: BlockData,
+    ... ) -> None:
+    ...     print(f"New block header: {response}\n")
+    ...     if response["number"] > 1234567:
+    ...         await sx.unsubscribe()
+
+    >>> async def pending_txs_handler(
+    ...     async_w3: AsyncWeb3,
+    ...     sx: EthSubscription,
+    ...     response: TxData,
+    ... ) -> None:
+    ...     ...
+
+    >>> async def log_handler(
+    ...     async_w3: AsyncWeb3,
+    ...     sx: EthSubscription,
+    ...     response: LogData,
+    ... ) -> None:
+    ...     ...
+
+    >>> async def sx_manager():
+    ...     local_w3 = await AsyncWeb3(AsyncIPCProvider(LOCAL_IPC, label="mainnet-ipc"))
+    ...     # subscribe to many subscriptions via the subscription manager with handlers
+    ...     await local_w3.subscription_manager.subscribe(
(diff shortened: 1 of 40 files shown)


fselmo:
This is ready for a review :)

fselmo:
I moved to a context-based args management for subscription handlers in order to facilitate custom args to be used within handlers. This is back on a freeze for reviews.

web3/_utils/module_testing/module_testing_utils.py
```diff
@@ -106,7 +106,6 @@ def assert_contains_log(
     emitter_contract_address: ChecksumAddress,
     txn_hash_with_log: HexStr,
 ) -> None:
-    assert len(result) == 1
```
reedsa, line 109: Removing this may cause an index out of range when retrieving the first result. Not sure of the intent here but it's fine with me if we can assume there's a result and potentially have more than one.

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,143 @@
...
+                await self.subscribe(sx)
+            return sx_ids
+        raise Web3TypeError("Expected a Subscription or a sequence of Subscriptions.")
+
+    async def unsubscribe(self, subscription: EthSubscription[Any]) -> bool:
+        """
+        Used to unsubscribe from a subscription that is being managed by the
+        subscription manager.
+
+        :param subscription: The subscription to unsubscribe from.
```
reedsa, line 135: Might include the type in the docs as well with `:type subscription: EthSubscription[Any]`

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,143 @@
...
+        while run_forever or len(self.subscriptions) > 0:
+            await self._w3.manager._get_next_message()
+
+    async def handle_subscriptions(self, run_forever: bool = False) -> None:
+        """
+        Used to process all subscriptions. It will run until all subscriptions are
+        unsubscribed from or, if `run_forever` is set to `True`, it will run
+        indefinitely.
+
+        :param run_forever: If `True`, the method will run indefinitely.
```
reedsa: Type for this could be included `:type run_forever: bool`

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,143 @@
...
+        Used to unsubscribe from all subscriptions that are being managed by the
+        subscription manager.
+
+        :return: None
+        """
+        unsubscribed = [await self.unsubscribe(sx) for sx in self.subscriptions.copy()]
+        if all(unsubscribed):
+            self.logger.info("Successfully unsubscribed from all subscriptions.")
+        else:
+            self.logger.warning(
```
reedsa: Could be useful to include which subscriptions are still active. Is it possible to `try` for each `self.unsubscribe` call and bubble up any exceptions that occur?

reedsa (commented):
Left a few ideas for consideration but didn't see anything that sticks out as a showstopper. Nice work!

wolovim (approved):
looking primarily at the public API -- this feels like a friendlier way to manage even one subscription. context-aware handlers and subscription type classes are great devex improvements. will be important to have a good reference guide in the docs to surface whats hidden behind context (follow-on 🤝). nice initiative!

reedsa (approved):

docs/internals.rst
```diff
@@ -430,10 +430,122 @@ request. The initial ``eth_subscribe`` request expects only one response, the
 subscription *id* value, but it also expects to receive many ``eth_subscription``
 messages if and when the request is successful. For this reason, the original request
 is considered a one-to-one request so that a subscription *id* can be returned to the
-user on the same line, but the ``process_subscriptions()`` method on the
+user on the same line. The many responses this call will produce can be handled in one
+of a few ways.
+
+The recommended way to handle one-to-many responses is to use the subscription manager
+API. The subscription manager API is a public API on the ``AsyncWeb3`` class, when
```
kclowes, line 437: nit:
```suggestion
API. The subscription manager API is a public API on the ``AsyncWeb3`` class. When
```

docs/internals.rst
```diff
@@ -430,10 +430,122 @@ request. The initial ``eth_subscribe`` request expects only one response, the
...
+subscribe to a subscription and handle the many responses asynchronously. The
+``subscription_manager`` instance is responsible for handling the many responses that
+come in over the socket connection, as long as handlers are passed to each subscription
+call. The subscription manager can also be used to unsubscribe from a subscription when
+the user is done with it.
+
+.. code-block:: python
+
+    >>> async def new_heads_handler(
+    ...     context: NewHeadsSubscriptionContext,
```
kclowes: is this `handler_context` now?

tests/core/providers/test_websocket_provider.py
```diff
@@ -270,11 +270,13 @@ async def test_listen_event_awaits_msg_processing_when_subscription_queue_is_ful
...
         method=RPCEndpoint("eth_subscribe"),
         params=["mock"],
-        response_formatters=(),
+        response_formatters=[[], [], []],
         subscription_id=sub_id,
     )
     async_w3.provider._request_processor._request_information_cache.cache(
-        "", sub_request_information
+        # cache key is the result of `generate_cache_key` with the sub_id as the arg
+        "0138b5d63d66121d8a6e680d23720fa7",
```
kclowes: may be worth just calculating that here 🤷 
```suggestion
        generate_cache_key(sub_id)
```

web3/eth/async_eth.py
```diff
@@ -714,17 +719,23 @@ async def subscribe(
                 bool,  # newPendingTransactions, full_transactions
             ]
         ] = None,
+        handler: Optional[EthSubscriptionHandler] = None,
+        handler_context: Dict[str, Any] = None,
```
kclowes: nit: 
```suggestion
        handler_context: Optional[Dict[str, Any]] = None,
```

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        :return: None
+        """
+        unsubscribed = [await self.unsubscribe(sx) for sx in self.subscriptions.copy()]
+        if all(unsubscribed):
+            self.logger.info("Successfully unsubscribed from all subscriptions.")
+        else:
+            if len(self.subscriptions) > 0:
+                self.logger.warning(
+                    "Failed to unsubscribe from all subscriptions. Some subscriptions "
+                    "are still active."
```
kclowes: Do you think there is a benefit in keeping both the debug log and the warning log here?

web3/providers/persistent/websocket.py
```diff
@@ -59,18 +59,19 @@ class WebSocketProvider(PersistentConnectionProvider):
...
         endpoint_uri: Optional[Union[URI, str]] = None,
         websocket_kwargs: Optional[Dict[str, Any]] = None,
         # `PersistentConnectionProvider` kwargs can be passed through
         **kwargs: Any,
     ) -> None:
+        # initialize the endpoint_uri before calling the super constructor
         self.endpoint_uri = (
             URI(endpoint_uri) if endpoint_uri is not None else get_default_endpoint()
         )
+        super().__init__(**kwargs)
```
kclowes, line 73: It looks like we do a fair amount of moving around the `super().__init__...` call throughout this PR - I'm curious about why we need to do that now, but not before?

web3/utils/subscriptions.py
```diff
@@ -0,0 +1,285 @@
...
+    """Wrap the handler to add bookkeeping and context creation."""
+    if handler is None:
+        return None
+
+    async def wrapped_handler(
+        context: EthSubscriptionContext[TSubscription, TSubscriptionResult],
+    ) -> None:
+        sx = context.subscription
+        sx.handler_call_count += 1
+        sx.manager.total_handler_calls += 1
```
kclowes: It's very possible I missed this, but other than testing, where does this get used?

web3/_utils/module_testing/persistent_connection_provider.py
```diff
@@ -8,245 +11,222 @@
...
+class SubscriptionHandlerTest:
+    passed: bool = False
+
+
+async def new_heads_handler(
+    context: NewHeadsSubscriptionContext,
+) -> None:
+    w3 = context.async_w3
+    sx = context.subscription
+    assert w3 is not None
```
kclowes: Maybe: 
```suggestion
    assert isinstance(w3, AsyncWeb3)
```

kclowes (approved):
Sweet! I didn't see anything major. This doesn't break the existing API, right? And then feel free to take or leave, but maybe `sub` or `subn` for an abbreviation of subscription instead of `sx` 😆 ?

docs/internals.rst
fselmo: The object is, but you can give it whatever name you want, it's not tied to the name `handler_context`. I can change it here to be more clear though.

tests/core/providers/test_websocket_provider.py
fselmo: A little more overhead for a test with an expected result, but makes it clearer 👍🏼

web3/providers/persistent/websocket.py
fselmo: I think it's cleaner to understand what arguments need to be instantiated before the `super.__init__()` call and describe them inline. There's some reliance on a few arguments needing to exist beforehand so I moved everything else after this init to make that distinction more clear for a reader.

I dedicated an isolated [commit](https://github.com/ethereum/web3.py/pull/3554/commits/4c313eac862d9d0c6eceafbe42f9228357b7c7b5) to these changes to make it a bit more clear.

web3/utils/subscriptions.py
fselmo: It's simply for keeping track. As a user I'd like to know these stats from a subscription manager if it's doing some background work that I can't see. If you think we should remove it, I can do so. Thoughts?

web3/providers/persistent/subscription_manager.py
fselmo: Debug includes the warning as well, but if you only set the log level to warning (warning and errors only) you will still see this log. Happy to remove though, thoughts?

fselmo: Maybe putting the debug content (the subscriptions that are still present) into the warning? 🤔 

edit: I went ahead and combined them into the warning. Let me know if you have another idea there.

fselmo:
> This doesn't break the existing API, right?

@kclowes, always good to ask 😅. The relevant API changes only [introduce more customization](https://github.com/ethereum/web3.py/pull/3554/files#diff-745353d44729acfd441365c5fe774d4b66affab2319ef3e1ab683dbed1885e38R722-R738). This shouldn't break the API. And if no handlers are passed in (new feature), you can still listen to free-flowing messages directly from the socket.

The typing in the socket stream was already `Optional` meaning a `None` could technically be returned as a "message". If a message has a handler a `None` is returned so that the iterator can move on to the next iteration. This is the only friction I see this PR introducing. I didn't see a clean way to move a users `async for message in socket` to the next message, if it was handled in a handler, without yielding a `None` but perhaps we can investigate that a bit more.

Thoughts on that?

docs/internals.rst
kclowes: Oh, I misread. No need to change unless you've already done it

kclowes (approved):
I like this! I didn't have time to go through with a fine tooth comb, but if you feel good about it and the tests, don't let me block your merge 👍

fselmo:
> I like this! I didn't have time to go through with a fine tooth comb, but if you feel good about it and the tests, don't let me block your merge 👍

It has been pretty battle tested but it's causing some irrelevant flaky replace transaction tests to fail pretty consistently, just on CI. I checked against `main` and it is specific to this branch. I'll wait a little longer to debug it and try to figure out what's causing it. The frustrating thing is that locally it isn't reproducible (these flaky tests haven't been reproducible locally all along). So I might push a commit on top of these until I can sniff it out and try to assess, roll the commit back, and handle appropriately.

My suspicion is it's not directly related to this PR. Ultimately those tests need to be properly fixed, so that might be what I end up doing in a separate PR.

Either way I'll be sure to preserve the current state of the new code (no force pushing on top of the previous commits) so that nothing that you all have reviewed changes.

web3/manager.py
```diff
@@ -513,20 +512,41 @@ async def socket_request(
...
-
+        async_w3 = cast("AsyncWeb3", self.w3)
+        middleware_onion = cast("MiddlewareOnion", self.middleware_onion)
+        send_func = await provider.send_func(
+            async_w3,
+            middleware_onion,
+        )
         self.logger.debug(
-            "Sending request to open socket connection: "
+            "Making request to open socket connection and waiting for response: "
```
fselmo: Note: This should be put back to its original message... got mixed with make_request somewhere along the way

web3/utils/subscriptions.py
```diff
@@ -0,0 +1,285 @@
...
+            f"    label: {sub.label}\n"
+            f"    call count: {sub.handler_call_count}\n"
+            f"    total handler calls: {sub.manager.total_handler_calls}"
+        )
+        await handler(context)
+
+    return wrapped_handler
+
+
+class EthSubscription(Generic[TSubscriptionResult]):
```
kclowes, line 100: This is pretty nitty, so feel free to take or leave, but what do you think about just calling this `Subscription`?

web3/utils/subscriptions.py
```diff
@@ -0,0 +1,285 @@
...
+
+class PendingTxSubscription(EthSubscription[Union[HexBytes, TxData]]):
+    def __init__(
+        self,
+        full_transactions: bool = False,
+        label: Optional[str] = None,
+        handler: Optional[PendingTxSubscriptionHandler] = None,
+        handler_context: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        self.full_transactions = full_transactions
```
kclowes, line 258: nit: are you using this anywhere?

reedsa:
Your docs are great and I wonder if they would be worth referencing in the [Eth.subscribe](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.subscribe) section and/or the Providers pages? Might help guide more advanced usage there.

fselmo:
> Your docs are great and I wonder if they would be worth referencing in the [Eth.subscribe](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.subscribe) section and/or the Providers pages? Might help guide more advanced usage there.

Good thought. We're going to do a big overhaul of these docs so this is just some basic stuff on usage for now. Don't want to put too much into this since it may change quite a bit. Thoughts?

web3/providers/persistent/persistent.py
```diff
@@ -72,9 +86,63 @@ def __init__(
...
+                    initialized = mw(async_w3)
+                    method, params = await initialized.async_request_processor(
+                        method, params
+                    )
+
+                return await self.send_request(method, params)
+
+            self._send_func_cache = (cache_key, send_function)
+
+        return self._send_func_cache[1]
```
kclowes, line 120: nit: maybe just `return send_function` here?

web3/providers/persistent/persistent.py
```diff
@@ -72,9 +86,63 @@ def __init__(
...
+                for mw in reversed(middleware):
+                    initialized = mw(async_w3)
+                    response = await initialized.async_response_processor(
+                        method, response
+                    )
+                return response
+
+            self._recv_func_cache = (cache_key, recv_function)
+
+        return self._recv_func_cache[1]
```
kclowes, line 146: same nit here, maybe:
```suggestion
        return recv_function
```

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,274 @@
...
+
+if TYPE_CHECKING:
+    from web3 import AsyncWeb3  # noqa: F401
+    from web3.providers.persistent import (  # noqa: F401
+        PersistentConnectionProvider,
+        RequestProcessor,
+    )
+
+
+class SubscriptionContainer:
```
kclowes: more nits: Do you think this `SubscriptionContainer` class deserves its own file?

web3/providers/persistent/subscription_manager.py
```diff
@@ -0,0 +1,274 @@
...
+        self.subscriptions_by_id: Dict[HexStr, EthSubscription[Any]] = {}
+        self.subscriptions_by_label: Dict[str, EthSubscription[Any]] = {}
+
+    def __len__(self) -> int:
+        return len(self.subscriptions)
+
+    def __iter__(self) -> Iterator[EthSubscription[Any]]:
+        return iter(self.subscriptions)
+
+    def add_subscription(self, subscription: EthSubscription[Any]) -> None:
```
kclowes: Do you think it's worth adding a tighter type here rather than the `Any`? Assuming I'm reading this right, there are well-defined params we allow here, right?

web3/_utils/module_testing/eth_module.py
```diff
@@ -2391,10 +2388,6 @@ async def test_async_eth_replace_transaction_gas_price_too_low(
         with pytest.raises(Web3ValueError):
             await async_w3.eth.replace_transaction(txn_hash, txn_params)
 
-    @flaky_with_xfail_on_exception(
```
kclowes, line 2394: 🎉

kclowes (approved):
This looks great! I had a some nits throughout, but feel free to take or leave 'em!

reedsa:
> > Your docs are great and I wonder if they would be worth referencing in the [Eth.subscribe](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.subscribe) section and/or the Providers pages? Might help guide more advanced usage there.
> 
> Good thought. We're going to do a big overhaul of these docs so this is just some basic stuff on usage for now. Don't want to put too much into this since it may change quite a bit. Thoughts?

SGTM! Looks good to go from my perspective.

web3/utils/subscriptions.py
fselmo: I'm into it 👍🏼

web3/utils/subscriptions.py
fselmo: No, this was a choice to set the properties on the class so it's available. May also be useful when handling subscriptions in the handlers.

web3/providers/persistent/persistent.py
fselmo: We need to return the cached function here. We don't always build it when we call `send_func` from the manager so this pulls it from the cache instead. It's how the `request_func` is set up as well. Thoughts?

web3/providers/persistent/persistent.py
fselmo: See other comment.

web3/providers/persistent/subscription_manager.py
fselmo: I could go either way. Do you? 😆 I don't think this file is too large but happy to abstract it out. Thoughts?

web3/providers/persistent/subscription_manager.py
fselmo: If I remember correctly, this allows us to still pass in and handle unknown subscriptions to our library, since we only define the execution-spec'd subscriptions. Alchemy has some [custom subs](https://docs.alchemy.com/reference/subscription-api) they define, for example. I'll see if that was the reason and if so, I think we can leave this be since we define tighter types on the defined subscription classes themselves where we know the results.

web3/utils/subscriptions.py
fselmo: oo I think I realized why I did this. We already had `EthSubscriptionParams` and `FormattedEthSubscriptionResponse` as types in web3.types. I added `EthSubscription` (class) and `EthSubscriptionResult` (type) in this PR. I'd still lean toward keeping it `Eth...` but I can be convinced that the class can be a different name. Thoughts there?

web3/providers/persistent/persistent.py
kclowes: Ahh, I see. That makes sense to me. Clever :)

web3/providers/persistent/subscription_manager.py
kclowes: Cool, I'm good with leaving it!

web3/providers/persistent/subscription_manager.py
kclowes: I could go either way too. I think I'd slightly prefer it, but not a strong opinion and we can always do it later if it gets unwieldy :)

fselmo: No prob, I'll add it in here 👍🏼

web3/utils/subscriptions.py
kclowes: I'm good with keeping it as `Eth` if that's the convention! 👍

web3/utils/subscriptions.py
kclowes: Gotcha, that makes sense. Maybe a comment (or test if there isn't one?) so it doesn't get accidentally removed along the way?

fselmo: I did miss it in the properties test. Logs checks the `address` and `topics` properties as well. I added this in the pending tx properties core test 👍🏼

PR #3557, opened 2024-12-13
Add Correct kwargs to AsyncHTTPProvider to allow graceful shutdown

### LINKED ISSUE #3524: Unclosed client session + Unclosed connector warnings
### What happened?

Creating two `AsyncWeb3` instances results in warnings on app exit.

```python
#!python3

import asyncio
from web3 import AsyncWeb3, HTTPProvider
from web3.providers import AsyncHTTPProvider

async def main():
    rpc = "https://mainnet.optimism.io"
    w3_1 = AsyncWeb3(AsyncHTTPProvider(rpc))
    await w3_1.eth.block_number
    w3_2 = AsyncWeb3(AsyncHTTPProvider(rpc))
    await w3_2.eth.block_number

asyncio.run(main())
```

### Full error output

```shell
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x103fc0ad0>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x103f5f9b0>, 8352140.600221)]']
connector: <aiohttp.connector.TCPConnector object at 0x103fc0c20>
```

_No response_

### web3 Version

7.4.0

### Python Version

3.13.0

### Operating System

osx

### Output from `pip freeze`

```shell
aiohappyeyeballs==2.4.3
aiohttp==3.10.10
aiosignal==1.3.1
annotated-types==0.7.0
asttokens==2.4.1
attrs==24.2.0
bitarray==3.0.0
certifi==2024.8.30
charset-normalizer==3.4.0
ckzg==2.0.1
cytoolz==1.0.0
decorator==5.1.1
eth-account==0.13.4
eth-hash==0.7.0
eth-keyfile==0.8.1
eth-keys==0.6.0
eth-rlp==2.1.0
eth-typing==5.0.1
eth-utils==5.1.0
eth_abi==5.1.0
executing==2.1.0
frozenlist==1.5.0
hexbytes==1.2.1
idna==3.10
ipython==8.28.0
jedi==0.19.1
matplotlib-inline==0.1.7
multidict==6.1.0
parsimonious==0.10.0
parso==0.8.4
pexpect==4.9.0
prompt_toolkit==3.0.48
propcache==0.2.0
ptyprocess==0.7.0
pure_eval==0.2.3
pycryptodome==3.21.0
pydantic==2.9.2
pydantic_core==2.23.4
Pygments==2.18.0
pyunormalize==16.0.0
regex==2024.9.11
requests==2.32.3
rlp==4.0.1
six==1.16.0
stack-data==0.6.3
toolz==1.0.0
traitlets==5.14.3
types-requests==2.32.0.20241016
typing_extensions==4.12.2
urllib3==2.2.3
wcwidth==0.2.13
web3==7.4.0
websockets==13.1
wheel @ file:///opt/homebrew/Cellar/python%403.13/3.13.0_1/libexec/wheel-0.44.0-py3-none-any.whl#sha256=f49b82715dce6365f75eddcb0a9bb47d0d46feaf14bbc739dfcbd677b7073f5b
yarl==1.17.0
```

kclowes:
### What was wrong?
The session wasn't being properly closed in the AsyncHTTPProvider if there was more than one w3 instance, and warnings were being emitted.

Closes #3524 

### How was it fixed?

Added an async context manager where needed. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -214,6 +214,8 @@ AsyncHTTPProvider
         >>> # If you want to pass in your own session:
         >>> custom_session = ClientSession()
         >>> await w3.provider.cache_async_session(custom_session) # This method is an async method so it needs to be handled accordingly
+        >>> # when you're finished, disconnect:
+        >>> w3.provider.disconnect()
 
     Under the hood, the ``AsyncHTTPProvider`` uses the python
     `aiohttp <https://docs.aiohttp.org/en/stable/>`_ library for making requests.
--- a/newsfragments/3557.bugfix.rst
+++ b/newsfragments/3557.bugfix.rst
@@ -1,0 +1,1 @@
+Add a ``disconnect`` method to the AsyncHTTPProvider that closes all sessions and clears the cache
--- a/tests/integration/go_ethereum/test_goethereum_http.py
+++ b/tests/integration/go_ethereum/test_goethereum_http.py
@@ -159,7 +159,45 @@ class TestGoEthereumAsyncNetModuleTest(GoEthereumAsyncNetModuleTest):
 
 
 class TestGoEthereumAsyncEthModuleTest(GoEthereumAsyncEthModuleTest):
-    pass
+    @pytest.mark.asyncio
+    async def test_async_http_provider_disconnects_gracefully_reuses(
+        self, async_w3, endpoint_uri
+    ) -> None:
+        w3_1 = async_w3
+
+        w3_2 = AsyncWeb3(AsyncHTTPProvider(endpoint_uri))
+        assert w3_1 != w3_2
+
+        await w3_1.eth.get_block("latest")
+        await w3_2.eth.get_block("latest")
+
+        w3_1_session_cache = w3_1.provider._request_session_manager.session_cache
+        w3_2_session_cache = w3_2.provider._request_session_manager.session_cache
+
+        for _, session in w3_1_session_cache.items():
+            assert not session.closed
+        for _, session in w3_2_session_cache.items():
+            assert not session.closed
+        assert w3_1_session_cache != w3_2_session_cache
+
+        await w3_1.provider.disconnect()
+        await w3_2.provider.disconnect()
+
+        assert len(w3_1_session_cache) == 0
+        assert len(w3_2_session_cache) == 0
+
+    @pytest.mark.asyncio
+    async def test_async_http_provider_reuses_cached_session(self, async_w3) -> None:
+        await async_w3.eth.get_block("latest")
+        session_cache = async_w3.provider._request_session_manager.session_cache
+        assert len(session_cache) == 1
+        session = list(session_cache._data.values())[0]
+
+        await async_w3.eth.get_block("latest")
+        assert len(session_cache) == 1
+        assert session == list(session_cache._data.values())[0]
+        await async_w3.provider.disconnect()
+        assert len(session_cache) == 0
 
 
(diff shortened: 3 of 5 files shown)


pacrob (approved):
:sheep:

fselmo:
This looks like this closes the session on every call. Is that what's going on? I think this negates the async session cache altogether. We have some logic that will look in the cache for a session and if it's closed, we have to recreate and replace the stale session.

Can we instead introduce this logic somewhere on the provider's garbage collection? Or we could encourage explicit usage of a `provider.disconnect()` which does the proper cleanup. That way we can still utilize open session caching. We could even implement this as a context manager on the provider instead of on the session, so when the context closes, we properly handle the closing on all open sessions.

Thoughts there?

fselmo:
I would think we have tests for making sure the session is re-used though... maybe I'm off here. But if we don't have these tests, we should add them here as well.

fselmo:
> I would think we have tests for making sure the session is re-used though... maybe I'm off here. But if we don't have these tests, we should add them here as well.

Looks like we don't have a test like this. This would be good to add. The following test on `test_go_ethereum_http.py` does fail and confirm we re-create the session every time with these changes. It passes on `main` 😅 .

```py
@pytest.mark.asyncio
    async def test_async_http_provider_reuses_cached_session(self, async_w3) -> None:
        await async_w3.eth.get_block("latest")
        session_cache = async_w3.provider._request_session_manager.session_cache
        assert len(session_cache) == 1
        session = list(session_cache._data.values())[0]

        await async_w3.eth.get_block("latest")
        assert len(session_cache) == 1
        assert session == list(session_cache._data.values())[0]
```

kclowes:
Aha, thanks for taking a look @fselmo. I _think_ only the persistent connection providers have a disconnect method, I'll fix that here.

alexferl:
@fselmo @kclowes so when I added my comment on the issue #3524, I took a quick look at the code a bit, to see if there was a `.close()` method or async context manager lifecycle methods on `AsyncWeb3`. Obviously there isn't a `close()` method, but the async context manager lifecycle methods are also limited to being used with persistent providers as you can see [here](https://github.com/ethereum/web3.py/blob/main/web3/main.py#L519). Unfortunately, that means that there's no quick fix to allow adding lifecycle methods with non-persistent providers without a bit of refactoring so that's why I came up with closing the session in the cache.

Perhaps having like a `close_all_sessions()` method on the `HTTPSessionManager` and then bubbling that up with a `close()` on `AsyncHTTPProvider` could be a simple solution?

kclowes:
Yeah, I came to the same conclusion @alexferl :) This now adds a `disconnect` method to the `AsyncHTTPProvider`, which will close all sessions and clear the cache. I'll add an issue to track down why exactly we need the timeout, but this is slightly better for now, I think.

kclowes:
Okay, all fixed. The timeout is actually linked to [in the docs]( https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown), but [this issue](https://github.com/aio-libs/aiohttp/issues/1925#issuecomment-2030109671) had an even better fix. Thanks for the assist @fselmo!

fselmo (approved):
lgtm!

reedsa (approved):

nikicat:
Hello, in my case underlying TCP connections are closed after every request, it creates delays for reestablishing connection for every request.

PR #3576, opened 2025-01-13
bugfix: Fix gethdev ipc path for auto.gethdev

fselmo:
### What was wrong?

There were some minor hiccups in establishing a connection to geth dev using `web3.auto.gethdev`. If the `WEB3_PROVIDER_URI` environment variable was set, it would automatically override the ipc path.

### How was it fixed?

- Check to make sure it points to a `geth.ipc`
- Else, build the path from the `TMPDIR` environment variable, defaulting to `/tmp/geth.ipc`.

`TMPDIR` is a unix and linux [convention](https://www.ibm.com/docs/en/idr/10.2.1?topic=windows-setting-tmpdir-environment-variable-linux-unix), so this was extended for linux in this PR. I think this is technically still a bugfix rather than a breaking change, given the intended purpose, but open to another interpretation as well.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`0_o`

### DIFF
--- a/newsfragments/3576.bugfix.rst
+++ b/newsfragments/3576.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a bug related to building the ipc path for connecting to a geth ``--dev`` instance via ``web3.auto.gethdev``.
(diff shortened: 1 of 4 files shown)


kclowes (approved):
:shipit:

PR #3579, opened 2025-01-16
Refactor for duplicate code in contract internal implementations

### LINKED ISSUE #3561: DRY up common code in contract classes
Working on #3540, I noticed a lot of the code that was cleaned up there can ideally be removed from the `contract.py` and `async_contract.py` into `base_contract.py` since the code is nearly identical. The biggest issue I can see to resolve may be the typing, which is why this warrants a separate issue since it's out of the scope of that PR.

Investigate the following, and perhaps other cases:

- refactor `__call__()` in `ContractFunction` (+ respective async code) to `BaseContractFunction`
- refactor `__getattr__()`, `__getitem__()` and `__iter__()` in `ContractEvents` and `ContractFunctions` (+ respective async code) to `BaseContractEvents` and `BaseContractFunctions`, respectively.

darwintree:
### What was wrong?

Closes #3561 

In internal contract implementations, there might be duplicate implementations repeating same code for async and non-async functions, for example, `ContractFunctions` and `AsyncContractFunctions`. 

### How was it fixed?

This case is because of:
- hardcoded class name
- type hint requirements

As for previous one, they can be resolved using self.__class__ to other techiniques to dynamically get the right class.
As for latter one, `Self` combined with `Generic` would help solve the issue

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/3579.internal.rst
+++ b/newsfragments/3579.internal.rst
@@ -1,0 +1,1 @@
+Move duplicate code into base class in (1) ContractFunction and AsyncContractFunction, (2) ContractEvents and AsyncContractEvents, and (3) ContractFunctions and AsyncContractFunctions.
--- a/newsfragments/3579.internal.rst
+++ b/newsfragments/3579.internal.rst
@@ -1,1 +1,1 @@
-Move duplicate code into base class in (1) ContractFunction and AsyncContractFunction, (2) ContractEvents and AsyncContractEvents, and (3) ContractFunctions and AsyncContractFunctions.
+Move duplicate code into ``BaseContract`` class from ``Contract`` and ``AsyncContract``. (1) ``ContractFunction`` and ``AsyncContractFunction`` (2) ``ContractFunctions`` and ``AsyncContractFunctions``, (3) ``ContractEvent`` and ``AsyncContractEvent``, and (4) ``ContractEvents`` and ``AsyncContractEvents``.
--- a/web3/contract/async_contract.py
+++ b/web3/contract/async_contract.py
@@ -32,7 +32,6 @@ from hexbytes import (
 
 from web3._utils.abi import (
     fallback_func_abi_exists,
-    get_name_from_abi_element_identifier,
     receive_func_abi_exists,
 )
 from web3._utils.abi_element_identifiers import (
@@ -86,11 +85,6 @@ from web3.contract.utils import (
     get_function_by_identifier,
 )
 from web3.exceptions import (
-    ABIEventNotFound,
-    ABIFunctionNotFound,
-    NoABIEventsFound,
-    NoABIFound,
-    NoABIFunctionsFound,
     Web3AttributeError,
     Web3TypeError,
     Web3ValidationError,
@@ -102,9 +96,6 @@ from web3.types import (
     StateOverride,
     TxParams,
 )
-from web3.utils.abi import (
-    _get_any_abi_signature_with_name,
-)
 
 if TYPE_CHECKING:
     from ens import AsyncENS  # noqa: F401
@@ -115,7 +106,7 @@ class AsyncContractEvent(BaseContractEvent):
     # mypy types
     w3: "AsyncWeb3"
 
-    def __call__(self, *args: Any, **kwargs: Any) -> "AsyncContractEvent":
+    def __call__(self, *args: Any, **kwargs: Any) -> Self:
         return copy_contract_event(self, *args, **kwargs)
 
     @combomethod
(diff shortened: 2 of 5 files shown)


web3/contract/base_contract.py
```diff
@@ -721,14 +766,124 @@ def __repr__(self) -> str:
...
+        )
+
+        if not len(function_abis_with_arg_count):
+            # Build an ABI without arguments to determine if one exists
+            function_abis_with_arg_count = [
+                ABIFunction({"type": "function", "name": function_name})
+            ]
+
+        # Check that arguments in call match a function ABI
+        num_attempts = 0
```
reedsa: I believe these `num_attempts` variables can be removed. They were mistakenly left in from a previous change.

web3/contract/base_contract.py
```diff
@@ -785,6 +938,46 @@ def __hasattr__(self, function_name: str) -> bool:
         except ABIFunctionNotFound:
             return False
 
+    def __iter__(self) -> Iterable[TContractFn]:
```
reedsa, line 944: The corresponding function `__iter__` in `ContractFunction` still exists and should be removed.

web3/contract/base_contract.py
```diff
@@ -785,6 +938,46 @@ def __hasattr__(self, function_name: str) -> bool:
...
+        if not hasattr(self, "_functions") or not self._functions:
+            return
+
+        for func in self._functions:
+            yield self[abi_to_signature(func)]
+
+    def __getattr__(self, function_name: str) -> TContractFn:
```
reedsa, line 951: The corresponding function `__getattr__` in `ContractFunction` still exists and should be removed.

web3/contract/base_contract.py
```diff
@@ -785,6 +938,46 @@ def __hasattr__(self, function_name: str) -> bool:
...
+                function_name, self._functions
+            )
+        else:
+            function_name = f"_{function_name}"
+
+        return super().__getattribute__(
+            function_name,
+        )
+
+    def __getitem__(self, function_name: str) -> TContractFn:
```
reedsa, line 981: The corresponding function `__getitem__` in `ContractFunction` still exists and should be removed.

web3/contract/base_contract.py
```diff
@@ -721,14 +766,124 @@ def __repr__(self) -> str:
...
+        self,
+        transaction: Optional[TxParams] = None,
+        block_identifier: Optional[BlockIdentifier] = None,
+        state_override: Optional[StateOverride] = None,
+        ccip_read_enabled: Optional[bool] = None,
+    ) -> Any:
+        """
+        Should be implemented by child class.
+        """
+        raise NotImplementedError("This method should be implemented by child class")
```
reedsa: Suggest keeping the wording consistent with other method overrides.
```suggestion
        raise NotImplementedError("This method should be implemented in the inherited class")
```

web3/contract/base_contract.py
```diff
@@ -721,14 +766,124 @@ def __repr__(self) -> str:
...
 
+    def call(
+        self,
+        transaction: Optional[TxParams] = None,
+        block_identifier: Optional[BlockIdentifier] = None,
+        state_override: Optional[StateOverride] = None,
+        ccip_read_enabled: Optional[bool] = None,
+    ) -> Any:
+        """
+        Should be implemented by child class.
```
reedsa:
```suggestion
        Implementation of ``call`` should create a callable contract function and execute it using the `eth_call` interface.
```

web3/contract/async_contract.py
```diff
@@ -122,7 +106,7 @@ class AsyncContractEvent(BaseContractEvent):
     # mypy types
     w3: "AsyncWeb3"
 
-    def __call__(self, *args: Any, **kwargs: Any) -> "AsyncContractEvent":
```
reedsa, line 125: I'm inclined to leave this as is since the returned object is a new copy of the event.
```suggestion
    def __call__(self, *args: Any, **kwargs: Any) -> "AsyncContractEvent":
```

web3/contract/base_contract.py
```diff
@@ -466,12 +474,13 @@ def __init__(
...
         abi: ABI,
         w3: Union["Web3", "AsyncWeb3"],
-        contract_event_type: Union[Type["ContractEvent"], Type["AsyncContractEvent"]],
+        contract_event_type: Type[TContractEvent],
         address: Optional[ChecksumAddress] = None,
     ) -> None:
         self.abi = abi
         self.w3 = w3
         self.address = address
+        self.contract_event_type = contract_event_type
```
reedsa, line 487: This seems handy, nice addition 👍

newsfragments/3579.internal.rst
```diff
@@ -0,0 +1 @@
+Move duplicate code into base class in (1) ContractFunction and AsyncContractFunction, (2) ContractEvents and AsyncContractEvents, and (3) ContractFunctions and AsyncContractFunctions.
```
reedsa:
```suggestion
Move duplicate code into ``BaseContract`` class from ``Contract`` and ``AsyncContract``. (1) ``ContractFunction`` and ``AsyncContractFunction`` (2) ``ContractFunctions`` and ``AsyncContractFunctions``, (3) ``ContractEvent`` and ``AsyncContractEvent``, and (4) ``ContractEvents`` and ``AsyncContractEvents``.
```

reedsa (commented):
These updates are looking great, thanks for your contributions. I've mentioned a few changes to additional cleanup and suggested exception messages. Should be good to go after that!

web3/contract/async_contract.py
darwintree: I think this is redundant as `Self` is used(if it work as I expected). Can check document: https://docs.python.org/3/library/typing.html#typing.Self

Code example from document:

```python
from typing import Self, reveal_type

class Foo:
    def return_self(self) -> Self:
        ...
        return self

class SubclassOfFoo(Foo): pass

reveal_type(Foo().return_self())  # Revealed type is "Foo"
reveal_type(SubclassOfFoo().return_self())  # Revealed type is "SubclassOfFoo"
```

darwintree: I think they are nearly the same and Self can keep consistency. Self is used to indicate type hints, but did not imply if it is `self` or a copy of it

darwintree:
@reedsa  Thanks for reviewing. 

Changes already made except for `AsyncContractEvent.__call__`. I think the mentioned problem is not an issue and it is ok to keep stay unchanged or using `Self` for consistent type hint. Would know your idea.

web3/contract/async_contract.py
reedsa: That's fine with me. Should this `__call__` and the one in the `ContractEvent` class move to base_contract? If not, I noticed the `ContractEvent.__call__` has the type `"ContractEvent"` rather than `self`, probably best to make them consistent.

web3/contract/base_contract.py
```diff
@@ -721,14 +766,125 @@ def __repr__(self) -> str:
...
+            # Build an ABI without arguments to determine if one exists
+            function_abis_with_arg_count = [
+                ABIFunction({"type": "function", "name": function_name})
+            ]
+
+        function_abi_matches = []
+        contract_function = None
+        for abi in function_abis_with_arg_count:
+            try:
+                pass
```
reedsa: Shouldn't need `pass` here.

web3/contract/async_contract.py
darwintree: Done. Moved into BaseContractEvent

darwintree:
@reedsa  I think the above mentioned issues are resolved now

reedsa (approved):

pacrob (approved):
lgtm!

PR #3583, opened 2025-01-17
Flaky test help

fselmo:
### What was wrong?

- We recently removed the specific flaky xfail markers from some replace transaction tests. They now sometimes fail with "nonce too low", which is better than the previous timeout issues. We should add the flaky_geth_dev_mining decorator to these so they can run again and ideally pass 1 out of 3 times.

- We also sometimes get performance issues with older Python versions on the multiple contract init core tests. Let's keep this a more conservative number for now. `1.5` -> `1.75` should do I think.

### Todo:

- [x] Clean up commit history
- [x] Re-run the CI multiple times to see if we get consistent results

#### Cute Animal Picture

`0_o`

### DIFF
--- a/newsfragments/3583.internal.rst
+++ b/newsfragments/3583.internal.rst
@@ -1,0 +1,1 @@
+Address flaky tests in CI runs.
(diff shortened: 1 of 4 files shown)


fselmo:
Seems to be [working well](https://app.circleci.com/pipelines/github/ethereum/web3.py?branch=pull%2F3583) now with just the geth dev mining decorator (flaky w/ `max_runs=3`). Unfortunately `flaky` won't tell us if a test ran twice or 3 times, just that it passed the 1 out of required 3 times. I think that's fine.

kclowes (approved):
LGTM after a newsfragment gets added!

PR #3590, opened 2025-01-26
Fix Typo in `Beacon.get_hash_root` Documentation

zeevick10:
**Description:**
This pull request addresses a minor typo in the documentation for the `Beacon.get_hash_root` method. The root hash value was incorrectly formatted in the example output.
**Changes Made:**
Corrected the formatting of the root hash in the example output to ensure consistency and accuracy.
This update enhances the clarity of the documentation and ensures that users have the correct information when utilizing the `get_hash_root` method.

### DIFF
--- a/docs/web3.beacon.rst
+++ b/docs/web3.beacon.rst
@@ -35,7 +35,7 @@ Methods
         >>> beacon.get_hash_root()
         {
           "data": {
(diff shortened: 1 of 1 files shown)


pacrob:
We do prefer double quotes over single, if only because `black` does. But there are hundreds of places just in that file alone where it could be updated. Please do a more a more thorough search if you are submitting formatting-only PRs.

PR #3595, opened 2025-01-30
Subscription manager bugfixes:

fselmo:
### What was wrong?

Some bugs shared via Discord conversation with @fubuloubu.

### How was it fixed?

- Fix ``run_forever`` logic so that it doesn't require subscriptions to start and doesn't end when no subscriptions are present.
- Fix the ``unsubscribe`` method so that it can accept subscription ids.
- Fix the ``unsubscribe`` method so it behaves like the ``subscribe`` method, accepting both single or multiple subscriptions or subscription ids. **Note this does change the signature from ``subscription`` to ``subscriptions``, but I think this can be a bugfix since the ``subscribe`` method accepts multiple subscriptions and ``unsubscribe`` should behave similarly.** Happy to revert back and change this in v8 if we don't want to break it here.

- Fix ``subscribe`` so that it returns a list of subscription ids when a list of subscriptions is provided.

Add tests:

- Add a test for ``run_forever`` to make sure it can start without subscriptions and that the task does not stop when no subscriptions are present.
- Add a test for ``unsubscribe`` to make sure it can accept multiple subscriptions as hexstr ids or as subscription objects.
- Update subscriptions tests to check that a list of ids is returned when multiple subscriptions are provided to the ``subscribe`` method.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-01-30 at 11 17 58

### DIFF
--- a/newsfragments/3585.breaking.rst
+++ b/newsfragments/3585.breaking.rst
@@ -1,0 +1,1 @@
+The bugfix to match ``unsubscribe`` to ``subscribe`` for multiple subscriptions breaks the function signature for ``unsubscribe``, changing the ``subscription`` argument to ``subscriptions``.
--- a/newsfragments/3595.bugfix.rst
+++ b/newsfragments/3595.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bugs related to subscription manager: ``run_forever`` can start with ``0`` subscriptions and remains alive, ``unsubscribe`` accepts single or multiple subs as objects or hexstrs, ``subscribe`` for many subs returns a list of hexstr ids.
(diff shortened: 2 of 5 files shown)


fselmo:
Of note here is `subscription` -> `subscriptions` in the `manager.unsubscribe()` signature. If we feel that is too breaking of a change, even this early on in the manager's existence, I can put this back and we can change the signature in `v8`. I think it could be chalked up to a bugfix and would significantly reduce confusion if both `subscribe` and `unsubscribe` had the same arg name.

kclowes (approved):
LGTM! 

> Note this does change the signature from subscription to subscriptions, but I think this can be a bugfix since the subscribe method accepts multiple subscriptions and unsubscribe should behave similarly.

Can we easilly alias `subscription`/`subscriptions`? and then remove `subscription` in v8? If not, I think it's okay to break, but we should put it in the release notes and make sure docs are up to date!

fselmo:
> Can we easilly alias subscription/subscriptions? and then remove subscription in v8? If not, I think it's okay to break, but we should put it in the release notes and make sure docs are up to date!

If you're good with it I'm happy to take the blame on any backlash 😆 . I'll update the release notes with a breaking newsfragment. I don't think the docs even highlight this yet but I'll double check.

PR #3597, opened 2025-01-31
docs: rename osx to macos

antazoey:
### What was wrong?

OSX renamed to macOS in 2016!


fselmo:
Yes... yes it was. Thanks!

PR #3620, opened 2025-02-16
Add flag to always convert websocket payloads to text frames

### LINKED ISSUE #3619: Some Websocket endpoints don't handle binary communication
### What happened?

During debugging of issues of websocket connections between Ankr (which works) and NodeReal (which doesn't) with the [Subscriptions Example](https://web3py.readthedocs.io/en/latest/subscriptions.html#an-example), I noticed in the debug logs that payloads were sent in binary, vs a `websocket` (library)-based script I built whose payloads I could see in plaintext.

Long story short, it appears not all Websocket servers can handle both.

For web3py's part, `WebSocketProvider.socket_send()` requires request data to be in bytes, which according to the [`websockets` docs](https://websockets.readthedocs.io/en/stable/reference/sync/client.html#websockets.sync.client.ClientConnection.send) is then sent as a binary frame. When I modified this method to `.decode()` that payload to text, the example code worked as expected for all RPC platforms I provided.

### Full error output

```shell
web3.manager.RequestManager DEBUG    Sending request to open socket connection: [NodeReal],
    method: eth_subscribe,
    params: ('logs', {'topics': ['0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ea']})
[...]

web3.providers.WebSocketProvider DEBUG    Caching request info:
    request_id=0,
    cache_key=cfcd208495d565ef66e7dff9f98764da,
    request_info={'method': 'eth_subscribe', 'params': ('logs', {'topics': ['0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ea']}), 'response_formatters': (<function apply_formatter_if at 0x1021ea7a0>, <cyfunction identity at 0x10206b6b0>, <cyfunction identity at 0x10206b6b0>), 'subscription_id': None, 'middleware_response_processors': []}

web3.manager.RequestManager DEBUG    Getting response for request from open socket connection:
    request: {'id': 0, 'jsonrpc': '2.0', 'method': 'eth_subscribe', 'params': ('logs', {'topics': ['0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ea']})}

[...]
Timed out waiting for response with request id `0` after 30.0 second(s). This may be due to the provider not returning a response with the same id that was sent in the request or an exception raised during the request was caught and allowed to continue.
```

While I intend to relay this to NodeReal for potential infrastructure-based solutions, it may be prudent to allow users to force `WebSocketClient` to use text frames, as we have little control over which platforms support binary communication.

This could be as simple as keeping it binary internally as to not disrupt internals too much, and simply converting `request_data` to text during `.socket_send()` if a flag is present.

### web3 Version

7.8.0

### Python Version

3.11

DefiDebauchery:
### What was wrong?

Some Websocket RPC platforms do not support binary frames, which is the internal default for communication, leading to connections that did not return data.

Closes #3619 

### How was it fixed?

Unfortunately this cannot be automatically determined without delaying the use of the websocket, so a flag was added on init that will inform the Provider to decode the binary payload into text so the underlying `websockets` library can send as such. And since all providers should handle text frames, setting this flag is not likely to have any negative impact.

### Todo:

- [X] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -291,7 +291,7 @@ AsyncIPCProvider
 WebSocketProvider
 +++++++++++++++++
 
-.. py:class:: web3.providers.persistent.WebSocketProvider(endpoint_uri: str, websocket_kwargs: Dict[str, Any] = {})
+.. py:class:: web3.providers.persistent.WebSocketProvider(endpoint_uri: str, websocket_kwargs: Dict[str, Any] = {}, use_text_frames: bool = False)
 
     This provider handles interactions with an WS or WSS based JSON-RPC server.
 
@@ -299,6 +299,8 @@ WebSocketProvider
       ``'ws://localhost:8546'``.
     * ``websocket_kwargs`` this should be a dictionary of keyword arguments which
       will be passed onto the ws/wss websocket connection.
+    * ``use_text_frames`` will ensure websocket data is sent as text frames
+      for servers that do not support binary communication.
 
     This provider inherits from the
     :class:`~web3.providers.persistent.PersistentConnectionProvider` class. Refer to
--- a/newsfragments/3619.feature.rst
+++ b/newsfragments/3619.feature.rst
@@ -1,0 +1,1 @@
+Add `use_text_frames` flag for `WebSocketProvider` to work around websocket servers that don't support binary frames
--- a/tests/core/providers/test_websocket_provider.py
+++ b/tests/core/providers/test_websocket_provider.py
@@ -55,6 +55,14 @@ class WSException(Exception):
     pass
 
 
+GET_BLOCK_JSON_MESSAGE = {
+    "id": 0,
+    "jsonrpc": "2.0",
+    "method": "eth_getBlockByNumber",
+    "params": ["latest", False],
+}
+
+
 def test_get_endpoint_uri_or_ipc_path_returns_endpoint_uri():
     provider = WebSocketProvider("ws://mocked")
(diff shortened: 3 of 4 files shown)


fselmo:
@DefiDebauchery what do you think about setting an internal encoding property (something like `self._message_encoding: Literal["utf-8", "bytes"]`) on the `AsyncJSONBaseProvider` and having this property always default to `bytes`, and having the control of the flag as you put it here on the `WebSocketProvider` that can change that property?

The way I see this is we already encode all messages in the base class to `bytes`, so if we have that property earlier on, instead of re-encoding as `utf-8`, we can be a bit more efficient and encode it the first time as text in the base class if we have that property at that step.

I have a commit locally that sort of works like this, if you wouldn't mind me pushing that up and we can discuss and move forward? Let me know your thoughts.

fselmo:
On second thought, that would touch the typing in way too many areas of the code just to accommodate this WebSocket-specific change. It's a bit less performant but I would say this is fairly negligible so I'm good with the websocket provider re-decoding into text as the message goes out.

fselmo:
I added a test and fixed the linting issues. Looks good on my end. If you're all set, I'm happy to merge this. Let me know 👀

DefiDebauchery:
~~Please do!~~ LGTM!

PR #3623, opened 2025-02-20
handle when a client returns the response id more than once (JSON-RPC spec error)

fselmo:
### What was wrong?

- Handle an edge case when a client sends an error response that isn't tied to any outstanding requests. A user reported a client returning the same response `id` more than once, the second with an error. This does not adhere to the JSON-RPC spec but was reported by a user on Discord as leaving the library hanging indefinitely, so we should try to fail hard under these edge cases.

### How was it fixed?

- After every request, if the request / response cache still has anything in it, check if there are any errors and if they are not tied to outstanding requests, raise them.
- Add a simple test to check that this logic is in place.
- Move the listener task exception check above looking for a response with `id` so we can fail hard and fast if an exception has already been raised in the listener.
- Some refactoring of formatters and validation methods out of `manager.py` and into `_utils.formatters.py` and `_utils/validation.py`, respectively.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

```
      /^-----^\
      V  o o  V
       |  Y  |
        \ Q /
       / - \
       |    \
       |     \     )
       || (___\==== 
```

### DIFF
--- a/newsfragments/3623.misc.rst
+++ b/newsfragments/3623.misc.rst
@@ -1,0 +1,1 @@
+Raise stray errors a client may send that is not tied to any outstanding requests.
--- a/tests/core/providers/test_async_ipc_provider.py
+++ b/tests/core/providers/test_async_ipc_provider.py
@@ -360,7 +360,7 @@ async def test_async_ipc_provider_write_messages_end_with_new_line_delimiter(
     async with AsyncWeb3(AsyncIPCProvider(pathlib.Path(jsonrpc_ipc_pipe_path))) as w3:
         w3.provider._writer.write = Mock()
         w3.provider._reader.readline = AsyncMock(
-            return_value=b'{"id": 0, "result": {}}\n'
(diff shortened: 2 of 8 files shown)


fselmo:
Actually, [the original approach in this PR] is a terrible idea as `id`s don't always come in order. I'm going to re-think this but since it's an issue on the client side I'm not 100% sure we need a great answer here. Maybe looking for stray `error` messages in the cache might be the best way to go about this. Disregard the review for now.

fselmo:
@pacrob @kclowes ok, ready for review. We can look for stray error responses in the cache that have no outstanding requests and raise if any are present. This still allows a single request with `id` to retrieve its response by `id` and raise more appropriately in the `manager.py`. But if we find any stragglers in the cache we still account for them.

Again, not really something that's expected, but better than leaving the library hanging like in the case that was reported.

kclowes (approved):
LGTM! Nice manager refactor!

PR #3629, opened 2025-02-27
Fix numerical formatting and byte encoding issues in test_utils_functions.py

vipocenka:
- Incorrect numerical formatting: `000` was used instead of `0`, which is unnecessary and could lead to confusion.  

- Incorrect byte encoding in test cases:  
  - `b"\x124"` was used instead of the correct representation `b"\x12\x34"`.  
  - `b"\x12444"` was corrected to `b"\x12\x34\x34\x34"`.  
  - `b"12"` was replaced with `b"\x12"` for consistency in hex representation.  

### **Todo:**  
- [ ] Clean up commit history  

### **Cute Animal Picture** 🐶🐱  
Cute Puppy

### DIFF
--- a/tests/core/filtering/test_utils_functions.py
+++ b/tests/core/filtering/test_utils_functions.py
@@ -20,7 +20,7 @@ from web3.exceptions import (
     (
         (
             pytest.param(
-                (-12345, 000, 111, Decimal(2) + Decimal(1) / Decimal(10)),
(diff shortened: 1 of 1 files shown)


fselmo:
The point of these tests is to test non-strict encoding. `b"\x124"` treats `4` as ASCII which represents `b"\x34"`. 

```py
>>> hex(int.from_bytes(b"\x124", "big"))
'0x1234'
```

Thanks for the thought but I am going to close this.

PR #3630, opened 2025-03-03
Grammar and Clarity Improvements in Documentation

Hopium21:
1. File Modified: docs/README-freebsd.md
Change Made: Corrected grammar in a sentence.
Old:
otherwise almost every Python 3 module will fail to install.
New:
otherwise almost every Python 3 module will fails to install.
Reason: This is a grammatical correction to ensure the correct verb form is used. The original sentence had a subject-verb agreement error, where the verb "fail" should be in the singular form to match "every Python 3 module."
2. File Modified: docs/troubleshooting.rst
Change Made: Changed "that" to "who" for correct usage of relative pronouns.
Old:
Some Windows users that do not have Microsoft Visual C++ version 14.0 or greater installed may see an error message.
New:
Some Windows users who do not have Microsoft Visual C++ version 14.0 or greater installed may see an error message.
Reason: The relative pronoun "that" is typically used for things or objects, whereas "who" is used for people. Since the sentence refers to people (Windows users), "who" is the grammatically correct choice.

### DIFF
--- a/docs/README-freebsd.md
+++ b/docs/README-freebsd.md
@@ -5,7 +5,7 @@
 ### Prerequisites
 
 Make sure you've UTF-8 defined for charset and lang in your [~/.login_conf](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/using-localization.html),
-otherwise almost every Python 3 module will fail to install.
+otherwise almost every Python 3 module will fails to install.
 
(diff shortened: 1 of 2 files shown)


pacrob:
Thanks @Hopium21 , but we won't merge this. The first actually introduces a grammatical error. The who/that distinction is not important here.

PR #3632, opened 2025-03-04
Fix incorrect exception types in event tests

vipocenka:
Some test cases in `test_construct_event_data_set.py` and `test_construct_event_topics.py` were using incorrect exception types for invalid input handling:

- `{"arg0": b"131414"}` was changed from `ValueOutOfBounds` to `EncodingTypeError`.

- `{"arg0": [""]}` was changed from `EncodingTypeError` to `ValueOutOfBounds`.

animal


kclowes:
Thanks for the PR, but it looks like those error types were correct. Closing.

PR #3633, opened 2025-03-07
fix error in function test_async_transaction.py

futreall:
tests/core/utilities/test_async_transaction.py
tranaction_fee - transaction_fee

_____________


futreall:
image

PR #3635, opened 2025-03-08
 Update type annotations in persistent provider and fix error in validation error message

Pricstas:
- The error message incorrectly stated:  
    `"The following array length does not the length specified"`  
  - It was missing the word **"match"**, leading to unclear error messaging.
  
- Incorrect Type Annotations:  
  - `_send_func_cache` and `_recv_func_cache` were annotated as `Tuple[int, Callable[..., Coroutine[Any, Any, RPCRequest]]]`, but were initialized with `(None, None)`, causing type inconsistency.  
  - This could lead to type-checking errors when using `mypy` or similar tools.

### DIFF
--- a/web3/_utils/validation.py
+++ b/web3/_utils/validation.py
@@ -131,7 +131,7 @@ def validate_abi_value(abi_type: TypeStr, value: Any) -> None:
                 )
             if specified_length != len(value):
                 raise Web3TypeError(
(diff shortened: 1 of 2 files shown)


Pricstas:
Funny Animal

kclowes (approved):
This looks good to me, thanks! I just fixed the linting. Will merge once tests pass.

PR #3637, opened 2025-03-11
Use ``-U`` for geth steps to install `py-geth`

fselmo:
### What was wrong?

- Issues with CI on main after pushing an update to the geth test fixture.

### How was it fixed?

- ``-U`` will update the version to the latest even if a *py-geth* version is already installed. This is required if we have just updated the test fixture and need to install the new geth version from the latest *py-geth* version.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


pacrob (approved):
:sheep:

PR #3615, opened 2025-02-15
Remove duplicated 'eth-account' entry in ETHEREUM_LIBRARIES

Bilogweb3:

PR #3644, opened 2025-03-14
fix duplicate install_pre_releases.py

futreall:
"eth-account", - `deleted`


fselmo:
duplicated by #3615 and not relevant, see that PR for info

PR #3652, opened 2025-03-25
feat: optimize with iterators

BobTheBuidler:
### What was wrong?

The formatters init quite a few lists each run, and can be made much more efficient using iterators. This PR does so.

I figured this was more important than the logger stuff in my last PR so finished this one up first.

Related to Issue # N/A
Closes # N/A

### How was it fixed?
creation of an IteratorProxy class that allows `map` objects to be passed thru the formatters instead of fully materializing a list every step of the way

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

)

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -7,6 +7,7 @@ from typing import (
     Collection,
     Dict,
     Iterable,
+    Iterator,
     NoReturn,
     Tuple,
     TypeVar,
@@ -35,7 +36,6 @@ from eth_utils.curried import (
     is_string,
     to_checksum_address,
     to_list,
-    to_tuple,
 )
 from eth_utils.toolz import (
     complement,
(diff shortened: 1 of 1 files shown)


web3/_utils/abi.py
```diff
@@ -611,17 +611,21 @@ def normalizer(datatype, data):
...
+        strip_abi_types,
+        list,
     )
 
-    return pipe(data, *pipeline)
-
 
 @curry
-def abi_data_tree(types: Sequence[TypeStr], data: Sequence[Any]) -> List[Any]:
+def abi_data_tree(types: Sequence[TypeStr], data: Sequence[Any]) -> "map[ABITypedData]":
```
BobTheBuidler: idea for more optimization: rip out this func and replace it with a lambda which would have less overhead than using curry

web3/_utils/abi.py
```diff
@@ -611,17 +611,21 @@ def normalizer(datatype, data):
     2. Recursively mapping each of the normalizers to the data
     3. Stripping the types back out of the tree
     """
-    pipeline = itertools.chain(
-        [abi_data_tree(types)],
-        map(data_tree_map, normalizers),
-        [partial(recursive_map, strip_abi_type)],
+    return pipe(
```
BobTheBuidler: this could be further optimized by caching the pipe per-method as a `compose` object

kclowes:
This looks like a combination of a few of your other PRs? Feel free to ping me once tests are passing, and let me know the order I should look at them. Thanks!

BobTheBuidler:
Yes, this was the original PR that ended up going beyond its scope. I wanted to make it easier on you so I'm rebasing this one as we handle the others and we can handle it towards the end.

BobTheBuidler:
I've rebased this on top of all the PRs we merged and this is now just one minor tweak, ready for review.

tldr we were initing tuples that weren't necessary and can just pass generators around instead

this does change the API but both functions are internal and the codebase handles the change just fine

PR #3659, opened 2025-04-10
Prague updates

### LINKED ISSUE #3650: Investigate pydantic model parsing in the json Serde class

The `FriendlyJsonSerde` has specific instructions for how to parse certain objects. More recently, I started tinkering with how we could standardize pydantic models across the libraries to serialize their "json" as JSON-RPC-spec-compatible objects.

I believe we can give the right instructions in the serde class to where if we define serialization options for all pydantic objects that expect to be serialized in some JSON-RPC call, we can directly use these objects in forming requests and have them be properly serialized, no matter their attributes or properties.

As an initial example, take the `SetCodeAuthorization` from eth-account in [this PR](https://github.com/ethereum/eth-account/pull/316/files#diff-1bad859d609bf23ecd2adc90abc7b9770c5c91a644ebf9253f7a9902ca236108R103-R149). This defines a way to serialize the object as camel-case and defines `_excludes` so that it only serializes what the JSON-RPC object would expect. In order to do this we must be specific when we model dump, as outlined here.

If we define a standard across our libraries and expect this standard when we serialize using the `FriendlyJsonSerde` in web3.py, we can remove the need for the [as_rpc_object()](https://github.com/ethereum/eth-account/pull/316/files#diff-1bad859d609bf23ecd2adc90abc7b9770c5c91a644ebf9253f7a9902ca236108R145-R149) method in this model (and deprecate it), and we can directly just use the pydantic model inside transactions and serialize them appropriately. To be clear, we use the logic that the `as_rpc_object()` method I linked is using directly in the Serde class itself so that it is implicitly converted as expected, rather than having to rely on some function to do this on every pydantic model.

---

Bonus: I think pydantic plays well with `orjson` so issue #3649 can also remain relevant here and both of these features could possible play quite nicely together.

fselmo:
### What was wrong?

We need to add some additional support for Prague-related fields.

### How was it fixed?

- Add formatters for authorization list fields and block requests field in block headers
- Add integration test for 7702 so both geth and py-evm interactions are tested with expected results

--- 
Significant change to integration tests:

Change integration tests so that each test has its own context. We are having way too many issues with muddied contexts and tests needing to wait on the right moment to test transactions between a mining state. This isolates each integration test into its own state which makes issues a lot easier to track and should give less CI headaches.

---

### Todo:

- [x] Clean up commit history
- [x] Add or update documentation related to these changes --> This probably needs to be better... future PR + blog post?
- [x] Add at least one good 7702 tx test to the integration tests so we can test across both geth and eth-tester <-> py-evm
- [x] Will need ethereum/eth-tester#317 and a release which is waiting on a *py-evm* Prague release. Update eth-tester lower pin when release is ready.

#### Cute Animal Picture

Screenshot 2025-04-23 at 13 19 00


fselmo:
I fixed the integration test issues 🙂  I think they are better this way... every test runs in its own context now and we can parallelize them with xdist in the CI which doesn't seem to cause any issues on the `xlarge` setup with `-n 10`.

This is ready for a review 👀

docs/contributing.rst
```diff
@@ -200,6 +201,16 @@ tests are written across different files within ``/web3/_utils/module_testing``.
...
 
+The integration tests are each run in insolation to prevent muddied contexts. Because
+they are run in isolation, they can be parallelized with ``pytest-xdist`` in order to
+speed up the test suite. To run the tests in parallel, you can use the ``-n`` flag
+with ``pytest``. For example, to run the tests in parallel with 4 workers, you can
+use the following command:
+
+.. code:: sh
+
+    $ pytest tests/integration/go_ethereum/path/to/module/or/test -n 4
```
kclowes, line 212: maybe:
```suggestion
    $ pytest tests/integration/go_ethereum/path/to/module/or/test -n auto
```

tox.ini
```diff
@@ -20,17 +20,18 @@ allowlist_externals=make,pre-commit
...
-    integration-ethtester: pytest {posargs:tests/integration/test_ethereum_tester.py}
+    ens: pytest {posargs:tests/ens --ignore=tests/ens/normalization/test_normalize_name_ensip15.py -n 10}
+    ensip15: pytest {posargs:tests/ens/normalization/test_normalize_name_ensip15.py -q -n 10}
+    integration-goethereum-ipc: pytest {posargs:tests/integration/go_ethereum/test_goethereum_ipc.py -k "not Async" -n 10}
+    integration-goethereum-ipc_async: pytest {posargs:tests/integration/go_ethereum/test_goethereum_ipc.py -k Async -n 10}
+    integration-goethereum-http: pytest {posargs:tests/integration/go_ethereum/test_goethereum_http.py -k "not Async" -n 10}
+    integration-goethereum-http_async: pytest {posargs:tests/integration/go_ethereum/test_goethereum_http.py -k Async -n 10}
+    integration-goethereum-legacy_ws: pytest {posargs:tests/integration/go_ethereum/test_goethereum_legacy_ws.py -n 10}
+    integration-goethereum-ws: pytest {posargs:tests/integration/go_ethereum/test_goethereum_ws -n 10}
+    integration-ethtester: pytest {posargs:tests/integration/test_ethereum_tester.py -n 10}
```
kclowes: I wonder if auto would be better here? then we're not hard coding a certain number.

tox.ini
```diff
@@ -20,17 +20,18 @@ allowlist_externals=make,pre-commit
 install_command=python -m pip install {opts} {packages}
 usedevelop=True
 commands=
+    # Note: `pytest-xdist` n=10 seems to cause connection issues with geth; keep it at 6 for now
```
kclowes: I think this can go?

web3/_utils/method_formatters.py
```diff
@@ -161,6 +164,12 @@ def type_aware_apply_formatters_to_dict(
     """
     Preserve ``AttributeDict`` types if original ``value`` was an ``AttributeDict``.
     """
+    # TODO: In v8, Use eth-utils 5.3.0 as lower pin where ``apply_formatters_to_dict``
```
kclowes, line 167: Do we want to pull CamelModel into web3 in this PR?

web3/_utils/module_testing/eth_module.py
```diff
@@ -3485,8 +3527,8 @@ def test_eth_replace_transaction_legacy(
         assert replace_txn["value"] == 1
         assert replace_txn["gas"] == 21000
         assert replace_txn["gasPrice"] == txn_params["gasPrice"]
+        print("yeeeeeeehhhboiiiiiiiii")
```
kclowes: 😆

kclowes (approved):
:shipit: Nothing major stood out to me. For my own learning, is the fixture scope from `module` -> `function` for ease of parallelization, or is there another reason? It looks like test times are down with the parallelization you added, but could we get more speedup if we scope the fixtures wider?

web3/_utils/module_testing/eth_module.py
fselmo: Can you tell how much joy this gave me? 😆

web3/_utils/method_formatters.py
fselmo: I suppose we can. Should I add it to this PR?

tox.ini
fselmo: My only hesitation was that `auto` could end up being _more_ than `10`. I was sort of trying to set some lower limit. But I guess if it can handle more than `10` that would be ok?

fselmo: Thoughts on that?

fselmo: I added it, let's see what auto does for the `xlarge` CI runs

web3/_utils/method_formatters.py
kclowes: I'll leave that up to you!

fselmo:
> For my own learning, is the fixture scope from `module` -> `function` for ease of parallelization, or is there another reason? 

I think there are maybe a few fixtures we can change back to `module` but this is so that each test has its own instance of geth. This is the change that makes them run in isolation as each function is a single test.

---
edit: Even still, the overhead of _not_ changing only one or two fixtures to `module` or `session` is probably pretty low

web3/_utils/module_testing/eth_module.py
kclowes: lol yes I can!

web3/_utils/method_formatters.py
fselmo: I added it in here 👍🏼

fselmo: Oh we can't. `eth-account` can't inherit from ``CamelModel`` so we can't support it here yet since ``SignedSetCodeAuthorization`` is coming from eth-account.

Putting it back.

fselmo:
@kclowes all changes from your comments accounted for. I don't think the ``dev.period`` is going to affect test times other than the ones that get a / wait for a transaction receipt, which should be 1 or 2 tests. Aside from those, tests should finish as soon as the test logic passes the test, which for most tests doesn't require block processing. I raised the ``dev.period`` to ``4`` in this PR since ``-n auto`` increased the number of workers to ~ ``36`` and slowed down processing time for each test, which ended up causing some issues due to mining.

In the end ``-n auto`` with increased ``dev.period`` seems like the right way to go. So far, running the CI with ``--dev.period=4`` seems to be plenty of time for the pending / replace transaction tests to run their logic before the block is processed.

I'm going to re-run the CI a few more times and make sure the success rate good and then I'll merge this.

reedsa (approved):
👍

pacrob (approved):
:sheep:

Issue #3613, opened 2025-02-12
New batching feature completely breaks thread safety


simone1999:
### What happened?

Once w3.batch_requests() gets called, the entire web3 object/provider is set into batching mode. Any other thread (or function called inside the batch context manager) is not able to use the web3 object after batching got initialized. All main methods return request descriptors meant for forwarding to batch.add rather than the expected responses. simply copying the web3 object with copy.copy also is not enough as the batch flag is set deep at the provider level rather than the web3 object itself. And copy.deepcopy does not work with web3.py anymore and throws TypeError: cannot pickle '_queue.SimpleQueue' object.

In short, the new batch feature breaks fundamental thread safety and often ends in hard to find race conditions.

### Code that produced the error

```python

```

### Full error output

```shell

```

### web3 Version

7.6.1

### Python Version

3.12

### Output from `pip freeze`

```shell

```

fselmo:
That's correct. For now, the Web3 provider instance enters a batching context while batching so that it doesn't make requests and you can use the same module API to add requests to the batch (e.g. `w3.eth.get_block("latest")` instead of building out the json).

We do plan on building on this so that this doesn't happen but that is lower priority at the moment. For now, I think you can get around it by instantiating a separate Web3 instance, even to the same uri, that is not meant to be used in the batching context.

I understand that's not ideal but it should technically be able to get you unstuck for now (I hope).

simone1999:
Thx a lot. I subclass the Web3 class itself with multiple added checks on init. This makes instantiating the object many times not very practical. But I will find a solution. Thx for the fast reply and the great work you and your team is doing with Web3.py

simone1999:
I tried to work around the issue with concatenated shallow copies like so:
w3 = copy.copy(self.w3)
w3.manager = copy.copy(w3.manager)
w3.manager.provider = copy.copy(w3.manager.provider)
w3.manager.provider._is_batching = False

But this does not resolve the issue. Honestly expected this to simply work :D

kclowes:
I suspect if you make a whole new `w3` instance it might work, instead of copying. But curious to hear if it doesn't

simone1999:
I too expect a new web3 instance to work. Sadly this does not work in my individual case as initializing a new instance of my custom web3.py subclass takes quite a bit. So trying to figure out which components I need to shallow copy to allow for thread save batching.

stas:
> I understand that's not ideal but it should technically be able to get you unstuck for now (I hope).

[...]

PR #3661, opened 2025-04-12
somewhat fixed Batching thread safety

simone1999:
The batching flag is set on the class, instead of likely as intended as an instance attribute. The style in which the attribute was set is common for other programming languages, but in Python this results in class level attributes, which are shared across all instances of that class. In this case the result was that everything breaks once web3.py was used in an concurrent (multi threaded) set up while using the new batching functionality. Even creating multiple w3 instances did not solve the issue as the batching flag was shared across all instances of JSONBaseProvider. There still is work required to make the batching feature thread save, but with this fix it should at least be possible to create multiple w3 instances to use each in it's own thread.
This is a step towards fixing #3613
Also the Async provider likely needs similar modifications.

### DIFF
--- a/newsfragments/3661.bugfix.rst
+++ b/newsfragments/3661.bugfix.rst
@@ -1,0 +1,1 @@
+Moves base providers' ``_is_batching`` and ``_batch_request_func_cache`` from class to instance attrs to help with thread safety.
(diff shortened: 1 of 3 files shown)


fselmo:
Yes, this is definitely one issue that's going on. Thanks for getting this started @simone1999 👍🏼

fselmo (approved):
lgtm 👍🏼 thanks again @simone1999!

callmephilip:
this is awesome. thanks @simone1999 🚀

PR #3663, opened 2025-04-14
check _request_information_cache value is a dict before .get-ing from it

### LINKED ISSUE #3642: 7.9.0: Can't make batch requests without `provider._is_batching`
### What happened?

in < 7.9.0 you could make both batch and single requests without any fuckery around `_is_batching`

```python
def _raise_stray_errors_from_cache(self) -> None:
        """
        Check the request response cache for any errors not tied to current requests
        and raise them if found.
        """
        if not self._is_batching:
            for (
                response
            ) in self._request_processor._request_response_cache._data.values():
                request = (
                    self._request_processor._request_information_cache.get_cache_entry(
                        generate_cache_key(response["id"])
                    )
                )
                if "error" in response and request is None:
                    # if we find an error response in the cache without a corresponding
                    # request, raise the error
                    validate_rpc_response_and_raise_if_error(
                        response, None, logger=self.logger
                    )
```

### Full error output

```shell
In: web3\providers\persistent\persistent.py
generate_cache_key(response["id"])
                       ~~~~~~~~^^^^^^
TypeError: list indices must be integers or slices, not str
```

revert whatever this is

### web3 Version

_No response_

### Python Version

_No response_

### Operating System

_No response_

### Output from `pip freeze`

```shell

```

pacrob:
### What was wrong?
Mixing batching and non-batching calls causing issues.

Closes #3642 

### How was it fixed?
 - Check that the value from cache is a `dict` before `get`ting from it.
 - Add `try/finally` blocks (via a decorator) to all `make_batch_request` functions to make sure `_is_batching` is `True` when called and reset to `False` afterwards.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/providers/test_http_provider.py
+++ b/tests/core/providers/test_http_provider.py
@@ -132,3 +132,21 @@ def test_http_empty_batch_response(mock_post):
 
     # assert that even though there was an error, we have reset the batching state
     assert not w3.provider._is_batching
+
+
+@patch(
+    "web3._utils.http_session_manager.HTTPSessionManager.make_post_request",
+    new_callable=Mock,
+)
+def test_sync_provider_is_batching_when_make_batch_request(mock_post):
+    def assert_is_batching_and_return_response(*_args, **_kwargs) -> bytes:
+        assert provider._is_batching
+        return b'{"jsonrpc":"2.0","id":1,"result":["0x1"]}'
+
+    provider = HTTPProvider()
+    assert not provider._is_batching
+
+    mock_post.side_effect = assert_is_batching_and_return_response
+
+    provider.make_batch_request([("eth_blockNumber", [])])
+    assert not provider._is_batching
--- a/tests/core/providers/test_websocket_provider.py
+++ b/tests/core/providers/test_websocket_provider.py
@@ -534,3 +534,19 @@ async def test_req_info_cache_size_can_be_set_and_warns_when_full(caplog):
             "behavior. Consider increasing the ``request_information_cache_size`` "
             "on the provider."
         ) in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_raise_stray_errors_from_cache_handles_list_response():
+    provider = WebSocketProvider("ws://mocked")
+    _mock_ws(provider)
+
+    bad_response = [
+        {"id": None, "jsonrpc": "2.0", "error": {"code": 21, "message": "oops"}}
+    ]
+    provider._request_processor._request_response_cache._data["bad_key"] = bad_response
+
+    try:
+        provider._raise_stray_errors_from_cache()
+    except Exception as e:
+        pytest.fail(f"{e}")
--- a/tests/core/providers/test_websocket_provider.py
+++ b/tests/core/providers/test_websocket_provider.py
@@ -537,7 +537,7 @@ async def test_req_info_cache_size_can_be_set_and_warns_when_full(caplog):
 
 
 @pytest.mark.asyncio
-async def test_raise_stray_errors_from_cache_handles_list_response():
+async def test_raise_stray_errors_from_cache_handles_list_response_without_error():
     provider = WebSocketProvider("ws://mocked")
     _mock_ws(provider)
 
@@ -546,7 +546,23 @@ async def test_raise_stray_errors_from_cache_handles_list_response():
     ]
     provider._request_processor._request_response_cache._data["bad_key"] = bad_response
 
-    try:
-        provider._raise_stray_errors_from_cache()
-    except Exception as e:
-        pytest.fail(f"{e}")
+    # assert no errors raised
+    provider._raise_stray_errors_from_cache()
+
+
+@pytest.mark.asyncio
+async def test_websocket_provider_is_batching_when_make_batch_request():
+    def assert_is_batching_and_return_response(*_args, **_kwargs) -> bytes:
+        assert provider._is_batching
+        return b'{"jsonrpc":"2.0","id":1,"result":["0x1"]}'
+
+    provider = WebSocketProvider("ws://mocked")
+    _mock_ws(provider)
+    provider._get_response_for_request_id = AsyncMock()
+    provider._get_response_for_request_id.side_effect = (
+        assert_is_batching_and_return_response
+    )
+
+    assert not provider._is_batching
+    await provider.make_batch_request([("eth_blockNumber", [])])
+    assert not provider._is_batching
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -30,6 +30,7 @@ from web3.types import (
 )
 
 from .._utils.batching import (
+    batching_context,
     sort_batch_response_by_response_ids,
 )
 from .._utils.caching import (
@@ -201,6 +202,7 @@ class IPCProvider(JSONBaseProvider):
         request = self.encode_rpc_request(method, params)
         return self._make_request(request)
 
+    @batching_context
     def make_batch_request(
         self, requests: List[Tuple[RPCEndpoint, Any]]
     ) -> List[RPCResponse]:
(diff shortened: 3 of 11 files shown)


web3/providers/persistent/persistent.py
```diff
@@ -318,17 +318,16 @@ def _raise_stray_errors_from_cache(self) -> None:
             for (
                 response
             ) in self._request_processor._request_response_cache._data.values():
-                request = (
-                    self._request_processor._request_information_cache.get_cache_entry(
+                if isinstance(response, dict):
```
fselmo, line 327: Looking again, I think this check is fine to keep. But ultimately, the issue seems to be that the provider's `make_batch_request` (or `async_make_batch_request`) is not setting the `_is_batching` state. This state is only being set within the [RequestBatcher](https://github.com/ethereum/web3.py/blob/6279d995b9f0a005eb15fe3ba9c66638144b2d2d/web3/_utils/batching.py#L102-L103).

Since those batch methods are public methods, and we should make sure we are within the batching state when we make those calls, we should ensure that before making the request, the state is set to `_is_batching=True`, and that, whether or not the call succeeds, we reset the state to `False`.

So I think a `try / finally` around the actual request seems appropriate.

Does that make sense? That would ultimately have resolved this issue and we shouldn't need to check that the response is a dict if we are not in a batching state (which we already had the check for in this code block).

pacrob:
@fselmo I've added the `try/finally` as suggested. Good as is, or is there a reason to wrap just the sending of the request itself?

I applied it across all instances of `make_batch_request` for consistency. Is there a reason to only apply it to the persistent provider?

web3/providers/ipc.py
```diff
@@ -204,10 +204,14 @@ def make_request(self, method: RPCEndpoint, params: Any) -> RPCResponse:
     def make_batch_request(
         self, requests: List[Tuple[RPCEndpoint, Any]]
     ) -> List[RPCResponse]:
-        self.logger.debug(f"Making batch request IPC. Path: {self.ipc_path}")
-        request_data = self.encode_batch_rpc_request(requests)
-        response = cast(List[RPCResponse], self._make_request(request_data))
-        return sort_batch_response_by_response_ids(response)
+        self._is_batching = True
```
fselmo: What do you think about creating a decorator that we can use on these methods that does the `try / finally` so we don't have to repeat this code all around? Just a thought, something like `@batching_context` / `@async_batching_context`?

tests/core/providers/test_websocket_provider.py
```diff
@@ -534,3 +534,19 @@ async def test_req_info_cache_size_can_be_set_and_warns_when_full(caplog):
             "behavior. Consider increasing the ``request_information_cache_size`` "
             "on the provider."
         ) in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_raise_stray_errors_from_cache_handles_list_response():
```
fselmo: Is this testing that we _don't_ raise errors from the cache if it's in a list form (batch request)? I think that's the case right? If so, we should clarify in the test name.

And we can remove the try / except because that would just raise an error naturally. Maybe have a comment saying 

```py
# assert no errors raised
provider._raise_stray_errors_from_cache()
```

fselmo (commented):
Looks good. I wanted to suggest the decorator as I think that could DRY things up and reduce common surface area to one place.

>  Is there a reason to only apply it to the persistent provider?

This should be applied to all providers that support batching (as you did) 👍🏼

tests/core/providers/test_websocket_provider.py
pacrob: Name updated to show that the function handles a list response without error and simplified logic.

web3/providers/ipc.py
pacrob: done!

web3/providers/async_base.py
```diff
@@ -66,6 +71,28 @@
     )
 
 
+class SupportsBatching(Protocol):
```
fselmo: This works well. It makes me think maybe we should have a base class for that all providers originally stem from before deviating into sync + async behavior, that houses all the shared attributes - just like we do with `BaseWeb3`, `BaseEth`, etc... Then we wouldn't need to create this ghost type.

That's a stray thought for another time. This looks good 👍🏼 

My one comment would be to move this and the sync decorator into `web3._utils.batching`, that way we only define these types and the `SupportsBatching` class once and share it. Plus, that would be a better place to house this logic imo.

web3/providers/base.py
```diff
@@ -52,6 +57,28 @@
     from web3 import Web3  # noqa: F401
 
 
+class SupportsBatching(Protocol):
+    _is_batching: bool
```
fselmo: See my comment above, we shouldn't define this twice. Was thinking we should move this into `web3._utils.batching.py` along with the async version.

web3/providers/base.py
```diff
@@ -52,6 +57,28 @@
...
+
+
+R = TypeVar("R")
+T = TypeVar("T", bound=SupportsBatching)
+
+
+def batching_guard(
```
fselmo: nit 

Curious on the "guard" here. We're setting the batching state with this wrapper, I think "guard" makes sense in our current context because I forgot to account for it in the original implementation (😅). I think something like `@batching_context`, `@batching_state`, `@set_batch_state`, or something more along the lines of a state context, makes more sense for the name. Thoughts?

fselmo (approved):
Code looks good, just some refactor ideas to think about. It would also be good to have a simple test that actually checks this state. Given that we also do the dict check (the first commit in this PR), I don't think any tests actually fail without this implementation and then pass after it is implemented (correct me if I'm wrong here).

It doesn't have to be a super complex test but something that tracks this change in case we ever meddle with it and accidentally turn it off would be ideal.

web3/providers/base.py
pacrob: I picked guard because it guards against `_is_batching` being in the wrong state.  Really no strong feelings - which would you prefer?

fselmo: It's a nit but I'd lean toward a state change word choice. I'm not the best at these, maybe we can get a @kclowes opinion 😄

kclowes: chatgpt suggests `@with_batching`. I'd vote for either that or `@batching_context`

pacrob: renamed to `batching_context`

pacrob:
> something that tracks this change 

Tests added for all providers (except legacy websocket), and verified a failure when the decorator is commented out. 

I decided to not test legacy websocket because even when I got its test to pass, I couldn't get rid of the `unclosed event loop` warnings, plus its being due to be removed sooner than later. Let me know if you'd like me to take another go at it.

tests/core/providers/test_http_provider.py
```diff
@@ -132,3 +135,16 @@ def test_http_empty_batch_response(mock_post):
 
     # assert that even though there was an error, we have reset the batching state
     assert not w3.provider._is_batching
+
+
+@pytest.mark.parametrize("initial_is_batching", [True, False])
+def test_http_provider_is_batching_is_reset_to_false_after_batching_call(
```
fselmo: What do you think of a test that checks both on and off behavior? Something like this:

```py
@patch(
    "web3._utils.http_session_manager.HTTPSessionManager.make_post_request",
    new_callable=Mock,
)
def test_http_provider_is_batching_when_make_batch_request(mock_post):
    def assert_is_batching_and_return_response(*_args, **_kwargs) -> bytes:
        assert provider._is_batching
        return b'{"jsonrpc":"2.0","id":1,"result":["0x1"]}'

    provider = HTTPProvider()
    assert not provider._is_batching

    mock_post.side_effect = assert_is_batching_and_return_response

    provider.make_batch_request([("eth_blockNumber", [])])
    assert not provider._is_batching
```

That way we check that it turns on while we are making the post request and that it turns back off after the call. We could add a similar test across all the providers.

pacrob: OK, all updated to the new pattern. Thanks!

fselmo:
lgtm once tests pass 👌🏼 (all good on the legacy ws, I agree on the deprecation reasoning)

Issue #3626, opened 2025-02-24
Add ABI properties as attributes of ContractFunction, ContractEvent


reedsa:
### What feature should we add?

There are a few lingering properties in the ABI spec that would be nice to use directly through the Contract classes.

ContractFunction - `stateMutability`, `payable` and `outputs`
ContractEvent - `anonymous`

For example, `contract.functions.myFunc.abi["stateMutability"] != "payable"` would also be available as `contract.functions.myFunc.state_mutability`.

PR #3664, opened 2025-04-17
feat: Add abi property for ContractFunction

tibor-reiss:
### What was wrong?
Before: `foo_contract.functions.setBar.abi["stateMutability"] == "nonpayable"`
After: `foo_contract.functions.setBar.state_mutability == "nonpayable"`

Related to Issue #3626 

### How was it fixed?
Add state_mutability property.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/web3/contract/base_contract.py
+++ b/web3/contract/base_contract.py
@@ -22,6 +22,8 @@ from eth_abi.exceptions import (
 )
 from eth_typing import (
     ABI,
+    ABIComponent,
+    ABIComponentIndexed,
     ABIElement,
     ABIEvent,
     ABIFunction,
@@ -218,7 +220,7 @@ class BaseContractEvent:
         return self.abi.get("anonymous")
 
     @property
-    def inputs(self) -> Optional[List]:
+    def inputs(self) -> Optional[Sequence["ABIComponentIndexed"]]:
         return self.abi.get("inputs")
 
     @combomethod
@@ -617,11 +619,11 @@ class BaseContractFunction:
         return "function"
 
     @property
-    def inputs(self) -> Optional[List]:
+    def inputs(self) -> Optional[Sequence["ABIComponent"]]:
         return self.abi.get("inputs")
 
     @property
-    def outputs(self) -> Optional[List]:
+    def outputs(self) -> Optional[Sequence["ABIComponent"]]:
         return self.abi.get("outputs")
 
     @combomethod
--- a/web3/contract/base_contract.py
+++ b/web3/contract/base_contract.py
@@ -209,6 +209,18 @@ class BaseContractEvent:
             self._topic = encode_hex(keccak(text=self.signature))
         return self._topic
 
+    @property
+    def type(self) -> str:
+        return "event"
+
+    @property
+    def anonymous(self) -> Optional[bool]:
+        return self.abi.get("anonymous")
+
+    @property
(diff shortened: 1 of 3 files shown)


web3/contract/base_contract.py
```diff
@@ -596,6 +596,12 @@ def __init__(self, abi: Optional[ABIFunction] = None) -> None:
         self.argument_names = tuple([input.get("name", None) for input in event_inputs])
         self.argument_types = tuple([input["type"] for input in event_inputs])
 
+    @property
+    def state_mutability(self) -> Optional[str]:
+        if self.abi is None:
+            return None
```
tibor-reiss: This or `raise Web3TypeError`?

web3/contract/base_contract.py
```diff
@@ -596,6 +596,12 @@ def __init__(self, abi: Optional[ABIFunction] = None) -> None:
         self.argument_names = tuple([input.get("name", None) for input in event_inputs])
         self.argument_types = tuple([input["type"] for input in event_inputs])
 
+    @property
+    def state_mutability(self) -> Optional[str]:
+        if self.abi is None:
```
tibor-reiss: On l595, there is `self.abi.get(...)` which will fail if `abi == None`. Is this intended please? Because if we accept that abi cannot be None, I do not need this check here.

tibor-reiss:
Hi @reedsa, is this implementation on the right track please? If yes, I can try to add the others, except `payable`, because that is deprecated according to https://github.com/ethereum/eth-typing/blob/main/eth_typing/abi.py#L78.

web3/contract/base_contract.py
reedsa: Great callout here, it is expected that the class instance will always have a value for `abi`. This works because the contract initialization uses the ContractFunction factory method. I think it should be fine to remove this check and implement better error handling in the `__init__` function itself.

Right now, if someone attempts to create a `ContractFunction()` without passing an abi, it will throw an `AttributeError`. While tangential to your change, it would be better to check if `self.abi is None` and raise a `Web3AttributeError` at that time.

reedsa:
> Hi @reedsa, is this implementation on the right track please? If yes, I can try to add the others, except `payable`, because that is deprecated according to https://github.com/ethereum/eth-typing/blob/main/eth_typing/abi.py#L78.

Looking good! Appreciate your contribution and it will be great to see all the other properties added.

tibor-reiss:
> > Hi @reedsa, is this implementation on the right track please? If yes, I can try to add the others, except `payable`, because that is deprecated according to https://github.com/ethereum/eth-typing/blob/main/eth_typing/abi.py#L78.
> 
> Looking good! Appreciate your contribution and it will be great to see all the other properties added.

Hi @reedsa, thanks for feedback, I added the non-deprecated attributes.

The failing tests work locally - could this be a CI problem?

web3/contract/base_contract.py
tibor-reiss: This seems to work in `__init__` - let me know please if I should include it in this PR:
```
if abi:
    self.abi = abi
if self.abi is None:
    raise Web3AttributeError
```

BobTheBuidler:
@tibor-reiss I had similar failing tests before #3665 and #3666 were merged but after rebasing my branch on their changes the failures went away

PR #3666, opened 2025-04-17
Limit hypothesis integers in contract data and topic tests to a max value of MAX_UINT_256

kclowes:
### What was wrong?

Tests on v6 were failing due to a hypothesis upgrade. 

### How was it fixed?

Added a max_value for hypothesis integers in filtering tests.

### Todo:

- [x] Clean up commit history
- [x] Add or update documentation related to these changes
- [x] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/main/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3667, opened 2025-04-17
feat: optimize utility_methods.py

BobTheBuidler:
This was originally part of my other PR but that is becoming quite cluttered and hard to review so I separated this one out. 

### What was wrong?
Previously, the code would create a new dict from `d` one time for each item in the checked iterable. Now the code creates just one new dict from `d` at the beginning and uses that same dict to check membership for each item.

Related to Issue #N/A
Closes #N/A

### How was it fixed?
Call `dict` one time prior to entering the genexp, instead of calling it repeatedly inside.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3667.performance.rst
+++ b/newsfragments/3667.performance.rst
@@ -1,0 +1,4 @@
+Optimize performance for:
+web3._utils.utility_methods.all_in_dict
+web3._utils.utility_methods.any_in_dict
+web3._utils.utility_methods.none_in_dict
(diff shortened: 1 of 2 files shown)


BobTheBuidler:
Now that you've merged #3665 and #3666 I rebased my branch on master and reran the tests, looks like everything passes now

The failure in py310-integration-goethereum-ws was due to a timeout and seems to be unrelated to the PR

pacrob (approved):
Thanks!

PR #3669, opened 2025-04-17
fix(types): update type hints in web3._utils.method_formatters.py

BobTheBuidler:
### What was wrong?
`get_result_formatters` had return type `Dict[str, Callable[..., Any]]` which is incorrect because the function returns a callable `toolz.compose` object

`get_error_formatters`, and `get_null_result_formatters` both had return type `Callable[..., Any]` which is techincally correct but could be made more specific for clarity for contributors.

Related to Issue #N/A
Closes #N/A

### How was it fixed?
I've changed the return type for all 3 functions to `Callable[[RPCResponse], Any]`.

### Todo:

- [x] Clean up commit history

I did not add an entry to [release notes](https://github.com/ethereum/web3.py/blob/main/newsfragments/README.md) as this change is not relevant for users.

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/method_formatters.py
+++ b/web3/_utils/method_formatters.py
@@ -105,6 +105,7 @@ from web3.types import (
     BlockIdentifier,
     Formatters,
     RPCEndpoint,
+    RPCResponse,
     SimulateV1Payload,
     StateOverrideParams,
     TReturn,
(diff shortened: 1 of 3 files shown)


web3/_utils/method_formatters.py
```diff
@@ -1214,7 +1211,7 @@ def get_result_formatters(
 
 def get_error_formatters(
     method_name: Union[RPCEndpoint, Callable[..., RPCEndpoint]]
```
BobTheBuidler: Is method_name actually supposed to be a `Union[RPCEndpoint, Callable[..., RPCEndpoint]]`?

I notice `Union[RPCEndpoint, Callable[..., RPCEndpoint]]` is used throughout the codebase but the code itself seems to assume method_name is always a string value, never callable. Should this be updated as well?

BobTheBuidler:
Now that you've merged #3665 and #3666 I rebased my branch on master and reran the tests, looks like everything passes now

web3/_utils/method_formatters.py
```diff
@@ -1190,7 +1187,7 @@ def filter_wrapper(
 
 @to_tuple
 def apply_module_to_formatters(
-    formatters: Tuple[Callable[..., TReturn]],
```
kclowes, line 1244: Curious about the reasoning behind this change?

web3/_utils/method_formatters.py
kclowes: Oh, nice catch. Yeah, these can only be RPCEndpoints once they're here.

web3/_utils/method_formatters.py
BobTheBuidler: This change just indicates to mypy that the function will work with any iterable of callables.

The original hint indicates that the function must take a tuple input, so while it didn't cause any type errors internally here it can cause unneeded type errors for downstream users.

BobTheBuidler:
@kclowes this one is good to go, and will help clear up some of the typing issues with #3670 once merged

kclowes (approved):
This looks good to me, thank you! Will add a newsfragment and merge!

web3/method.py
```diff
@@ -182,7 +182,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
     @property
     def method_selector_fn(
         self,
-    ) -> Callable[..., Union[RPCEndpoint, Callable[..., RPCEndpoint]]]:
+    ) -> Callable[..., RPCEndpoint]:
```
fselmo, line 185: 👌🏼

Issue #3534, opened 2024-11-20
A transaction class that can encode and decode from an RPC transaction dict


fselmo:
### What feature should we add?

There should be an easy way to take a transaction directly as a JSON-RPC dict and put it into an object that can `encode()` and `decode()` the transaction. This would behave much like the `TypedTransaction` class does for typed transactions in *eth-account* - and in fact this class can probably carry a lot of the heavy load here - but it also needs to clean up certain fields based on the type in order for it to even be plugged into the `TypedTransaction` class.

I'm thinking something like:

```py
>>> from web3.utils import Web3Transaction

>>> txn = Web3Transaction.from_dict(w3.eth.get_transaction("0x..."))
>>> txn.encode().hex()
02ed01808504a817c8008504a817c800830186a09400000000000000000000000000000000000000008080c0808080...

>>> # maybe...
>>> txn.max_fee_per_gas
20000000000

>>> # etc... ???
```

Or, maybe it's a `HexBytes` as the result of `encode()`. Either way, I think the utility here is fairly clear. Take a transaction straight from a result, put it into a utility class, and be able to manipulate and have useful methods available on the transaction.

PR #3673, opened 2025-04-20
feat: Add Web3Transaction

tibor-reiss:
### What was wrong?
"There should be an easy way to take a transaction directly as a JSON-RPC dict..."

Related to Issue #3534

### How was it fixed?
Add Web3Transaction thin wrapper for better UX.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/utilities/test_transaction_utils.py
+++ b/tests/core/utilities/test_transaction_utils.py
@@ -1,0 +1,111 @@
+import pytest
+
+from hexbytes import (
+    HexBytes,
+)
+
+from web3.utils.transaction import (
+    Web3Transaction,
+)
+
+ACCESS_LIST_TRANSACTION_TEST_CASE = {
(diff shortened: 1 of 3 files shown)


tibor-reiss:
Hi @fselmo, could you please check if this is according to your vision?
I made a single class with ValueErrors. Other option would be to have a class hierarchy with `Web3DynamicFeeTransaction`, `Web3AccessListTransaction`, ...

BobTheBuidler:
might I recommend defining __slots__ to make the wrapper even thinner?

tibor-reiss:
> might I recommend defining **slots** to make the wrapper even thinner?

Hi @BobTheBuidler, could you please provide some example? E.g. should only `typed_transaction` be part of slots, or all the attributes from `_dict`? To me, this would make sense with a class hierarchy, because some of the attributes do not exist on specific types. Let me what you think.

BobTheBuidler:
I was just thinking about `__slots__ = ("typed_transaction, "_dict")`, not necessarily slots for each attr

Though that *would* be even thinner if you did it!

fselmo:
Thanks for the brainstorming here @tibor-reiss 🙂 . I'd like to think on this some more and how we could maybe leverage this transaction API as more of a "first class citizen" of the library, maybe as a pydantic model for the next major version - possibly replacing `TxParams` altogether. That may require quite a bit of work though. This isn't a priority at the moment but I think it's good to get the conversation going on it at least, so thanks for getting it started 👍🏼

PR #3684, opened 2025-04-25
 Fix list numbering in Windows setup instructions

davidjsonn:
### What was wrong?
Change second step label from `1.` to `2.` for correct sequence


kclowes:
This displays correctly when rendered.

PR #3685, opened 2025-04-26
Deprecate eth.get_uncle* methods

### LINKED ISSUE #3683: Deprecate get_uncle* endpoints

Use the `DeprecatedMethod` class to communicate to users that these methods will be removed in v8.

tibor-reiss:
### What was wrong?

Closes #3683

### How was it fixed?
Warn users that eth.get_uncle* will be dropped in v8.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

t18AHHmB

### DIFF
--- a/docs/web3.eth.rst
+++ b/docs/web3.eth.rst
@@ -398,6 +398,8 @@ The following methods are available on the ``web3.eth`` namespace.
 
 .. py:method:: Eth.get_uncle_by_block(block_identifier, uncle_index)
 
+    .. warning:: Deprecated. Will be removed in v8.
+
     * Delegates to ``eth_getUncleByBlockHashAndIndex`` or
       ``eth_getUncleByBlockNumberAndIndex`` RPC methods
 
@@ -443,6 +445,8 @@ The following methods are available on the ``web3.eth`` namespace.
 
 
 .. py:method:: Eth.get_uncle_count(block_identifier)
+
+    .. warning:: Deprecated. Will be removed in v8.
 
     * Delegates to ``eth_getUncleCountByBlockHash`` or
       ``eth_getUncleCountByBlockNumber`` RPC methods
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -2807,18 +2807,24 @@ class EthModuleTest:
     def test_eth_getUncleCountByBlockHash(
         self, w3: "Web3", empty_block: BlockData
     ) -> None:
-        uncle_count = w3.eth.get_uncle_count(empty_block["hash"])
+        with pytest.warns(
+            DeprecationWarning, match=r"All get_uncle\* methods have been deprecated"
+        ):
+            uncle_count = w3.eth.get_uncle_count(empty_block["hash"])
 
-        assert is_integer(uncle_count)
-        assert uncle_count == 0
+            assert is_integer(uncle_count)
+            assert uncle_count == 0
 
     def test_eth_getUncleCountByBlockNumber(
         self, w3: "Web3", empty_block: BlockData
     ) -> None:
-        uncle_count = w3.eth.get_uncle_count(empty_block["number"])
+        with pytest.warns(
+            DeprecationWarning, match=r"All get_uncle\* methods have been deprecated"
+        ):
+            uncle_count = w3.eth.get_uncle_count(empty_block["number"])
 
-        assert is_integer(uncle_count)
-        assert uncle_count == 0
+            assert is_integer(uncle_count)
+            assert uncle_count == 0
 
     def test_eth_get_code(
         self, w3: "Web3", math_contract_address: ChecksumAddress
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -2172,19 +2172,31 @@ class AsyncEthModuleTest:
     async def test_eth_getUncleCountByBlockHash(
         self, async_w3: "AsyncWeb3", async_empty_block: BlockData
     ) -> None:
-        uncle_count = await async_w3.eth.get_uncle_count(async_empty_block["hash"])
+        with pytest.warns(
+            DeprecationWarning,
+            match=r"get_uncle_count is deprecated: all get_uncle\* "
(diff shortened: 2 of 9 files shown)


tibor-reiss:
wdyt @kclowes?

docs/web3.eth.rst
```diff
@@ -397,6 +397,7 @@ The following methods are available on the ``web3.eth`` namespace.
 
 
 .. py:method:: Eth.get_uncle_by_block(block_identifier, uncle_index)
+    .. warning:: Deprecated. Will be removed in v8.
```
kclowes, line 401: This needs a newline above to make it render correctly
```suggestion

    .. warning:: Deprecated. Will be removed in v8.
```

docs/web3.eth.rst
```diff
@@ -443,6 +444,7 @@ The following methods are available on the ``web3.eth`` namespace.
 
 
 .. py:method:: Eth.get_uncle_count(block_identifier)
+    .. warning:: Deprecated. Will be removed in v8.
```
kclowes, line 449: Same here, needs a newline above.

web3/_utils/module_testing/eth_module.py
```diff
@@ -2807,18 +2807,20 @@ def test_eth_getBlockTransactionCountByNumber_block_with_txn(
     def test_eth_getUncleCountByBlockHash(
         self, w3: "Web3", empty_block: BlockData
     ) -> None:
-        uncle_count = w3.eth.get_uncle_count(empty_block["hash"])
+        with pytest.warns(DeprecationWarning):
```
kclowes: We like to match on the message for deprecation warnings, so this would just need a `match=r"All get_uncle\* methods have been deprecated"`

web3/_utils/module_testing/eth_module.py
```diff
@@ -2807,18 +2807,20 @@ def test_eth_getBlockTransactionCountByNumber_block_with_txn(
...
-        assert is_integer(uncle_count)
-        assert uncle_count == 0
+            assert is_integer(uncle_count)
+            assert uncle_count == 0
 
     def test_eth_getUncleCountByBlockNumber(
         self, w3: "Web3", empty_block: BlockData
     ) -> None:
-        uncle_count = w3.eth.get_uncle_count(empty_block["number"])
+        with pytest.warns(DeprecationWarning):
```
kclowes: Same here, add match keyword

kclowes (commented):
Thank you for the PR! There were just a few nits that I left that should be addressed. We'll also need to add the warning to the AsyncEth module as well. Let me know if you want me to pick it up.

tibor-reiss:
> Thank you for the PR! There were just a few nits that I left that should be addressed. We'll also need to add the warning to the AsyncEth module as well. Let me know if you want me to pick it up.

Thanks for the feedback. Idk how I missed the async method, my bad. Is the double underscore version ok, or shall I just add the warning into the async function please?

Afais the failing tests are not connected to the change, but let me know if otherwise.

web3/eth/async_eth.py
```diff
@@ -660,14 +661,20 @@ async def sign_typed_data(
...
         mungers=[default_root_munger],
     )
+    _get_uncle_count = DeprecatedMethod(
+        __get_uncle_count,
+        old_name="__get_uncle_count",
+        new_name="_get_uncle_count",
+        msg="All get_uncle* methods have been deprecated",
+    )
 
     async def get_uncle_count(self, block_identifier: BlockIdentifier) -> int:
```
kclowes, line 676: I actually think that instead of using the `DeprecatedMethod` class here, we can write a new `async_deprecated_for` method and use that as a decorator instead of the double underscore above. The sync `deprecated_for` can be found [here](https://github.com/ethereum/web3.py/blob/main/web3/_utils/decorators.py#L42-L63). I didn't test this, but I think the function should look something like: 
```
def async_deprecated_for(msg):
    def decorator(to_wrap):
        @wraps(to_wrap)
        async def wrapper(*args, **kwargs):
            warnings.warn(
                msg,
                category=DeprecationWarning,
                stacklevel=2,
            )
            return await to_wrap(*args, **kwargs)

        return wrapper

    return decorator
```

kclowes (commented):
Thanks for the updates! I left a comment on the async methods to give you some direction there. Let me know if that doesn't work, or if this is turning into more work than you signed up for and you want to hand it off!

web3/eth/async_eth.py
tibor-reiss: Sounds good to me, I have replaced the class with the decorator. Since it is a removal and not replacement, I added the custom message as optional, similar to the cherry picked commit. But this feels clunky: 2 args instead of 1. Would you be happy with the following solution?

```
def async_deprecated_for(msg: str):
    ...
    message = f"{to_wrap.__name__} is deprecated: {msg}"
    ...

@async_deprecated_for("all get_uncle* methods have been deprecated")  # for removal
async def get_uncle_count(self, block_identifier: BlockIdentifier) -> int:

@async_deprecated_for("use get_aunt_count instead")  # for replacement
async def get_uncle_count(self, ...

```

If this is something which you like, I would also adjust the sync decorator to be consistent. WDYT?

kclowes: I think we can just go with a singular message arg and then whatever message can get passed in. If we start to use it somewhere else and we want to standardize the message or get a little more clever, we can update it. Thanks!

kclowes:
Actually, I'm looking at this, and I think we can just use the `deprecated_for` decorator. Sorry for the misdirection. If you want to be fancy, you could update the message so that it makes a little more sense, but I'm also fine if you just do something like: `@deprecated_for('nothing')`.

tibor-reiss:
> Actually, I'm looking at this, and I think we can just use the `deprecated_for` decorator. Sorry for the misdirection. If you want to be fancy, you could update the message so that it makes a little more sense, but I'm also fine if you just do something like: `@deprecated_for('nothing')`.

No problem, I have replaced now with `deprecated_for`.

M66rr (approved):

kclowes (approved):
Looks good to me! Thanks @tibor-reiss!

M66rr (commented):
Done ✅

PR #3695, opened 2025-05-09
Address flaky tests for now

fselmo:
### What was wrong?

Use `blockNumber` for ``eth_getCode`` and ``eth_call``

  - For now, use the `blockNumber` from the transaction receipt to make ``eth_getCode`` and ``eth_call`` calls. Recent geth versions don't seem to be returning the latest state when called with ``latest`` as the block identifier.
  
  - Added TODOs to remove the `blockNumber` workaround after this is fixed.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-05-09 at 13 12 25

### DIFF
--- a/newsfragments/3695.internal.rst
+++ b/newsfragments/3695.internal.rst
@@ -1,0 +1,1 @@
+Address some flaky tests due to a geth bug in state synchronization.
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -723,6 +723,8 @@ class AsyncEthModuleTest:
         keyfile_account_pkey: HexStr,
         async_math_contract: "AsyncContract",
     ) -> None:
+        # TODO: remove blockNumber block_id from eth_call and eth_getCode calls once
+        #  geth behavior for "latest" seems stable again.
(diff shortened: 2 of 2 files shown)


fselmo:
I will run the CI a few more times but these are the tests that got flaky and seem to be stable now

fselmo:
OK changes look good with 5 CI runs. I'm actually going to try to just raise the `dev.period` by `1` to see if the `wait_for_transaction_receipt` test is less flaky. I think the issues we're seeing are related to block processing times so if that works it would be better than `xfail(strict=False)`.

I will add the change as a commit on top of the current state of things so that if we need to drop that commit we know that the CI is good at commit 8069b5f07939b91929e9528d19319130b4a72ced.

fselmo:
No such luck. I dropped that commit attempt. It looks like from time to time we may have some [issues with logs](https://app.circleci.com/pipelines/github/ethereum/web3.py/8511/workflows/e9f97c38-4bcf-40bb-a08f-7ffb2119deaa/jobs/637981) too, more rarely. Something to keep an eye on.

kclowes (approved):
🚢

PR #3697, opened 2025-05-14
feat: optimize map_abi_data

BobTheBuidler:
### What was wrong?

Title says it all, this is ready for review.

Related to Issue #
Closes #

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/web3/_utils/abi.py
+++ b/web3/_utils/abi.py
@@ -73,7 +73,6 @@ from eth_utils import (
 )
 from eth_utils.toolz import (
     curry,
-    partial,
     pipe,
 )
 
@@ -115,21 +114,21 @@ def receive_func_abi_exists(contract_abi: ABI) -> Sequence[ABIReceive]:
     return filter_abi_by_type("receive", contract_abi)
 
 
-def get_indexed_event_inputs(event_abi: ABIEvent) -> Sequence[ABIComponentIndexed]:
+def get_indexed_event_inputs(event_abi: ABIEvent) -> List[ABIComponentIndexed]:
     return [arg for arg in event_abi["inputs"] if arg["indexed"] is True]
 
 
-def exclude_indexed_event_inputs(event_abi: ABIEvent) -> Sequence[ABIComponentIndexed]:
+def exclude_indexed_event_inputs(event_abi: ABIEvent) -> List[ABIComponentIndexed]:
     return [arg for arg in event_abi["inputs"] if arg["indexed"] is False]
 
 
-def filter_by_types(types: Collection[str], contract_abi: ABI) -> Sequence[ABIElement]:
+def filter_by_types(types: Collection[str], contract_abi: ABI) -> List[ABIElement]:
     return [abi_element for abi_element in contract_abi if abi_element["type"] in types]
 
 
 def filter_by_argument_name(
     argument_names: Collection[str], contract_abi: ABI
-) -> Sequence[ABIElement]:
+) -> List[ABIElement]:
     """
     Return a list of each ``ABIElement`` which contains arguments matching provided
     names.
@@ -186,7 +185,7 @@ def get_name_from_abi_element_identifier(
 
 def get_abi_element_signature(
     abi_element_identifier: ABIElementIdentifier,
-    abi_element_argument_types: Optional[Sequence[str]] = None,
+    abi_element_argument_types: Optional[Iterable[str]] = None,
 ) -> str:
     element_name = get_name_from_abi_element_identifier(abi_element_identifier)
     argument_types = ",".join(abi_element_argument_types or [])
@@ -585,9 +584,9 @@ def normalize_event_input_types(
 
 @curry
 def map_abi_data(
-    normalizers: Sequence[Callable[[TypeStr, Any], Tuple[TypeStr, Any]]],
-    types: Sequence[TypeStr],
-    data: Sequence[Any],
+    normalizers: Iterable[Callable[[TypeStr, Any], Tuple[TypeStr, Any]]],
+    types: Iterable[TypeStr],
+    data: Iterable[Any],
 ) -> Any:
     """
     Applies normalizers to your data, in the context of the relevant types.
@@ -611,17 +610,21 @@ def map_abi_data(
     2. Recursively mapping each of the normalizers to the data
     3. Stripping the types back out of the tree
     """
-    pipeline = itertools.chain(
-        [abi_data_tree(types)],
-        map(data_tree_map, normalizers),
-        [partial(recursive_map, strip_abi_type)],
+    return pipe(
+        data,
+        # 1. Decorating the data tree with types
+        abi_data_tree(types),
+        # 2. Recursively mapping each of the normalizers to the data
+        *map(data_tree_map, normalizers),
+        # 3. Stripping the types back out of the tree
+        strip_abi_types,
     )
-
-    return pipe(data, *pipeline)
 
 
 @curry
-def abi_data_tree(types: Sequence[TypeStr], data: Sequence[Any]) -> List[Any]:
+def abi_data_tree(
+    types: Iterable[TypeStr], data: Iterable[Any]
+) -> List["ABITypedData"]:
     """
     Decorate the data tree with pairs of (type, data). The pair tuple is actually an
     ABITypedData, but can be accessed as a tuple.
@@ -631,10 +634,7 @@ def abi_data_tree(types: Sequence[TypeStr], data: Sequence[Any]) -> List[Any]:
     >>> abi_data_tree(types=["bool[2]", "uint"], data=[[True, False], 0])
     [("bool[2]", [("bool", True), ("bool", False)]), ("uint256", 0)]
     """
-    return [
-        abi_sub_tree(data_type, data_value)
-        for data_type, data_value in zip(types, data)
-    ]
+    return list(map(abi_sub_tree, types, data))
 
 
 @curry
@@ -721,6 +721,10 @@ def strip_abi_type(elements: Any) -> Any:
         return elements.data
     else:
         return elements
+
+
+def strip_abi_types(elements: Any) -> Any:
+    return recursive_map(strip_abi_type, elements)
 
 
 def build_non_strict_registry() -> ABIRegistry:
--- a/newsfragments/3697.performance.rst
+++ b/newsfragments/3697.performance.rst
@@ -1,0 +1,1 @@
+optimize map_abi_data


web3/_utils/abi.py
```diff
@@ -115,21 +114,21 @@ def receive_func_abi_exists(contract_abi: ABI) -> Sequence[ABIReceive]:
     return filter_abi_by_type("receive", contract_abi)
 
 
-def get_indexed_event_inputs(event_abi: ABIEvent) -> Sequence[ABIComponentIndexed]:
+def get_indexed_event_inputs(event_abi: ABIEvent) -> List[ABIComponentIndexed]:
```
kclowes, line 117: Even though it doesn't really feel like it, this is a breaking change since we're going to something more specific, so we'll have to wait until v8 for these `Sequence` -> `List` changes

web3/_utils/abi.py
```diff
@@ -611,17 +610,21 @@ def normalizer(datatype, data):
...
+        strip_abi_types,
     )
 
-    return pipe(data, *pipeline)
-
 
 @curry
-def abi_data_tree(types: Sequence[TypeStr], data: Sequence[Any]) -> List[Any]:
+def abi_data_tree(
+    types: Iterable[TypeStr], data: Iterable[Any]
```
kclowes, line 626: I'm curious about these type changes? What's the reasoning for going from `Sequence` -> `Iterable`

web3/_utils/abi.py
```diff
@@ -611,17 +610,21 @@ def normalizer(datatype, data):
...
+) -> List["ABITypedData"]:
```
kclowes, line 627: Unfortunately this will have to wait until v8 too :-/

kclowes (commented):
Thanks for the PR! There are a couple breaking typing changes here that I listed, and a quick benchmark of `test_map_abi_data` shows minimal improvement on only one test case and either equal or worse performance with the other 3, so I'm not quite convinced on this one. If you can show me data that this is more performant though I'd be happy to revisit!

web3/_utils/abi.py
BobTheBuidler: Because Iterable[X] signifies "object with __iter__method that yields X objects" Sequence[X] signifies "Iterable[X], but with __getitem__ defined for indexing" which this function does not require. I think it also signifies a __reversed__ method but I'm not certain and in any case that is not relevant here either. 

Its not exactly 'wrong' but I was taught that type hints for input args should be as permissive as possible when describing what inputs can be passed to a function without breaking it, and type hints for the outputs should be as specific as possible to describe the actual object returned. This helps the type checker find more issues earlier on than it otherwise might be able to. So that's the principle I keep in mind when I write my own type hints or suggest changes of this nature

As an example, when I was writing my iterator PR I was trying to pass a Generator[T] into a function that required an Iterable[T] (ie my input was good) but which had type hints indicating it required a Sequence[T]. A generator is an iterable but is not a sequence, so mypy was screaming at me. But in that same case, the code ran fine and the tests all passed, which means the type hint was incorrectly defined for the function.

web3/_utils/abi.py
BobTheBuidler: Is there an appropriate place for me to make that PR now?

BobTheBuidler:
How are you benchmarking? Is there a standard method that should be used? 

The current code does exactly the same stuff as the new code, same func calls and all, but also creates and populates two lists and creates and iterates thru an intermediary chain object which the new code does not do.

I didn't anticipate you'd see a substantial difference, since this is a microoptimization after all, but a slowdown doesn't make any sense to me when the new PR is "call all the same functions with all the same inputs, minus a few that didn't need calling". We didn't modify the instructions we're sending to the cpu, we just removed a couple.

I'd love to look into this further to figure out what's going on.

web3/_utils/abi.py
BobTheBuidler: On second thought, I would argue that instead of a breaking change, this is a bugfix, just like #3669 . The function returns a list and can only ever return a list. We're not seeking to modify this behavior, only to clarify a fact pattern that already exists in the code for the type checker. Any calling function that expects a Sequence will still pass the type checks, mypy checking will not turn red neither in your repo nor in repos downstream. I'm curious what type of things you had in mind that might break from this change

fselmo: I believe these methods only exist under `/_utils` and are not re-exported anywhere else, so either way it's safe for `v7`. But I do think this counts as a bugfix 👍🏼

web3/_utils/abi.py
BobTheBuidler: I think it might have actually been this function haha

BobTheBuidler:
I did this benchmark with empty args so we're truly comparing apples to apples and not polluting the results with the time it takes to actually format all the data

```python
from time import time
import itertools
from functools import partial
from web3._utils.abi import abi_data_tree, data_tree_map, recursive_map, strip_abi_type
from cytoolz import pipe


def old(data, types, normalizers):
    pipeline = itertools.chain(
        [abi_data_tree(types)],
        map(data_tree_map, normalizers),
        [partial(recursive_map, strip_abi_type)],
    )

    return pipe(data, *pipeline)

def new(data, types, normalizers):
    return pipe(
        data,
        # 1. Decorating the data tree with types
        abi_data_tree(types),
        # 2. Recursively mapping each of the normalizers to the data
        *map(data_tree_map, normalizers),
        # 3. Stripping the types back out of the tree
        strip_abi_types,
    )

def strip_abi_types(elements):
    return recursive_map(strip_abi_type, elements)



for i in range(5):
    normalizers = (int,) * i

    start = time()
    for _ in range(100_000):
        old((), (), normalizers)
    print(f"old: {time() - start}")

    start = time()
    for _ in range(100_000):
        new((), (), normalizers)
    print(f"new: {time() - start}")
```

```
old: 1.5397992134094238
new: 1.442474603652954
old: 2.885958194732666
new: 2.764631509780884
old: 4.140100479125977
new: 4.019676208496094
old: 5.37146520614624
new: 4.730041265487671
old: 5.9497997760772705
new: 5.8652191162109375
```

web3/_utils/abi.py
kclowes: nvm, this is in _utils too. We're good!

web3/_utils/abi.py
kclowes: doh, nice catch!

web3/_utils/abi.py
kclowes: Got it, thanks for the explanation!

kclowes (commented):
We don't have any benchmarking at this granular of a level. I just grabbed the test cases from [this test](https://github.com/ethereum/web3.py/blob/2f3edcc20dc004d1094df0b8f8a8c849b289027d/tests/core/utilities/test_abi.py#L329) and then ran the each test case 1000 times and compared main to this branch. Here's the script I was using: 
```
import time
from typing import List, Tuple, Any, Callable
from web3._utils.abi import map_abi_data
from web3._utils.normalizers import BASE_RETURN_NORMALIZERS, addresses_checksummed, abi_string_to_text

def benchmark_map_abi_data():
    test_cases = [
        # Case 1: Simple bool array and int
        {
            "types": ["bool[2]", "int256"],
            "data": [[True, False], 9876543210],
            "funcs": [
                lambda typ, dat: ((typ, "Tru-dat") if typ == "bool" and dat else (typ, dat)),
                lambda typ, dat: (typ, hex(dat)) if typ == "int256" else (typ, dat),
            ],
        },
        # Case 2: Address normalization
        {
            "types": ["address"],
            "data": ["0x5b2063246f2191f18f2675cedb8b28102e957458"],
            "funcs": BASE_RETURN_NORMALIZERS,
        },
        # Case 3: Address array normalization
        {
            "types": ["address[]"],
            "data": [["0x5b2063246f2191f18f2675cedb8b28102e957458"] * 2],
            "funcs": BASE_RETURN_NORMALIZERS,
        },
        # Case 4: Complex tuple with addresses
        {
            "types": ["(address,address)[]"],
            "data": [[
                (
                    "0x5b2063246f2191f18f2675cedb8b28102e957458",
                    "0xebe0da78ecb266c7ea605dc889c64849f860383f",
                )
            ] * 2],
            "funcs": BASE_RETURN_NORMALIZERS,
        },
        # Case 5: String and address array
        {
            "types": ["(string,address[])"],
            "data": [(
                b"a string",
                [b"\xf2\xe2F\xbbv\xdf\x87l\xef\x8b8\xae\x84\x13\x0fOU\xde9["],
            )],
            "funcs": [addresses_checksummed, abi_string_to_text],
        },
    ]

    for i, test_case in enumerate(test_cases, 1):
        # Warm up
        for _ in range(100):
            map_abi_data(test_case['funcs'], test_case['types'], test_case['data'])
        
        # Actual benchmark
        start_time = time.time()
        for _ in range(1000):
            result = map_abi_data(test_case['funcs'], test_case['types'], test_case['data'])
        end_time = time.time()
        
        total_time = end_time - start_time
        print(f"Total time for 1000 runs: {total_time:.4f} seconds")

if __name__ == "__main__":
    benchmark_map_abi_data()
```

Here is the data I was seeing: 
| Test Case | Types | Main Branch Run Time (s) | Feature Branch Run Time Yesterday (s) | Feature Branch Run Time Today (Run 1) (s) | Feature Branch Run Time Today (Run 2) (s) |
|-----------|-------|--------------------------|--------------------------------------|------------------------------------------|------------------------------------------|
| 1 | ['bool[2]', 'int256'] | 0.0933 | 0.0668 | 0.0907 | 0.0972 |
| 2 | ['address'] | 0.0316 | 0.0321 | 0.0317 | 0.0318 |
| 3 | ['address[]'] | 0.0604 | 0.0607 | 0.0602 | 0.0604 |
| 4 | ['(address,address)[]'] | 0.1217 | 0.1268 | 0.1204 | 0.1200 |
| 5 | ['(string,address[])'] | 0.0855 | 0.0862 | 0.0898 | 0.0847 |

Running a few more times on this branch today, I'm seeing slight improvements on the feature branch numbers above. I think the type changes are good, and I think the performance change is minimal but doesn't impact readability so I'm okay merging but curious if you have an opinion @fselmo?

fselmo (approved):
> curious if you have an opinion `fselmo`?

I like it. Seems to be a slight improvement and doesn't hurt readability. I would certainly like for us to squash the 7 commits on a `+24 -19` before merging though!

kclowes (approved):
Thanks @BobTheBuidler!

kclowes:
Will squash and merge!

BobTheBuidler:
wait wait undo this

https://github.com/ethereum/web3.py/compare/5c0124e921d1c932ede559c9a2b63ca3a3b18f73..af103725241e73820726909944a13edc59c296c7

BobTheBuidler:
That should still be Sequence because it will do a __len__ check at  `abi_element_argument_types or []`

BobTheBuidler:
no I lied, sorry

we're good

PR #3699, opened 2025-05-16
feat: optimize passing data thru formatters

BobTheBuidler:
### What was wrong?

In the current code, data_tree_map is called multiple times in rapid succession. This occurs for every call that is formatted.

data_tree_map creates and populates a new list each time its called, which is quite wasteful and not necessary for our use case, especially when we're calling it recursively

This PR changes the codebase so, instead of creating and populating a new list object at each and every stage of formatting, now we pass `map` objects thru the formatters and only create and populate one list when the formatting is complete. 

The only api that changed was in an internal module, and the codebase was adapted accordingly. This small change makes a much bigger difference than the past few PRs did, so I'm hoping this API change doesn't block us from merging. 

@kclowes you can use the same benchmark script you used for the last PR, you will see a tangible difference on that benchmark and you should also in the overall testing times

Related to Issue #
Closes #

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/utilities/test_abi.py
+++ b/tests/core/utilities/test_abi.py
@@ -262,7 +262,7 @@ def test_get_tuple_type_str_parts(
 def test_abi_data_tree(
     types: List[str], data: Tuple[List[bool], bytes], expected: List[Any]
 ) -> None:
-    assert abi_data_tree(types, data) == expected
+    assert list(abi_data_tree(types, data)) == expected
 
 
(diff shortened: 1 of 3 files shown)


BobTheBuidler:
@kclowes your same benchmark script you used for the last PR will show a tangible difference with this PR

I can put together a description later if you need me to

PR #3703, opened 2025-05-19
feat: cache error and null result formatters

BobTheBuidler:
### What was wrong?

When using the default formatters, we can cache the callable for re-use instead of rebuilding it every time it is used.

Related to Issue #
Closes #

### How was it fixed?

A simple dict-based cache

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


BobTheBuidler:
I didn't implement this for the successful result formatters since the key is a little more complex and I'm not entirely sure how Module works here, but thinking I should do that as well. Thoughts?

PR #3705, opened 2025-05-20
thread safety: context vars for batching context

### LINKED ISSUE #3613: New batching feature completely breaks thread safety

fselmo:
### What was wrong?

Closes #3613

### How was it fixed?

- [bugfix](https://github.com/ethereum/web3.py/pull/3705/commits/2e3559741feba1e305c8ab99ce809de35603900a) Use `ContextVar` rather than global flag to toggle batching context. This should better isolate the batching context to its own thread-safe context.
- Add test for concurrent execution of many requests at the same time as a batch request
- [bugfix](https://github.com/ethereum/web3.py/pull/3705/commits/30f76faa5a0c99c53c96178f48bbd44c0b0f88ff): Found an issue with `PersistentConnectionProvider` batching requests where we hadn't yet separated the send and receive functions to deterministically match responses with their requests by `id`. Fixed that in this PR as well.[](https://github.com/ethereum/web3.py/pull/3705/commits/30f76faa5a0c99c53c96178f48bbd44c0b0f88ff)
- [bugfix](https://github.com/ethereum/web3.py/pull/3705/commits/9cf645dd5cf341810cc8af6b1ec0e8c8dd12ad62): threading id means nothing for async, use loop id
- [test fix](https://github.com/ethereum/web3.py/pull/3705/commits/1542d9df4ac5cf2f0978c05f4d203558705a4245): gather sessions deterministically to check order

---
🎉 🎉  Bonus 🎉 🎉 

I was able to switch our core tests to run via `pytest-xdist` insanely fast by sorting any `set` parametrized params or using `tuple`s / `list`s instead, where appropriate. I am kind of shocked we hadn't gone this route before but this is a pretty significant update.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-05-20 at 18 59 53

### DIFF
--- a/tests/core/providers/test_http_provider.py
+++ b/tests/core/providers/test_http_provider.py
@@ -15,9 +15,6 @@ from web3 import (
     Web3,
     __version__ as web3py_version,
 )
-from web3._utils.batching import (
-    is_batching_context,
-)
 from web3.eth import (
     Eth,
 )
@@ -130,9 +127,9 @@ def test_http_empty_batch_response(mock_post):
     )
     w3 = Web3(HTTPProvider())
     with w3.batch_requests() as batch:
-        assert is_batching_context()
+        assert w3.provider._is_batching
         with pytest.raises(Web3RPCError, match="empty batch"):
             batch.execute()
 
     # assert that even though there was an error, we have reset the batching state
-    assert not is_batching_context()
+    assert not w3.provider._is_batching
--- a/tests/core/providers/test_http_provider.py
+++ b/tests/core/providers/test_http_provider.py
@@ -15,6 +15,9 @@ from web3 import (
     Web3,
     __version__ as web3py_version,
 )
+from web3._utils.batching import (
+    is_batching_context,
+)
 from web3.eth import (
     Eth,
 )
@@ -127,26 +130,9 @@ def test_http_empty_batch_response(mock_post):
     )
     w3 = Web3(HTTPProvider())
     with w3.batch_requests() as batch:
+        assert is_batching_context()
         with pytest.raises(Web3RPCError, match="empty batch"):
             batch.execute()
 
     # assert that even though there was an error, we have reset the batching state
-    assert not w3.provider._is_batching
-
-
-@patch(
-    "web3._utils.http_session_manager.HTTPSessionManager.make_post_request",
-    new_callable=Mock,
-)
-def test_sync_provider_is_batching_when_make_batch_request(mock_post):
-    def assert_is_batching_and_return_response(*_args, **_kwargs) -> bytes:
-        assert provider._is_batching
-        return b'{"jsonrpc":"2.0","id":1,"result":["0x1"]}'
-
-    provider = HTTPProvider()
-    assert not provider._is_batching
-
-    mock_post.side_effect = assert_is_batching_and_return_response
-
-    provider.make_batch_request([("eth_blockNumber", [])])
-    assert not provider._is_batching
+    assert not is_batching_context()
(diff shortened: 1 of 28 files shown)


tox.ini
pacrob, line 1: why adding all the space around `=`s?

fselmo: oh weird, must've been my local version of `black`? I'll put this back. We should look to see if we need to update the version of `black` in `make lint` / precommit at some point.

fselmo: wait this is `.ini` so not `black`... something must've updated in my PyCharm formatter ... will fix :/

fselmo: yep this was my fault, I turned something on recently in PyCharm that I didn't think was this but it was. Fixed 👍🏼

tests/core/providers/test_http_provider.py
```diff
@@ -127,26 +127,9 @@ def test_http_empty_batch_response(mock_post):
...
 
     # assert that even though there was an error, we have reset the batching state
     assert not w3.provider._is_batching
-
-
-@patch(
-    "web3._utils.http_session_manager.HTTPSessionManager.make_post_request",
-    new_callable=Mock,
-)
-def test_sync_provider_is_batching_when_make_batch_request(mock_post):
```
fselmo, line 141: ~~@pacrob I'm going to put all these back. The original implementation for thread safety didn't lend itself well to this test anymore but I put back `_is_batching` as the `ContextVar` holder / flag so I will put these back now!~~

fselmo: Oh nevermind. So this is actually not required for `provider.make_batch_request()` because the payload that is passed into that is the broken up request information list (list of `("eth_getBlockByNumber", ("latest", False))`), not requests via the API (`w3.eth.get_block("latest")`).

In the recent PR where [we added this](https://github.com/ethereum/web3.py/pull/3663), I was wrong to think we need the batching state while making these requests. We really only need it when we're suppressing the call to the method while calling via the API `w3.eth.get_block()`, etc...

This method should work just fine and your test for `_raise_stray_errors_from_cache` from that PR is all we need I believe.

tests/integration/go_ethereum/test_goethereum_legacy_ws.py
```diff
@@ -56,7 +56,10 @@ def w3(start_geth_process_and_yield_port):
 
 
 class TestGoEthereumWeb3ModuleTest(GoEthereumWeb3ModuleTest):
-    pass
+    def test_batch_requests_concurrently_with_regular_requests(
+        self, w3: "Web3"
+    ) -> None:
+        pytest.skip("LegacyWebSocketProvider does not support concurrent requests")
```
kclowes: nit: maybe use the decorator here to match `test_ethereum_tester` below, and so that it skips before running the test?

kclowes (approved):
LGTM! I left one nit, feel free to take or leave!

reedsa (approved):
lgtm

PR #3706, opened 2025-05-21
update 7702 test to test both send and send raw tx

fselmo:
### What was wrong?

- We aren't really testing `eth_sendTransaction` for set code transactions at the moment. We should make sure we have a test for this.

### How was it fixed?

- This PR changes the resetting of the code in the set code test to use `eth_sendTransaction`, whereas the previous step of setting the code uses `eth_sendRawTransaction`.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

🐍

### DIFF
--- a/web3/_utils/module_testing/eth_module.py
+++ b/web3/_utils/module_testing/eth_module.py
@@ -37,6 +37,7 @@ from eth_utils import (
 )
 from eth_utils.toolz import (
     assoc,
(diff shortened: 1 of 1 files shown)


kclowes (approved):
:shipit:

reedsa (approved):
Nice work, glad you spotted this one!

PR #3710, opened 2025-05-26
chore: fix typos

Shresth79:
### What was wrong?

There was a typo in the class name `PersistantSocket` in web3/providers/ipc.py. The correct spelling should be `PersistentSocket` (with an 'e' instead of 'a').

### How was it fixed?

Fixed the typo by renaming the class from `PersistantSocket` to `PersistentSocket` throughout the file.

#### Cute Animal Picture

### DIFF
--- a/newsfragments/3710.misc.rst
+++ b/newsfragments/3710.misc.rst
@@ -1,0 +1,1 @@
+Fix typo in class used for private API within ``IPCProvider``.
--- a/web3/providers/ipc.py
+++ b/web3/providers/ipc.py
@@ -59,7 +59,7 @@ def get_ipc_socket(ipc_path: str, timeout: float = 2.0) -> socket.socket:
         return sock
 
 
-class PersistantSocket:
+class PersistentSocket:
     sock = None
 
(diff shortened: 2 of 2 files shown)


fselmo:
This is a breaking change and this provider has been deprecated and will be removed. I'm going to close this.

fselmo:
Actually, I apologize on both of my comments and will re-open 😅.

I was under the impression the `PersistantSocket` was for the old WebSocket provider (now, `LegacyWebSocketProvider`). I don't think we changed this because it seems like a breaking change but it actually sets the socket on a private, underscored `_socket` which technically means it can and should be corrected - as this is not a public API.

I'm going to let others get eyes on this to make sure I'm not missing anything but I think this is a fine change to make, and I believe it to be a non-breaking change. Do you want to add a newsfragment for this? If not we can. And thanks for the PR!

fselmo:
@kclowes If you're good with this as non-breaking I can add a newsfragment and merge 👀

kclowes (approved):
Yep, LGTM!

fselmo:
Thanks @Shresth79 👍🏼

PR #3712, opened 2025-05-29
bugfix: Fix batching for AutoProvider

### LINKED ISSUE #3711: AutoProvider has no attribute `_batching_context`
### What happened?

We can no longer use AutoProvider without getting an AttributeError for `_batching_context`

```python

```

### Full error output

```shell

```

_No response_

### web3 Version

_No response_

### Python Version

_No response_

### Operating System

_No response_

### Output from `pip freeze`

```shell

```

fselmo:
### What was wrong?

Closes #3711

### How was it fixed?

- We can't set the provider, we must let `AutoProvider` decide when it makes proxy calls using the "active" provider.

- For the reason above, we need to implement a proxy batch request so that it can instead pull the active provider and make the batch request using it.

- Make sure current tests fail with the old implementation and pass with the new one.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-05-29 at 09 30 31

### DIFF
--- a/web3/providers/auto.py
+++ b/web3/providers/auto.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    List,
     Optional,
     Sequence,
     Tuple,
@@ -21,9 +22,9 @@ from web3.exceptions import (
     CannotHandleRequest,
 )
 from web3.providers import (
-    BaseProvider,
     HTTPProvider,
     IPCProvider,
+    JSONBaseProvider,
     LegacyWebSocketProvider,
 )
 from web3.types import (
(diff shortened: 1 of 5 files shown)


web3/providers/auto.py
```diff
@@ -35,7 +36,7 @@
 WS_SCHEMES = {"ws", "wss"}
 
 
-def load_provider_from_environment() -> BaseProvider:
+def load_provider_from_environment() -> Optional[JSONBaseProvider]:
```
fselmo, line 39: This narrows scope by a small margin, but I'd argue it's a bugfix. Thoughts here? Batching _requires_ `JSONBaseProvider`. All of web3 does at the moment, for that matter.

fselmo:
cc: @antazoey

web3/providers/auto.py
kclowes: I think that's fine

kclowes (approved):
LGTM! Nice test updates!

PR #3723, opened 2025-06-18
Added high-level diagrams

ivanmilevtues:
### What was wrong?

This PR introduces diagram representation of the web3.py codebase. You can see how it renders here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/web3.py/on_boarding.md

The goal of the diagram is to give an overview of the codebase, where it shows the main components and how they interact with eachother. We believe visualization is the best way to get to know new codebase, as it gives you context. When a new contributor starts their first task they can use the docs to narrow down the component in which they have to work and then find the relevant source files (which are linked), then when starting the task they'd have context on how what they are doing fits in the bigger picture of the codebase. Let me know what do you think of diagram first documentation generation!

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people. Further we have introduced free github action, to keep these docs up-to-date so that we don't introduce another maintainance task for the team :)

### How was it fixed?

Not relevant

### Todo:

Not super relevant

- [x] Clean up commit history

#### Cute Animal Picture

I love the American Fuzzy Lop :)

### DIFF
--- a/.codeboarding/Blockchain_Interaction_Layer_Eth_Contract_.md
+++ b/.codeboarding/Blockchain_Interaction_Layer_Eth_Contract_.md
@@ -1,0 +1,269 @@
+```mermaid
+
+graph LR
+
+    Eth["Eth"]
+
+    AsyncEth["AsyncEth"]
+
+    Contract["Contract"]
+
+    AsyncContract["AsyncContract"]
+
+    BaseContract["BaseContract"]
+
+    ABI_Encoding_Utilities["ABI & Encoding Utilities"]
+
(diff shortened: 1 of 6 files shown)


kclowes:
The diagram/docs you added aren't really the API we want to encourage users to use, and this information is handled already in our docs.

ivanmilevtues:
Hi @kclowes, thanks for answering!

The idea of these docs is for them to be read by contributors. So they share the insight structure of the codebase and a new contributor can find the component he wants to contribute to and start working on it immediately having context on how the whole system integrates!

PR #3726, opened 2025-06-25
removed (almost) unreachable code

simone1999:
checking for None after using dict.get with an default argument is never true except if the key is present in the dict and the value is explicitly None. This check seems to be non intentional, especially as it results in an unexpected behaviour where any RPC response containing response["error"]["data"] == None is treated as an ContractLogicError, no matter what error it actually is. Simply removing this code part makes the code both more readable and more correct.


kclowes:
This is working as intended.

PR #3732, opened 2025-07-15
fix: replace generic Exception with specific RequestException in exce…

reallesee:
### What was wrong?
Replace the overly broad `except Exception` with the more specific `except requests.RequestException` in the handle_offchain_lookup function. This change improves error handling by only catching expected network-related exceptions while allowing other unexpected errors to propagate properly, making the code more robust and easier to debug

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

image


kclowes:
Thanks for the PR, but we want to catch any exception here.

PR #3737, opened 2025-08-04
Add transaction tests for type 1 and type 2 with accessList (Fixes #3707)

### LINKED ISSUE #3707: Add more transaction tests
We don't seem to be testing certain transactions. It isn't really our job to test clients via our library, but we should be testing at least one case of each of these to make sure formatters are working correctly for every transaction field. 

Some that come to mind are `eth_sendTransaction` and `eth_sendRawTransaction` tests for each of:

- Sending access lists in type 1 transactions
- Sending access lists in type 2 transactions
- Sending access lists in type 4 transactions - perhaps just creating access lists locally for the transactions we are already sending in this test and attaching them to each of the transactions we send (both raw and send). This should test the full formatters for type 4 txs

darkiss7:
### What was wrong?
Lack of test coverage for certain eth_sendTransaction use cases involving accessList.

Related to Issue: #3707
Closes #3707

### How was it fixed?
Added two new tests in test_transactions.py:

✅ test_send_type1_transaction_with_access_list

✅ test_send_type2_transaction_with_access_list

These tests validate Web3.py's transaction formatters for:

Type 1 transactions (Legacy, with access list)

Type 2 transactions (EIP-1559, with access list)

All tests pass using the EthereumTester backend. Code is formatted with black, passes flake8, and other pre-commit hooks.

### Todo:
- [x] Clean up commit history

- [x] Add or update documentation related to these changes (N/A for tests)

- [ ] Add entry to the release notes (optional)

### DIFF
--- a/tests/core/eth-module/test_transactions.py
+++ b/tests/core/eth-module/test_transactions.py
@@ -588,3 +588,48 @@ async def test_async_send_set_code_transaction(async_w3, async_math_contract):
 
     reset_code = await async_w3.eth.get_code(acct.address)
     assert reset_code == HexBytes("0x")
+
+
+def test_send_type1_transaction_with_access_list(w3):
+    txn = {
+        "from": w3.eth.accounts[0],
+        "to": w3.eth.accounts[1],
+        "gas": 50000,
+        "gasPrice": w3.to_wei(1, "gwei"),
+        "accessList": [
+            {
+                "address": w3.eth.accounts[2],
+                "storageKeys": [
+                    "0x0000000000000000000000000000000000000000000000000000000000000000"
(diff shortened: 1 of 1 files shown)


darkiss7:
Hi, I pushed my changes but the CI fails on `test_key_generation_is_deterministic` with a Hypothesis `FailedHealthCheck: Input generation is slow`. 

This test passes fine locally but fails on CI due to input generation speed.

I’m considering adding `@settings(suppress_health_check=[HealthCheck.too_slow])` to fix this, but wanted to check if that’s the preferred approach or if you have other suggestions.

Thanks for your help!

PR #3745, opened 2025-08-28
chore: Clarify log wording

Shresth79:
## Description

`qsize` reflects the length of _subscription_response_queue (number of pending responses/messages), not the number of active subscriptions. The previous wording “has %s subscriptions” was misleading.


kclowes:
Thanks for the PR, but I don't think this adds much.

Issue #3679, opened 2025-04-24
v8 Breaking Changes Meta Issue


kclowes:
A place to keep track of breaking changes we might want to add to v8.

fselmo:
Can't add as sub-issue 😩 : #3580

kclowes:
Can't add this to the nice sub-issue dropdown because it's already a PR, but #3581 belongs on here too

Issue #3530, opened 2024-11-14
move `websockets` bottom pin to `>=14`


pacrob:
### What feature should we add?

Use `websockets>=14` and migrate away from the legacy path. See what features we can make use of.

kclowes:
Closed via #3779

PR #3749, opened 2025-09-02
Fix websockets deprecation warning

mdalp:
### What was wrong?
Importing `from web3 import Web3` was causing a DeprecationWarning to be raised.
This can be easily replicated by doing:
```shell
$ pytest tests/core/providers/test_legacy_websocket_provider.py -W error::DeprecationWarning
ImportError while loading conftest '/Users/mdalp/coding/web3.py/conftest.py'.
conftest.py:7: in <module>
    from tests.utils import (
tests/utils.py:8: in <module>
    from websockets.legacy.client import (
.venv/lib/python3.13/site-packages/websockets/legacy/__init__.py:6: in <module>
    warnings.warn(  # deprecated in 14.0 - 2024-11-09
E   DeprecationWarning: websockets.legacy is deprecated; see https://websockets.readthedocs.io/en/stable/howto/upgrade.html for upgrade instructions
```
This PR aims at resolving that once for all.

### How was it fixed?
1. lazy loading `websockets.legacy` in the legacy provider
2. using the recommended implementation from the `websockets` package
3. running tests to ensure they still pass

### Todo:

- [x] Clean up commit history

### Notes:
The pre-commit hook changed quite a lot the file, I ended up committing those changes in a separate commit so that the relevant changes can be seen in isolation.

#### Cute Animal Picture
Cute baby xenomorph

### DIFF
--- a/web3/providers/legacy_websocket.py
+++ b/web3/providers/legacy_websocket.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
 import asyncio
 import json
 import logging
 import os
+import typing
 from threading import (
     Thread,
 )
@@ -21,10 +23,6 @@ from typing import (
 from eth_typing import (
     URI,
 )
-from websockets.legacy.client import (
-    WebSocketClientProtocol,
-    connect,
-)
 
 from web3._utils.batching import (
     sort_batch_response_by_response_ids,
@@ -42,6 +40,11 @@ from web3.types import (
     RPCEndpoint,
     RPCResponse,
 )
+
+if typing.TYPE_CHECKING:
+    from websockets.legacy.client import (
+        WebSocketClientProtocol,
+    )
 
 RESTRICTED_WEBSOCKET_KWARGS = {"uri", "loop"}
 DEFAULT_WEBSOCKET_TIMEOUT = 30
@@ -72,6 +75,7 @@ class PersistentWebSocket:
 
     async def __aenter__(self) -> WebSocketClientProtocol:
         if self.ws is None:
+            from websockets.legacy.client import connect
             self.ws = await connect(uri=self.endpoint_uri, **self.websocket_kwargs)
         return self.ws
 
--- a/web3/providers/legacy_websocket.py
+++ b/web3/providers/legacy_websocket.py
@@ -12,8 +12,8 @@ from threading import (
 from types import (
     TracebackType,
 )
-import typing
 from typing import (
+    TYPE_CHECKING,
     Any,
     List,
     cast,
@@ -40,7 +40,7 @@ from web3.types import (
     RPCResponse,
 )
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from websockets.legacy.client import (
         WebSocketClientProtocol,
     )
--- a/web3/providers/legacy_websocket.py
+++ b/web3/providers/legacy_websocket.py
@@ -1,22 +1,21 @@
-from __future__ import annotations
+from __future__ import (
+    annotations,
+)
+
 import asyncio
 import json
 import logging
 import os
-import typing
 from threading import (
     Thread,
 )
 from types import (
     TracebackType,
 )
+import typing
 from typing import (
     Any,
     List,
-    Optional,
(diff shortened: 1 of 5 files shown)


web3/providers/persistent/websocket.py
```diff
@@ -136,7 +146,7 @@ async def _provider_specific_connect(self) -> None:
 
     async def _provider_specific_disconnect(self) -> None:
         # this should remain idempotent
-        if self._ws is not None and not self._ws.closed:
+        if self._ws is not None:
```
kclowes, line 149: I think we still need to figure out how to check for `self._ws.closed`, even if it's not that same method name.

web3/providers/legacy_websocket.py
```diff
@@ -43,6 +40,11 @@
     RPCResponse,
 )
 
+if typing.TYPE_CHECKING:
```
kclowes: If you import `TYPE_CHECKING` from the typing lib above this becomes: 
```suggestion
if TYPE_CHECKING:
```

web3/providers/legacy_websocket.py
```diff
@@ -8,23 +12,16 @@
 from types import (
     TracebackType,
 )
+import typing
```
kclowes: We prefer to import what we need from typing instead of importing the whole library. Elsewhere we import `TYPE_CHECKING` directly, so let's do the same here.

kclowes (commented):
Thanks for this PR! This is an (admittedly pretty small) breaking change with the type from `WebSocketClientProtocol` -> `ClientConnection`, so we'll need to wait until v8 to get this in. I am going to start working on v8 changes in the next few weeks, so will be able to get these changes in then. I left a couple comments that I can either address when I pull it in under v8, or if you want/have bandwidth to update, feel free!

I didn't know that you could use the Union `|` before Python 3.10, that's awesome!

mdalp:
Hey @kclowes ! I agree this is a potentially breaking change, we might even not need it in this shape, depending on when it is planned to sunset the `legacy_websocket` provider implementation. But definitely it makes sense to wait until v8.

About your comments, I can definitely take care of them 💪 , thanks for those.

web3/providers/legacy_websocket.py
mdalp: Upsie, you're right. It must have been my IDE autocompleting that.
Fixed here [b473927](https://github.com/ethereum/web3.py/pull/3749/commits/b4739277339e21c38c00f64f08a0028bed2de4a7).

web3/providers/persistent/websocket.py
mdalp: Hey, I did check this and I found out that `self._ws.close()` is idempotent already, what I didn't realise though is that web3.py still supports websockets `>=10.0,<13.0`, meaning me importing directly from `websockets.asyncio` would break that compatibility.
So I would need to rework my PR to take that into account.

I also noticed this issue https://github.com/ethereum/web3.py/issues/3679 which plans to [move websockets bottom pin to >=14 #3530](https://github.com/ethereum/web3.py/issues/3530), which would essentially remove this problem completely.
Should I just do that?

mdalp: I could either go down the path of supporting all the variations (aka <13; >=13,<14; >=14), i.e.
```python
import websockets
websockets_version = tuple(int(x) for x in websockets.__version__.split(".") if x.isdigit())

if websockets_version < (13, 0):
    from websockets.legacy.client import (
        WebSocketClientProtocol as ClientConnection,  # we are safe to steal the name as the scope of interface we use is the same
        connect,
    )
else:
    # python3.8 supports up to version 13,
    # which does not default to the asyncio implementation yet.
    # For this reason connect and ClientConnection need to be imported
    # from asyncio.client explicitly.
    # When web3.py stops supporting python3.8,
    # it'll be possible to use `from websockets import connect, ClientConnection`.
    from websockets.asyncio.client import (
        ClientConnection,
        connect,
    )
```

or simply drop support for at least version <13.

Given the amount of usage of the websockets library interface either solution is fine, not a big deal.

Issue #3681, opened 2025-04-24
Drop Python 3.9 support


kclowes:
Python 3.9 EOL is Oct 31, 2025. We'll be working on v8 at that point, and so can drop python 3.9 then.

PR #3774, opened 2025-10-29
Drop Support for Python 3.8 & 3.9

### LINKED ISSUE #3680: Drop Python 3.8 support

kclowes:
### What was wrong?

Python 3.8 has been EOL for a year, and Python 3.9 just reached its EOL, so it's time to drop our support as well. 

Closes #3680, #3681 

### How was it fixed?
Dropped support in setup.py, and ran pyupgrade from 3.9 -> 3.9, and subsequently 3.9 -> 3.10.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
Amazing, lgtm 🚀 ! Checked for `3.8` and `3.9` in a find and replace... surprised there weren't more comments related to this upgrade. Love the type hinting updates, finally 😅. TIL about `pyupgrade` but makes perfect sense this exists... really nice 👌🏼 

One thing I just thought about is starting to aggregate the breaking changes into a migration section. I think this may be best to do along the way rather than retroactively but open to doing this later / in a separate PR as well. So far I think it's just this and the `LegacyWebSocketProvider` removal.

kclowes:
> One thing I just thought about is starting to aggregate the breaking changes into a migration section. I think this may be best to do along the way rather than retroactively but open to doing this later / in a separate PR as well. So far I think it's just this and the LegacyWebSocketProvider removal.

Yeah, we def should. I'll add those two in a separate PR and put up an issue to track. Going forward I think it would be good to add at the same time though 👍

PR #3513, opened 2024-10-15
Silence some warnings related to copying itertools objects

fselmo:
### What was wrong?

Related to Issue #3511

### How was it fixed?

Don't deepcopy the `itertools.count()` counter object for the request mocker. This should get rid of most warnings related to this in the test suite. It'll be a bit more difficult to tackle the `PersistentConnectionProvider` case (in its request information cache) of copying the request counter. We may have to switch to a different object for the counter itself in order to avoid copying an itertools object.

### Todo:

- [ ] Clean up commit history
- [ ] Add or update documentation related to these changes
- [ ] Add entry to the [release notes](https://github.com/ethereum/web3.py/blob/main/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

Issue #3511, opened 2024-10-14
Address warnings present in test suite


fselmo:
We are getting a lot of warnings in our test suite related to:

```
  /Users/...../.pyenv/versions/3.12.3/lib/python3.12/copy.py:151: DeprecationWarning: Pickle, copy, and deepcopy support will be removed from itertools in Python 3.14.
```

We should silence these and see if there are any other warnings coming through that need silencing.

fselmo:
#3513 addresses the main case of this. The `PersistentConnectionProvider` request information cache also copies the counter. This may be more difficult to tackle.

fselmo:
Closed by [#3554](https://github.com/ethereum/web3.py/commit/82d67320bd77dc984a2c60cdf3c238e6fd7cd838#diff-12b1a6ab0cb9e0aa9c2611374a311c54677b6ccf915cde7b4a25ae690e2cd4a9L136-L142) and #3779

PR #3779, opened 2025-11-05
Add Python 3.14 Support

kclowes:
### What was wrong?
Python 3.14 is now out and stable. Added those runs to our CI. 

Related to Issue #3679 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/providers/test_websocket_provider.py
+++ b/tests/core/providers/test_websocket_provider.py
@@ -357,21 +357,17 @@ async def test_async_iterator_pattern_exception_handling_for_requests():
             raise_exception=ConnectionClosed(None, None)
         ),
     ):
-        agen = AsyncWeb3(WebSocketProvider("ws://mocked")).__aiter__()
-        try:
-            async for w3 in agen:
-                try:
-                    await w3.eth.block_number
-                except ConnectionClosed:
-                    if iterations == 3:
-                        break
-                    else:
-                        iterations += 1
-                        continue
+        async for w3 in AsyncWeb3(WebSocketProvider("ws://mocked")):
+            try:
+                await w3.eth.block_number
+            except ConnectionClosed:
+                if iterations == 3:
+                    break
+                else:
+                    iterations += 1
+                    continue
 
-                pytest.fail("Expected `ConnectionClosed` exception.")
-        finally:
-            await agen.aclose()
+            pytest.fail("Expected `ConnectionClosed` exception.")
 
         assert iterations == 3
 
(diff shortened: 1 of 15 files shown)


tests/core/providers/test_websocket_provider.py
```diff
@@ -42,12 +45,12 @@
 
 def _mock_ws(provider):
     provider._ws = AsyncMock()
-    provider._ws.closed = False
+    provider._ws.state = State.OPEN
```
kclowes, line 48: The `websocket` interface changed, and in the new world, there is a check for `State.OPEN` instead of `conn.closed`. I updated these mock providers so that they more closely match the websocket providers.

tests/utils.py
```diff
@@ -65,7 +66,7 @@ async def _async_wait_for_transaction_fixture_logic(async_w3, txn_hash, timeout=
 def async_partial(f, *args, **kwargs):
     async def f2(*args2, **kwargs2):
         result = f(*args, *args2, **kwargs, **kwargs2)
-        if asyncio.iscoroutinefunction(f):
+        if inspect.iscoroutinefunction(f):
```
kclowes, line 69: DeprecationWarning fix

tests/core/providers/test_websocket_provider.py
```diff
@@ -379,19 +386,23 @@ async def test_async_iterator_pattern_exception_handling_for_subscriptions():
...
-                    pass
-            except ConnectionClosed:
-                if iterations == 3:
-                    break
-                else:
-                    iterations += 1
-                    continue
-
-            pytest.fail("Expected `ConnectionClosed` exception.")
+        agen = AsyncWeb3(WebSocketProvider("ws://mocked")).__aiter__()
```
kclowes: We were getting some warnings in tests around a `ConnectionNotClosed` error. The `finally` block fixes this.

web3/_utils/module_testing/utils.py
```diff
@@ -113,12 +110,11 @@ def __init__(
                 "AsyncMakeRequestFn", "MakeRequestFn"
             ] = w3.provider.make_request
 
+        self._mock_request_counter = 1
+
     def _build_request_id(self) -> int:
-        request_id = (
-            next(copy.deepcopy(self.w3.provider.request_counter))
```
kclowes, line 118: This was raising a: `TypeError: cannot pickle 'itertools.count'` because itertools objects can no longer be deepcopied or pickled in 3.14. So I opted to just manually track it here. The logic in the WebSocket Providers remains the same.

tests/core/providers/test_websocket_provider.py
fselmo: I'd rather not make this change. It looks like updating `pytest-asyncio` (which is where the issue was coming from) fixed the warning for me locally.

If we end up testing this `.__aiter__()` pattern instead of the `async for` directly, it feels like too much of a deviation to the actual behavior of the API imo. I think we should lower-pin `pytest-asyncio>=1.3.0` if we can. Can you double check that fixes it on your end?

web3/_utils/module_testing/utils.py
fselmo: Yeah I think that works well. There's an issue to track this [here](https://github.com/ethereum/web3.py/issues/3511). I think the fact that we just use our own counter shouldn't interfere with anything right? I had used the provider's request counter directly (no need to copy really) in [this PR](https://github.com/ethereum/web3.py/pull/3513/changes). Thoughts?

And I think what was mentioned in that issue with the `PersistentConnectionProvider` was fixed a while ago [here](https://github.com/ethereum/web3.py/commit/82d67320bd77dc984a2c60cdf3c238e6fd7cd838#diff-12b1a6ab0cb9e0aa9c2611374a311c54677b6ccf915cde7b4a25ae690e2cd4a9L136-L142). So I think both that issue and my stale PR can be closed :)

fselmo (commented):
Just a comment that I'd rather not change the way the tests with the warnings work, just to silence the warning. I want to make sure that the intended API design doesn't issue a warning when it shouldn't (connection closes properly). I looked into it and it did look correct. I updated `pytest-asyncio` and it seems to have fixed the issue for me.

Would be good to double check the update works and I wouldn't mind just pinning the higher version.

web3/_utils/module_testing/utils.py
kclowes: Oh, sweet! IIRC, it was a 0 vs 1 indexed problem. I can't remember which was which, but either this or the provider started at 0, and the other started at 1, so we had to update the tests. It shouldn't impact the provider at all. But I'm happy to go with either solution!

fselmo (approved):
lgtm 🚀

PR #3787, opened 2025-12-08
Upgrade geth to 1.16.7

kclowes:
### What was wrong?

Updates the geth binary that we run web3.py testing on. Had to update the URL that go binaries get downloaded from. I think that circleci was using cached go binaries which was why that change didn't get caught until now. 

### How was it fixed?
Ran the upgrade script, upgraded go version (lmk if you think this should be a separate PR, I don't think it really matters to this PR) and updated the place the go binary is sourced from.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
lgtm!

PR #3811, opened 2026-03-01
fix: change r and s types from HexBytes to int in SetCodeAuthoriz…

### LINKED ISSUE #3716: Serialization bug with authorization signatures
I checked, and it uses `HexBytes`, which is wrong. So you need to bring it up with them to change the encoding.

https://github.com/ethereum/web3.py/blob/0daa470f9bd7f3eaf0316684f2eaa8fdcb24bb43/web3/types.py#L106

_Originally posted by @fjl in https://github.com/ethereum/go-ethereum/issues/31967#issuecomment-2948404441_

ozge-devops:
Fixes #3716

r and s fields in SetCodeAuthorizationData were incorrectly typed as HexBytes. Changed to int as per the EIP-7702 specification.

### What was wrong?
r and s signature fields were typed as HexBytes instead of int.

### How was it fixed?
Changed both r and s field types from HexBytes to int in SetCodeAuthorizationData TypedDict.

### DIFF
--- a/web3/types.py
+++ b/web3/types.py
@@ -111,8 +111,8 @@ class SetCodeAuthorizationData(TypedDict):
     address: ChecksumAddress
     nonce: Nonce
     yParity: int
-    r: HexBytes
-    s: HexBytes
+    r: int
+    s: int
 
 
 # syntax b/c "from" keyword not allowed w/ class construction


fubuloubu:
Closed in favor of #3864

PR #3832, opened 2026-05-01
Fix operator precedence on evicted-session close timeout

genisis0x:
## What was wrong?

`HTTPSessionManager.cache_and_return_session` (sync) and `async_cache_and_return_session` (async) schedule a close-evicted-sessions task to fire after the in-flight request would have timed out, plus a small `0.1`s slack so any request that started right before the eviction has time to finish before the underlying session is closed.

Both call sites used:

```python
# sync, web3/_utils/http_session_manager.py:99
request_timeout or DEFAULT_HTTP_TIMEOUT + 0.1
# async, web3/_utils/http_session_manager.py:271
request_timeout.total or DEFAULT_HTTP_TIMEOUT + 0.1
```

`+` binds tighter than `or` in Python, so the expression actually parses as `request_timeout or (DEFAULT_HTTP_TIMEOUT + 0.1)`. The intended `+ 0.1` slack only applied when the caller fell back to the default. Any caller that supplied a non-default timeout (e.g. `request_timeout=5`) had its close timer fire at exactly `5.0`s, racing the in-flight request:

```python
>>> request_timeout = 5
>>> DEFAULT_HTTP_TIMEOUT = 30
>>> request_timeout or DEFAULT_HTTP_TIMEOUT + 0.1
5
>>> (request_timeout or DEFAULT_HTTP_TIMEOUT) + 0.1
5.1
```

The comments above each call site already describe the intended behavior (`wait over the default timeout` / `bit more than the request_timeout`).

## How was it fixed?

Parenthesize the `or` so the slack is always applied to the chosen timeout:

```python
(request_timeout or DEFAULT_HTTP_TIMEOUT) + 0.1
```

Same change in both the sync and async paths.

## Cute Animal Picture

cat

#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

#### Test plan

Added focused regression tests in `tests/core/utilities/test_http_session_manager.py`:
- `test_session_manager_evicted_close_timer_uses_request_timeout_plus_slack` — patches `threading.Timer`, fills a 1-entry cache, and asserts the captured interval is `request_timeout + 0.1`.
- `test_session_manager_evicted_close_timer_falls_back_to_default` — same path with `request_timeout=None`, asserts the interval is `DEFAULT_HTTP_TIMEOUT + 0.1`.
- `test_async_evicted_close_task_uses_request_timeout_plus_slack` — async equivalent that patches `_async_close_evicted_sessions` and asserts it was scheduled with `ClientTimeout.total + 0.1`.

The pre-existing `test_async_session_manager_cache_does_not_close_session_before_call` already exercised the async eviction path and continues to pass with the change (it waited for `_timeout + 0.1` regardless and is not sensitive to the bug).

### DIFF
--- a/newsfragments/3832.bugfix.rst
+++ b/newsfragments/3832.bugfix.rst
@@ -1,0 +1,6 @@
+Fix an operator-precedence bug in ``HTTPSessionManager`` so the
+``+ 0.1`` slack on the evicted-session close timer is applied to the
+caller-supplied ``request_timeout`` rather than only to the
+``DEFAULT_HTTP_TIMEOUT`` fallback. Previously a non-default timeout
+caused the close timer to fire at exactly the request timeout, racing
+the in-flight request.
--- a/tests/core/utilities/test_http_session_manager.py
+++ b/tests/core/utilities/test_http_session_manager.py
@@ -26,6 +26,9 @@ from requests.adapters import (
 from web3._utils.caching import (
     generate_cache_key,
 )
+from web3._utils.http import (
+    DEFAULT_HTTP_TIMEOUT,
+)
 from web3._utils.http_session_manager import (
     HTTPSessionManager,
 )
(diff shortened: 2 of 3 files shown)


genisis0x:
Hi @fselmo @kclowes — also ~18 days open. Single-line operator-precedence fix on the evicted-session close timeout (current `a or b * c` reads wrong). Would one of you have a moment? Thanks.

PR #3838, opened 2026-05-13
Replace Wei and Gwei NewType aliases with arithmetic-preserving int subclasses

### LINKED ISSUE #3833: `Wei` and `Gwei` lose their type through arithmetic — `+=` / `-=` rejected by mypy
## Problem

`Wei` and `Gwei` are currently defined as `NewType` aliases over `int`:

```python
Wei = NewType("Wei", int)
Gwei = NewType("Gwei", int)
```

`NewType` creates a distinct type for the type checker, but all arithmetic operations return the *base* type (`int`), not the `NewType`. This means any arithmetic on `Wei` or `Gwei` values immediately "escapes" the type and mypy flags the result as `int`, not `Wei`/`Gwei`.

In practice, augmented assignment (`+=`, `-=`) is rejected outright:

```python
x = Wei(5)
x += Wei(1)   # error: Incompatible types in assignment (expression has type "int", variable has type "Wei")
```

Even plain addition requires an explicit cast every time:

```python
total: Wei = Wei(0)
for tx in block["transactions"]:
    total = Wei(total + tx["value"])   # cast needed on every accumulation
```

This affects real-world usage patterns — accumulating gas costs, summing balances, computing fee estimates — and makes well-typed Ethereum code significantly more verbose.

## Proposed fix

Replace the `NewType` aliases with `int` subclasses that override arithmetic operators to preserve the return type:

```python
class Wei(int):
    def __add__(self, other: "Wei") -> "Wei": ...
    def __sub__(self, other: "Wei") -> "Wei": ...
    def __mul__(self, other: int) -> "Wei": ...
    # etc.

class Gwei(int):
    ...
```

Because the classes subclass `int`, they are fully backward-compatible: anywhere `int` is accepted, `Wei`/`Gwei` still work. The `isinstance` check also becomes meaningful, which opens the door for formatters to return typed values instead of plain `int`.

As a natural follow-up, the result formatters in `method_formatters.py` can wrap monetary fields (`gasPrice`, `value`, `maxFeePerGas`, `effectiveGasPrice`, `baseFeePerGas`, `eth_getBalance`, etc.) with `Wei(...)` and withdrawal `amount` with `Gwei(...)`, so that values coming out of RPC calls already carry the correct type at the boundary.

## Backward compatibility

- `Wei` and `Gwei` are `int` subclasses, so all existing code that treats them as `int` continues to work without modification.
- `NewType` values cannot be used with `isinstance`; the new classes can — this is strictly additive.
- The formatter changes are internal and do not affect the public API.

## Prior art

The `eth_typing` library uses a similar pattern for `BlockNumber` (also a `NewType`), and the same limitation applies there. This proposal is scoped to `Wei` and `Gwei` since those are most commonly involved in arithmetic.

genisis0x:
## What was wrong?

Closes #3833.

\`Wei\` and \`Gwei\` were \`NewType(\"Wei\", int)\` / \`NewType(\"Gwei\", int)\` aliases. \`NewType\` is a typing-only construct, so any arithmetic on a typed value immediately returns bare \`int\`. Strict-mypy users hit this on every accumulator:

\`\`\`python
total: Wei = Wei(0)
for tx in block[\"transactions\"]:
    total = Wei(total + tx[\"value\"])  # cast required on every step
\`\`\`

and augmented assignment is rejected outright:

\`\`\`python
x: Wei = Wei(5)
x += Wei(1)
# error: Incompatible types in assignment (expression has type \"int\", variable has type \"Wei\")
\`\`\`

## How was it fixed?

Replace the \`NewType\` aliases with \`int\` subclasses (via a shared \`_PreservingInt\` base) that override the arithmetic dunders to return \`type(self)\`.

**Operators that preserve the type:** \`+\`, \`-\`, \`*\`, \`//\`, \`%\`, unary \`+\`, unary \`-\`, \`abs()\`.

**Operators that don't (intentional):** \`/\` (true division) keeps the \`int\` base implementation, since dividing a quantity by a scalar yields \`float\`.

### Runtime impact is contained

- \`Wei(123)\` / \`Gwei(123)\` keep working (class is callable with an int).
- \`isinstance(Wei(5), int)\` is \`True\`.
- \`Wei(5) == 5\` is \`True\`; \`hash(Wei(5)) == hash(5)\`.
- JSON-serialises as the underlying int.
- Mixed-type arithmetic (\`Wei + Gwei\`) follows Python's standard left-operand rule for int subclasses. The goal of #3833 is **type preservation**, not cross-unit conversion in the type system.

I grepped for any runtime \`isinstance(..., Wei)\` / \`isinstance(..., Gwei)\` checks in \`web3/\`, \`tests/\`, \`ens/\`, and \`docs/\` — there are none — so existing callers that rely on the NewType identity-function behaviour are unaffected.

## Tests

New \`tests/core/utilities/test_wei_gwei_types.py\` exercises each branch under \`@pytest.mark.parametrize(\"cls\", [Wei, Gwei])\`:

- value is an \`int\` subclass and an instance of the concrete class
- \`+\`, \`-\`, \`*\`, \`//\`, \`%\` preserve the type from both sides
- augmented \`+=\` preserves the type
- unary \`+\`, \`-\`, \`abs()\` preserve the type
- \`/\` returns \`float\` (documented behaviour, not preserved)
- \`json.dumps\` emits the integer value
- \`Wei\` and \`Gwei\` are distinct types and follow the left-operand rule for mixed arithmetic

\`python3 -m py_compile\` clean on both files. (Local environment is Python 3.9 so I couldn't run the full pytest suite — happy to iterate on anything CI flags.)

## News fragment

\`newsfragments/3833.feature.rst\` added.

## Checklist

- [x] Added or updated unit tests
- [x] Updated relevant documentation — the docstrings on \`Wei\` / \`Gwei\` now describe the new arithmetic behaviour
- [x] Added an entry to \`newsfragments/\` describing the fix

### DIFF
--- a/newsfragments/3833.feature.rst
+++ b/newsfragments/3833.feature.rst
@@ -1,0 +1,6 @@
+``Wei`` and ``Gwei`` are now ``int`` subclasses that preserve their own type
+through arithmetic operations (``+``, ``-``, ``*``, ``//``, ``%``, unary
+``+``/``-``, ``abs()``). Previously they were ``NewType`` aliases, so any
+arithmetic — including ``+=`` and ``-=`` — escaped back to bare ``int`` and
+forced an explicit cast on every accumulator step. Strict-mypy code can now
+write ``total: Wei = Wei(0); total += tx_value`` without a cast.
--- a/tests/core/utilities/test_wei_gwei_types.py
+++ b/tests/core/utilities/test_wei_gwei_types.py
@@ -1,0 +1,85 @@
+import json
+
+import pytest
+
+from web3.types import (
+    Gwei,
+    Wei,
+)
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
+def test_value_is_int_subclass(cls):
+    value = cls(5)
+    assert isinstance(value, int)
+    assert isinstance(value, cls)
+    assert value == 5
+    assert hash(value) == hash(5)
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
+def test_addition_preserves_type(cls):
+    assert type(cls(2) + cls(3)) is cls
+    assert type(cls(2) + 3) is cls
+    assert type(3 + cls(2)) is cls
+
+    total = cls(0)
+    total += cls(7)
+    assert total == 7
+    assert type(total) is cls
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
+def test_subtraction_preserves_type(cls):
+    assert type(cls(5) - cls(2)) is cls
+    assert type(cls(5) - 2) is cls
+    assert type(10 - cls(2)) is cls
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
+def test_multiplication_by_scalar_preserves_type(cls):
+    assert type(cls(5) * 3) is cls
+    assert type(3 * cls(5)) is cls
+    assert cls(5) * 3 == 15
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
+def test_floor_division_and_mod_preserve_type(cls):
+    assert type(cls(10) // 3) is cls
+    assert type(cls(10) % 3) is cls
+    assert cls(10) // 3 == 3
+    assert cls(10) % 3 == 1
+
+
+@pytest.mark.parametrize("cls", [Wei, Gwei])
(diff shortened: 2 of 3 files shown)


genisis0x:
Opened against wrong fork; reopening on ethereum/web3.py.

PR #3877, opened 2026-09-13
fix(subscriptions): enforce bool|None type contract for parallelize

zuyua9:
## What changed
- `AsyncEth.subscribe()` and `EthSubscription.__init__()` now raise `Web3TypeError` when `parallelize` is neither a `bool` nor `None`
- the check runs before the provider-type check in `subscribe()`, so argument errors are independent of the provider
- regression tests cover both entry points plus `None`/`bool` acceptance

## Why
Both entry points annotate `parallelize: bool | None` but never enforce it at runtime. `SubscriptionManager` dispatch only matches `is True` / `is None` (`subscription_manager.py`), so a truthy non-bool value such as `"yes"` silently disables handler parallelization instead of failing loudly — the exact kind of annotation/runtime drift that is easy to hit in typed-but-unvalidated keyword arguments.

## Tests
`python -m pytest tests/core/subscriptions/test_subscription_manager.py` → 17 passed (14 pre-existing + 3 new)


zuyua9:
Closing — opened against the wrong repository due to a pre-existing fork-lineage mixup in this account namespace (this fork tracks ApeWorX/web3.py, not ethereum/web3.py). The intended PR has been filed against ethereum/web3.py. Apologies for the noise.

zuyua9:
(Self-note: this PR was briefly closed and reopened while sorting out a fork-network mixup on my side — the head branch was restored unchanged at the same commit `a61a155a`. No code changes.)

